PR #167, opened 2016-10-26
PLAT-29542: Create storybook sample for flexbox

sugardave:

Added (to qa-stories) an example of using flexbox.
### Resolution

Since `Fittables` are no more, it is good to have at least a simple example of flexbox usage.
### Links

PLAT-29542

Enyo-DCO-1.1-Signed-off-by: Dave Freeman dave.freeman@lge.com


webOS101:
Don't need an example of browser features.  We also have a sample of the new `Layout` module.

PR #599, opened 2017-03-09
ENYO-3970: Ensure Spottable forwards enter keyup events

germboy:
When a spottable control is selected, as a result of selecting it with the enter key (via 5-way), and focus is immediately changed programmatically to another control (while the enter key is still pressed), the selected control does not forward a keyup event. This is problematic, since it means certain controls (such as Button) is unable to remove the pressed CSS class.

### Resolution
We can keep a reference to the enter key's `keydown` event, then when the component is updated, we can determine if the component is no longer spotted and if the down event still exists, we can forward a `keyup` event.

### Additional Considerations
In the case of `Button`, when it is selected, it needs to animate (via an applied "pressed" class). If focus is removed from the `Button` immediately upon pressing it, there is not enough time to animate the `Button`. I've introduced a `selectionAnimationDuration` prop to `Spottable`'s config that will guarantee this animation can complete in these instances.


ryanjduffy:
Replaced by #691

PR #632, opened 2017-03-23
ENYO-3774: Fix containers as they relate to Popup-ContextualPopup

germboy:
This PR resolves issues related to spotlight-navigation using `Popup` and `ContextualPopup`. There are some underling spotlight container changes that were needed in order to add this functionality.

### Resolution
This PR introduces spotlight behavior that allows/denies navigation based on an active container `_lastContainerId`. Measures are taken to only change the active container under specific scenarios. Also, spottable components can no longer be spotted via a pointer if they exist outside of a container using a `restrict: 'self-only'` container config.

In these cases we naturally prevent forwarding click events, since the component isn't spotted/active. A byproduct of this is - in our test suite when attempting to call `component.simulate('click')`. The click event isn't being forwarded because the component isn't spotted, unless you to call `component.simulate('focus')` first.

### Additional Considerations
There are some edge-case scenarios to keep in mind when dealing with 5-way behavior in `Popup`. Behaviors can be dictated by the following individual (or combination of) props:
`noAutoDismiss`, `scrimType`, `spotlightRestrict`, then behavior is further affected if the `Popup` contains (or does not contain) spottable components.

### Links
There are a series of in-progress PRs that this PR was dependent upon in order to work. The following PRs were incorporated into this fix:
ENYO-3864 - https://github.com/enyojs/enact/pull/594
ENYO-3987 - https://github.com/enyojs/enact/pull/605
ENYO-3970 - https://github.com/enyojs/enact/pull/599
ENYO-3905 - https://github.com/enyojs/enact/pull/600

### DIFF
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -232,6 +232,7 @@ const ButtonFactory = factory(css => {
 				{className: componentCss.marquee},
 				Pressable(
 					Spottable(
+						{selectionKeyUpDelay: 200},
 						Base
 					)
 				)
--- a/packages/moonstone/Button/Button.less
+++ b/packages/moonstone/Button/Button.less
@@ -146,8 +146,7 @@
 
 	// Button-non-disabled rules
 	&:not([disabled]) {
-		&.pressed,
-		&:global(.spottable):active {
+		&.pressed {
 			.bg {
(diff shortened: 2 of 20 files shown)


ryanjduffy:
Replaced by #691

PR #942, opened 2017-07-06
ENYO-4372: Add ApplicationCloseDecorator

viodragon2:
`ApplicationCloseButton` is part of `Panels`, which makes a single-panel-structured app to wrap in Panels if they want to use it.

### Resolution
Add `ApplicationCloseDecorator` which will add `AplicationCloseButton` at the application level. This allows us to have application close buttons to non-panel type apps as well.

Enyo-DCO-1.1-Signed-off-by: Stephen Choi


lishammel:

PR #965, opened 2017-07-13
Add doc parsing to Travis

webOS101:

### Resolution

### Additional Considerations
This could add a lot of overhead to the travis process.  I think it could be simplified to only install
the dependencies needed to parse the docs, which would speed up install.

### Links
PLAT-43172


lishammel:
@webOS101  - is this still needed?

webOS101:
No, let's close this.  We could test it again with the module caching but it will need to be redone against `develop` and with the latest options.

PR #976, opened 2017-07-17
ENYO-4248: P&V: VideoPlayer mini feedback

sugardave:
"Mini" feedback should be displayed when controlling the video playback via remote control and only when the playback controls are not visible.

### Resolution

### Additional Considerations
I can't get left/right (skip back/forward) to function while the feedback/slider are displayed, so I can't verify the changes I made to the feedback icons are correct.

### Links
ENYO-4248

### Comments
I'm anticipating comments about the styles :D


ryanjduffy:
Superseded by #1126

PR #992, opened 2017-07-26
PLAT-43488: Force to update when jumping to 0 index

ybsung:

When jumping to the target item far from a current item, we added a placeholder in `VirtualList` to keep focus in it. Then we deleted it when the target item got to focus. But when jumping to 0 index, there was still a placeholder even though the 0 index item got to focus.

### Resolution

When getting to focus the 0 index item after jumping, this.nodeIndexToBeFocused became null. But render() in `VirtualListBase` was not called. On the other hand, when jumping to another index item, this.nodeIndexToBeFocused became null, the `firstIndex` of `VirtualList`'s state changed, then the render() was called 1 or 2 times more. Finally, the placeholder could be deleted correctly.

So I tried to force render() when this.nodeIndexToBeFocused became null.

### Additional Considerations

### Links
PLAT-43488

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -699,6 +699,9 @@ class VirtualListCore extends Component {
 
 			this.focusOnNode(item);
 			this.nodeIndexToBeFocused = null;
+			if (index === 0) {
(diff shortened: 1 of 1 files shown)


ybsung:
I created the new PR #1026 instead of this PR.

PR #1008, opened 2017-08-01
VideoPlayer: support multiple video tags

rundmt:
We need to support multiple video tags to make switching to the next video faster by loading up everything the next video in the background.

### Resolution
Went with an approach to swap video tag props from an active video to a non-active video. Essentially, we stop and hide the current video then play and unhide the next video. 

Originally tried to just change the source of multiple video tags, but the problem is that videos don't cache the same way so it wasn't a reliable way of preloading videos.

### Additional Considerations
I made a qa-sample to test this. When you go to the sample let it load for a bit. 

Then click on `Next Video` the next video should load without seeing a spinner. You could also turn off the internet to prove that you have a bit of the video downloaded.

### Links
ENYO-4162

Enact-DCO-1.0-Signed-off-by: Derek Tor derek.tor@lge.com


ryanjduffy:
Replaced by #1535

PR #1060, opened 2017-08-21
ENYO-4582 Allow browser to determine text directionality

webOS101:
Corrects text direction detection by not using `isTextRtl()` and instead relying on browser to
detect direction.

### Resolution
Using `window.getComputedStyle` and `dir="auto"`.

### Additional Considerations
Sample text: "English,한국어,العربية"

### Links
ENYO-4582


lishammel:

PR #1135, opened 2017-09-14
ENYO-4516: Add Popup prop onActivatorFocusFail

germboy:
Need a way to direct focus in the case where a `Popup` closes and the activating control doesn't exist anymore. Currently, we simply call `Spotlight.focus()` and allow spotlight to perform a default focus behavior. This sets focus back to the app, but sometimes the focus is somewhere unexpected or a developer wants to direct focus in a particular way.

### Resolution
I'm adding an `onActivatorFocusFail` prop to `Popup` that will execute when the popup's activating control cannot gain focus after the popup closes. This hook allows a developer to include a function that can include focus-change behavior when the popup cannot re-spot the activating control.

This option seems less intrusive and more flexible than other proposed solutions:
- add a `activator` prop to `Popup`, allowing a dev to provide a node to set focus to when the activator disappears
- query for all container id's of the activator during mount and attempt to set focus to them recursively when the activator no longer exists


lishammel:

PR #1169, opened 2017-09-25
ENYO-4501: Fix navigating via 5-way from paging controls

ryanjduffy:
* Collapse spotlight containers for Scrollable consumers
* Add `data-spotlight-overflow` support to allow spotlight to escape overflow containers via 5-way


ryanjduffy:
Closing in favor of #1553

PR #1200, opened 2017-10-03
ENYO-4772: TooltipDecorator add tooltipHidden prop

teckliew:
Tooltip is not removed when VideoPlayer control panel is hidden

### Resolution
Tooltip visibility is based on its wrapped component DOM events. In the case that wrapped component is both disabled, un-focusable, _and_ hidden by its parent, there is no way `TooltipDecorator` knows when to hide tooltip... because there is no DOM event that `TooltipDecorator` can receive. Adding the prop `tooltipHidden` allows for control over hiding tooltip. When `tooltipHidden = false`, tooltip will show and hide as normal.

### Additional Considerations
We could also guard DOM eventlisteners with `tooltipHidden`

### Links
ENYO-4772


lishammel:

PR #1223, opened 2017-10-12
Remeasure ContextualPopupDecorator when props change

webOS101:

If the `popupComponent` or `popupProps` changes while a contextual
popup is open, it will not be laid out correct as it is never remeasured.

### Resolution
Detect props changes and trigger a remeasure afterwards

### Additional Considerations
This will causes three updates to the popup in this case.  It should be rare, though.
One of the updates is calling `setState` in `componentDidUpdate`, which will
also trigger another `setState` in `componentDidUpdate`.  :feelsbadman:

### Links
ENYO-4808


webOS101:
Solved app-side

PR #1239, opened 2017-10-17
ENYO-4811: Picker Marquee Fix

teckliew:
1. Picker item doesn't marquee after hovering from picker button to picker item.
2. Picker item doesn't marque when picker button becomes `disabled`

### Resolution
There was an issue in focus and mouse handling between `MarqueeController` and `MarqueeDecorator`.
Unhovering picker button and then hovering picker item results in `MarqueeController`'s `onBlur` handler overriding the default picker item's `MarqueeDecorator` handling for mouse event. Solution is to separate focus events for 5-way and mouse events for pointer mode, so that `MarqueeController` is in control of pointer events because picker item is non spottable.

### Links
ENYO-4811
ENYO-4882


teckliew:
closed in favor of #1301

PR #1295, opened 2017-11-09
ENYO-4628: Handle page up and down keys properly with disabled items in VirtualList/VirtualListNative

ybsung:

There is a VirtualList/VirtualListNative with disabled items.

Case 1) If a current focused item was not stick to the first visible line or the last visible line of VirtualList, the VirtualList did not jump to the first or last visible line and did scroll without animation.

Case 2) A VirtualList focused the wrong item when starting scrolling without animation.

Case 3) A VirtualList did not focus an item properly via page up and down keys.

### Resolution

Case 1) Try to stick to the first or last line first, then try to scroll

Case 2) When starting scrolling, the VirtualList tried to focus the item with index. If not needed, skip focusing an item.

Case 3) Instead of finding an enabled item linearly, try to find an enabled item which is close to a current item

### Additional Considerations

### Links

ENYO-4628

### Comments

### DIFF
--- a/packages/moonstone/Scroller/Scrollable.js
+++ b/packages/moonstone/Scroller/Scrollable.js
@@ -894,7 +894,9 @@ const ScrollableHoC = hoc((config, Wrapped) => {
 			this.animator.stop();
 			this.isScrollAnimationTargetAccumulated = false;
 			this.childRef.setContainerDisabled(false);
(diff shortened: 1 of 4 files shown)


ybsung:
#1310 was merged for this issue.

PR #1303, opened 2017-11-13
ENYO-4825: Add imperative API to unmount all floating layers

ryanjduffy:
When FloatingLayerDecorator is used within a subtree of the app, the
app/component may want to unmount all floating layers it owns (e.g. if it
is hiding all layers as in moonstone/VideoPlayer).

This change adds a new imperative API available, unmountFloatingLayers, via
context to support this.


ryanjduffy:
Closing in favor of #1666

PR #1364, opened 2018-01-04
PLAT-48442: Remove <source> slot from VideoPlayer

viodragon2:
In react 16, newly introduced FiberNodes contain extra information which prevents us from a deep comparison of two nodes to check equality. `moonstone/VideoPlayer` compares `<source>` slot and it will fail to check correctly.

### Resolution
- Remove `<source>` slot and expose `src` and `type` props instead

### Comment
We could keep `<source>` slot and fix the comparison logic to compare `source.props.src`, and etc. It is subject to change depending on what we decide on how to support multiple sources.


ryanjduffy:
Replaced by #1366

PR #1380, opened 2018-01-11
Fix a bug that returning wrong time when starting DST

Seonghyup:
### Issue Resolved
 When the Timer picker is in the DST adjustment range, the date picker can't be changed to the next day from the DST date.

### Resolution
iLib RataDie was based on old javascript engine, So I fixed a bug that returning wrong time when starting DST.
- Align with the way to applying DST in javascript native engine when it is started on starting boundary time.


lishammel:
@Seonghyup  - should this PR be closed out since https://github.com/enactjs/enact/pull/1388 was used instead? Thanks.

ryanjduffy:
Closing in favor of #1388

PR #1395, opened 2018-01-20
ENYO-4883: Hide an opposite scrollbar's thumb if the thumb of a current scrollbar is located at the first or at the end

ybsung:

Even though the thumb of a scroll bar is located at the end or at the first, the thumb is not hidden.

### Resolution

Hide an opposite scrollbar's thumb if the thumb of a current scrollbar is located at the first or at the end

### Additional Considerations

### Links

### Comments


ybsung:
We decided not to fix this issue.

PR #1439, opened 2018-02-13
PLAT-50258: Update scroll button disable/enable status and scroll thumb size when the data size of VirtualList is changed

ybsung:

Due to the edge case in ENYO-4603 and ENYO-4620, we prevented updating scrollbar visibility, scroll button disable/enable status, and scroll thumb size when the data size of `VirtualList` is changed. So even though its data size is very small, the `ScrollButton` in it is enabled.

### Resolution

When changing the data size of `VirtualList`, I updated not scrollbar visibility but scrollbar visibility, scroll button disable/enable status.

### Additional Considerations

### Links

PLAT-50258

### Comments

### DIFF
--- a/packages/moonstone/Scroller/Scrollable.js
+++ b/packages/moonstone/Scroller/Scrollable.js
@@ -1104,7 +1104,7 @@ const ScrollableHoC = hoc((config, Wrapped) => {
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


ybsung:
This PR was replaced with #1454.

PR #1444, opened 2018-02-15
ENYO-5018: Add moonstone/MediaOverlay component

xbinary:
Implement PIG(Picture in graphic) component for webOS TV only

### Resolution
Some App support PIG function. Each App implement this.
Therefore, it is supported as component

### Additional Considerations

### Links
ENYO-5018


packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,174 @@
...
+			...rest} = this.props;
+
+		delete rest.onLoadedMetadata;
+		delete rest.onVideoClick;
+
+		const containerClassName = classNames(className, css.pictureInGraphics);
+		const textOverlayStyle = {color: textOverlayColor};
+
+		return (
+			<Container {...rest} className={containerClassName} spotlightDisabled={spotlightDisabled}>
```
viodragon2: It seems like the focus border only extends to the video and not to the caption. But by wrapping `Spottable` to the outer most div, it will get focused when hovered over the caption. This seems awkward.

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,174 @@
...
+import classNames from 'classnames';
+import equals from 'ramda/src/equals';
+
+import Slottable from '@enact/ui/Slottable';
+import Spottable from '@enact/spotlight/Spottable';
+import {forward} from '@enact/core/handle';
+
+import Image from '../Image';
+import Skinnable from '../Skinnable';
+import {MarqueeController, MarqueeText} from '../Marquee';
```
viodragon2: Note the path has changed in 2.0

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,174 @@
...
+	static propTypes = {
+		/**
+		 * Any children `<source>` tag elements will
+		 * be sent directly to the `<video>` element as video sources.
+		 *
+		 * @type {Node}
+		 * @public
+		 */
+		source: PropTypes.node.isRequired,
+		/**
```
viodragon2: nitpick: spaces between doc and next prop

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,174 @@
...
+						className={css.video}
+						autoPlay
+						controls={false}
+						ref={this.setVideoRef}
+						onLoadedMetadata={this.handleLoadedMetadata}
+					>
+						{source}
+					</video>
+					{imageOverlaySrc ? <Image placeholder={placeholder} className={css.image} src={imageOverlaySrc} /> : null}
+					{textOverlayContent ? (<MarqueeText alignment="center" className={css.textOverlay} style={textOverlayStyle}>{textOverlayContent}</MarqueeText>) : null}
```
viodragon2: nitpick: extra `(`

packages/moonstone/PictureInGraphics/PictureInGraphics.js
viodragon2: If it is indeed needed to wrap everything, instead of composing `MarqueeController`, and `Spottable` separately you can compose them to decorate the base component.
e.g.
```
const PictureInGraphics = MarqueeController({marqueeOnFocus: true},
	Spottable(
		Slottable(
			{slots: ['captionComponent', 'source']},
			Skinnable(
				PictureInGraphicsBase
			)
		)
	)
);
```

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,178 @@
...
+		videoComponent: 'video'
+	}
+
+	constructor (props) {
+		super(props);
+		this.video = null;
+	}
+
+	componentDidMount () {
+		this.video.load();
```
viodragon2: we don't need to `load` video on component mount

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,178 @@
...
+		 * @public
+		 */
+		textOverlayContent: PropTypes.string,
+
+		/**
+		 * Video component to use. The default (`'video'`) renders an `HTMLVideoElement`. Custom
+		 * video components must have a similar API structure, exposing the following APIs:
+		 *
+		 * Properties:
+		 * * `currentTime` {Number} - Playback index of the media in seconds
```
viodragon2: we only require `load()` method for it to work in PIG.

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,178 @@
...
+		imageOverlaySrc: PropTypes.string,
+
+		/**
+		 * By default, the video will start playing immediately after it's loaded, unless this is set.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		noAutoPlay: PropTypes.bool,
```
viodragon2: Is there a use case where PIG wants to not autoplay upon its load? We do not offer to control the video state in any case, so I'm not sure how this can be ever useful when it is set to `true`.

packages/moonstone/PictureInGraphics/PictureInGraphics.less
```diff
@@ -0,0 +1,56 @@
...
+		display: block;
+		width: 100%;
+		height: 100%;
+		margin: 0;
+	}
+
+	.image {
+		box-sizing: border-box;
+		display: block;
+		position: absolute;
```
viodragon2: Whenever image placeholder is declared, the video won't be visible. Is this intentional? My guess is for locked/restricted channels may need that, but wanted to confirm. If so let's document it.

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,163 @@
...
+					{source}
+				</Media>
+				{imageOverlaySrc ? <Image className={css.image} placeholder={placeholder} sizing="fit" src={imageOverlaySrc} /> : null}
+				{textOverlayContent ? (<Marquee alignment="center" className={css.textOverlay} marqueeOn="render">{textOverlayContent}</Marquee>) : null}
+			</div>
+		);
+	}
+}
+
+const PictureInGraphics = Pure(
```
viodragon2: Let's use `ramda/src/compose` and call this `PictureInGraphicsDecorator` to be consistant with other components.
You should assign `const PictureInGraphics = PictureInGraphicsDecorator(PictureInGraphicsBase)` and export it as default.

packages/moonstone/PictureInGraphics/PictureInGraphics.js
```diff
@@ -0,0 +1,163 @@
...
+		}
+	}
+
+	setVideoRef = (video) => {
+		this.video = video;
+	}
+
+	render () {
+		let p = Object.assign({}, this.props);
+		if (this.renderStyles) p = this.renderStyles(p);
```
viodragon2: Same here. Do we need to check for `this.renderStyles`?

packages/moonstone/PictureInGraphics/PictureInGraphics.less
```diff
@@ -0,0 +1,72 @@
...
+		.textOverlay,
+		.image {
+			right: 0;
+			left: auto;
+		}
+	}
+}
+
+:not(:global(.spottable))&.pictureInGraphics {
+	padding: 0;
```
viodragon2: Since we allow `.pictureInGraphics`, do you think this still should remain as component's requirement?

ryanjduffy:
:tada: Great collaboration on this ticket. Thanks to @xbinary @viodragon2 and @Djspaceg! :tada:

PR #1459, opened 2018-02-23
ENYO-4998: Add support for custom media components in VideoPlayer

viodragon2:
App wants to use custom video component for a non-standard video that's not supported via HTML video element.

### Resolution
- Add `component` prop to `VideoPlayer` to support custom components
- Add `ui/Media` component to represent a HTMLMediaElement which forwards standard HTMLMediaEvents and custom media events
- Removed forwarding media states in standard media events (currentTime, duration, etc.)


viodragon2:
Updated per review and some more.
The documentation added to the `videoComponent` seem a bit messy. I'd like your feedback on this on what you think.

In addition to things you've pointed out, I've found that VideoPlayer had some unused/unnecessary internal states that we could get rid of. (i.e. `buffered`, `readyState`, `mute`, `volume`)

I didn't address the `customMediaEventsMap` issue you've pointed because I think I found a way to get rid of `getNode()` from `Media`. I was never a big fan of this at first, and I think it is better if we can remove it. 

Let me know what you think

ryanjduffy:
I think this is a good start. Not sure what is a better way to handle the docs for `videoComponent` right now. I'm okay with what you have for the time being.

packages/ui/Media/Media.js
```diff
@@ -0,0 +1,238 @@
+/**
+ * Provides a represetation of HTMLMediaElement component with an ability to attach a custom event
+ * handler.
+ *
+ * @module ui/Media
+ * @exports handledMediaEventsMap
+ * @exports readyState
+ * @exports Media
```
webOS101: either `Media` should be `@public` or it should not appear in `@exports`.

packages/ui/Media/Media.js
```diff
@@ -0,0 +1,238 @@
+/**
+ * Provides a represetation of HTMLMediaElement component with an ability to attach a custom event
+ * handler.
+ *
+ * @module ui/Media
+ * @exports handledMediaEventsMap
+ * @exports readyState
```
webOS101: Where is `readyState` exported?

PR #1467, opened 2018-02-27
ENYO-4978: Add Tooltip to ProgressBar

teckliew:
Add `tooltip` to `ProgressBar`.

`tooltip` follows the progress bar and shows the percentage of the progress bar.

`forceTooltipSide` is also added to support for cases where Tooltip needs to be switched to the other side when `vertical` progress bar is at the right edge of the container.

### Additional Considerations
`SliderTooltip` is pulled out to be in `Tooltip` as `ProgressBarTooltip`

### Links
ENYO-4978

### Comments


viodragon2:
Everything looks good. Can you resolve :mergeconflict:?

ryanjduffy:
@teckliew Can you elaborate on why you placed `ProgressBarTooltip` into `Tooltip`?

ryanjduffy:
@teckliew + @viodragon2 - WDYT of moving `ProgressBarTooltip` "back" to `moonstone/ProgressBar` and also exporting it as `SliderTooltip` from `moonstoneSlider`? I think that would make the API more predictable (both behave the same) and easier to use (only have to import the module for the component you're using).

ryanjduffy:
Here's a branch with the change for consideration: https://github.com/enactjs/enact/tree/feature/ENYO-4978-ryan

viodragon2:
In general, I agree with your approach. The only downside I see with that approach is that `moonstone/Slider` is still importing a module that doesn't belong to the component. It would make much more sense if slider utilizes the progress bar as a core component, which I believe it's postponed for now.

ryanjduffy:
I agree. The advantage is it avoids the deprecation and likely positions us for the future in which we do reuse ProgressBar within Slider.

teckliew:
I am still unsure if having `ProgressBarTooltip` in `Tooltip` is the best idea, but because this is a component that is based on `Tooltip` that is used by both `Slider` and `ProgressBar`, I felt this makes sense. Like what @viodragon2 said, it also makes sense set up `ProgressBar` as the core component of `Slider`. I am indifferent so let me know what you guys decided on.

Djspaceg:
Can we also merge Ryan's changes if they're satisfactory?

PR #1469, opened 2018-02-28
PLAT-47085: Incorporated a mini-scrim into Input

Djspaceg:
`Input` fields need to be the only thing clickable when one of them is focused.

### Resolution
Add a scrim, local to `Input`, which when present, appears above all other elements but behind input fields.

Due to a browser implementation detail of GPU accelerated layers, it was not possible to use our existing `FloatingLayer` component. GPU rendered floating layers cannot be drawn between standard composited layers. Inputs being inside of a GPU accelerated Panel could not be pulled apart from that rendering stack to be interleaved with the FloatingLayer's `z-index`.

### DIFF
--- a/packages/moonstone/Input/Input.js
+++ b/packages/moonstone/Input/Input.js
@@ -235,29 +235,31 @@ const InputBase = kind({
 		value: ({value}) => typeof value === 'number' ? value : (value || '')
 	},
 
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Planning to pursue a different approach but keeping the branch around for now.

PR #1473, opened 2018-03-06
PLAT-49792: Add spotlight/Pause to manage pause state

ryanjduffy:
The framework has conflicts within itself and with apps when pausing and resuming spotlight. We've build pause/resume pairs into the framework in several places to avoid user action during transitions (Panels, Expandable) or to "lock" the user to a component (Input, EditableIntegerPicker). However, sometimes these can interfere with each other or with apps that wish to prevent auto focus behaviors by pausing spotlight.

In one such case (ENYO-4982), the app wants to defer focusing something until a remote service call completes to populate the UI. The app set `autoFocus="none"` on the Panel but Spotlight also attempts to re-establish focus when the Scroller's paging controls become active when nothing is focused (#1023) subverting the apps pause which was "cancelled" by `moonstone/Panels` when the panel finished transition.

### Resolution
* Add `spotlight/Pause` module to acts as a binary semaphore to manage the pause state such that only the last "pauser" (or `Spotlight.resume()`) can resume.
* Adopt `spotlight/Pause` within the framework to ensure that only the initiator of a pause will resume it. This does not prevent the application from pausing or resuming and, in fact, ensures that the framework does not resume an app-initiated pause.


packages/spotlight/Pause/Pause.js
```diff
@@ -53,6 +53,26 @@ function isPaused () {
  * @public
  */
 class Pause {
+
+	constructor (name) {
```
webOS101: We should document the constructor, particularly since it takes an argument.

PR #1474, opened 2018-03-07
PLAT-51917: TooltipDecorator: Want to investigate if there are ways we can reduce how much work is done per instance

rundmt:
`TooltipDecorator` is spending too much time mounting. When we don't have a tooltip.

### Resolution
Removed forward and handle from TooltopDecorator. Those functions slowing down the mounting time because they use ramda's `curry`.

### Additional Consideration
I did not remove `forProp`. It seems like it wouldn't run unless we actually had a `tooltipText` so I figured it would skip in the case we want to speed up.

### Links
PLAT-51917


lishammel:

PR #1483, opened 2018-03-09
ENYO-4988: VideoPlayer update

viodragon2:
Updated `VideoPlayer` various per various requests and bug reports

### Resolution
New props
- `noSpinner`: no spinner is displayed during demos
- ~`mediaControlsProps`: allows to pass down props to media controls. This is required for passing down `data-component-id` in P&V~ Removed in favor of `data-media-controls`

Bug fixes
- correctly adjust spaces when the number of components changes in `leftComponents` and `rightComponents`: use `countReactChildren` instead of `React.Children.count()`
- read out audio guidance every time `source` changes: reset flag on prop change
- display custom thumbnail node: Correctly pass node
- hide more icon when right components are removed: use `countReactChildren` instead of `React.Children.count()`


packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -2078,7 +2098,6 @@ const VideoPlayerBase = class extends React.Component {
 					// This captures spotlight focus for use with 5-way.
 					// It's non-visible but lives at the top of the VideoPlayer.
 					className={css.controlsHandleAbove}
-					onClick={this.showControls}
```
ryanjduffy: Why was this removed? I found the offending commit (489bff1b978332c738d81c755ad5b160c608cef1) but not sure why it was included.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -286,6 +283,14 @@ const VideoPlayerBase = class extends React.Component {
 		 */
 		loading: PropTypes.bool,
 
+		/**
+		 * An object containing properties to be passed to media controls component.
+		 *
+		 * @type {Object}
+		 * @public
+		 */
+		mediaControlsProps: PropTypes.Object,
```
ryanjduffy: I'd rather not expose this. It doesn't make a lot of sense as a public API when `MediaControls` itself isn't public so there's no way to know what to pass without diving into the private source. If we feel it's truly necessary to support testware deep within VideoPlayer as the ticket requests, I'd rather define a static attribute (e.g. `data-media-controls`) on the node they could use to find it.

packages/moonstone/VideoPlayer/VideoPlayer.js
viodragon2: It seems unnecessary to me. `SpottableDiv` is non-clickable as it's height is 0. `Overlay` handles `onVideoClick` to show controls.

PR #1484, opened 2018-03-09
ENYO-4972: Scroller/IconButton spacing fix

teckliew:
Scroller doesn't scroll all the way down when `IconButton` is at the bottom.

### Resolution
`IconButton` uses pseudo element to increase the tap area of the button. `Scrollable` doesn't include the dimensions of the pseudo element when calculating scroll on focus and there is no performant way to have `Scrollable` include the calculations of children of focused items that are bigger than their parent. Easy fix here is to aim for the last-child of `Scrollable` and make sure they have `overflow: hidden`.
Suppressing the invisible tap area for `IconButton` and hiding children that are bigger than the focused item will be the side-effects of this fix.

### Links
ENYO-4972

### Comments

### DIFF
--- a/packages/moonstone/Scroller/Scrollable.less
+++ b/packages/moonstone/Scroller/Scrollable.less
@@ -32,5 +32,9 @@
 		position: relative;
 		overflow: hidden;
 		flex: 1 1 100%;
+
+		> :last-child {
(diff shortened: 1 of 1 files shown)


lishammel:

PR #1489, opened 2018-03-12
PLAT-49383: Add simplified internal FP library

ryanjduffy:
Ramda is more full-featured than we generally need in much of the framework internals. We've also over-used it in places leading to a compounding performance impact in hot paths.

### Resolution
Add a simplified internal FP library to core and remove unnecessary ramda usage

### Additional Considerations
There are a few places remaining we could alter to eliminate ramda altogether but the primary impact has been mitigated in this PR.
I've also omitted a changelog for this since there is no "bug" fixed nor feature added and all changes are internal. It falls into the general "performance improvements" of any release.


ryanjduffy:
Closing for now as the perf benefits don't appear to outweigh the cost of maintaining our own library for this

PR #1495, opened 2018-03-13
ENYO-4559: Adjust spotlight partitions and priorities

germboy:
At times, spotlight fails to select the expected "most natural" spottable element when using 5-way. This is due to how spotlight partitions and prioritizes its "next" spottable elements when 5-way is used.

### Resolution
Currently, spotlight simply selects the first eligible spottable element based on group (placement in the control grid/matrix) and distance. The problem is that "distance" is more or less "alignment" rather than distance - meaning that eligible controls are selected based on how they align with the currently-spotted control. In certain cases, a "next" control can fit within the alignment requirement priority and be returned as "next" when it is actually farther away from the currently spotted control than another eligible control.

I've introduced a new behavior that calculates actual distance from the currently-spotted control - to better match with enyo spotlight. This appears to give a more natural 5way behavior and fixes the scenarios where 5-way "jumps over" more expected 5-way targets.

### Additional Considerations
While this more closely matches enyo spotlight behavior, enyo spotlight had other nuances and behaviors (intended or not), where there are certain pixel tolerance amounts that can allow eligibility and prioritization. I've added info in the JIRA on this, but have not added this behavior to this fix - although I have a branch that does include some of that behavior, for consideration if needed.


germboy:
I've added an algorithm to more intelligently directionally-straight elements vs oblique ones that we could probably test/play with to get a feel for how we should weight the valid "next" elements. Currently, I'm using a 1/5 ratio which - in specific "nextdoor/touching" scenarios - will ultimately match enyo's 5px "tolerance". That number scales appropriately as the distance between elements grows.

PR #1508, opened 2018-03-17
ENYO-5056: InputHeader support

teckliew:
`inputHeader` is supported to replace h1 title with h1 styled `Input` component provided by the consumer, allowing the title in `Header` to be editable.

### Additional Considerations
I don't recommend that we support `inputHeader` for `compact` `Header` if there are no requirements for it.
The width or the size of `input` needs to be explicitly specified and it doesn't adjust its width to the contents of `input`. Because `input` is inlined with `titleBelow`, adding `input` support won't be consistent with the normal `compact` `Header`.

### Links
ENYO-5056

### Comments


teckliew:
I don't recommend supporting `inputMode` for `compact` Header at all.
Based on the requirement, @Djspaceg and I agreed that there is no Spotlight styling, the blinking cursor is enough, but I can see why that might not be enough. In the case that we need Spotlight styling, I think we need to ask for a more concrete UX design.

packages/moonstone/Panels/Header.less
```diff
@@ -92,6 +112,24 @@
 		border-top-color: @moon-header-border-color;
 		border-bottom-color: @moon-header-bottom-border-color;
 
+		.inputTitle {
+
+			&:focus {
+				border-color: transparent;
```
viodragon2: Do we need this?

teckliew:
The `large-text` height bug is because the the `SwitchItem` below `Header` somehow triggered an overlap with `Layout`.

teckliew:
Header's `title` are both plugged into input's `placeholder`. Should I have it in `value` instead?
I am assuming that if we open up `value` or `defaultValue`, then `value` will have precedence over Header's `title`?

webOS101:
There appeared to be some discussion about not supporting input mode on small headers.  Should we throw an error if this combination is specified?

ryanjduffy:
Little concerned about the prop expansion required for this feature along with the conflicts @webOS101 mentioned. Would it be a better DX to export specialized versions from `moonstone/Header` rather than (only) a generalized version with all the props?

Djspaceg:
Placeholder is different from value. A value is submittable text in the input, while a placeholder is just a hint to the user as to what the field is for or an example input. An input must support a pre-set value. An example of why this would be used is someone typing into a field, navigating away from the field, then navigating back and seeing their text already in the field. The app would be pre-populating their value into the field on return.

Djspaceg:
The qa-stories for Header were lacking many states that we should be checking for, so I added them, along with documentation in the recent commits.

PR #1509, opened 2018-03-20
ENYO-5074: Replace `vertical` with `orientation`

Djspaceg:
`vertical` is a 1-dimensional property without room for extension over time.

### Resolution
`vertical` has been replaced with `orientation` in all Slider and ProgressBar iterations. `orientation` accepts a customizable string, which will allow for easy future development.

### DIFF
--- a/packages/moonstone/IncrementSlider/IncrementSlider.js
+++ b/packages/moonstone/IncrementSlider/IncrementSlider.js
@@ -324,6 +324,7 @@ const IncrementSliderBase = kind({
 		 * down. Must be either `'horizontal'` or `'vertical'`.
 		 *
 		 * @type {String}
+		 * @default 'horizontal'
 		 * @public
 		 */
 		orientation: PropTypes.oneOf(['horizontal', 'vertical']),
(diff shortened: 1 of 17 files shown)


packages/moonstone/IncrementSlider/IncrementSlider.js
```diff
@@ -319,6 +319,15 @@ const IncrementSliderBase = kind({
 		 */
 		onSpotlightUp: PropTypes.func,
 
+		/**
+		 * Sets the orientation of the slider, whether the slider moves left and right or up and
+		 * down. Must be either `'horizontal'` or `'vertical'`.
+		 *
+		 * @type {String}
+		 * @public
```
webOS101: Need `@default` here.  I can fix it.

packages/ui/ProgressBar/ProgressBar.js
```diff
@@ -57,29 +57,29 @@ const ProgressBar = kind({
...
 		 */
-		vertical: PropTypes.bool
+		progress: PropTypes.number
 	},
 
 	defaultProps: {
 		backgroundProgress: 0,
-		progress: 0,
-		vertical: false
+		progress: 0
```
webOS101: Should we have a default here, as we do in Moonstone?

Djspaceg: Yes, probably. At least until we open up `orientation` in UI.

PR #1514, opened 2018-03-21
ENYO-5079: Fix spotlight pause-resume behavior for animated Popups

germboy:
When using an animated `Popup`, spotlight could become unresponsive after opening or closing the `Popup`.

### Resolution
`Popup` uses `ui/Transition` internally as its dom wrapper, which it passes `onShow` and `onHide` props. These props ultimately get called as part of `ui/Transition`'s `onTransitionEnd` method. There was a [recent change](#1481 ) to `ui/Transition` that altered the dom of the component. Within `Popup`'s `onShow` and `onHide` methods (which return the event of `onTransitionEnd`), we were [incorrectly] validating against `ev.target` instead of `ev.currentTarget`. Prior to #1481, both prop values were the same so it wasn't a problem until now.

Enact-DCO-1.1-Signed-off-by: Jeremy T...

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -26,9 +26,6 @@ The following is a curated list of changes in the Enact moonstone module, newest
 - `moonstone/ContextualPopup` prop `popupContainerId` to `popupSpotlightId`
 - `moonstone/Popup` prop `containerId` to `spotlightId`
 - `moonstone/VideoPlayer` prop `containerId` to `spotlightId`
-
-### Changed
-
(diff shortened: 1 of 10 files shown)


ryanjduffy:
I think we should pull this into alpha.6 so i'm going to retarget to the release branch 🤞

ryanjduffy:
Looks like it pulled in a couple extra commits but they're no-ops so no worries.

ryanjduffy:
@germboy Excellent analysis on this, by the way. It's great to call out the both the cause of and the source PR for the regression and really helps the review process.

PR #1521, opened 2018-03-22
PLAT-52473: Add webos/speech with voice control support

changgilee:
VoiceControl component create div and declare attribute, handler to dom node.

Web Engine will be refer this node and perform voice control behavior.

### Resolution

### Additional Considerations

### Links


packages/webos/VoiceControl/VoiceControl.js
```diff
@@ -0,0 +1,53 @@
...
+				data-voice-disabled={voiceDisabled}
+				data-voice-intent={voiceIntent}
+				data-voice-label={voiceLabel}
+				data-voice-title={voiceTitle}
+				{...props}
+			/>
+		);
+	}
+}
+
```
seunghoh: Removed all unnecessary props. What about this? and I am not sure this is really needed or not. Too simple...
```
class VoiceControl extends React.Component {
           static propTypes = {
                     onVoice: PropTypes.func.isRequired
           }

           componentDidMount () {
                     findDOMNode(this).addEventListener('onvoice', this.onVoice);
           }

           componentWillUnmount () {
                     findDOMNode(this).removeEventListener('onvoice', this.props.onVoice);
           }

           render = () => (this.props.children)
}
```

packages/moonstone/Button/Button.js
```diff
@@ -103,6 +103,7 @@ const ButtonBase = kind({
 
 		return (
 			<UiButtonBase
+				webos-voice-intent="Select"
```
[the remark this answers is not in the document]
webOS101: No, I think it's fine to use `webos-`.

packages/webos/VoiceControl/VoiceControl.js
```diff
@@ -0,0 +1,25 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import {findDOMNode} from 'react-dom';
+
+class VoiceControl extends React.Component {
```
ryanjduffy: Architecturally, I have some concerns about this approach. It's HOC-like but not a HOC which makes it a bit out of place. I think I'd rather see it use `core/hoc`. Do we also need a solution for app-level event handling by listening to events on `document` rather than the node?

changgilee: I did change to HOC and it seems more appropriate structure. Thanks~
Additionally, in current status, web engine does not support event handling base on document object.(Because they are searching for information base on dom node) But I think app scenario can be covered using by node because app developer can create node to top level of application and it can be handle voice event globally.

packages/webos/VoiceControlDecorator/VoiceControlDecorator.js
```diff
@@ -0,0 +1,35 @@
...
+	return class extends React.Component {
+		static displayName = 'VoiceControlDecorator'
+
+		static propTypes = {
+			onVoice: PropTypes.func.isRequired
+		}
+
+		componentDidMount () {
+			this.node = findDOMNode(this);
+			if (!(this.node.hasAttribute('data-voice-event-target') || this.node.hasAttribute('data-webos-voice-intent'))) {
```
webOS101: Should this be `data-webos-voice-event-target`?

changgilee: I will change it to data-webos-voice-event-target. (This attribute is not defined by web engine, but it is deeply relate to webOSVoice event. So, it looks better in terms of consistency as you mentioned.) Thanks~

webOS101:
I created a PR (#1582 ) that consolidates the speech APIs.  Let me know how you like this and merge it into this branch if you approve.  You can expand upon the documentation.

packages/moonstone/LabeledItem/LabeledItem.js
```diff
@@ -96,7 +96,7 @@ const LabeledItemBase = kind({
 	},
 
 	render: ({children, css, disabled, label, titleIcon, ...rest}) => (
-		<Controller disabled={disabled} {...rest} css={css}>
+		<Controller disabled={disabled} data-webos-voice-intent="Select" {...rest} css={css}>
```
ryanjduffy: This isn't needed. `Item` already has the prop set.

packages/moonstone/SlotItem/SlotItem.js
```diff
@@ -63,6 +63,7 @@ const SlotItemBase = kind({
 	render: (props) => {
 		return (
 			<UiSlotItemBase
+				data-webos-voice-intent="Select"
```
ryanjduffy: This isn't needed. `SlotItem` uses `Item` which already has this prop applied.

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -216,7 +216,15 @@ const ExpandableItemBase = kind({
...
-		spotlightDisabled: PropTypes.bool
+		spotlightDisabled: PropTypes.bool,
+
+		/**
+		 * Set voice control property to LabeledItem
+		 *
+		 * @type {Object}
+		 * @public
+		 */
+		voiceProps: PropTypes.object
```
ryanjduffy: What are `voiceProps`? Are there a defined set which we should declare individually here?

packages/moonstone/Input/Input.js
```diff
@@ -210,7 +210,15 @@ const InputBase = kind({
...
-		value: PropTypes.oneOfType([PropTypes.string, PropTypes.number])
+		value: PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+
+		/**
+		 * Add voice control property to input tag
+		 *
+		 * @type {Object}
+		 * @public
+		 */
+		voiceProps:  PropTypes.object
```
ryanjduffy: Similar question to `ExpandableItem` on the contents of `voiceProps`.

packages/webos/speech/VoiceControlDecorator.js
```diff
@@ -0,0 +1,49 @@
...
+import hoc from '@enact/core/hoc';
+
+/**
+ * Adds a callback for voice events
+ *
+ * @class VoiceControlDecorator
+ * @memberof webos/speech
+ * @hoc
+ */
+const VoiceControlDecorator = hoc({}, (config, Wrapped) => {
```
ryanjduffy: No need to include an empty object for the default config. It can be omitted and just pass the callback as the first argument.

packages/webos/speech/VoiceControlDecorator.js
```diff
@@ -0,0 +1,49 @@
...
+			 * @type {Function}
+			 * @required
+			 * @public
+			 */
+			onVoice: PropTypes.func.isRequired
+		}
+
+		componentDidMount () {
+			this.node = findDOMNode(this);	// eslint-disable-line 
+			if (!(this.node.hasAttribute('data-webos-voice-event-target') || this.node.hasAttribute('data-webos-voice-intent'))) {
```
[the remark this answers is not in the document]
changgilee: I will add guard code about this. Thanks~

packages/moonstone/IncrementSlider/IncrementSlider.js
```diff
@@ -179,6 +188,15 @@ const IncrementSliderBase = kind({
...
 		incrementIcon: PropTypes.string,
 
+		/**
+		* Sets the voice control string to increment button.
+		*
+		* @default 'increment'
+		* @type {String}
+		* @public
+		*/
+		incrementVoiceLabel: PropTypes.string,
```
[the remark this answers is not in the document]
changgilee: For example, Zoom Slider of PhotoVideo App, IconButton's aria-label is usded like this("130% press ok button to increase the value") base on audio guidance scenario. But when user try to voice control, they will just said "zoom in". In this case, it seems that additional attribute(data-webos-voice-label) has to declared to IconButton of IncrementSlider.

packages/moonstone/LabeledItem/LabeledItem.js
changgilee: I will remove it.

packages/moonstone/SlotItem/SlotItem.js
changgilee: I will remove it.

packages/moonstone/IncrementSlider/IncrementSlider.js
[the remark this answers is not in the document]
ryanjduffy: I should have been more clear. I understand why they're needed but I don't think they should be. Your scenario is exactly the incongruity to which I was referring. The a11y text and voice control text do not line up as I think they should.

packages/webos/speech/VoiceControlDecorator.js
changgilee: I will fix it~

packages/moonstone/ExpandableItem/ExpandableItem.js
changgilee: voiceProps is used to customize voice control properties of the ExpandableItem.
For example, if app developer want to set custom voice control label, they will declare it to ExpandableItem Component.
In current status, the props will be delegate to ContainerDiv(not LabeledItem)
About this modification, I guess we need more discussion, So, I will check again and make seperate ticket.(In this PR, I will remove voiceProps related code)

packages/moonstone/Input/Input.js
changgilee: Same to ExpandableItem. I will make seperate ticket.

packages/moonstone/IncrementSlider/IncrementSlider.js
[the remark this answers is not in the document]
changgilee: I also agree your concern and I will check again with UX (In this PR, I will remove code and make seperate ticket if need)

PR #1524, opened 2018-03-23
PLAT-52443: Deprecate `data` prop and add `isDisabledItem` in VirtualList

ybsung:

- We will deprecate the `data` prop in `VirutalList` to eliminate the misunderstanding caused by the ambiguity of `data`
- We will add the `isDisabledItem` prop in `VirutalList` because `data` prop is used only for handling disabled item inside of the VirtualList, so we'd like to provide the way to do it after removing `data` prop

### Resolution

- Deprecate `data` prop and add `isDisabledItem` in VirtualList

### Additional Considerations

### Links

PLAT-52443

### Comments


packages/moonstone/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
+- `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` prop `data` to be removed
```
MikyungKim: `data`,?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -119,6 +119,14 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			initUiChildRef: PropTypes.func,
 
+			/**
+			 * Check if the item with index is disabled
+			 *
+			 * @param {Number} index
```
MikyungKim: @type {Function}?
@private seems to be @public

packages/moonstone/CHANGELOG.md
MikyungKim: Should add log for adding `isDisabledItem`

packages/moonstone/docs/virtual-list-scroller.md
```diff
@@ -58,19 +53,20 @@ This document describes VirtualList, VirtualGridList, and Scroller.
 
     ```
     //.js
-    renderItem = ({data, index, ...rest}) => {
+    items = []
```
MikyungKim: ; ?

packages/sampler/stories/moonstone-stories/VirtualList.js
```diff
@@ -15,8 +15,9 @@ const Config = mergeComponentMetadata('VirtualList', VirtualList, VirtualListBas
 
 const
 	items = [],
+	isDisabledItem = (index) => (items[index].disabled),
```
MikyungKim: I guess sampler doesn't have a disabled example.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -733,7 +713,6 @@ const VirtualListBaseFactory = (type) => {
 			if (type === Native) {
 				containerClass = (this.isPrimaryDirectionVertical) ? css.vertical : css.horizontal;
 			}
-
```
MikyungKim: ?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -119,6 +119,14 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			initUiChildRef: PropTypes.func,
 
+			/**
+			 * Check if the item with index is disabled
```
MikyungKim: An example would be helpful.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -119,6 +114,30 @@ const VirtualListBaseFactory = (type) => {
...
+			 * 		/>
+			 * 	);
+			 * }
+			 * ```
+			 *
+			 * @type {Function}
+			 * @param {Number} index
+			 * @public
+			 */
+			isDisabledItem: PropTypes.func,
```
webOS101: This property is not required yet there is no default or fallback specified.  Would it make sense to assume no items are disabled if this is not supplied and document it that way?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -375,13 +396,14 @@ const VirtualListBaseFactory = (type) => {
...
-				{cbScrollTo, data, dataSize} = this.uiRef.props,
+				{isDisabledItem} = this.props,
+				{cbScrollTo, dataSize} = this.uiRef.props,
 				{firstIndex, numOfItems} = this.uiRef.state,
 				{isPrimaryDirectionVertical} = this.uiRef,
 				rtl = this.props.rtl,
 				currentIndex = Number.parseInt(target.getAttribute(dataIndexAttribute));
 
-			if (!data || !Array.isArray(data) || !data[currentIndex] || data[currentIndex].disabled) {
+			if (!isDisabledItem || isDisabledItem(currentIndex)) {
```
webOS101: Would it be better to assume that it's not disabled if the method is not supplied here?   `isDisabledItem && ...`

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -401,12 +423,12 @@ const VirtualListBaseFactory = (type) => {
 
 			if (isForward) {
 				// See if the next item is spottable then delegate scroll to onFocus handler
-				if (currentIndex < dataSize - 1 && !data[currentIndex + 1].disabled) {
+				if (currentIndex < dataSize - 1 && !isDisabledItem(currentIndex + 1)) {
```
webOS101: Is there a guard against `isDisabledItem` being undefined here and the cases below?

packages/moonstone/docs/virtual-list-scroller.md
```diff
@@ -8,26 +8,22 @@ This document describes VirtualList, VirtualGridList, and Scroller.
 
 ### Basic usage of VirtualList and VirtualGridList
 
-*   At least four props below are required to show a list properly.
+*   At least three props below are required to show a list properly.
```
webOS101: Should we have a discussion of `isDisabledItem`?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -119,6 +114,30 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			initUiChildRef: PropTypes.func,
 
+			/**
+			 * The Function that returns `true` if the item at the index is disabled
+			 *
+			 * Usage:
+			 * ```
+			 * isDisabledItem = (index) => (this.items[index].disabled)
```
webOS101: Would `isItemDisabled` make more sense?  It reads more as a sentence than the other name.

packages/moonstone/VirtualList/VirtualListBase.js
ybsung: The guard to check if `isDisabledItem` is undefined is the following line that you commented above. So `isDisabledItem` is always not undefined.
https://github.com/enactjs/enact/pull/1524/files/6da91f4e11b5d41c934a6696d5ee2d8dfd26d83d#diff-b6579b4fbd9d50a0d236c66680c0570dR406

webOS101: I see now.  Might revisit this in relation to the question at the comment above.

packages/moonstone/VirtualList/VirtualListBase.js
ybsung: @webOS101 
Do you mean that it is better to wrap the code in `jumpToSpottableItem()` with the following condition?

```
if (
	isItemDisabled &&
	!isItemDisabled(currentIndex)
) {
	// Moved the code here
}
```

packages/moonstone/docs/virtual-list-scroller.md
ybsung: Yes. I added the document related with `isItemDisabled`.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -375,13 +398,14 @@ const VirtualListBaseFactory = (type) => {
...
-				{cbScrollTo, data, dataSize} = this.uiRef.props,
+				{isItemDisabled} = this.props,
+				{cbScrollTo, dataSize} = this.uiRef.props,
 				{firstIndex, numOfItems} = this.uiRef.state,
 				{isPrimaryDirectionVertical} = this.uiRef,
 				rtl = this.props.rtl,
 				currentIndex = Number.parseInt(target.getAttribute(dataIndexAttribute));
 
-			if (!data || !Array.isArray(data) || !data[currentIndex] || data[currentIndex].disabled) {
+			if (!isItemDisabled || isItemDisabled(currentIndex)) {
```
webOS101: I think I had misunderstood this code.  Some in-line comments might help clarify the intent of this code.  If I understand it correctly now, if there is no function for checking disabled or the current item is disabled, then allow Spotlight to handle the spotting directly.  Otherwise, we need to check to see if there are disabled items above or below (depending on direction) to figure out the next spottable item.  If this is the case, then I think this code is fine.

ybsung: You understand it correctly. That's the way how the `Virtuallist` manages spotlight with disabled items. I added the in-line comment to describe what is the mean of the condition

```
if (
	!isItemDisabled || // It is the case that there is no disabled items in a list.
	isItemDisabled(currentIndex) // If the currnet index item is disabled, it means that all items in a list are disabled.
)
```

packages/moonstone/docs/virtual-list-scroller.md
ybsung: We thought this code could be defined outside of class. But it looks this code was defined inside of class. So we're going not to fix it.

packages/moonstone/docs/virtual-list-scroller.md
```diff
@@ -58,19 +53,20 @@ This document describes VirtualList, VirtualGridList, and Scroller.
...
     ```
     //.js
-    renderItem = ({data, index, ...rest}) => {
+    items = []
+    ...
+    renderItem = ({index, ...rest}) => {
         return (
             <div {...rest}>
-                {data[index].name}
+                {this.items[index].name}
```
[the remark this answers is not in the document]
ybsung: We thought this code might be defined outside of class. But it looks this code was defined inside of class. So we need `this` and we're going not to fix it.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -85,6 +88,12 @@ const VirtualListBaseFactory = (type) => {
 			 * ```
 			 *
 			 * @type {Function}
+			 * @param {Object} event
+			 * @param {Number} event.data-index It is required for Spotlight 5-way navigation. Pass to the root element in the component.
+			 * @param {Number} event.index The index number of the component to render
+			 * @param {Number} event.key It MUST be passed as a prop to the root element in the component for DOM recycling.
+			 *
+			 * @required
```
ybsung: I had to define `@param`s here. Thanks~

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -117,7 +127,7 @@ const VirtualListBaseFactory = (type) => {
 			/**
 			 * The Function that returns `true` if the item at the index is disabled.
 			 * It is used to navigate a list properly with 5 way keys, page up key,
-			 * and page down key. If it is not supplied, it assumes that no items are disabeld.
+			 * and page down key. If it is not supplied, it assumes that no items are disabled.
```
ybsung: Oops. My type error.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -55,6 +55,14 @@ const
 	JS = 'JS',
 	Native = 'Native';
 
+/**
+ * The base version of the Moonstone virtual list component.
+ *
+ * @class VirtualListBase
```
ybsung: We already defined `VirtualListBase` in this file.

```
/**
 * A Moonstone-styled base component for [VirtualList]{@link moonstone/VirtualList.VirtualList} and
 * [VirtualGridList]{@link moonstone/VirtualList.VirtualGridList}.
 *
 * @class VirtualListBase
 * @memberof moonstone/VirtualList
 * @extends ui/VirtualList.VirtualListBase
 * @ui
 * @private
 */
const VirtualListBase = VirtualListBaseFactory(JS);
VirtualListBase.displayName = 'VirtualListBase';
```

webOS101: I missed that.  It's private, I guess that's why the docs didn't show up.  We can revert that, I guess.  I'll take another stab at it later.

ybsung: I made `VirtualListCore`, `VirtualListBase` public. Then the `role`, `itemRenderer`, and `isItemDisabled` of `VirtualList` became visible in Enact API documentation.

ybsung: `VirtualListBase` and `VirtualListBaseNative` are based on `VirtualListCore`.
`VirtualList` and `VirtualGridList` are based on `VirtualListBase`,
`VirtualListNative` and `VirtualGridListNative` are based on `VirtualListBaseNative`.

So all common virtual list related props should be defined in `VirtualListCore`.

PR #1501, opened 2018-03-15
PLAT-52178: Deprecate `component` in `VirtualList`, which will be replaced by `itemRenderer` (release 1.x)

ybsung:
### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)

In Enact 2.0 (https://github.com/enyojs/enact/pull/1500), we'd like to rename the `component` in a `VirtualList` as `itemRenderer`.

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)

Add `dprecate` logs and JS docs

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)

PLAT-52178

### Comments

Enact-DCO-1.0-Signed-off-by: YB Sung (yb.sung@lge.com)

PR #1528, opened 2018-03-23
PLAT-52443: Add the log to deprecate `data` prop in `VirtualList` (release 1.x)

ybsung:

In Enact 2.0 (https://github.com/enactjs/enact/pull/1524), we'd like to remove the `data` prop in a `VirtualList`.

### Resolution

Add the log to deprecate `data` prop in `VirtualList`

### Additional Considerations

### Links

PLAT-52443

### Comments

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -51,6 +51,11 @@ const deprecateComponent = deprecate(
 	{name: 'component', replacedBy: 'itemRenderer', until: '2.0.0'}
 );
 
+const deprecateData = deprecate(
+	() => {},
+	{name: 'data', until: '2.0.0'}
+);
+
 /**
  * {@link moonstone/VirtualList.VirtualListBase} is a base component for
  * {@link moonstone/VirtualList.VirtualList} and
@@ -137,6 +142,7 @@ class VirtualListCore extends Component {
 		 *
 		 * @type {Any}
 		 * @default []
+		 * @deprecated will be removed in 2.0.0
 		 * @public
 		 */
 		data: PropTypes.any,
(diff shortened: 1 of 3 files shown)


MikyungKim:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -51,6 +51,11 @@ const deprecateComponent = deprecate(
 	{name: 'component', replacedBy: 'itemRenderer', until: '2.0.0'}
 );
 
+const deprecateData = deprecate(
+	() => {},
+	{name: 'data', until: '2.0.0'}
```
webOS101: It's not necessary to pass an empty method to deprecate.  Not sure how I missed this before but if you pass a single argument it will print the message immediately. Can do same with `deprecateComponent` as well.

packages/moonstone/VirtualList/VirtualListBaseNative.js
```diff
@@ -53,6 +53,11 @@ const deprecateComponent = deprecate(
 	{name: 'component', replacedBy: 'itemRenderer', until: '2.0.0'}
 );
 
+const deprecateData = deprecate(
```
webOS101: Same as above

packages/moonstone/VirtualList/VirtualListBase.js
ybsung: In https://github.com/enactjs/enact/pull/1501/commits/318af46e4d35579c09c136f7adb536056c9e4c82, @ryanjduffy tried to print the message once. According to the definition of `deprecate`, we need to pass the function as the first argument and the configuration as the second argument which could not be omitted to do it once. If I missed something, please let me know.

packages/moonstone/VirtualList/VirtualListBaseNative.js
ybsung: Same as above.

PR #1530, opened 2018-03-23
ENYO-5088: Refactor ProgressBarTooltip to use CSS variables for positioning

Djspaceg:
`ProgressBarTooltip` (formerly `SliderTooltip`) used some pretty outdated and non-functional techniques.

### Resolution
Essentially now, the `ProgressBarTooltip` component just transmits its proportion value directly to CSS. CSS then does calculations and positioning for every orientation of the tooltip. `Slider` can define a moonstone CSS "offset" variable which is then factored into the `ProgressBarTooltip` positioning calculation so it too is properly positioned.

### Additional Considerations
Does Storybook not support CSS variables? My Storybook was erroring and it didn't make sense as to why. My installation may be out of date.


teckliew:
I added an extra guard for `(!context.rtl && forceSide && side === 'before')` to make sure a progress bar with `tooltipSide='before'` and `tooltipForceSide` shows up normally in `LTR` 
We should probably add a changelog for adding `tooltipSide`.

PR #1531, opened 2018-03-26
PLAT-51654: Add timeout support to webos/LS2Request

sugardave:
- `onTimeout` and `timeout` props are added to `LS2Request`
- default `onFailure` and `onTimeout` handlers to `LS2Request`

### Additional Considerations
If a `timeout` is not specified, the `onTimeout` handler will not be called.

### Links
PLAT-51654

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,6 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
-## [unreleased]
-
-### Added
-
-- `onTimeout` and `timeout` props to `webos/LS2Request`
-- default `onFailure` and `onTimeout` handlers to `webos/LS2Request`
-
 ## [2.0.0-alpha.6] - 2018-03-22
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `onTimeout` and `timeout` props to `webos/LS2Request`
+
 ## [2.0.0-alpha.6] - 2018-03-22
 
 ### Removed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact project, newest changes 
 ### Added
 
 - `onTimeout` and `timeout` props to `webos/LS2Request`
+- default `onTimeout` handler to `webos/LS2Request`
 
 ## [2.0.0-alpha.6] - 2018-03-22
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ The following is a curated list of changes in the Enact project, newest changes 
 ### Added
 
 - `onTimeout` and `timeout` props to `webos/LS2Request`
-- default `onTimeout` handler to `webos/LS2Request`
+- default `onFailure` and `onTimeout` handlers to `webos/LS2Request`
 
 ## [2.0.0-alpha.6] - 2018-03-22
 
--- a/packages/webos/CHANGELOG.md
+++ b/packages/webos/CHANGELOG.md
@@ -7,7 +7,7 @@ The following is a curated list of changes in the Enact webos module, newest cha
 ### Added
 
 - `onTimeout` and `timeout` props to `LS2Request`
-- default `onTimeout` handler to `LS2Request`
+- default `onFailure` and `onTimeout` handlers to `LS2Request`
 
 ## [2.0.0-alpha.6] - 2018-03-22
 
--- a/packages/webos/CHANGELOG.md
+++ b/packages/webos/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact webos module, newest cha
 ### Added
 
 - `onTimeout` and `timeout` props to `LS2Request`
+- default `onTimeout` handler to `LS2Request`
 
 ## [2.0.0-alpha.6] - 2018-03-22
 
--- a/packages/webos/CHANGELOG.md
+++ b/packages/webos/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact webos module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `onTimeout` and `timeout` props to `LS2Request`
+
 ## [2.0.0-alpha.6] - 2018-03-22
 
 No significant changes.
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
```
webOS101: We update the root level changelog only when we're preparing a release.

packages/webos/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact webos module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `onTimeout` and `timeout` props to `LS2Request`
```
webOS101: Follow style of putting component name first and props after.

packages/webos/LS2Request/LS2Request.js
```diff
@@ -55,10 +66,22 @@ export default class LS2Request {
 		this.bridge.onservicecallback = this.callback.bind(this, onSuccess, onFailure, onComplete);
 		this.bridge.call(adjustPath(service) + method, JSON.stringify(parameters));
 
+		// timer will be cleared when the callback is run, so this should only fire if the timeout threshold is reached
```
webOS101: Could `this.bridge.call` be synchronous for any calls?  If so, timer will be clear when `callback` is called and the timer will be set up after.  Would it be safer to reorder these lines?

packages/webos/LS2Request/LS2Request.js
```diff
@@ -55,10 +66,22 @@ export default class LS2Request {
...
+		this.timer = setTimeout(() => {
+			// only call onTimeout handler if there was a timeout threshold to beat
+			if (timeout) {
+				onTimeout({errorCode: -1, errorText: `LS2Request timed out after ${timeout} ms.`, returnValue: false});
```
viodragon2: I think the `errorCode` should be different for timeout. I'd suggest `-2`.

packages/webos/LS2Request/LS2Request.js
```diff
@@ -55,10 +66,22 @@ export default class LS2Request {
...
+				// cancel the request
+				this.cancel();
+			}
+		}, timeout ? timeout : 0);
```
viodragon2: Can we avoid calling timer when `timeout` is set to `0`?

packages/webos/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact webos module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `webos/LS2Request` props `onTimeout` and `timeout`
```
viodragon2: Not sure if they are qualified as `props` as it's not a React.Component.

packages/webos/LS2Request/LS2Request.js
```diff
@@ -53,15 +65,27 @@ export default class LS2Request {
 		// eslint-disable-next-line no-undef
 		this.bridge = new PalmServiceBridge();
 		this.bridge.onservicecallback = this.callback.bind(this, onSuccess, onFailure, onComplete);
+		if (timeout) {
+			// timer will be cleared when the callback is run, so this should only fire if the timeout threshold is reached
+			this.timer = new Job(() => {
```
viodragon2: Let's declare `Job` outside so that we don't initiate every time `send()` is invoked. Something like this

```
timeOutJob = new Job( ... )
...
send () {
  ... 
  if (timeout) {
    timeOutJob.startAfter(timeout);
  }
  ...
}
```

webOS101: Can there only be one outstanding LS2 request at a time?

webOS101: Or did you mean at the class level?

viodragon2: at class level

sugardave: The issue I see is that `onTimeout` is passed to `send()` so at the class level, I can't make the job function use it.

viodragon2: You could pass arguments for the callback function as a second parameter in `startAfter`. So something like this: 

```
timeOutJob = new Job(({onTimeout, timeout}) => {
...
this.timeOutJob.startAfter(timeout, {onTimeout, timeout});
```

packages/webos/LS2Request/LS2Request.js
```diff
@@ -24,12 +30,18 @@ export default class LS2Request {
 		onSuccess = null,
 		onFailure = null,
 		onComplete = null,
-		subscribe = false
+		onTimeout = timeoutHandler,
```
viodragon2: nit: should it be `onTimeOut` with capital "O"?

sugardave: I think "timeout" is correct.

packages/webos/LS2Request/LS2Request.js
sugardave: :genius:

packages/webos/LS2Request/LS2Request.js
```diff
@@ -54,14 +71,20 @@ export default class LS2Request {
...
 		return this;
 	}
 
 	callback (onSuccess, onFailure, onComplete, msg) {
 		if (this.cancelled) {
 			return;
 		}
+
+		// remove timeout job
+		this.timeoutJob.stop();
```
ryanjduffy: This is a bit tangential but this line brought it to mind. It appears that we're not resetting the `cancelled` flag when sending a new request. As a result, if you cancel the first request, the second won't fire the callback nor cancel the timeout job (since it'll return early just before this line) causing `onTimeout` to fire for a successful second call.

@viodragon2 + @sugardave - Are you reading the same? If so, let's add that change here to proactively fix the bug I invented above.

sugardave: I think you're right, I'll fix it up.

packages/webos/LS2Request/LS2Request.js
```diff
@@ -10,26 +18,60 @@ const adjustPath = (path) => {
...
+const failureHandler = ({errorText}) => console.error(`LS2Request: ${errorText}`);
+const timeoutHandler = ({errorText}) => console.warn(`LS2Request: ${errorText}`);
+
 export default class LS2Request {
+	timeoutJob = new Job(({onTimeout, timeout}) => {
+		onTimeout({errorCode: -2, errorText: `Request timed out after ${timeout} ms.`, returnValue: false});
+		// cancel the request
+		this.cancel();
+	});
+
```
webOS101: Need `@class` declaration here.

PR #1534, opened 2018-03-28
ENYO-5094: Display overflowing children in slide transition

viodragon2:
Transition applies `overflow: hidden` at the same level as `className` prop in `slide` type transition. In cases where the user wants to set `position: absolute`, children won't be visible.
We see this problem in `moonstone/Notification` where it does not display when `open` is set to true. 

### Resolution
Partial revert from #1481 to apply `overflow: hidden` at an outer `div`.

### DIFF
--- a/packages/ui/Transition/tests/Transition-specs.js
+++ b/packages/ui/Transition/tests/Transition-specs.js
@@ -84,7 +84,7 @@ describe('Transition Specs', () => {
 				);
 
 				const expected = key;
-				const actual = wrapped.find('div').at(1).prop('className');
+				const actual = wrapped.find('div').at(0).prop('className');
 
 				expect(actual).to.contain(expected);
 			});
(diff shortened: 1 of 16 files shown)


packages/ui/Transition/tests/Transition-specs.js
```diff
@@ -84,7 +84,7 @@ describe('Transition Specs', () => {
 				);
 
 				const expected = key;
-				const actual = wrapped.find('div').at(0).prop('className');
+				const actual = wrapped.find('div').at(1).prop('className');
```
viodragon2: Should we be concerned that it only tests for `type: slide`?

viodragon2:
The y-position of notification container needs to be anchored at 324px from the bottom, so I've used `translateY(50%)` to address the issue. Also, `pointer-events` were removed from `ui/Trasition` and moved to `moonstone/Popup`.

ryanjduffy:
Good fix, @viodragon2!

Djspaceg:
*A detail clarification.* The Notification's center should be anchored 324px from the bottom. Any changes in height will expand outward from that anchor, both up and down, which is why `translateY(50%)` is used. The bottom of the Notification will never be 324px from the bottom of the screen.

PR #1535, opened 2018-03-29
ENYO-4162: Add preload support to moonstone/VideoPlayer

rundmt:
Video player loads the next video a bit slow. We need to add another tag to preload the source.

### Resolution
This is the current API. 
```
<VideoPlayer
    {...}
    preloadSource={<source src={this.state.preloadSrc} type="video/mp4" />}
>
    <source src={this.state.src} type="video/mp4" />
</VideoPlayer>
```

### Additional Considerations
Not sure if this only needs to be for local videos, if that's the case something like Ajaxing could be better. http://dinbror.dk/blog/how-to-preload-entire-html5-video-before-play-solved/

### Links
ENYO-4162


packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -755,9 +764,17 @@ const VideoPlayerBase = class extends React.Component {
...
 
-		const {source} = this.props;
+		const {source, preloadSource} = this.props;
 		const {source: nextSource} = nextProps;
 
+		if (preloadSource && compareSources(preloadSource, nextSource)) {
+			this.preloadSourcePlaying = !this.preloadSourcePlaying;
+			const currentVideoSource = this.video;
+			this.video = this.preloadVideo;
+			this.preloadVideo = currentVideoSource;
```
webOS101: Why swap the current video to preload?  And, do we need to call `load()` on it if it was already loaded?

webOS101: What happens when the preload video is different from the previous preload video?  Do we need to handle that case?

rundmt: The reason I swap the current video to preload is because of the refs. Refs don't re run on render so we have to swap it out then call load on it. This way we can reuse the ref for loading the next video.

rundmt: Also I reworked the loaded stuff. I think we should only call it after componentDidUpdate. This way it will load after the rendering. I think that makes most sense.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -766,13 +777,25 @@ const VideoPlayerBase = class extends React.Component {
...
+
+		if (preloadSource && compareSources(preloadSource, nextSource)) {
+			this.preloadSourcePlaying = !this.preloadSourcePlaying;
+			const currentVideoSource = this.video;
+			this.video = this.preloadVideo;
+			this.preloadVideo = currentVideoSource;
+			this.isVideoPreloaded = true;
+		}
+
+		if (nextPreloadSource) {
```
webOS101: Do we need to compare against the previous preload source?  This will happen every time, will it not?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1896,6 +1924,20 @@ const VideoPlayerBase = class extends React.Component {
 		return null;
 	}
 
+	// These methods are here because on webOS TVs we can't play a video until after second video
+	// player is loaded
+	onLoadStartInitialPlaying = () => {
+		if (this.isPlayerMounted && this.isVideoPreloaded) {
```
webOS101: Do we need to test for `autoPlay` here?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1962,6 +2005,32 @@ const VideoPlayerBase = class extends React.Component {
 		const moreDisabled = !(this.state.more);
 		const controlsAriaProps = this.getControlsAriaProps();
 
+		const initialPlayingProps = {
```
webOS101: Could you add some documentation to describe the difference between these two sets of props?  Also, should we apply keys to either of these in case the app doesn't set a preload video immediately?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1952,17 +1964,29 @@ const VideoPlayerBase = class extends React.Component {
...
-					autoPlay={!noAutoPlay}
-					className={css.video}
-					component={videoComponent}
-					controls={false}
-					onUpdate={this.handleEvent}
-					ref={this.setVideoRef}
-				>
-					{source}
-				</Media>
+				{this.props.preloadSource ?
```
Djspaceg: I'm curious why this check happens here, instead of inside the `VideoWithPreload` component, and more fundamentally, why `VideoWithPreload` is different from `Media`. Shouldn't `Media` just have a `preloadSource` prop, which then invokes that line of logic inside `Media`? One component able to support preloading or normal.

PR #1545, opened 2018-04-03
PLAT-52790: Fix the a11y property of VirtualList

Baekwoo-s:
Add 'role=list' in VirtualList
Remove 'role=region' in SpotlightPlaceholder

Reasons to add 'role=list' to VirtualList
- The list role is a semantic role, so putting it in the list does not change the behavior. However, for the XofY feature, we need 'role=list' just above the 'aria-posinset' and 'aria-setsize' properties.
I will include the list role in the VirtualList so that the App does not have to pass the list role to the VirtualList as prop.

Reasons to remove 'role=region' from spotlightPlaceholder
- Currently, in the 1553 Master Image, there is no issue in the existing behavior even if there is no region role in the spotlightPlaceholder.
This is because Chethan has modified Blink to not read an empty div. (Related ticket: https://jira2.lgsvl.com/browse/PLAT-44268)

### Links
PLAT-52790

### Comments

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -731,7 +731,7 @@ VirtualListBase.displayName = 'VirtualListBase';
 const VirtualListBaseNative = VirtualListBaseFactory(Native);
 VirtualListBaseNative.displayName = 'VirtualListBaseNative';
 
-const ScrollableVirtualList = ({role, ...rest}) => ( // eslint-disable-line react/jsx-no-bind
+const ScrollableVirtualList = ({...rest}) => ( // eslint-disable-line react/jsx-no-bind
 	<Scrollable
 		{...rest}
 		childRenderer={(props) => ( // eslint-disable-line react/jsx-no-bind
@@ -739,7 +739,7 @@ const ScrollableVirtualList = ({role, ...rest}) => ( // eslint-disable-line reac
 				{...props}
 				itemsRenderer={({cc, primary, needsScrollingPlaceholder, initItemContainerRef, handlePlaceholderFocus}) => ( // eslint-disable-line react/jsx-no-bind
 					[
-						cc.length ? <div key="0" ref={initItemContainerRef} role={role}>{cc}</div> : null,
+						cc.length ? <div key="0" ref={initItemContainerRef} role="list">{cc}</div> : null,
 						primary ?
 							null :
 							<SpotlightPlaceholder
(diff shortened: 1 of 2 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -770,23 +770,22 @@ VirtualListBase.displayName = 'VirtualListBase';
 const VirtualListBaseNative = VirtualListBaseFactory(Native);
 VirtualListBaseNative.displayName = 'VirtualListBaseNative';
 
-const ScrollableVirtualList = ({role, ...rest}) => ( // eslint-disable-line react/jsx-no-bind
+const ScrollableVirtualList = ({...rest}) => ( // eslint-disable-line react/jsx-no-bind
```
ybsung: Please define `props` instead of `{...rest}`.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -770,23 +770,22 @@ VirtualListBase.displayName = 'VirtualListBase';
...
 	<Scrollable
 		{...rest}
```
ybsung: Then, `{...rest} should be `{...props}`.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -796,33 +795,22 @@ const ScrollableVirtualList = ({role, ...rest}) => ( // eslint-disable-line reac
...
-	 * Aria role.
-	 *
-	 * @type {String}
-	 * @public
-	 */
-	role: PropTypes.string
-};
-
-const ScrollableVirtualListNative = ({role, ...rest}) => (
+const ScrollableVirtualListNative = ({...rest}) => (
```
ybsung: Same above.

Please remind this our coding rule.

packages/moonstone/VirtualList/VirtualListBase.js
webOS101: Or, more simply, `(props)`.

0x64:
I updated the description to make it clearer.

Baekwoo-s:
I don't know why the travis CI build failed.

packages/moonstone/VirtualList/VirtualListBase.js
Baekwoo-s: I missed that point. I changed from 'rest' to 'props'.

packages/moonstone/VirtualList/VirtualListBase.js
Baekwoo-s: I missed that point. I changed from 'rest' to 'props'.

0x64:
LGTM

PR #1551, opened 2018-04-05
PLAT-49774: Add wrap prop to VirtualList

0x64:
For Easy TV, `VirtualList` and `VirtualGridList` are required to support wrap-around navigation by 5-way keys.

In this PR, `VirtualListEasy` and `VirtualGridListEasy` are added to moonstone. The new prop `wrap` is added. The acceptable value should be one of `true`, `false`, and `'noAnimation'`
- false (default): the spotlight deos not move to the opposite side.
- true: the spotlight moves to the opposite side with an animation when it reaches the starting point or the ending point.
- 'noAnimation': a list is not animated but it jumps when the spotlight moves in wraparound manner.

### Links
PLAT-49774

### Comments


packages/moonstone/VirtualList/VirtualList.js
```diff
@@ -176,12 +176,154 @@ const VirtualGridListNative = kind({
...
+ * @extends moonstone/VirtualList.VirtualList
+ * @ui
+ * @public
+ */
+class VirtualListEasy extends Component {
+	static displayName = 'VirtualListEasy'
+
+	static propTypes = /** @lends moonstone/VirtualList.VirtualListEasy.prototype */ {
+		/**
+		 * When `true`, if the spotlight focus moves in wraparound manner to the other side,
```
ybsung: How about removing `if`?
```
When `true`, the spotlight focus moves in wraparound manner to the other side,
```

0x64: Oh, I got your point. Thanks!

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -441,20 +520,12 @@ const VirtualListBaseFactory = (type) => {
 
 		jumpToSpottableItem = (keyCode, target) => {
 			const
-				{cbScrollTo, dataSize, isItemDisabled} = this.props,
+				{cbScrollTo, dataSize, isItemDisabled, wrap, wrapAnimated} = this.props,
```
ybsung: You could add `rtl` to get this.props.rtl instead of `rtl = this.props.rtl,`.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -441,20 +520,12 @@ const VirtualListBaseFactory = (type) => {
...
+				{findSpottableExtent, findSpottableItem, getExtentIndex} = this,
```
ybsung: Please define variables from this.uiRef, then variable from this.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -466,66 +537,107 @@ const VirtualListBaseFactory = (type) => {
 					null
 				);
 
-			let nextIndex = -1;
+			// If the currently focused item is disabled, we assume that all items in a list are disabled.
+			if (isItemDisabled(currentIndex) || (!isForward && !isBackward)) {
```
ybsung: Please check if `isItemDisabled === isItemDisabledDefault` first.

packages/moonstone/VirtualList/VirtualListBase.js
0x64: 👍

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -441,20 +520,12 @@ const VirtualListBaseFactory = (type) => {
...
-			if (
-				!isItemDisabled || // It is the case that there is no disabled items in a list.
-				isItemDisabled(currentIndex) // If the currnet index item is disabled, it means that all items in a list are disabled.
-			) {
-				return false;
-			}
-
-			const
+				currentIndex = Number.parseInt(target.getAttribute(dataIndexAttribute)),
 				isForward = (
```
ybsung: Please move the code defining `isForward` and `isBackward` after cheking `if (isItemDisabled(currentIndex)` because in the most of the cases, disabled items are used and it could make VirtualList better navigating performance with all enabled items.

packages/moonstone/VirtualList/VirtualListBase.js
0x64: Okay

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -466,66 +537,107 @@ const VirtualListBaseFactory = (type) => {
...
+			if (nextIndex !== -1) {
+				if (firstIndex <= nextIndex && nextIndex < firstIndex + numOfItems) {
+					this.focusOnItem(nextIndex);
+				} else {
+					if (isWrapped) {
+						// In case of 'wrapping-around',
+						// we need to blur the current focus immediately
+						// since it can be a very long scroll (from one edge to the other edge)
+						// and definitely it's not a case of changing "pointer" mode to "5way key" mode.
+						target.blur();
```
ybsung: Please blur a DOM element after calling Spotlight.pause().

ybsung: I'm not sure, but if blurring a target directly, Spotlight could give it focus again while switching from "pointer" mode to "5way key" mode. But I could not produce this issue with your PR. I think that it's find now, but we need to keep thinking of it.

packages/moonstone/VirtualList/VirtualListBase.js
0x64: Calculations are not so hard and I think it's okay regarding key event rate and calculation time.

packages/moonstone/VirtualList/VirtualListBase.js
0x64: Good point. I will fix it.

0x64: Thank you for remind. Sure, we need to.

packages/moonstone/VirtualList/VirtualListBase.js
0x64: Okay.

ybsung:
LGTM

ybsung:
@0x64
I forgot that we needed to update CHANGELOG.md. Could you do it?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -466,82 +527,129 @@ const VirtualListBaseFactory = (type) => {
...
 			if (getDirection(keyCode)) {
 				ev.preventDefault();
 				this.setSpotlightContainerRestrict(keyCode, target);
-				this.isScrolledBy5way = this.jumpToSpottableItem(keyCode, target);
+				this.isScrolledBy5way = this.jumpToSpottableItem(keyCode, repeat, target);
+				if (this.shouldPreventScrollByFocus()) {
+					ev.stopPropagation();
+				}
 			}
-			forward('onKeyDown', ev, this.props);
```
ybsung: If removing this code, page up/down key handling code won't be called. Please check it again.

ybsung:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -300,9 +323,78 @@ const VirtualListBaseFactory = (type) => {
...
+				-1 <= indexTo && indexTo <= dataSize &&
+				0 <= position && position < dimensionToExtent) {
+				const
+					{isItemDisabled} = this.props,
+					direction = (indexFrom < indexTo) ? 1 : -1,
+					delta = direction * dimensionToExtent,
+					diffPosition = (indexFrom % dimensionToExtent) - position,
+					// When direction is 1 (forward) and diffPosition is positive, add dimensionToExtent.
+					// When direction is -1 (backward) and diffPosition is negative, substract dimensionToExtent.
+					startIndex = indexFrom - diffPosition + ((direction * diffPosition > 0) ? (direction * dimensionToExtent) : 0);
```
MikyungKim: (direction * dimensionToExtent) -> Isn't this delta?

0x64: YES! 😅

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -482,82 +546,129 @@ const VirtualListBaseFactory = (type) => {
...
-			const {keyCode, target} = ev;
+			const {keyCode, repeat, target} = ev;
 
 			this.isScrolledBy5way = false;
 			if (getDirection(keyCode)) {
 				ev.preventDefault();
 				this.setSpotlightContainerRestrict(keyCode, target);
-				this.isScrolledBy5way = this.jumpToSpottableItem(keyCode, target);
+				this.isScrolledBy5way = this.jumpToSpottableItem(keyCode, repeat, target);
+				if (this.shouldPreventScrollByFocus()) {
```
MikyungKim: Scrollable already does this. It's redundant.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -482,82 +546,129 @@ const VirtualListBaseFactory = (type) => {
...
 
-			if (nextIndex !== -1 && (firstIndex > nextIndex || nextIndex >= firstIndex + numOfItems)) {
-				// When changing from "pointer" mode to "5way key" mode,
-				// a pointer is hidden and a last focused item get focused after 30ms.
-				// To make sure the item to be blurred after that, we used 50ms.
-				setTimeout(() => {
-					target.blur();
-				}, 50);
+			if (nextIndex !== -1) {
+				if (firstIndex <= nextIndex && nextIndex < firstIndex + numOfItems) {
```
MikyungKim: Shouldn't this be visible indices? please test the case.

0x64: It works okay. `jumpToSpottableItem` returns `false` in this case.

packages/moonstone/VirtualList/VirtualListBase.js
0x64: We need to prevent spotlight from a 'keydown' event which makes another focus moving.

MikyungKim:
LGTM

PR #1552, opened 2018-04-05
PLAT-52106: Add closeButtonAriaLabel prop to moonstone/Panels

seunghoh:
There's no way to override a11y string on Exit App button of Panels

### Resolution
Add Panels property `exitAppAriaLabel` to override readout string on 

### Additional Considerations

### Links
PLAT-52106

### Comments
Enact-DCO-1.0-Signed-off-by: Seungho Park <seunghoh.park@lge.com>

### DIFF
--- a/packages/moonstone/Panels/ApplicationCloseButton.js
+++ b/packages/moonstone/Panels/ApplicationCloseButton.js
@@ -59,14 +59,13 @@ const ApplicationCloseButton = kind({
 	},
 
 	computed: {
-		closeButtonAriaLabel: ({'aria-label': ariaLabel}) => (ariaLabel == null) ? $L('Exit app') : ariaLabel
+		'aria-label': ({'aria-label': ariaLabel}) => ariaLabel == null ? $L('Exit app') : ariaLabel
 	},
 
-	render: ({backgroundOpacity, closeButtonAriaLabel, onApplicationClose, ...rest}) => {
+	render: ({backgroundOpacity, onApplicationClose, ...rest}) => {
 		return (
 			<IconButton
 				{...rest}
-				aria-label={closeButtonAriaLabel}
 				backgroundOpacity={backgroundOpacity}
 				onTap={onApplicationClose}
 				small
--- a/packages/moonstone/Panels/ApplicationCloseButton.js
+++ b/packages/moonstone/Panels/ApplicationCloseButton.js
@@ -23,6 +23,15 @@ const ApplicationCloseButton = kind({
 
 	propTypes: /** @lends moonstone/Panels.ApplicationCloseButton.prototype */ {
 		/**
+		* Sets the hint string read when focusing the application close button.
+		*
+		* @type {String}
+		* @default 'Exit app'
+		* @public
+		*/
+		'aria-label': PropTypes.string,
+
+		/**
 		 * The background-color opacity of this button; valid values are `'opaque'`, `'translucent'`,
 		 * `'lightTranslucent'` and `'transparent'`.
 		 *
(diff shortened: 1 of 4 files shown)


packages/moonstone/Panels/Panels.js
```diff
@@ -69,6 +70,15 @@ const PanelsBase = kind({
...
 		closeButtonBackgroundOpacity: PropTypes.oneOf(['opaque', 'translucent', 'lightTranslucent', 'transparent']),
 
+		/**
+		* Sets the hint string read when focusing the application close button.
+		*
+		* @type {String}
+		* @default 'Exit app'
+		* @public
+		*/
+		exitAppAriaLabel: PropTypes.string,
```
ryanjduffy: Suggest using `closeButtonAriaLabel` for consistency with the other related prop names

packages/moonstone/Panels/Panels.js
```diff
@@ -137,12 +147,14 @@ const PanelsBase = kind({
 		className: ({noCloseButton, styler}) => styler.append({
 			hasCloseButton: !noCloseButton
 		}),
-		applicationCloseButton: ({closeButtonBackgroundOpacity, id, noCloseButton, onApplicationClose}) => {
+		exitAppAriaLabel: ({exitAppAriaLabel}) => (exitAppAriaLabel == null) ? $L('Exit app') : exitAppAriaLabel,
```
ryanjduffy: This won't cascade into the `applicationCloseButton` computed prop as you're expected. In order to keep these predictable and optimizable, each computed property receives the same source props and not the results of "previous" computed executions. As a result, in the `applicationCloseButton` computed prop below, `exitAppAriaLabel` would always equal the value from props and not this computed value.

That all said, since you've implemented the same logic within the `ApplicationCloseButton` component, I suggest omitting this logic from this component and letting the `closeButtonAriaLabel` prop flow untouched to `ApplicationCloseButton`.

packages/moonstone/Panels/ApplicationCloseButton.js
```diff
@@ -49,11 +58,15 @@ const ApplicationCloseButton = kind({
 		className: 'applicationCloseButton'
 	},
 
-	render: ({backgroundOpacity, onApplicationClose, ...rest}) => {
+	computed: {
+		closeButtonAriaLabel: ({'aria-label': ariaLabel}) => (ariaLabel == null) ? $L('Exit app') : ariaLabel
+	},
+
+	render: ({backgroundOpacity, closeButtonAriaLabel, onApplicationClose, ...rest}) => {
 		return (
```
seunghoh: @ryanjduffy one question. Shall I add `delete rest['aria-label']` here or will be OK since override below?

packages/moonstone/Panels/ApplicationCloseButton.js
```diff
@@ -49,11 +58,15 @@ const ApplicationCloseButton = kind({
 		className: 'applicationCloseButton'
 	},
 
-	render: ({backgroundOpacity, onApplicationClose, ...rest}) => {
+	computed: {
+		closeButtonAriaLabel: ({'aria-label': ariaLabel}) => (ariaLabel == null) ? $L('Exit app') : ariaLabel
+	},
+
+	render: ({backgroundOpacity, closeButtonAriaLabel, onApplicationClose, ...rest}) => {
```
seunghoh: @ryanjduffy one question. Shall I add `delete rest['aria-label']` here or will be OK since override below?

ryanjduffy: Normally, if we're overriding a prop, we use that prop as the computed value to avoid the confusion. This does work fine as is but I'll update for consistency.

PR #1553, opened 2018-04-06
ENYO-4501: Fix navigating via 5-way from paging controls

MikyungKim:
Currently, Scroller/VirtualList has two separate spotlight containers. One for Scrollable and the other for Scroller/VirtualList itself.
This structure leads to unwanted behaviors that if a user press "up" key from "up-side" paging controls in VirtualListNative from the bottom of the list, the focus supposed to move to out of the list but it goes inside of the list. Also, when a user presses 5way to go into the list from paging controls, spotlight focuses the last focused element instead of the nearest element.

### Resolution
To resolve above issues, @ryanjduffy initially proposed merging two spotlight containers to one and had made a PR for this. (https://github.com/enactjs/enact/pull/1169)
- Collapse spotlight containers for Scrollable consumers
- Add data-spotlight-overflow support to allow spotlight to escape overflow containers via 5-way

In this PR, I rebased the above PR and made some changes to work.

### Additional Considerations
enact-samples/pattern-virtuallist-preserving-focus and photo and video app have been tested.

### Links
ENYO-4501

### Comments


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -488,6 +507,7 @@ class Scrollable extends Component {
 									...childComponentProps,
 									cbScrollTo: scrollTo,
 									className: componentCss.scrollableFill,
+									spotlightId,
```
ybsung: Alphabetically please.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -192,7 +184,24 @@ const VirtualListBaseFactory = (type) => {
...
 			 * @public
 			 */
-			spacing: PropTypes.number
+			spacing: PropTypes.number,
+
+			/**
+			 * Spotlight Id. It would be the same with [Scrollable]{@link ui/Scrollable.Scrollable}'s.
+			 *
+			 * @type {String}
+			 * @private
```
ybsung: Please define default value.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -855,12 +855,10 @@ class ScrollableBaseNative extends Component {
 
 	scrollTo = (opt) => {
 		if (!this.deferScrollTo) {
-			const
-				{left, top} = this.getPositionForScrollTo(opt),
-				{scrollTo} = this.props;
+			const {left, top} = this.getPositionForScrollTo(opt);
```
ybsung: Do we need to use `this.props.scrollTo` instead of `props.scrollTo`?

packages/spotlight/SpotlightContainerDecorator/SpotlightContainerDecorator.js
```diff
@@ -116,6 +116,14 @@ const SpotlightContainerDecorator = hoc(defaultConfig, (config, Wrapped) => {
 	const forwardMouseLeave = forward(leaveEvent);
 	const {navigableFilter, preserveId, ...containerConfig} = config;
 
+	const add = ({spotlightId}) => {
```
ybsung: `add` is not good for the function name here.
How about `initializeState`?

packages/spotlight/SpotlightContainerDecorator/SpotlightContainerDecorator.js
```diff
@@ -116,6 +116,14 @@ const SpotlightContainerDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+		const id = Spotlight.add(spotlightId);
+		return {
+			id,
+			preserveId: preserveId && id === spotlightId
+		};
```
ybsung: Do we need to execute the following code if changing the state of `preserveId`?

```
			if (preserveId) {
				Spotlight.unmount(this.state.id);
			} else {
				Spotlight.remove(this.state.id);
			}
```

packages/moonstone/Scrollable/Scrollable.js
MikyungKim: okay

packages/ui/Scrollable/ScrollableNative.js
MikyungKim: The previous way, the context was broken so I had to change this way.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -26,41 +26,24 @@ const
...
- * A Moonstone-styled native component that provides horizontal and vertical scrollbars.
- *
- * @class ScrollableNative
- * @memberof moonstone/ScrollableNative
- * @extends ui/Scrollable.ScrollableBaseNative
- * @ui
- * @private
- */
-class ScrollableNative extends Component {
+class ScrollableBaseNative extends Component {
```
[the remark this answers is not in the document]
MikyungKim: :okay:

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -476,10 +494,11 @@ class Scrollable extends Component {
 					touchableProps,
 					verticalScrollbarProps
 				}) => (
-					<ScrollableSpotlightContainer
+					<div
```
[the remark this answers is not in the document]
ybsung: @MikyungKim and I discussed about the comment above. We're going not to fix the comment because <div> element has customized attributes and scrollbars. So it is better to remain the code.

packages/spotlight/SpotlightContainerDecorator/SpotlightContainerDecorator.js
ybsung: @MikyungKim and I have no idea for naming. If 2nd-reviewer has the same opinion to rename it, please let @MikyungKim know.

ybsung:
LGTM

packages/moonstone/CHANGELOG.md
```diff
@@ -23,6 +27,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/TimePicker` to show `meridiem` correctly in all locales
+- `moonstone/Scrollable/ScrollButtons` to read out out audio guidance when button down.
```
webOS101: We should not reference non-public components in the CHANGELOG.  Reference the public components affected by the change instead.

packages/spotlight/SpotlightContainerDecorator/SpotlightContainerDecorator.js
webOS101: Maybe `stateFromProps`?

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -252,6 +260,33 @@ class ScrollButtons extends Component {
...
+		const
+			{rtl, vertical} = this.props,
+			{prevButtonDisabled, nextButtonDisabled} = this.state,
+			{keyCode, target} = ev,
+			direction = getDirection(keyCode),
+			fromNextToPrev = (vertical && direction === 'up') || (!vertical && direction === (rtl ? 'right' : 'left')),
+			fromPrevToNext = (vertical && direction === 'down') || (!vertical && direction === (rtl ? 'left' : 'right'));
+
+		// manually focus the opposite scroll button when 5way pressed
+		if ((fromNextToPrev && target === this.nextButtonNodeRef && !prevButtonDisabled) ||
```
webOS101: Do we need to do all these calculations?  It seems a bit complicated.  Should we just allow spotlight to fail to focus the opposite button when it's disabled?  Is there any other way to simplify this?

packages/moonstone/CHANGELOG.md
MikyungKim: Oh, I should've removed this line. Forgot to remove after auto-merging. Sorry!

packages/moonstone/Scrollable/ScrollButtons.js
MikyungKim: I guess we could remove the condition of seeing disabled but I'm afraid we cannot remove the rest.
`focusOnOppositeScrollButton` should be only called when the target is button node and it should be matched with the desired direction otherwise, it behaves wrong. If we remove those conditions, pressing 'left' key from 'up' button, it goes to 'down' button which is we don't want to.

packages/spotlight/SpotlightContainerDecorator/SpotlightContainerDecorator.js
MikyungKim: Well, this is not from my head so I'd like to hear from the original author, @ryanjduffy .

ryanjduffy: Might want to proactively migrate to a static `getDerivedStateFromProps` (or something similar) in prep for React 16.4. If we wanted to be safe, we could define it as a `const` as it is now and migrate it to a static member when we complete the 16.3/4 adoption.

PR #1556, opened 2018-04-10
ENYO-5134: Fix ExpandableItem from Applying Open Class When Disabled

teckliew:
Prevent applying `open` class when `disabled`

### Links
ENYO-5134 ENYO-5119

### Comments

### DIFF
--- a/packages/moonstone/ExpandableItem/ExpandableItem.js
+++ b/packages/moonstone/ExpandableItem/ExpandableItem.js
@@ -271,7 +271,7 @@ const ExpandableItemBase = kind({
 	},
 
 	computed: {
-		className: ({open, styler}) => (styler.append({open})),
+		className: ({disabled, open, styler}) => disabled ? null : (styler.append({open})),
 		label: ({label, noneText}) => (label || noneText),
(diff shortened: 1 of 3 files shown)


packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -271,7 +271,7 @@ const ExpandableItemBase = kind({
 	},
 
 	computed: {
-		className: ({open, styler}) => (styler.append({open})),
+		className: ({disabled, open, styler}) => disabled ? null : (styler.append({open})),
```
ryanjduffy: Returning `null` here will overwrite both the `kind`-managed and author-provided classes. We should continue to use `styler` but conditionally apply `open` based on the combined state of `open` and `disabled` as we do in the computed `open` prop below.

packages/moonstone/ExpandableItem/ExpandableItem.less
```diff
@@ -24,9 +24,9 @@
 	&.open {
 		.labeledItem {
 			&.auto {
-				.label {
-					display: none;
-				}
+					.label {
```
ryanjduffy: Based on the GH preview, these changes appear to be adding extra tabs and should be reverted

PR #1557, opened 2018-04-11
PLAT-52108: Accessibility: Add props for overriding inner focusable component - VirtualList, Scroller

ybsung:

There's no way to override a11y string on scrollbar scroll buttons.

### Resolution

Add Scrollable/ScrollableNative props - horizontalScrollbarNextButtonAriaLabel, horizontalScrollbarPreviousButtonAriaLabel, verticalScrollbarNextButtonAriaLabel, verticalScrollbarPreviousButtonAriaLabel to override readout string on

### Additional Considerations

### Links

PLAT-52108

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,8 +7,8 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
-- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` props 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the aria-label set on scroll buttons in the scrollbars
-- `moonstone/Panels` property `closeButtonAriaLabel` to configure the label set on application close button 
+- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` props `scrollRightAriaLabel`, `scrollLeftAriaLabel`, `scrollDownAriaLabel`, and `scrollUpAriaLabel` to configure the aria-label set on scroll buttons in the scrollbars
+- `moonstone/Panels` property `closeButtonAriaLabel` to configure the label set on application close button
 
 ### Changed
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
+- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` property 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the label set on scroll buttons in the scrollbars
 
 ### Changed
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,7 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
-- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` property 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the label set on scroll buttons in the scrollbars
+- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` property 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the aria-label set on scroll buttons in the scrollbars
 
 ### Changed
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,7 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
-- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` property 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the aria-label set on scroll buttons in the scrollbars
+- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` props 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the aria-label set on scroll buttons in the scrollbars
 
 ### Changed
 
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -122,7 +122,7 @@ class Scrollable extends Component {
 		* Sets the hint string read when focusing the next button in the horizontal scroll bar.
 		*
 		* @type {String}
-		* @default $('scroll right')
+		* @default $L('scroll right')
 		* @public
 		*/
 		horizontalScrollbarNextButtonAriaLabel: PropTypes.string,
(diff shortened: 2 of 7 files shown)


packages/moonstone/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
+- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` property 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the label set on scroll buttons in the scrollbars
```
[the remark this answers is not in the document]
ybsung: I fixed it.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -494,9 +553,26 @@ class Scrollable extends Component {
 									ref: this.initChildRef
 								})}
 							</TouchableDiv>
-							{isVerticalScrollbarVisible ? <Scrollbar {...verticalScrollbarProps} {...this.scrollbarProps} disabled={!isVerticalScrollbarVisible} /> : null}
+							{isVerticalScrollbarVisible ?
+								<Scrollbar
+									{...verticalScrollbarProps}
+									{...verticalScrollbarA11yProps}
```
Baekwoo-s: Please change the prop order alphabetically.

packages/moonstone/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
+- `moonstone/VirtualList`, `moonstone/VirtualGridList` and `moonstone/Scroller` property 'horizontalScrollbarNextButtonAriaLabel', 'horizontalScrollbarPreviousButtonAriaLabel', 'verticalScrollbarNextButtonAriaLabel', and 'verticalScrollbarPreviousButtonAriaLabel' to configure the aria-label set on scroll buttons in the scrollbars
```
0x64: Are they properties or props?

packages/moonstone/Scrollable/Scrollable.js
Baekwoo-s: I think it should be as follows.

{...this.scrollbarProps}
{...verticalScrollbarA11yProps}
{...verticalScrollbarProps}

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -95,11 +97,51 @@ class ScrollableNative extends Component {
...
 		 * @public
 		 */
-		focusableScrollbar: PropTypes.bool
+		focusableScrollbar: PropTypes.bool,
+
+		/**
+		* Sets the hint string read when focusing the next button in the horizontal scroll bar.
+		*
+		* @type {String}
+		* @default 'scroll right'
```
Baekwoo-s: Is there no need to mention $L?

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -502,7 +544,24 @@ class ScrollableNative extends Component {
 	}
 
 	render () {
-		const {focusableScrollbar, childRenderer, ...rest} = this.props;
+		const
+			{
+				focusableScrollbar,
+				childRenderer,
```
Baekwoo-s: Please change the variable order alphabetically.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -546,9 +605,26 @@ class ScrollableNative extends Component {
 									ref: this.initChildRef
 								})}
 							</TouchableDiv>
-							{isVerticalScrollbarVisible ? <Scrollbar {...verticalScrollbarProps} {...this.scrollbarProps} disabled={!isVerticalScrollbarVisible} /> : null}
+							{isVerticalScrollbarVisible ?
```
Baekwoo-s: If isVerticalScrollbarVisible is false, do you need to put arialabel in verticalScrollbarA11yProps above?

packages/moonstone/Scroller/tests/Scroller-specs.js
```diff
@@ -68,4 +68,78 @@ describe('Scroller', () => {
...
 	});
+
+	describe('Scrollbar accessibility', () => {
+		it('should set "aria-label" to previous scroll button in the horizontal scroll bar', function () {
+			const label = 'custom button aria label';
+			const subject = mount(
+				<Scroller
+					horizontalScrollbar="visible"
+					verticalScrollbar="visible"
+					horizontalScrollbarPreviousButtonAriaLabel={label}
```
Baekwoo-s: Do we have to do this in alphabetical order?
If yes, please correct it.

packages/moonstone/VirtualList/tests/VirtualList-specs.js
```diff
@@ -281,4 +281,82 @@ describe('VirtualList', () => {
...
+		it('should set "aria-label" to previous scroll button in the horizontal scrollbar', function () {
+			const label = 'custom button aria label';
+			const subject = mount(
+				<VirtualList
+					clientSize={clientSize}
+					dataSize={dataSize}
+					direction="horizontal"
+					itemRenderer={renderItem}
+					itemSize={30}
+					horizontalScrollbarPreviousButtonAriaLabel={label}
```
Baekwoo-s: Please change the prop order alphabetically.

packages/moonstone/Scrollable/ScrollableNative.js
ybsung: No. There is no DOM element to put aria-label.

packages/moonstone/Scrollable/Scrollable.js
ybsung: I don't think we need to order them alphabetically because some props could override `this.scrollbarProps`. For spread operation, we don't need to do it.

packages/moonstone/Scrollable/ScrollableNative.js
Baekwoo-s: Do we need to create a variable(horizontalScrollbarA11yProps, verticalScrollbarA11yProps) that does not apply to the DOM element?

horizontalScrollbarA11yProps = {
	nextButtonAriaLabel: horizontalScrollbarNextButtonAriaLabel,
	previousButtonAriaLabel: horizontalScrollbarPreviousButtonAriaLabel
},
verticalScrollbarA11yProps = {
	nextButtonAriaLabel: verticalScrollbarNextButtonAriaLabel,
	previousButtonAriaLabel: verticalScrollbarPreviousButtonAriaLabel
};

packages/moonstone/CHANGELOG.md
ybsung: `props` would be good.

packages/moonstone/Scrollable/Scrollable.js
Baekwoo-s: Ok, I understand this case. Thank you.

ybsung:
I fixed all issues in the comments above.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -131,7 +131,7 @@ class Scrollable extends Component {
 		* Sets the hint string read when focusing the previous button in the horizontal scroll bar.
 		*
 		* @type {String}
-		* @default 'scroll left'
+		* @default $('scroll left')
```
Baekwoo-s: Please fix $('scroll left') -> $L('scroll left')

packages/moonstone/Scrollable/ScrollableNative.js
ybsung: `isVerticalScrollbarVisible` and `isHorizontalScrollbarVisible` variables were passed as ui `Scrollable`'s `containerRenderer` parameters. That means that we need to define `horizontalScrollbarA11yProps`, `verticalScrollbarA11yProps` in  containerRenderer(). In my opinion, it would be better to define them as I did then to define them in containerRenderer().

0x64:
LGTM

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -449,7 +491,24 @@ class Scrollable extends Component {
...
+			{
+				childRenderer,
+				focusableScrollbar,
+				horizontalScrollbarNextButtonAriaLabel,
+				horizontalScrollbarPreviousButtonAriaLabel,
+				verticalScrollbarNextButtonAriaLabel,
+				verticalScrollbarPreviousButtonAriaLabel,
+				...rest
+			} = this.props,
+			horizontalScrollbarA11yProps = {
```
ryanjduffy: Let's include these inline within the JSX. In general, we try to keep all props defined within the JSX. There are some cases in which is makes sense to bundle them separately (e.g. there are props that are defined externally such as in some HOCs or props are conditionally passed) but this case causes an extra object allocation and affects readability by adding another property bag to be mentally parsed.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -95,11 +97,51 @@ class ScrollableNative extends Component {
...
+		* @default $L('scroll up')
+		* @public
+		*/
+		verticalScrollbarPreviousButtonAriaLabel: PropTypes.string
 	}
 
 	static defaultProps = {
-		focusableScrollbar: false
+		focusableScrollbar: false,
+		horizontalScrollbarNextButtonAriaLabel: $L('scroll right'),
```
ryanjduffy: `$L` shouldn't be used in `defaultProps` because it will only be evaluated at component definition time and not reflect changes to the locale. For now, I recommend include a null check inline in render. When async ilib lands, we'll likely have a different solution but that is yet to be defined.

```
nextButtonAriaLabel={horizontalScrollbarNextButtonAriaLabel == null ? $L('scroll right') : horizontalScrollbarNextButtonAriaLabel}
```

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -114,11 +116,51 @@ class Scrollable extends Component {
...
+		* @default $L('scroll up')
+		* @public
+		*/
+		verticalScrollbarPreviousButtonAriaLabel: PropTypes.string
 	}
 
 	static defaultProps = {
-		focusableScrollbar: false
+		focusableScrollbar: false,
+		horizontalScrollbarNextButtonAriaLabel: $L('scroll right'),
```
ryanjduffy: Inline `null` check here as well

seunghoh:
@ryanjduffy If I change the prop name to `scrollUpAriaLabel` without button,  isn't this confused with overriding reading text when press up button? - we read "up" when press the up button to tell user we are doing scrolling up and `scrollUpAriaLabel` looks like overriding this text for me. What do you think? It will be OK since we don't support override this sting? `scrollUpButtonAriaLabel` still looks long?

PR #1558, opened 2018-04-11
PLAT-53640: Allow to force to re-render VirtualList items if needed

ybsung:

Even though calling this.forceUpdate() app side, VirtualList items are not re-rendered because we cached them to enhance performance by blocking re-rendering them.

### Resolution

While scrolling a VirtuslList, we still use the cached items. But while updating it, we re-render the items.

### Additional Considerations

### Links

PLAT-53640

### Comments


MikyungKim:
LGTM

PR #1563, opened 2018-04-12
PLAT-49780: Optimize localized font loading performance

germboy:
Currently too much (unnecessary) work is happening when attempting to load localized fonts into the dom. Localized font definitions are being generated on render regardless of whether they are already inserted in the dom. Also we're needlessly specifying miso-museo font unicode-range definitions, which add to overall load time bloat.

### Resolution
1) Optimize font loading by only generating font definitions when necessary: when they have not been loaded already or the locale has changed.
2) Remove the miso-museo font unicode-range definition rules.

### DIFF
--- a/packages/ui/internal/localized-fonts/localized-fonts.js
+++ b/packages/ui/internal/localized-fonts/localized-fonts.js
@@ -53,13 +53,19 @@ const buildFontSet = function (fontName, fonts, strLang, bitDefault) {
 	return strOut;
 };
 
-function fontGenerator (locale) {
+function fontGenerator (locale, localeChanged) {
+	const styleId = 'localized-fonts';
+	let styleElem = document.getElementById(styleId);
+
+	if (styleElem && !localeChanged) {
+		return;
+	}
+
 	const
 		matchLang = locale.match(/\b([a-z]{2})\b/),
 		language = matchLang && matchLang[1],
 		matchReg = locale.match(/\b([A-Z]{2}|[0-9]{3})\b/),
-		region = matchReg && matchReg[1],
-		styleId = 'localized-fonts';
+		region = matchReg && matchReg[1];
 
 	let fontDefinitionCss = '';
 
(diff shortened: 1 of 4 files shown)


packages/ui/internal/localized-fonts/localized-fonts.js
```diff
@@ -53,13 +53,19 @@ const buildFontSet = function (fontName, fonts, strLang, bitDefault) {
 	return strOut;
 };
 
-function fontGenerator (locale) {
```
JayCanuck: Enact CLI directly runs `@enact/moonstone/MoonstoneDecorator/fontGenerator` directly to prerender the localized font style block in the localized prerendered index html files. That directly calls `@enact/ui/internal/localized-fonts`. So if I'm understanding this change, Enact CLI would be updated to support passing `localeChanged` argument?

packages/ui/internal/localized-fonts/localized-fonts.js
```diff
@@ -53,13 +53,19 @@ const buildFontSet = function (fontName, fonts, strLang, bitDefault) {
...
+function fontGenerator (locale, localeChanged) {
+	const styleId = 'localized-fonts';
+	let styleElem = document.getElementById(styleId);
```
JayCanuck: A `if (typeof document !== 'undefined') {` guard is required for `document` usage so it doesn't error on prerendering in Node environment.

packages/ui/internal/localized-fonts/localized-fonts.js
germboy: Pasting response from slack: CLI shouldn't have to be updated, since failing to pass a 2nd arg (`localeChanged`) - while resulting in a truthy value in the added check (!undefined = true) - will still fail the 1st check, resulting in executing the code underneath since the `<style>` tag doesn't exist. The 2nd arg is only for runtime use - to force an update to the `<style>` block when it exists.

lishammel:

PR #1566, opened 2018-04-16
PLAT-52103: Add aria label props to Picker, DatePicker, and TimePicker

Baekwoo-s:
There's no way to override a11y string on Picker, RangePicker, DatePicker, TimePicker, EditableIntegerPicker

### Resolution
Added a11y props in Picker, DatePicker, TimePicker
- Picker
  - aria-label
  - decrementAriaLabel
  - incrementAriaLabel
- DataComponentPicker
  - accessibilityHint
- DateComponentRangePicker
  - accessibilityHint
- DatePicker
  - dayAccessibilityHint
  - dayAriaLabel
  - monthAccessibilityHint
  - monthAriaLabel
  - yearAccessibilityHint
  - yearAriaLabel
- TimePicker
  - hourAccessibilityHint
  - hourAriaLabel
  - meridiemAccessibilityHint
  - meridiemAriaLabel
  - minuteAccessibilityHint
  - minuteAriaLabel

RangePicker, EditableIntegerPicker does not add props because it will pass all prop with the picker.

### Links
PLAT-52103

### Comments


MikyungKim:
Could you add unit tests for setting this props like other PRs?

packages/moonstone/internal/DateComponentPicker/DateComponentRangePicker.js
```diff
@@ -45,6 +45,15 @@ const DateComponentRangePickerBase = kind({
 		 */
 		value: PropTypes.number.isRequired,
 
+		/**
+		 * Sets the hint string read when focusing the picker.
+		 *
+		 * @type {String}
+		 * @default label
```
MikyungKim: Let's remove this default value.

packages/moonstone/internal/DateComponentPicker/DateComponentPicker.js
```diff
@@ -41,6 +41,15 @@ const DateComponentPickerBase = kind({
 		 */
 		value: PropTypes.number.isRequired,
 
+		/**
+		 * Sets the hint string read when focusing the picker.
+		 *
+		 * @type {String}
+		 * @default label
```
MikyungKim: same here

packages/moonstone/internal/DateComponentPicker/DateComponentRangePicker.js
Baekwoo-s: I removed this.

packages/moonstone/internal/DateComponentPicker/DateComponentPicker.js
Baekwoo-s: I removed this.

Baekwoo-s:
> Could you add unit tests for setting this props like other PRs?

Unit test will be implemented on the PLAT-54096 ticket.

MikyungKim:
LGTM

Baekwoo-s:
@ryanjduffy 
> On DatePicker and TimePicker, let's use xxxAccessibilityHint and xxxAriaLabel for each date component picker to maintain consistency with the accessibilityHint on Picker.

I changed the prop name of all buttons to xxxAccessibilityHint.
Because it is a prop that changes the String after {Value}.
If we use the name aria-label or xxxAriaLabel, Perhaps the App developers are misunderstood as changing from reading all string "{Value} {Hint}" when focused.

Let me know if you have any other comments.

+ If you think this change is correct, we should also change the prop name of the IncrementSlider.
(AS-IS: IncrementAriaLabel, TO-BE: IncrementAccessibilityHint)

packages/moonstone/DatePicker/DatePickerBase.js
```diff
@@ -210,13 +246,16 @@ const DatePickerBase = kind({
 	},
 
 	render: ({
+		['aria-label']: ariaLabel,
```
ryanjduffy: The brackets aren't required here; quoting the prop name is sufficient. There are a few other instances of this as well.

```
render: ({
  'aria-label': ariaLabel,
```

packages/moonstone/DatePicker/DatePickerBase.js
```diff
@@ -248,11 +288,14 @@ const DatePickerBase = kind({
 						const isLast = index === order.length - 1;
 						const isLeft = isFirst && !rtl || isLast && rtl;
 						const isRight = isFirst && rtl || isLast && !rtl;
+
 						switch (picker) {
 							case 'd':
 								return (
 									<DateComponentRangePicker
+										aria-label={ariaLabel}
```
ryanjduffy: I don't think I was clear on my intent. I wouldn't expect to use the same label for all pickers. Instead, this would use `dayAriaLabel` and we'd used `dayAccessibilityHint` (a new prop) below.

packages/moonstone/TimePicker/TimePickerBase.js
```diff
@@ -329,6 +363,7 @@ const TimePickerBase = kind({
 								case 'a':
 									return (
 										<DateComponentPicker
+											aria-label={ariaLabel}
```
ryanjduffy: Is there a reason we wouldn't support individual props (`meridiemAriaLabel` and `meridiemAccessibilityHint`) for meridiem too?

packages/moonstone/internal/Picker/Picker.js
```diff
@@ -670,9 +697,14 @@ const PickerBase = class extends React.Component {
 	}
 
 	calcButtonLabel (next, valueText) {
+		const
+			{decrementAriaLabel, incrementAriaLabel, joined} = this.props,
```
ryanjduffy: We can simplify this a bit using default values. Also moved the `$L` calls within the conditional to avoid the unnecessary function call for joined pickers.

```js
if (!this.props.joined) {
    const {decrementAriaLabel = $L('previous item'), incrementAriaLabel = $L('next item')} = this.props;

    return `${valueText} ${next ? incrementAriaLabel : decrementAriaLabel}`;
}
```

packages/moonstone/internal/Picker/Picker.js
```diff
@@ -685,9 +717,11 @@ const PickerBase = class extends React.Component {
...
 
 	calcJoinedLabel (valueText) {
-		const {orientation} = this.props;
-		const hint = orientation === 'horizontal' ? $L('change a value with left right button') : $L('change a value with up down button');
-		return `${valueText} ${hint}`;
+		const
+			{['aria-label']: ariaLabel, orientation} = this.props,
+			hint = orientation === 'horizontal' ? $L('change a value with left right button') : $L('change a value with up down button');
+
+		return `${valueText} ${(ariaLabel == null) ? hint : ariaLabel}`;
```
ryanjduffy: There's a small, hidden bug in here we should address as well. In `render()`, if the picker isn't joined `aria-label` will always be set to `null`. To avoid this, we should always call `calcJoinedLabel` (which should probably be renamed to `calcAriaLabel`) and when `joined`, run this logic and when `!joined` simply return `aria-label`.

ryanjduffy: Also, I think that if `aria-label` is set, we should use it "as is" without prepending the value. What do you think?

The resulting logic (with the above comment) would be:

```
if (!joined || ariaLabel != null) {
  return ariaLabel;
}

// existing valuText + hint logic
```

packages/moonstone/DatePicker/DatePickerBase.js
Baekwoo-s: I will apply this.

packages/moonstone/DatePicker/DatePickerBase.js
Baekwoo-s: I will apply this.

packages/moonstone/TimePicker/TimePickerBase.js
```diff
@@ -110,6 +110,24 @@ const TimePickerBase = kind({
...
 		title: PropTypes.string.isRequired,
 
+		/**
+		 * Sets the hint string read when focusing the joined picker.
+		 *
+		 * @type {String}
+		 * @default 'change a value with up down button'
+		 * @public
+		 */
+		'aria-label': PropTypes.string,
```
webOS101: This needs an `@memberof ...` to keep documented properly.

packages/moonstone/TimePicker/TimePickerBase.js
Baekwoo-s: I did't add TimePicker meridiem labels because there was no scenario to read.
I'll add meridiemAriaLabel and meridiemAccessibilityHint.

packages/moonstone/internal/Picker/Picker.js
Baekwoo-s: I will apply this.
Thanks.

packages/moonstone/internal/Picker/Picker.js
Baekwoo-s: @ryanjduffy 
That's a good point.
I changed the name from calcJoinedLabel to calcAriaLabel, and I put some of your code.
I have one question.

if (!joined || ariaLabel != null) {
  return ariaLabel;
}

In the above code, ariaLabel is unconditionally set if aria-label is not null (TimePicker, DatePicker). I don't think this is expected.

What do you think of the code below?

if (!joined && !(ariaLabel == null)) {
  return ariaLabel;
}

packages/moonstone/TimePicker/TimePickerBase.js
Baekwoo-s: I have renamed the aria-label prop of TimePickerBase to another prop.
Do you have a prop that requires @memerof to be appended?

packages/moonstone/CHANGELOG.md
```diff
@@ -7,6 +7,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Added
 
 - `moonstone/VirtualList.VirtualList` and `moonstone/VirtualList.VirtualGridList` `role="list"`
+- `moonstone/Panels` prop `closeButtonAriaLabel` to configure the aria-label set on application close button
+- `moonstone/Picker` props `aria-label` , `decrementAriaLabel`, `incrementAriaLabel` to configure the label set on each button
```
MikyungKim: please add `and` between `, `incrementAriaLabel``
Also for DatePicker and TimePicker

Baekwoo-s: I added `and` between last `PROP`

ryanjduffy:
Made a few prop naming changes which I'm going to go ahead and merge but we can revisit later if need be. In short, instead of using `xxxAccessibilityHint` which mirrored the label below the picker, I changed them to `xxxLabel` and used the same value for both the visual label and the aural label.

Also added some aria debugging CSS complements of @Djspaceg. Now, you can add `.debug.aria` to a node and its descendants will display a tooltip with the current `aria-label` and/or `aria-valuetext`. :tada:

PR #1570, opened 2018-04-17
PLAT-48639: Add focused + disabled support

teckliew:
Add focused + disabled support for `IconButton`'s used in `Scroller`, `IncrementSlider`, and `Picker`.

### Additional Considerations
Removed `DisappearSpotlightDecorator`

### Links
PLAT-48639

### Comments


teckliew:
The only cases I can think of for `DisappearSpotlightDecorator` is when apps use it to redirect spotlight after disabling a component.
So if we are reverting to supporting focused and disabled where we don't need to redirect Spotlight, I don't see where `DisappearSpotlightDecorator` is needed, at least for framework components afaik.
We could keep it for possible future uses WDYT?

germboy:
I don't see a real benefit for keeping `DisappearSpotlightDecorator` currently, since it was fairly specific specific for internal use for directing focus elsewhere when focused `Spottable` components became disabled. Since we now want to maintain focus when a component becomes disabled, I don't see much of a use for it.

As for the general implementation of the fix, I've pushed various changes to the original PR that address performance (only re-rendering when necessary instead of each time focus is gained/lost), and re-added some general `Spottable` behavior that appears to have been inadvertently removed originally.

Djspaceg:
I was thinking about cases where a spotted component completely disappears, like it's removed from the DOM. If that's not already implemented, maybe it could use `DisappearSpotlightDecorator` to handle that? Just a thought.

packages/spotlight/Spottable/Spottable.js
```diff
@@ -176,58 +174,26 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
...
-
-					if (!Spotlight.getPointerMode()) {
-						getContainersForNode(this.node).reverse().reduce((found, id) => {
-							return found || Spotlight.focus(id);
-						}, false);
-					}
-				}
-			}
-
-			// if the component became enabled, notify spotlight to enable restoring "lost" focus
```
ryanjduffy: I don't think we can remove this block of code which is handling the inverse (changing from disabled to enabled) of the change proposed. Doing so would likely cause regressions of #1254 and #1023.

packages/spotlight/Spottable/Spottable.js
```diff
@@ -176,58 +174,26 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
...
+			};
 		}
 
-		componentDidMount () {
-			// eslint-disable-next-line react/no-find-dom-node
-			this.node = ReactDOM.findDOMNode(this);
+		componentWillReceiveProps () {
+			this.setState((state, nextProps) => {
+				const spottableDisabled = this.isFocused && nextProps.disabled || nextProps.spotlightDisabled;
+				return !this.state.spottableDisabled && spottableDisabled ? {spottableDisabled} : null;
```
ryanjduffy: Should be accessing `state` rather than `this.state` within a `setState` callback

packages/moonstone/Picker/SpottablePicker.js
```diff
@@ -11,18 +11,21 @@ const SpottablePicker = hoc(null, (config, Wrapped) => {
...
+			onSpotlightDisappear: PropTypes.func,
 			onSpotlightDown: PropTypes.func,
 			onSpotlightLeft: PropTypes.func,
 			onSpotlightRight: PropTypes.func,
 			onSpotlightUp: PropTypes.func
 		}
 
 		render () {
-			const {onSpotlightDown, onSpotlightLeft, onSpotlightRight, onSpotlightUp, ...rest} = this.props;
+			const {onSpotlightDisappear, onSpotlightDown, onSpotlightLeft, onSpotlightRight, onSpotlightUp, ...rest} = this.props;
```
ryanjduffy: I don't think it makes sense to forward this on to the buttons. The only case in which `onSpotlightDisappear` fires is on unmount now. In the case of a joined picker, this will be handled correctly by Spottable. In the case of separate picker buttons, we should generate the `onSpotlightDisappear` at the picker boundary rather than emitting two events (one from each button). We likely need similar logic for IncrementSlider as well.

packages/spotlight/Spottable/Spottable.js
```diff
@@ -288,9 +254,14 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
...
 		}
 
+		isActionable = () => {
+			return !this.props.disabled && !this.props.spotlightDisabled;
+		}
+
 		handle = handle.bind(this)
 
 		handleKeyDown = this.handle(
+			this.isActionable,
```
ryanjduffy: Should we still allow the spotlight directional events (e.g. `onSpotlightLeft`) when focused but disabled? Seems feasible that consumers would want the same support for managing 5-way navigation from a component regardless of its `disabled` state.

Djspaceg:
> it's not apparent that the components are still focused but disabled; they just look disabled.

I think I am not understanding what you're saying. It's not apparent that items are disabled+spotted when the contents are translucent while the background is full opacity? I could understand if you'd said "they just look focused/spotted", but not the other way around. Could you clarify?

ryanjduffy:
I meant the other components don't appear focused and disabled. The IconButtons for scroller/picker/slider look fine but things like `Item` just look disabled because the spotlight background color has its opacity reduced.

Djspaceg:
Well there's yer problem. You expected everything's style to be updated when we only got specs for one component.

I propose this: 
A. We add a simple change to the appearance of `.spottable[disabled]:focus` to _*generaically*_ support this new format of disabled+focused, which will cause visual regressions in customized components like `FormCheckbox`, and then go back later and update any component we can find with what we assume the component should look like.
B. We ignore _other types_ of focused+disabled elements for the time being, until we get real-world bugs to fix them. Moonstone doesn't internally invoke `disabled` on any components other than `Button` derivatives, so it _could_ be unlikely that any actually show up.

PR #1571, opened 2018-04-17
ENYO-5459: Disable lockfile generation for packages

JayCanuck:
* Currently we have no lockfiles. Historically, NPM lockfiles have been buggy and we've disabled globally on our systems via [`npm config set package-lock false`](https://docs.npmjs.com/misc/config#package-lock).
* With NPM 5.8, lockfiles appear stable now and while we could go 'all in' and add lockfiles to each package, it feels more apt to test the waters with select lockfiles (in cli, docs, and a root one here).

### Resolution
* Add a root package-lock.json
* Update Lerna config to disable package lock usage on the `./packages`

### Additional Considerations
* This only affects users who have the `package-lock` setting `true`. Anybody who has disabled it (via the command mentioned earlier) will see zero change.

### Links
* https://docs.npmjs.com/misc/config#package-lock


lishammel:
@JayCanuck  - does this need a jira to get it merged?

PR #1573, opened 2018-04-18
PLAT-53627: Consolidate ProgressBarTooltip positioning props

ryanjduffy:

We have two props to control positioning a `ProgressBarTooltip` which could be consolidated into one.

### Resolution

* Consolidated into `side` which supports `before`, `after`, `left`, and `right`
* Converted `tooltip` to be either `bool`/`element`/`func` for consistency with the new Slider rewrite


packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -1,12 +1,34 @@
...
 import Tooltip from '../TooltipDecorator/Tooltip';
 
 import css from './ProgressBarTooltip.less';
 
+const memoizedPercentFormatter = memoize((/* locale */) => new NumFmt({
+	type: 'percentage',
+	useNative: false
+}));
+
+const getSide = (orientation, side) => {
```
teckliew: Should we also have a guard like `getOrientation` for `orientation` as well? When `orientation` gets  a non recognizable value, the tooltip doesn't position correctly.

ryanjduffy: `PropTypes` should be a sufficient guard for that. In that light, we could (should?) move this logic into a custom PropType function instead which would make them consistent.

packages/moonstone/ProgressBar/ProgressBar.js
```diff
@@ -69,45 +71,37 @@ const ProgressBarBase = kind({
...
 		 */
-		tooltipSide: PropTypes.string
+		tooltip: PropTypes.oneOfType([PropTypes.bool, PropTypes.object, PropTypes.func])
 	},
 
 	defaultProps: {
-		tooltip: false,
-		tooltipSide: 'before',
-		tooltipForceSide: false
+		tooltip: false
```
teckliew: Should we have `progress` default to 0?

teckliew:
Looking awesome but a few loose ends:
Currently tooltip positions outside of the `ProgressBar` when `progress` is out of range; should we handle the tooltip position when `progress` is less than min and larger than max?
Not sure if the off by one error is in this PR or another one but I noticed that `IncrementSlider` skips `29%` and `58%`; and repeats `56%` twice

teckliew:
About the `IncrementSlider`, it only reproduces in sampler for me 🤔

packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -94,30 +117,48 @@ const ProgressBarTooltipBase = kind({
 	contextTypes,
 
 	computed: {
-		className: ({forceSide, knobAfterMidpoint, orientation, side, styler}) => styler.append(orientation, {ignoreLocale: forceSide, knobAfterMidpoint}, side),
-		arrowAnchor: ({knobAfterMidpoint, orientation}) => {
+		children: ({children, proportion, percent}) => {
+			if (percent) {
+				const formatter = memoizedPercentFormatter(ilib.getLocale());
+
+				return formatter.format(Math.floor(proportion * 100));
```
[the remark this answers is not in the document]
ryanjduffy: Good catch. Running into some JS precision problems. Switching to `Math.round()` is a safer alternative.

ryanjduffy:
Not too worried about `progress` being out of bounds. We already have dev mode warnings for the range in `ui/ProgressBar` and that's probably sufficient for now

ryanjduffy:
Rebased this on `develop` since it previously included all the slider refactor commits.

PR #1575, opened 2018-04-19
ENYO-5186: Fix memoize to forward all args to provided function

ryanjduffy:
While experimenting with using memoized event handlers with `core/kind`, I discovered that `core/util.memoize` doesn't send all args to provided function.

### Resolution

* Fixed `core/util.memoize` pass all args to provided function
* Added test for the above
* Changed the generated handlers from `core/kind` to return the result of the handler. This won't affect the majority of cases because returned values are ignored by React (afaik) but will enable the memoized handlers with which I was experimenting.

### DIFF
--- a/packages/core/util/util.js
+++ b/packages/core/util/util.js
@@ -191,7 +191,7 @@ const memoize = (fn) => {
 		if (n in cache) {
 			return cache[n];
 		} else {
-			let result = fn(n);
+			let result = fn(...args);
 			cache[n] = result;
 			return result;
 		}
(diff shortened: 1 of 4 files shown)


packages/core/util/util.js
```diff
@@ -191,7 +191,7 @@ const memoize = (fn) => {
 		if (n in cache) {
```
viodragon2: should we use a different value for the `key` instead of `args[0]`? maybe `JSON.stringify(args)`?

ryanjduffy: I don't think we want to go so far to stringify the args. That could be an expensive operation, particularly if used with `handlers` which receive the props and context objects. We may want a more robust memoization function (likely using a third party module) in the future but I'm okay with this implementation for now.

PR #1583, opened 2018-04-23
PLAT-49780: Optimize localized font loading performance

germboy:
Currently too much (unnecessary) work is happening when attempting to load localized fonts into the dom. Localized font definitions are being generated on render regardless of whether they are already inserted in the dom. Also we're needlessly specifying miso-museo font unicode-range definitions, which add to overall load time bloat.

### Resolution
1) Optimize font loading by only generating font definitions when necessary:
-a) generate required static/unchanging font css definitions when they have not been loaded already
-b) generate locale-specific font css definitions only when the current locale needs them
2) Remove the miso-museo font unicode-range definition rules.


Djspaceg:
How do you feel about adding some documentation to the fontGenerator component on what's happening here and why?

PR #1584, opened 2018-04-23
ENYO-5184: Restore focus after mounting VirtualList

ybsung:

When `componentDidUpdate` in `VirtualListBase` is called, we tried to restore focus in `VirtualList`. But if the number of items are small, `componentDidUpdate` was not called. So focus was not restored.

### Resolution

When `componentDidMount` in `VirtualListBase` is called, I tried to restore focus after generating all DOM elements.

### Additional Considerations

The solution in this PR depends on timing. If VirtualList renders asynchronously. the approach in this PR would have a problem. But currently, there is no alternative solution. So I hope that we could merge the PR first. Then we could find the best solution for it later.

### Links

ENYO-5184

### Comments


0x64:
It looks not so good to call `timeout` in `componentDidMount` regarding complex life cycle of lists, but I agree that it is our best for now. We cannot gaurantee `componentUpdate` cycle to be called at the first rendering or not for various scenarios.

0x64:
LGTM

MikyungKim:
I think we should revisit this issue later.

MikyungKim:
LGTM..

PR #1585, opened 2018-04-23
PLAT-47085: Change moonstone/Input to lock pointer when focused

ryanjduffy:

New UX requirement that in addition to pausing Spotlight, the user should not be able to click on other components but still select and position the carat within the input.

### Resolution

Added a pointer lock submodule to `moonstone/Input` which uses the capture phase and some module-scoped state to prevent click and "tap" actions when the input has focus.

### DIFF
--- a/packages/moonstone/Input/pointer.js
+++ b/packages/moonstone/Input/pointer.js
@@ -1,0 +1,46 @@
+import {handle, preventDefault, returnsTrue, stop} from '@enact/core/handle';
+
+let active;
+let capturedDown = false;
+
+// if there is a focused input (there should be) and the target is not the input
+const shouldCapture = ({target}) => active && target !== active;
+const isCapturing = () => capturedDown && active;
+const setCapturing = (capturing) => returnsTrue(() => (capturedDown = capturing));
+
+const handlePointerDown = handle(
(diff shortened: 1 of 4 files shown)


packages/moonstone/Input/pointer.js
```diff
@@ -0,0 +1,46 @@
...
+	stop,						// prevent the click event from propagating
+	setCapturing(false),		// clear the capturing flag
+	() => active.blur()			// and blur the active node
+);
+
+// Lock the pointer from emitting click events until released
+const lockPointer = (target) => {
+	active = target;
+	document.addEventListener('mousedown', handlePointerDown, {capture: true});
+	document.addEventListener('touchstart', handlePointerDown, {capture: true});
```
germboy: Since we're now preventing pointer/mouse events when focusing an `Input`, I feel it may make sense to prevent `mouseup`/`touchend` events as well. It might feel inconsistent that those events can still be fired on background elements while their originating events (`mousedown`/`touchstart`) are prevented. Thoughts?

germboy:
We're deciding to mark the behavior "where a spottable background component does not gain focus while hovered when the VKB is dismissed" as a separate issue to be fixed later.

PR #1587, opened 2018-04-24
ENYO-5141: VirtualGridList: Fix showing blank when direction prop changed

MikyungKim:
VirtualGridList shows blank when `direction` prop changed after scroll.

### Resolution
This issue only occurs in "JS" version of VirtualLists.
After syncing the DOM structure of "JS" list with "Native" list, we made contentRef for "JS" list and transform it when the scroll occurs.
But when the `direction` prop changed after scroll position changed, we didn't reset the transform value for contentRef.
So added reset transform code for contentRef.

### Additional Considerations

### Links
ENYO-5141

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


0x64:
LGTM

PR #1588, opened 2018-04-24
PLAT-49775: Add DayPicker dayNameLength Prop

teckliew:
Add`dayNameLength` prop that supports different length for names of days in `DaySelector` and `DayPicker`'s label.

### Resolution

### Additional Considerations

### Links
PLAT-49775

### Comments


packages/moonstone/DayPicker/DayPicker.js
```diff
@@ -278,14 +156,12 @@ const DayPickerBase = class extends React.Component {
...
-const DayPicker = Pure(
-	Expandable(
-		Changeable(
-			{prop: 'selected', change: 'onSelect'},
-			Subscription(
-				{channels: ['i18n'], mapMessageToProps: (channel, {locale}) => ({locale})},
-				DayPickerBase
-			)
+
+const DayPicker = Expandable(
```
rundmt: Looks like you removed Pure. Was this intentional?

teckliew: Nope. Added back in!

rundmt:
LGTM

ryanjduffy:
This is good work, @teckliew! I appreciate that you took the opportunity while adding a feature to do some refactoring to clean up the code base and reduce duplication.

I made a couple changes I wanted to call out:

* I changed `DayPicker` to only use `DaySelectorDecorator.js` rather than the entire HOC stack because it resulted in fewer layers due to removing an extra instance of `Pure`. The duplication introduced (a configured `Changeable`) was pretty minor and worth the tradeoff, imo.
* I adopted the `DayPickerDecorator` convention with `compose` we've used elsewhere but did not add it as an export at this time. When we revisit the expandable UI migration, we can address all of these components together.
* I moved the i18n subscription into `DaySelectorDecorator` to reduce the number of layers plus add dynamic locale changing to `DaySelector` which it previously lacked.
* I refactored `initIlib()` to generate a state object rather than setting a bunch of member variables. Member variables are generally bad anyway because they can be set out of band of state updates and we needed a state change to reflect the new locale so it bundled nicely.

Please give it a look and verify everything still behaves as you expected and we'll get it merged.

teckliew:
Everything looks 👌 to go. Definitely looks a lot better.

PR #1590, opened 2018-04-25
ENYO-5197: Fix slider tooltip visibility condition

viodragon2:
`moonstone/Slider`'s tooltip visible condition is dictated by `focused` prop. `moonstone/VideoPlayer.MediaSlider` which uses `moonstone/Slider` as a base has its own visibility condition but gets ignored. Fixed it to only apply `focused` condition for the default tooltip.

### DIFF
--- a/packages/moonstone/Slider/Slider.js
+++ b/packages/moonstone/Slider/Slider.js
@@ -251,12 +251,13 @@ const SliderBase = kind({
 			activateOnFocus,
 			active
 		}),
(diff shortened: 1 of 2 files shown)


teckliew:
:LGTM: Just needs to update the unit test because it looks like `'should set the tooltip to visible when focused'` didn't catch the regression.

ryanjduffy:
Closing in favor of https://github.com/enactjs/enact/commit/8541498adabb1d32a8137df0d34b8e7c3217983a

PR #1592, opened 2018-04-25
ENYO-5089: Skip to handle a key when changing from pointer mode to 5 way key mode in Scrollable

ybsung:

If pressing a page up or down key in VirtualList with pointer mode, the focus in it changes because `pointerHide` event (keyCode: 1537) and page up/down key comes together and the VirtualList handle it.
If pressing any key in pointer mode, we should not handle any key except `pointerHide` event.

### Resolution

As we did in spotlight.js, I skip to handle a page up/down key when changing from pointer mode to 5 way key mode in Scrollable.

### Additional Considerations

Even though I did not add `onKeyUp` event handler, the handler called. So I did not do it additionally.

### Links

ENYO-5089

### Comments


ybsung:
#1629 PR was merged instead of this PR.

PR #1594, opened 2018-04-25
ENYO-5196 Rename Tooltip exports

webOS101:

Renamed Tooltip exports from `ProgressBar` and the sliders

### Resolution

### Additional Considerations
I am not super happy about the 'name' of `ProgressBar/Tooltip` being `'ProgressBarTooltip'`.
I considered other names but recall that `/` and `.` don't work.  Considered `-`, which is OK
but not great.  Cannot set to `'Tooltip'` as this is ambiguous with its internal `Tooltip` component.

### Links
ENYO-5196

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -19,15 +19,17 @@ The following is a curated list of changes in the Enact moonstone module, newest
 - `moonstone/Popup` property `closeButtonAriaLabel` to configure the label set on popup close button
 - `moonstone/MediaOverlay` component
 
+### Changed
+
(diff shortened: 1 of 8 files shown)


Djspaceg:
Please do not rename a component-specific extension of a component (`ProgressBarTooltip`) to match the standard component's name (`Tooltip`). Having two different exports that perform different functions be named the same thing adds unnecessary confusion.

ryanjduffy:
While I agree that it's redundant, the change would introduce a scenario in which an author would have to rename an import to use both a "regular" tooltip and a progress bar tooltip in the same module. That, coupled with the potential for confusion that @Djspaceg pointed out, outweighs the simplification that the name change brings.

PR #1596, opened 2018-04-26
ENYO-5201: Update ui/Transition docs

Djspaceg:
The docs sorely need updating for `Transition`
Also, there's no way to customize the transition duration.

### Resolution
Fixed the docs
and
Updated duration value to accept not only the preset values, but also custom values, like numbers and strings.

### DIFF
--- a/packages/ui/Transition/Transition.js
+++ b/packages/ui/Transition/Transition.js
@@ -29,7 +29,7 @@ const forwardOnShow = forward('onShow');
 const forwardOnHide = forward('onHide');
 
 /**
- * The stateless structure of the component, in case you want toprovide all of the state yourself.
+ * The stateless structure of the component, in case you want to provide all of the state yourself.
  * In general, you'll probably want to use the [stateful version]{@link ui/Transition.Transition}.
  *
  * @class TransitionBase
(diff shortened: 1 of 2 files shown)


packages/ui/Transition/Transition.js
```diff
@@ -19,9 +29,8 @@ const forwardOnShow = forward('onShow');
 const forwardOnHide = forward('onHide');
 
 /**
- * {@link ui/Transition.TransitionBase} is a stateless component that allows for applying
- * transitions to its child items via configurable properties and events. In general, you want to
- * use the stateful version, {@link ui/Transition.Transition}.
+ * A minimally styled toggle item without any behavior, ripe for extension.
```
rundmt: Looks like we have some toggle item cruft here.

packages/ui/Transition/Transition.js
```diff
@@ -177,6 +204,11 @@ const TransitionBase = kind({
 				if (visible && (direction === 'up' || direction === 'down')) {
 					style.height = clipHeight;
 				}
+				// If duration isn't a know named string, assume it is a CSS duration value
```
rundmt: I think this is either an un-needed "know" or should possibly be "known"

PR #1597, opened 2018-04-26
PLAT-49776: Allow more flexible usage of Notification and SwitchItem

Djspaceg:
`Notification` requires the `buttons` prop, which may not be necessary in all cases. Requiring it adds unnecessary restriction.

`SwitchItem` is restrictive in its implementation and should be expanded to support other use cases.

### Resolution
`Notification` no longer requires `buttons` and `SwitchItem` has a default-assigned, but overridable, `Switch` component to allow for more flexible usage.

### DIFF
--- a/packages/moonstone/SwitchItem/SwitchItem.js
+++ b/packages/moonstone/SwitchItem/SwitchItem.js
@@ -39,7 +39,12 @@ const SwitchItemBase = kind({
 		 * @type {Object}
 		 * @public
 		 */
-		css: PropTypes.object
+		css: PropTypes.object,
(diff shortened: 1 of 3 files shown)


packages/moonstone/SwitchItem/SwitchItem.js
```diff
@@ -39,7 +39,12 @@ const SwitchItemBase = kind({
 		 * @type {Object}
 		 * @public
 		 */
-		css: PropTypes.object
+		css: PropTypes.object,
+		iconComponent: PropTypes.object
```
viodragon2: Add docs here

PR #1598, opened 2018-04-26
ENYO-5039:  Add flag to ignore to scroll on focus when jumping

ybsung:

When jumping over the disabled items in VirtualList, the focus in it gone.

### Resolution

If calling `setState` or `forceUpdate` app side while jumping, the focus information that we have to give gone because we get focus, scroll the list, then initialize the focus information. So for the case, we add a flag  not to scroll on focus

### Additional Considerations

### Links

ENYO-5039

### Comments

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -186,7 +186,7 @@ class ScrollableBase extends Component {
 	}
 
 	componentDidUpdate () {
-		if (this.uiRef.scrollToInfo === null && this.childRef.ignoreToScrollOnFocus === false) {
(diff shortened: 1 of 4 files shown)


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -186,7 +186,7 @@ class ScrollableBase extends Component {
 	}
 
 	componentDidUpdate () {
-		if (this.uiRef.scrollToInfo === null) {
+		if (this.uiRef.scrollToInfo === null && this.childRef.ignoreToScrollOnFocus === false) {
```
MikyungKim: How about Scroller case?

ybsung: I updated the condition for Scroller

MikyungKim:
LGTM

PR #1607, opened 2018-05-01
ENYO-5212: Prevent Spottable Disabled OnClick

teckliew:
Check for `disabled` and `isHovered`

### Links
ENYO-5212

### Comments

### DIFF
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -316,16 +316,16 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		handleFocus = (ev) => {
-			if (!this.props.disabled) {
-				if (ev.currentTarget === ev.target) {
-					this.isFocused = true;
-				}
+			if (this.props.disabled) return;
 
-				if (Spotlight.isMuted(ev.target)) {
-					ev.stopPropagation();
-				} else {
-					forward('onFocus', ev, this.props);
-				}
+			if (ev.currentTarget === ev.target) {
+				this.isFocused = true;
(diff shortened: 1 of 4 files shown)


teckliew:
I thought it would be a universal case to cover for `disabled` and `isHovered`, but I agree it'll be safer to limit the scope of this fix to checking for `disabled` in `SliderBehaviorDecorator`.

viodragon2:
I just realized this problem can happen to all `disabled` components. When a disabled component rerenders, then `Spottable`'s internal state `isFocused` will be updated to `true` which in turn will add spottable class and hightlight at render time. I wonder if it's better to stop propagation in `onFocus` from `Spottable`?

ryanjduffy:
Good analysis, @viodragon2! Appears that the exposure is limited to components that handle `onFocus` to cause a state change and therefore a prop change to Spottable triggering its state change. The new slider is, of course, one of those instances: SliderBehaviorDecorator handles `onFocus` to pass the focused state down to show the tooltip.

packages/spotlight/Spottable/Spottable.js
```diff
@@ -316,14 +316,16 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		handleFocus = (ev) => {
-			if (ev.currentTarget === ev.target) {
-				this.isFocused = true;
-			}
+			if (!this.props.disabled) {
```
ryanjduffy: Let's update this to be an early bail out to minimize the diff and keep the code as readable as possible.

```js
handleFocus = (ev) => {
  if (this.props.disabled) return;
  ...
}
```

PR #1608, opened 2018-05-01
ENYO-5221: Automatic Image src swapping during resolution change

Djspaceg:
An image on the screen doesn't update when the screen is resized and new imagery should be swapped in for the source. A refresh is required.

### Resolution
Images now automatically detect when a screen size change occurs and re-renders only if a new image source (src) is available and appropriate.

### Additional Considerations
The Sample was also updated to be more helpful. It now has a border and uses appropriate scaling of images (1x,2x,4x -> 2x,3x,6x) to more accurately represent the relative scale between 720, 1080, and 4k

### DIFF
--- a/packages/moonstone/Image/Image.js
+++ b/packages/moonstone/Image/Image.js
@@ -98,11 +98,15 @@ const ResponsiveImageDecorator = hoc((config, Wrapped) => {
 		}
 
 		handleResize = () => {
-			const src = selectSrc(this.props.src);
-			if (src !== this.state.src) {
+			this.setState((state, props) => {
+				const src = selectSrc(props.src);
 				// Trigger a render and save the currently selected src for later comparisons
-				this.setState({src});
-			}
+				if (src !== state.src) {
+					return {src};
+				}
+
+				return null;
+			});
 		}
 
 		render () {
--- a/packages/moonstone/Image/Image.js
+++ b/packages/moonstone/Image/Image.js
@@ -11,8 +11,10 @@
  */
 
 import kind from '@enact/core/kind';
+import hoc from '@enact/core/hoc';
 import UiImage from '@enact/ui/Image';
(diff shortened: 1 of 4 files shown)


packages/moonstone/Image/Image.js
```diff
@@ -62,6 +64,53 @@ const ImageBase = kind({
...
+		componentDidMount () {
+			window.addEventListener('resize', this.handleResize);
+		}
+
+		componentWillUnmount () {
+			window.removeEventListener('resize', this.handleResize);
+		}
+
+		handleResize = () => {
+			const src = selectSrc(this.props.src);
```
ryanjduffy: Since resize events can happen repeatedly, I suggest rewriting this to use the functional setState.

```js
this.setState((state, props) => {
  const src = selectSrc(props.src);
  if (src != state.src) {
    return {src};
  }

  return null;
});
```

Djspaceg: I like it. Done.

packages/moonstone/Image/Image.js
```diff
@@ -62,6 +64,57 @@ const ImageBase = kind({
...
+// This induces a render when there is a screen size change that has a corosponding image src value
+// associated with the new screen size. The render is kicked off by remembering the new image src.
+//
+// This hoc could (should) be rewritten at a later time to use a smarter context API and callbacks,
+// or something like pub/sub; each of which would be hooked together from the resolution.js that
+// would coordinate any screen size/orientation changes and emit events from there.
+//
+// This is ripe for refactoring, and could probably move into UI to be generalized, but that's for
+// another time. -B 2018-05-01
+const ResponsiveImageDecorator = hoc((config, Wrapped) => {
```
webOS101: Should this same decorator be applied to `GridlistImageItem`?

Djspaceg: I suppose it could.

PR #1611, opened 2018-05-02
ENYO-5228: Fix to scroll properly with all disabled items in VirtualList

ybsung:

Even though pressing a page up/down key in VirtualList with all disabled items, the list does not scroll.

### Resolution

We tried to cache the value of the scroll distance when mounting a VirtualList.
But at that time after updating React version, the list's client height equals 0 when `componentDidMount` called, then the cached value also equals 0 and finally the list does not scroll.

Sometimes the client size and scroll direction could be changed dynamically. So I think that it would be better not to cache it instead of checking if they are changed or not.

### Additional Considerations

### Links

ENYO-5228

### Comments

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -596,8 +596,7 @@ class ScrollableBase extends Component {
 		const
 			bounds = this.getScrollBounds(),
 			canScrollVertically = this.canScrollVertically(bounds),
-			direction = isPageUp(keyCode) ? -1 : 1,
-			pageDistance = direction * (canScrollVertically ? bounds.clientHeight : bounds.clientWidth) * paginationPageMultiplier;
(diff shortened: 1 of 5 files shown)


ybsung:
I'll update `CHANGELOG.md` later after being reviewed.

0x64:
LGTM

packages/ui/Scrollable/Scrollable.js
```diff
@@ -600,7 +596,8 @@ class ScrollableBase extends Component {
 		const
 			bounds = this.getScrollBounds(),
 			canScrollVertically = this.canScrollVertically(bounds),
-			pageDistance = isPageUp(keyCode) ? (this.pageDistance * -1) : this.pageDistance;
+			direction = isPageUp(keyCode) ? -1 : 1,
```
MikyungKim: Let's not make `direction` here.

MikyungKim:
Please add ChangeLog.

MikyungKim:
LGTM

PR #1615, opened 2018-05-03
ENYO-5210: Fix Marquee to always animate when marqueeOn="render"

viodragon2:
`Marquee` stops the animation when some of the props value updates, however it should always marquee if `marqueeOn` value is set to `render`. 
Updated condition to always start the animation for `marqueeOn="render"` and restart when `stop` is called.

### Additional consideration
I'm not 100% sure if marquee should restart when `forceDirection` is updated because of the rare chance of apps needing to change at runtime. Given that, some props change (e.g. `marqueeDelay`, `marqueeResetDelay`) won't update the animation state even with the fix. Should we check for them as well?

In regards to recent react 16.3.x update, we should consider refactoring animation check logic in `componentWillReceiveProps` to  `componentDidUpdate`.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,6 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/Item` to use its natural width rather than imposing a 100% width allowing inline Items to be the correct width.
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Tacked on a commit to remove `forceDirection` when `marqueeDisabled`. We missed it and it emits a invalid prop warning.

PR #1616, opened 2018-05-03
ENYO-5109: Fix to correctly reset marquee animation when children updates

viodragon2:
`MarqueeDecorator` resets animation when `children` is updated which was checked by strict equality operator (`===`). This won't work for array type children. Proposing it to compare using `shallowEqual` to handle the case.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,8 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/Item` to use its natural width rather than imposing a 100% width allowing inline Items to be the correct width.
+- `ui/Item` to use its natural width rather than imposing a 100% width allowing inline Items to be the correct width
+- `ui/MarqueeDecorator` to correctly reset animation when children updates
 
 ## [2.0.0-beta.1] - 2018-04-29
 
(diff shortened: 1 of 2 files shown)


ryanjduffy:
How does this reconcile with #1449 which removed our custom `childrenEquals` method which was trying to solve the same (class of) problem (iirc)?

@germboy - any thoughts on this one?

viodragon2:
Doesn't it only resolve the string type children, though? `ui/ToggleItem` technically takes `node` type children, so it would fail to check if it contained `` or `<span>` as part of the text.

ryanjduffy:
That's a good point. What appears to allow this change to work is that React will return the same element if its props haven't changed. As a result, it passes `shallowEqual` even though they both objects because they're the *same* (strictly equal) objects. This might be change from prior versions of React.

For the common cases of Marquee, this is probably okay. Ultimately, `MarqueeDecorator` can't know from props if it needs to revalidate its bounds. A change in children doesn't necessarily equate to a change in dimensions. I think the only failsafe approach would be a combination of remeasuring on every update along with an optional method to trigger remeasures when wrapped component changes its state resulting in a change in dimensions.

We could investigate such a design but it's not necessary for this ticket. Would like to get @germboy's perspective on this before moving forward to validate we've avoid regressions.

PR #1617, opened 2018-05-03
ENYO-5143: IconButton is now visually customizable

Djspaceg:
IconButton isn't visually customizable the way Button is, which is _just wrong_.

### Resolution
Added ComponentOverride to ui/IconButton so moonstone/IconButton could send its `css` into the Button component for customization. Also had to include the two customizable classes in the moonstone/IconButton.less file so they're available for override externally.

### DIFF
--- a/packages/ui/IconButton/IconButton.js
+++ b/packages/ui/IconButton/IconButton.js
@@ -15,6 +15,7 @@ import PropTypes from 'prop-types';
 import React from 'react';
 
 import Touchable from '../Touchable';
+import ComponentOverride from '../ComponentOverride';
 
 import componentCss from './IconButton.less';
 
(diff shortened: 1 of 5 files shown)


packages/ui/IconButton/IconButton.js
```diff
@@ -156,11 +157,17 @@ const IconButtonBase = kind({
 			children = null;
 		}
 
+		console.log('css:', css);
```
ryanjduffy: (oops)

packages/ui/IconButton/IconButton.js
```diff
@@ -156,11 +157,17 @@ const IconButtonBase = kind({
...
 		return (
-			<Button {...rest} small={small} minWidth={false}>
+			<ComponentOverride
```
ryanjduffy: We need to update the prop type to support objects as well now that we've added `ComponentOverride`.

ryanjduffy:
Went ahead and made the changes since they were so minor. :lgtm:

Djspaceg:
Thanks for the fix! I didn't notice this till just now.

PR #1618, opened 2018-05-03
ENYO-4914: Fix Scroller bounds when an ExpandableItem is opens

rundmt:
When the scrollbar updates on an expandable item the height doesn't adjust properly.

### Resolution
We can forceUpdate once we know a scrollbar appears in order to recalculate the measurements properly.

### Links
ENYO-4914

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -385,10 +385,6 @@ class ScrollableBase extends Component {
 			this.updateScrollbars();
 		}
 
-		if (isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible) {
-			this.forceUpdate();
-		}
-
 		if (this.scrollToInfo !== null) {
 			if (!this.deferScrollTo) {
 				this.scrollTo(this.scrollToInfo);
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -375,15 +375,19 @@ class ScrollableBase extends Component {
 		this.clampScrollPosition();
 
 		this.addEventListeners();
+		console.log('did up', isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible);
 		if (
 			hasDataSizeChanged === false &&
 			(isHorizontalScrollbarVisible && !prevState.isHorizontalScrollbarVisible || isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible)
 		) {
 			this.deferScrollTo = false;
 			this.isUpdatedScrollThumb = this.updateScrollThumbSize();
-			this.enqueueForceUpdate();
 		} else {
 			this.updateScrollbars();
+		}
+
+		if (isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible) {
+			this.enqueueForceUpdate();
 		}
 
 		if (this.scrollToInfo !== null) {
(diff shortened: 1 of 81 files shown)


packages/ui/Scrollable/Scrollable.js
```diff
@@ -388,6 +388,10 @@ class ScrollableBase extends Component {
 			this.updateScrollbars();
 		}
 
+		if (isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible) {
```
ybsung: In my opinion, at least we need to check if it is for Scroller or not because all items in a VirtualList could be re-rendered again by calling `forceUpdate`, then it make VirtualList rendering performance lower.

CC: @MikyungKim , @0x64

rundmt: Yeah I've been thinking about this as well. That's why this is just a placeholder PR. I'm still trying to figure out the best way to handle this. 

I'm not as familiar with `VirtualList` real world implementations, but VirtualList may not be hit as badly in normal scenarios. Most `VirtualList`'s would have a `ScrollBar` if it had enough items. So I assume that it would only be in a rare case.

ybsung: @rundmt 
When launching enact-internal-samples/qa-virtualgridlistnative, 53 items in the VirtualGridList are rendered. Then those items are rendered again by calling your `forceUpdate()` in your PR.

When rendering a VirtualList at the first time, we do not add a vertical or horizontal scroll bar. Then we render it again if we decide that we need any scroll bar because the item layout in VirtualGridList should be updated not to overlap the scrollbars and the items. At that time `componeneDidUpdate` function is called. If calling `forceUpdate()` in it, all items will be rendered again. All VirtualList and VirtualGridList works with this process when launching. So I would like to say that it is not in a rare case.

So that is the reason why I want to call `forceUpdate()` if it is a Scroller. By doing it, we could prevent increasing the VirtualList launch time.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -382,6 +382,10 @@ class ScrollableBaseNative extends Component {
 			this.updateScrollbars();
 		}
 
+		if (isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible) {
+			this.forceUpdate();
+		}
+
```
ybsung: Do we need this code in ui/Scrollable/Scrollable.js?

rundmt: Whoops forgot to delete this.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -769,6 +769,7 @@ const VirtualListBaseFactory = (type) => {
 			delete rest.rtl;
 			delete rest.spacing;
 			delete rest.updateStatesAndBounds;
+			delete rest.isVerticalScrollbarVisible;
```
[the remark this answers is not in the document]
rundmt: Yeah I tried to narrow it to only use `Scroller`. You can check in `Scrollable` for the name of `ScrollerBase`, but they may be too restrictive. I think it would be better for the Wrapped component to opt in so they have the option to fix the problem themselves. 

I also tried to use React's new context API, but that causes our unit tests to fail because Enzyme hasn't updated yet.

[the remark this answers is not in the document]
ybsung: I got it.

PR #1621, opened 2018-05-04
ENYO-5111: Update Group QA samples to apply disabled prop

teckliew:
Apply `disabled` prop to `<Group>` as well

### Links
ENYO-5111

### Comments

### DIFF
--- a/packages/sampler/stories/qa-stories/SelectableItem.js
+++ b/packages/sampler/stories/qa-stories/SelectableItem.js
@@ -95,6 +95,7 @@ storiesOf('SelectableItem', module)
 					disabled: boolean('disabled', false)
 				}}
 				selectedProp="selected"
+				disabled={boolean('disabled', false)}
 				onSelect={action('onSelect')}
 			>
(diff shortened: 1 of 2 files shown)


ryanjduffy:
The proposed change looks good. Part of the issue here is the lack of support to prevent emitting events when a component is disabled. We'll get some of that support when we add explicit click handling to `Touchable`. Another issue is we should be mapping the correct, managed event to `Group` rather than relying on `onClick`. In the case of `ToggleItem` variants, we should use `onToggle` with `Group`,

I'd like to extend this to include another QA sample that validates individually disabled items as well with the right selection event.

Something like:
```js
			<Group
				childComponent={SelectableItem}
				childSelect="onToggle"
				selectedProp="selected"
				onSelect={action('onSelect')}
			>
				{[
					{children: '1', disabled: true},
					{children: '2', disabled: false},
					{children: '3', disabled: true},
					{children: '4', disabled: false}

				]}
			</Group>
```

PR #1626, opened 2018-05-04
ENYO-5244: Add Video component for preload support

ryanjduffy:
Adding onto the work in #1535 to further encapsulate preloading support and refactor related components

### Resolution

* Added `moonstone/VideoPlayer.Video` to support the `preloadSource` prop and fully encapsulate source switching from `VideoPlayer`
* Added `ui/ForwardRef` to support maintaining the correct `ref` to the media
* Changed `ui/Media` to auto-load media and marked the component `@public` in docs
* Updated `moonstone/MediaOverlay` to leverage `ui/Media`'s loading support

### Additional Considerations

* Validate both the VideoPlayer sample  and QA sample
* Validate VideoPlayer accessibility tests
* Validate the MediaOverlay sample

### Links

### Comments


ryanjduffy:
Sadly, had to duplicate the logic from `ComponentOverride` into VideoPlayer in one place to support passing a `ref` without using `forwardRef` which isn't supported by enzyme 3 yet.

packages/moonstone/VideoPlayer/Video.js
```diff
@@ -0,0 +1,297 @@
...
+			setMedia(this.video);
+		}
+	}
+
+	autoPlay () {
+		if (this.props.noAutoPlay || !this.canPlay()) return;
+
+		this.video.play();
+
+		forward('onLoadStart', this.loadStartEvent, this.props);
```
viodragon2: It seems like `loadstart` event was fired from the preload source would be still forwarded but with the payload from the other event. Any specific reason we want to do this?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -955,6 +924,16 @@ const VideoPlayerBase = class extends React.Component {
 		return true;
 	}
 
+	handleLoadStart = () => {
+		this.firstPlayReadFlag = true;
```
viodragon2: I'm not sure if we want to wait until `loadstart` event is fired to reset states. Can we reset when `source` prop updates?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1639,38 +1601,47 @@ const VideoPlayerBase = class extends React.Component {
...
+		delete mediaProps.thumbnailUnavailable;
+		delete mediaProps.titleHideDelay;
+		delete mediaProps.videoPath;
+
+		mediaProps.autoPlay = !noAutoPlay;
+		mediaProps.className = css.video;
+		mediaProps.controls = false;
+		mediaProps.mediaComponent = 'video';
+		mediaProps.onPlay = this.handlePlayEvent;
+		mediaProps.onLoadStart = this.handleLoadStart;
```
viodragon2: This should be added to `videoComponent` documentation

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1639,38 +1601,47 @@ const VideoPlayerBase = class extends React.Component {
...
+		delete mediaProps.seekDisabled;
+		delete mediaProps.setApiProvider;
+		delete mediaProps.thumbnailUnavailable;
+		delete mediaProps.titleHideDelay;
+		delete mediaProps.videoPath;
+
+		mediaProps.autoPlay = !noAutoPlay;
+		mediaProps.className = css.video;
+		mediaProps.controls = false;
+		mediaProps.mediaComponent = 'video';
```
viodragon2: This seems redundant.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -527,10 +518,10 @@ const VideoPlayerBase = class extends React.Component {
 		 * component as a child node.
 		 *
 		 * @type {Component}
-		 * @default 'video'
+		 * @default {@link ui/Media.Media}
```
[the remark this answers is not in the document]
ryanjduffy: We could since they're API-compatible but it does add some overhead that isn't necessary in the common case so I went with the "leaner" option.

packages/moonstone/VideoPlayer/Video.js
```diff
@@ -0,0 +1,297 @@
...
+
+import PropTypes from 'prop-types';
+
+/**
+ * Adds support for preloading a video source for `VideoPlayer`.
+ *
+ * @class VideoBase
+ * @memberof moonstone/VideoPlayer
+ * @ui
+ * @private
```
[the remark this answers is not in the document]
ryanjduffy: Shouldn't be `@private`. It's a component for all the people. 👐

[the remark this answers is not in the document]
ryanjduffy: Actually, I remember now! `VideoBase` is set to `@private` since it is only exported from the submodule. `Video` (with `Slottable` applied) is marked as `@public` and exported from `moonstone/VideoPlayer`.

I did update the docs for the public export to illustrate the advanced features the component enables.

packages/ui/Media/Media.js
```diff
@@ -12,7 +14,69 @@ import {forward} from '@enact/core/handle';
...
+ * @returns {String}                   Key representing sources
+ * @function
+ * @memberof ui/Media
+ * @public
+ */
+const getKeyFromSource = (source = '') => {
+	if (React.isValidElement(source)) {
+		return React.Children.toArray(source)
+			.filter(s => !!s)
+			.map(s => s.props.src)
```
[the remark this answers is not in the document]
ryanjduffy: I don't think the `type` will be relevant for determining uniqueness. For it to matter, two URIs would need to result in different physical sources. While this is possible, including the `type` wouldn't help us disambiguate them because (afaik) the `type` is only a hint to the browser and isn't sent to server hosting the video so the browser would get the same source for the same URI regardless of `type`.

packages/sampler/stories/qa-stories/VideoPlayer.js
```diff
@@ -61,15 +59,12 @@ class VideoSourceSwap extends React.Component {
 				{button('Next Preload Video without changing preload', this.nextVideoKeepPreload)}
 				{button('Change Preload without changing video', this.nextPreloadVideoKeepVideo)}
 				{button('Reset Sources', this.resetSources)}
-				<VideoPlayer
-					muted
-					preloadSource={<source src={this.state.playlist[this.state.preloadCursor]} type="video/mp4" />}
-				>
-					<source src={this.state.playlist[this.state.cursor]} type="video/mp4" />
+				<VideoPlayer muted>
```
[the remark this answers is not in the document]
ryanjduffy: It's declared on `VideoPlayer` and included (by not explicitly removing it) in `mediaProps` so this should work. It would _also_ work being declared on `Video` though since `VideoPlayer` doesn't explicitly handle it in any way.

packages/moonstone/VideoPlayer/Video.js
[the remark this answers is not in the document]
ryanjduffy: There's probably a problem here but I need to think on how to resolve it. It's designed to only send `onLoadStart` for the primary video and not the preload *and* not until both have loaded but I think `noAutoPlay` condition will interfere.

Also noticed that I'm not cleaning up the `loadStartEvent` anywhere which should be done.

packages/moonstone/VideoPlayer/Video.js
```diff
@@ -0,0 +1,297 @@
...
+		// if there's source and it has changed and it's not the prior preloaded source
+		if (source && key !== prevKey) {
+			// if it wasn't the preload key
+			if (key !== prevPreloadKey) {
+				// flag it as unloaded
+				this.loaded.video = false;
+			} else {
+				// if the preload video is switched to the current, try to play it now because it
+				// won't fire an onLoadStart events since it has already loaded.
+				this.autoPlay();
```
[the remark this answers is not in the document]
ryanjduffy: Good question! I'll have to try this out. This specific case address switching to a preloaded video which previously did not have `autoPlay` but now does. I don't know if it will auto play if the video was already loaded when `autoPlay` was not set on load.

packages/moonstone/VideoPlayer/Video.js
ryanjduffy: I don't think that will happen. If the preload video fires `onLoadStart` first, `canPlay` will prevent this line from being reached because `this.loaded.video` will be `false`. If the video fires `onLoadStart` first, its event will be cached in `loadStartEvent` and will be forwarded when the preload video fires `onLoadStart`.

Do you read it the same way?

packages/moonstone/VideoPlayer/VideoPlayer.js
ryanjduffy: I wasn't able to test the a11y features because my board died. I guess the question is how it should behave for a video that hasn't loaded yet (either because it's in progress or failed). I'm not sure what the right answer for that is but I'll investigate.

packages/moonstone/VideoPlayer/Video.js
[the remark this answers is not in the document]
viodragon2: Gotcha. One more thing. Let's add the `@exports Video` at the top of `VideoPlayer.js`

packages/sampler/stories/qa-stories/VideoPlayer.js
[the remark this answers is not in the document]
viodragon2: Yeah. I was hoping to remove `Media`/`Video` specific props from `VideoPlayer` if possible. I have a similar concern with `noAutoPlay` prop, too. But `noAutoPlay` affects the initial value of one of VP's internal member currently, so I'm not super concerned to fix that in this PR.

It would look something like this when changed:
```
<VideoPlayer>
  <Video muted noAutoPlay />
</VideoPlayer

File: packages/moonstone/VideoPlayer/Video.js (hunk 1 above)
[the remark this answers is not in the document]
Comment (viodragon2) [reply packages/moonstone/VideoPlayer/Video.js]: If we're not allowing any other event handlers to be attached to the preload source, shouldn't we only forward `loadstart` for primary video? Should `VideoPlayer` be concerned whether preload source has been loaded or not?

File: packages/moonstone/VideoPlayer/VideoPlayer.js (hunk 2 above)
Comment (viodragon2) [reply packages/moonstone/VideoPlayer/VideoPlayer.js]: I just realized that `firstPlayReadFlag` has been working incorrectly for a very long time. Even with this fix, the first play button click or play remote button click wouldn't read out "play" as it should. I'll file a separate bug for this.

File: packages/moonstone/VideoPlayer/Video.js (hunk 9 above)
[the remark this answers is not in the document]
Comment (ryanjduffy) [reply packages/moonstone/VideoPlayer/Video.js]: Okay. So it seems that adding `autoplay` after the fact _might_ work. It does in firefox but doesn't in chrome. See https://codepen.io/anon/pen/bMvzEb. The chrome failure may be unique to a newer version of chrome for UX reasons (to prevent sites from auto playing unwanted videos).

That all said, we can't call `load()` on a video that was already loaded without throwing away the preload. Barring other suggestions, I'm going to leave this as is for now.

File: packages/moonstone/VideoPlayer/Video.js (hunk 1 above)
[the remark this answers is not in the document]
Comment (ryanjduffy) [reply packages/moonstone/VideoPlayer/Video.js]: We are effectively only forwarding for the primary video (the payload is cached from it only) but are delaying sending it until both have started. This is building on the work @rundmt started (assuming I've understood it correctly) to work around platform issues with pausing a video when another loads. Perhaps this deferral isn't required, though, and we could dispatch the `onLoadStart` from the primary video whenever it happens.

Hoping @rundmt can chime in on this! :)

File: packages/ui/Media/Media.js [hunk 10]
```diff
@@ -124,12 +206,28 @@ class Media extends React.Component {
...
+			this.autoLoad();
+		}
 	}
 
 	componentWillUnmount () {
 		this.detachCustomMediaEvents();
 	}
 
+	autoLoad () {
+		if (this.props.noAutoLoad) return;
```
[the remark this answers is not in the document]
ryanjduffy: Good catch! It prevents *reloading* but not loading. At best, it's poorly named and should only be invoked in `componentDidUpdate`. I'll make a change.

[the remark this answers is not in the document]
ryanjduffy: This could be cruft from earlier refactoring. In testing in Chrome, I've found that the "auto loading" doesn't have any effect. In the usage by `Video`, `componentDidUpdate` never fires with a changed source because the `key` generated for each `Media` component ensures that a change in source results in a new component.

For now, I've removed this logic but will need to verify on the TV hardware to ensure that there aren't regressions.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -218,6 +220,7 @@ const VideoPlayerBase = class extends React.Component {
 		 * * `onJumpBackwardButtonClick` - Called when the jump backward button is pressed
 		 * * `onJumpForwardButtonClick` - Called when the jump forward button is pressed
 		 * * `onKeyDown` - Called when a key is pressed
+		 * * `onLoadStart` - Called when the video starts to load
```
viodragon2: This shouldn't be in `mediaControlsComponent`, but in  `videoComponent`

packages/ui/ForwardRef/ForwardRef.js
```diff
@@ -0,0 +1,67 @@
+/**
+ * Provides a HOC interface for `ref` forwarding.
```
Djspaceg: Is the intent of this component to provide a more convenient way to simulate the upcoming React forwardRef feature that Enzyme doesn't support yet?

ryanjduffy: To be clear, it's not _simulating_ `React.forwardRef`; it's using it from within the HOC.

Djspaceg: I like it. And just for verbose clarification, this _will_ work with the version of Enzyme we're using?

ryanjduffy: Sadly, no. Hence the gymnastics included here

https://github.com/enactjs/enact/pull/1626/files#diff-598514a33c478cdac5391092c58009feR1662

ryanjduffy: There's no way to simulate `forwardRef`. By attaching `ref` to a component instance, you will always get a reference to that instance and not any underlying component. The best we've accomplished previously is `ApiDecorator` which still returns a `ref` to itself by proxies method and property access.

`forwardRef` adds the internal plumbing necessary to take a `ref` prop in and send another prop out.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -955,6 +926,16 @@ const VideoPlayerBase = class extends React.Component {
 		return true;
 	}
 
+	handleLoadStart = () => {
+		this.firstPlayReadFlag = true;
+		this.setState({
+			announce: AnnounceState.READY,
```
viodragon2: Since `loadstart` event does not fire for the "preload" source, title won't be read.

ryanjduffy: Based on our conversation on disabling playback, should we adjust `Video` to emit the `onLoadStart` event "manually" when switching to a preload source that was already loaded?

viodragon2: We could. One question with that is what do you think about handling other loading events? I'm not sure if we have a good reason to only handle `loadstart` event when there are other loading related events like `loadeddata` or `loadedmetadata` and so on. Should we fire those as well?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -955,6 +926,16 @@ const VideoPlayerBase = class extends React.Component {
 		return true;
 	}
 
+	handleLoadStart = () => {
+		this.firstPlayReadFlag = true;
```
viodragon2: Found a bug related to resetting internal values when a source has been updated.
In P&V app scenario, app loads the next video when it reaches the end of the video. So we need to reset `prevCommand`, and `speedIndex` so that ff rate does not show when the next video gets loaded when it reaches the end of the video.
Other internal values might need to be reset, but it didn't seeme necessary when I tested it.

ryanjduffy:
Summary of changes:

* `ui/Media` fires `onLoadStart` and then (the new) `onReady`
* VideoPlayer handles `onReady` to reset new video state
* `Video` fires `onReady` manually in `componentDidUpdate` when preload switches to active. No need to manually fire for a non-preloaded video since it will receive that automatically from `ui/Media` via `...rest`
* `Video` fires `onUpdate` after `onReady` to handle the error case and updating other internal state

packages/moonstone/VideoPlayer/Video.js
```diff
@@ -0,0 +1,279 @@
...
+		 */
+		source: PropTypes.oneOfType([PropTypes.string, PropTypes.node])
+	}
+
+	static defaultProps = {
+		mediaComponent: 'video'
+	}
+
+	componentDidMount () {
+		this.setMedia();
```
viodragon2: Do we need this when media is set on mount via `setVideoRef`?

packages/moonstone/VideoPlayer/Video.js
```diff
@@ -0,0 +1,279 @@
...
+				// since it won't receive the onLoadStart or onError event
+				forward('onUpdate', {type: 'onUpdate'}, this.props);
+
+				this.autoPlay();
+			} else if (key !== prevKey) {
+				// if there's source and it has changed.
+				this.autoPlay();
+			}
+
+			this.autoPlay();
```
viodragon2: This will prevent from pausing :sadpanda:

ryanjduffy: Fortunately, it's just cruft from when I consolidated the above blocks. Removing! :boom:

packages/moonstone/VideoPlayer/Video.js
ryanjduffy: Nope, I don't this so. Good catch!

packages/moonstone/VideoPlayer/Video.js
```diff
@@ -0,0 +1,273 @@
...
+			this.clearMedia(prevProps);
+			this.setMedia();
+		}
+
+		if (source) {
+			if (key === prevPreloadKey) {
+				// if there's source and it was the preload source
+
+				// if the preloaded video didn't error, notify VideoPlayer it is ready to reset
+				if (!this.video.error) {
```
viodragon2: If this is the case, shouldn't we be firing `onReady` when `canplaythrough` event is fired instead of `loadstart`?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1644,38 +1608,47 @@ const VideoPlayerBase = class extends React.Component {
...
+		delete mediaProps.onSeekFailed;
+		delete mediaProps.pauseAtEnd;
+		delete mediaProps.playbackRateHash;
+		delete mediaProps.seekDisabled;
+		delete mediaProps.setApiProvider;
+		delete mediaProps.thumbnailUnavailable;
+		delete mediaProps.titleHideDelay;
+		delete mediaProps.videoPath;
+
+		mediaProps.autoPlay = !noAutoPlay;
```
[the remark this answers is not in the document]
viodragon2: I am not sure why, but after properly setting `noAutoPlay` prop to `Video`, it seems like loading gets "suspended" if `autoPlay=false`. Calling `load()` on mount and src update seem to resolve the problem. Where is the `suspend` event coming from??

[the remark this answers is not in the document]
ryanjduffy: Good catch on the prop. I elected to keep this line and change Video to use `autoPlay` instead of `noAutoPlay`. That keeps Video consistent with Media (on which it's based).

Regarding the issue you're seeing, I'm not following what the problem is. I see the `suspend` event firing but it doesn't appear to cause problems nor necessarily indicate a problem based on the docs: https://developer.mozilla.org/en-US/docs/Web/Events/suspend.

PR #1628, opened 2018-05-07
ENYO-4747: Fix spotlight handling 5-way events when in pointer mode

sugardave:
Fixed case where spotlight focus was erroneously remaining on the incorrect spottable item.

### Resolution
Spotlight no longer forces `_pointerMoveDuringKeyPress = true;` when handling pointer events because wheeling isn't necessarily moving.

### Links
ENYO-4747

Enact-DCO-1.0-Signed-off-by: Dave Freeman (dave.freeman@lge.com)

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -358,11 +358,7 @@ const Spotlight = (function () {
 		const direction = getDirection(keyCode);
 		const pointerHandled = notifyKeyDown(keyCode, handlePointerHide);
 
-		if (pointerHandled) {
-			return;
-		}
-
-		if (!(direction || isEnter(keyCode))) {
+		if (pointerHandled || !(direction || isEnter(keyCode))) {
 			return;
 		}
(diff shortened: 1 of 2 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -359,7 +359,6 @@ const Spotlight = (function () {
 		const pointerHandled = notifyKeyDown(keyCode, handlePointerHide);
 
 		if (pointerHandled) {
-			_pointerMoveDuringKeyPress = true;
```
webOS101: I would have to go back and look, but I _think_ this flag is supposed to be set in the block below.  The intention being that if you were holding a key down and moved the pointer, it's not supposed to activate pointer mode.  It seems when we split the handling before we removed this from the conditional.  If this is the case we don't wish to set this anymore then we should remove the logic associated with it, since it's no longer possible to have this flag set.  We can further simplify by recombining this `if` with the one below.

PR #1634, opened 2018-05-10
ENYO-5146: Fix spotlight handling when leaving ExpandableInput

ryanjduffy:
There were several lingering issues with navigating from the `Input` within `ExpandableInput` via 5-way:

* Pressing down would result in the element below `ExpandableInput` receiving focus instead of the label (ENYO-5146)
* Unmounting an `ExpandableInput` while the `Input` has focus would result in the pointer being locked (ENYO-5249)
* Pressing left or right to navigate out of an `ExpandableInput` inside a "last focused" container (such as `moonstone/Panels.Panel`) would correctly navigate but would not update the last focused causing spotlight to land inside the `Input` when the container obtained focus.

### Resolution

* Change `moonstone/Input` to allow consumers to call `ev.preventDefault()` from `onKeyDown` handlers to prevent its default behavior (in this case, calling `Spotlight.move()` on 5-way up and down)
* Add `componentWillUnmount` callback to resume spotlight in `ExpandableInput`
* Refactor logic in `moonstone/ExpandableItem.ExpandableSpotlightDecorator` to fix the last focused element for either the entire tree (when nothing is focused) or just the immediate expandable spotlight container (when something else is focused) to ensure focus restores correctly.

### Additional Considerations

* Also removed the "enter" handler in `ExpandableInput` since the `preventDefault` was now preventing `Input`'s `dismissOnEnter` handling (which was needed) and the "problem" it was avoiding was no longer reproducing (I think because of changes in Input to not blur when `dismissOnEnter` until keyup).

### DIFF
--- a/packages/moonstone/ExpandableInput/ExpandableInput.js
+++ b/packages/moonstone/ExpandableInput/ExpandableInput.js
@@ -7,7 +7,7 @@
  */
 
 import Changeable from '@enact/ui/Changeable';
-import {forKey, forward, oneOf, preventDefault, stopImmediate} from '@enact/core/handle';
+import {adaptEvent, call, forKey, forward, handle, oneOf, preventDefault, stopImmediate} from '@enact/core/handle';
 import deprecate from '@enact/core/internal/deprecate';
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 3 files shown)


teckliew:
:wow:

PR #1636, opened 2018-05-10
PLAT-49795: Add overscroll effect for moonstone

0x64:
Added overscroll effect for moonstone which makes a user notice the end of the list.

### Resolution
Supporting overscroll effects on wheeling and paging control buttons. For performance and implementation reasons, the effect is based on `CSS transition`. Basically, there is no difference between wheeling and paging control buttons, but a user cannot see any _repeating_ effect on clicking paging control buttons because buttons will be _disabled_ as soon as the list reached the end.

### Links
PLAT-49795

### Comments


0x64:
@Djspaceg, Thank you for your detail review.

I wanted to get your opinion before I finish further works because the logic itself may be updated (or even re-written) if the current behavior does not meet our requirements and expectations. So some code is not ready yet.

1. Of course. It depends on UX and logic, so I deferred clearing for your review. I think you agree on the current behavior, so I will clear up.

2. I got it. I will update.
3. pass!
4. 'forth' is a temporal name due to naming collision in my code, and my plan was to use 'forward' anyway after getting your initial review and agreement. (don't worry about it. :) ) And, I like your suggestion. Boolean props looks pretty in `Scrollable`, but the 'meaning' is not clear. I will update.
5. Sure.
6. Oh, I missed the CSS way. Thanks!
7. No. the effect SHOULD NOT applied to scrollbars. I considered it on very first design but scrollbars can be hidden explicitly with the prop `verticalScrollbar` and `horizontalScrollbar` even a list or a scroller can scroll in that way. That is why I should implment effects on contents div with additional DOM nodes. One altanative approach is to implment on scrollbars and handles the edge cases - 'invisible scrollbars' but I think it is an anti-pattern. Well, if we don't need to display effects when a scrollbar is hidden, we can change the implementation. My personal preference is to display regardless scrollbar visibility. Please let me know if you have other opinion.
8. I want also less refs, but I don't have any idea to reduce refs because API-calls of self-controlled overscroll effects are the most safe and simple way to support dynamically requests of the effect. My first try is, you know, to use keyframe animation and I thought we could control the animation only with classes (which can be controlled by `Scrollable` only with props). But after I tested several scenarios including repeating wheeling, I concluded that the effect element itself should have a status _flow_ and we cannot invoke that flow with a class or a CSS attribute. If we only need to support a discreted effect (eg. displays always whole pair of expanding-shrinking and repeating from the expanding again), I think there are many chances to control it.

UX design affects a lot of structures and logic. If you have any objection to my comment (starting with *Notes*), please let me know. I welcome your idea always.

Djspaceg:
screen shot 2018-05-15 at 4 09 06 pm
To help with my description of bullet 7, I've included the above diagram. This represents how I believe the scroller structure works.

It sounds like my intent with 7 was interpreted differently than I expected. I was suggesting that the `::before` and `::after` pseudo-elements be attached to the scrollbar's sibbling element, the vertical and horizontal content areas, not the actual bar. This way, like you described, the scroll bar itself could be hidden, and the related content area would still be visible, and the overscroll pseudo-elements would also still be available. More specifically, for the top and bottom overscroll elements would be attached to the before and after elements (respectively) of the vertical content area. The left and right overscroll elements would be attached to the before and after elements (respectively) of the horizontal content area. 
We'll need to do a little bit of trickery when positioning the `::after` of the horizontal content area, since it's edge is to the _right_ of the vertical scroll bar (to indicate when the vertical scroll bar is present or not), but that can relatively easily be solved with a set of classes on the outer scroller element which indicate the current state of the scroller: whether the vertical scroll bar is visible, the horizontal is visible, both are, or neither are: `className="vertical"`, `className="horizontal"`, `className="vertical horizontal"`, and `className=""` respectively.

I think I noticed an additional comment/question in your response: If there are no scroll bars, because there is nowhere to scroll, no overscroll edge effect should display, even if someone uses the scroll wheel. The intended effect is that if a user can't scroll at the moment, they don't need to be aware that the scrolling container is even present at all.

packages/moonstone/Scrollable/OverscrollEffect.less
```diff
@@ -0,0 +1,97 @@
+// OverscrollEffect.less
+//
+@import '~@enact/ui/Scrollable/Scrollable.less';
+
+@import '../styles/mixins.less';
+@import '../styles/variables.less';
+@import '../styles/skin.less';
+
+:root {
+	--moon-scrollable-overscroll-ratio-horizontalbefore: 0;
```
Djspaceg: These CSS variables also don't need to be prefixed with `moon-` since they're part of a moonstone component already. We _sort of_ decided recently that CSS variables defined inside a scope, only need include that scope name for as broad of an audience it's hitting. So, a theme-scoped variable _should_ include both the theme name and the component name it's targeting, but if it's specific to that component, defined only in that component's files, it can just include the component name. Furthermore, if it's defined even narrower than that, like on a child of a component, it can just refer to that name too.

A crummy but ok example:

```
.moonstone {
	--moon-scroller-overscroll-scale: 0.5;

	.scroller {
		--scroller-overscroll-scale: 0.5;

		.overscroll {
			--overscroll-scale: 0.5;
		}
	}
}
```

0x64:
I updated LESS and more. Touch inputs are not handled properly yet and now I'm working on it.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -485,6 +531,55 @@ class ScrollableBase extends Component {
...
+
+	clearAllOverscrollEffects = () => {
+		this.clearOverscrollEffect('horizontal');
+		this.clearOverscrollEffect('vertical');
+	}
+
+	playOverscrollEffect = (nodeRef, orientation, position, ratio, auto) => {
+		const prefix = '--scrollable-overscroll-ratio-';
+
+		nodeRef.style.setProperty(prefix + orientation + position, ratio);
```
Djspaceg: As of React 15.6.0, we no longer need to use `ref` to set CSS variables. These can be done much more simply by setting `style`, e.g.:

```jsx
<Scrollable
	style={{prefix + orientation + position: ratio}}
```
or a more literal example:
```jsx
<Scrollable
	style={{'--scrollable-overscroll-ratio-vertical-before': ratio}}
```

That should simplify things quite a lot.

0x64: Those CSS variables are very frequently updated especially on dragging, so I have a concern about the performance impact when we put these into JSX. I don't want to push Scrollable to go into any heavy life cycle.

Djspaceg: That's a fair point. Though, I don't think they need to be updated at all during scrolling, just once at the very end of scrolling. If it's more than that, it's unnecessary.

0x64: So the dragging is the key case. :) To be natural, an effect reacts ASAP along with the pointer moving. Scrollable's `componentDidUpdate` has a lot of jobs and it looks not good to invoke an additional React update cycle for a style.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -210,6 +225,13 @@ class ScrollableBase extends Component {
 	indexToFocus = null
 	nodeToFocus = null
 
+	// overscroll
+	overscrollRefs = {['horizontal']: null, ['vertical']: null}
```
ybsung: It looks better to use `overscrollRefs = {horizontal: null, vertical: null}`.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -210,6 +225,13 @@ class ScrollableBase extends Component {
...
+	overscrollJobs = {
+		['horizontal']: {['before']: null, ['after']: null},
+		['vertical']: {['before']: null, ['after']: null}
```
ybsung: Same above.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -240,15 +262,26 @@ class ScrollableBase extends Component {
 		if (pos) {
 			const
 				{top, left} = pos,
-				bounds = this.uiRef.getScrollBounds();
+				bounds = this.uiRef.getScrollBounds(),
+				horizontalMove = bounds.maxLeft > 0 && left !== this.uiRef.scrollLeft,
+				verticalMove = bounds.maxTop > 0 && top !== this.uiRef.scrollTop;
```
ybsung: How about `canMoveHorizontally`, `canMoveVertically` for `horizontalMove `, `verticalMove ` like `canScrollVertically`?

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -240,15 +262,26 @@ class ScrollableBase extends Component {
...
-			if ((bounds.maxTop > 0 && top !== this.uiRef.scrollTop) || (bounds.maxLeft > 0 && left !== this.uiRef.scrollLeft)) {
+			if (horizontalMove || verticalMove) {
 				this.uiRef.start({
 					targetX: left,
 					targetY: top,
 					animate: (animationDuration > 0) && this.animateOnFocus,
 					duration: animationDuration
 				});
+
+				if (this.childRef.shouldPreventOverscrollEffect ? !this.childRef.shouldPreventOverscrollEffect() : true) {
```
ybsung: It would be better to use the following condition instead of `?`.
```
!this.childRef.shouldPreventOverscrollEffect || !this.childRef.shouldPreventOverscrollEffect()
```

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -399,8 +432,27 @@ class ScrollableBase extends Component {
 
 	onKeyDown = (ev) => {
 		this.animateOnFocus = true;
-		if (!Spotlight.getPointerMode() && (isPageUp(ev.keyCode) || isPageDown(ev.keyCode)) && !ev.repeat && this.hasFocus()) {
-			this.scrollByPage(ev.keyCode);
+
+		if (!Spotlight.getPointerMode() && !ev.repeat && this.hasFocus()) {
```
ybsung: I don't think that we need `!Spotlight.getPointerMode()` condition because it should scroll when pressing a page up or down key in pointer mode.

packages/moonstone/Scrollable/OverscrollEffect.less
```diff
@@ -0,0 +1,92 @@
...
+		&::after {
+			top: 50%; // Center the element
+			width: @moon-scrollable-overscroll-thickness;
+			height: 70%;
+		}
+
+		&::before {
+			transform: translateY(-50%) scale3d(~"calc((2 - var(--scrollable-overscroll-ratio-horizontalbefore)) / 2)", var(--scrollable-overscroll-ratio-horizontalbefore), 1);
+			left: 0;
+			z-index: 1;
```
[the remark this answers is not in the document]
0x64: Required. The before pseudo element is covered by contents without `z-index`.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -399,8 +432,27 @@ class ScrollableBase extends Component {
...
+			if (isPageUp(ev.keyCode) || isPageDown(ev.keyCode)) {
+				this.scrollByPage(ev.keyCode);
+			} else {
+				const
+					direction = getDirection(ev.keyCode);
```
ybsung: Please define `direction` in one line.

packages/moonstone/Scrollable/Scrollable.js
0x64: Right.

packages/moonstone/Scrollable/Scrollable.js
0x64: Good!

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -485,6 +537,56 @@ class ScrollableBase extends Component {
...
+		this.clearOverscrollEffect('horizontal', 'before');
+		this.clearOverscrollEffect('horizontal', 'after');
+		this.clearOverscrollEffect('vertical', 'before');
+		this.clearOverscrollEffect('vertical', 'after');
+	}
+
+	applyOverscrollEffect = (orientation, position, type, ratio) => {
+		const
+			nodeRef = this.overscrollRefs[orientation],
+			prefix = '--scrollable-overscroll-ratio-';
```
ybsung: How about defining `prefix` as global in this file?

packages/moonstone/Scrollable/Scrollable.js
0x64: It is shorter than the current code, but it looks little unclear.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -530,6 +594,56 @@ class ScrollableBaseNative extends Component {
...
+
+			if (type === overscrollTypes.scrolling) {
+				this.overscrollJobs[orientation][position].start(orientation, position, overscrollTypes.none, 0);
+			}
+		}
+	}
+
+	newOverscrollJob = (orientation, position) => {
+		if (!this.overscrollJobs[orientation][position]) {
+			this.overscrollJobs[orientation][position] = new Job(this.applyOverscrollEffect.bind(this), overscrollTimeout);
```
ybsung: Do we need `bind` here because we defined it with arrow function for lexical scope?

packages/moonstone/Scrollable/Scrollable.js
0x64: So, was it a bug that calling `this.scrollByPage` only when 5-way key mode?

packages/moonstone/Scrollable/Scrollable.js
0x64: Sure.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -548,6 +662,22 @@ class ScrollableBaseNative extends Component {
 		}
 	}
 
+	initHorizontalOverscrollRef = (ref) => {
+		if (ref) {
+			this.overscrollRefs['horizontal'] = ReactDOM.findDOMNode(ref); // eslint-disable-line react/no-find-dom-node
```
ybsung: We could simply use `this.overscrollRefs.horizontal`.

packages/moonstone/Scrollable/Scrollable.js
0x64: For easy testing, I needed to define here. I will move it.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -548,6 +662,22 @@ class ScrollableBaseNative extends Component {
...
+			this.newOverscrollJob('horizontal', 'before');
+			this.newOverscrollJob('horizontal', 'after');
+		}
+	}
+
+	initVerticalOverscrollRef = (ref) => {
+		if (ref) {
+			this.overscrollRefs['vertical'] = ref;
```
ybsung: Did you intended to use `ref` instead of `ReactDOM.findDOMNode(ref)`?

packages/moonstone/Scrollable/ScrollableNative.js
0x64: I will do it if the code exists after cleaning up. ;)

packages/moonstone/Scrollable/ScrollableNative.js
0x64: `ref` is not a React component but a `div` node actually here. So, yes, we don't need `findDOMnode`.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -604,19 +737,19 @@ class ScrollableBaseNative extends Component {
...
 						data-spotlight-container={spotlightContainer}
 						data-spotlight-container-disabled={spotlightContainerDisabled}
 						data-spotlight-id={spotlightId}
 						ref={initUiContainerRef}
 						style={style}
 					>
-						<div className={componentCss.container}>
-							<TouchableDiv {...touchableProps}>
+						<div className={classNames(componentCss.container, overscrollCss.overscrollFrame, overscrollCss.vertical, isHorizontalScrollbarVisible ? overscrollCss.horizontalScrollbarVisible : null)} ref={this.initVerticalOverscrollRef}>
+							<TouchableDiv className={classNames(touchableClassName, overscrollCss.overscrollFrame, overscrollCss.horizontal)} ref={this.initHorizontalOverscrollRef} {...restTouchableProps}>
```
ybsung: Hmmm..
I'm not sure...but...
Is it better to define a variable for `classNames(componentCss.container, overscrollCss.overscrollFrame, overscrollCss.vertical, isHorizontalScrollbarVisible ? overscrollCss.horizontalScrollbarVisible : null)` at 743 line?

packages/moonstone/Scrollable/Scrollable.js
0x64: Oh, it was the condition inside 'if' statement. I will fix it.

packages/moonstone/Scrollable/Scrollable.js
0x64: I will check this when I update with the latest `develop` branch.

packages/moonstone/Scrollable/Scrollable.js
0x64: I will change them to `scrollHorizontally' and 'scrollVertically' since they are not for scrollability but for actual scrolling.

packages/moonstone/Scrollable/ScrollableNative.js
0x64: We need.

packages/moonstone/Scrollable/ScrollableNative.js
0x64: As discussed, no benefit.

0x64:
Note: I tested with #1719 to prevent overscroll logics from unhappy scrolling by focus after #1703 and #1720 are merged.

packages/moonstone/Scrollable/Scrollable.js
0x64: Okay, it looks like a different issue from this PR. It will be handled with a new ticket.

ybsung:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -643,12 +654,10 @@ const VirtualListBaseFactory = (type) => {
 		onKeyDown = (ev) => {
 			const {keyCode, repeat, target} = ev;
 
-			this.isScrolledBy5way = false;
 			if (getDirection(keyCode)) {
 				ev.preventDefault();
 				this.setSpotlightContainerRestrict(keyCode, target);
-				this.isScrolledBy5way = this.jumpToSpottableItem(keyCode, repeat, target);
-				if (this.isScrolledBy5way) {
+				if (this.jumpToSpottableItem(keyCode, repeat, target)) {
```
MikyungKim: I think we need to clean up return values in `jumpToSpottableItem`

PR #1646, opened 2018-05-14
PLAT-54708: Fixed `storybook` lists to show items properly when `dataSize` is greater than 1000.

0x64:
In `VirtualList` and `VirtualGridList` samples (for both `ui` and `moonstone`), if `dataSize` is set to any number grater than 1000, no item is rendered after 'Item 999'. 

### Resolution
Data were initially generated with the size of 1000 and was never updated even if `dataSize` was changed.
To support any number of items, items will be re-generated whenever `dataSize` is changed. It is a natural behavior since the change of  `dataSize` means the change of data in real apps.

### Additional Considerations
Heading '0's for list subjects will be changed along with the number of digits to show. (For example, there will be no '0' if a list shows items from '0' to '9')

### Links
PLAT-54708

### Comments

### DIFF
--- a/packages/sampler/stories/moonstone-stories/VirtualGridList.js
+++ b/packages/sampler/stories/moonstone-stories/VirtualGridList.js
@@ -54,7 +54,7 @@ const
 
 const updateDataSize = (dataSize) => {
 	const
-		itemNumberDigits = ((dataSize - 1) + '').length,
+		itemNumberDigits = dataSize > 0 ? ((dataSize - 1) + '').length : 0,
 		headingZeros = Array(itemNumberDigits).join('0');
 
 	items.length = 0;
--- a/packages/sampler/stories/moonstone-stories/VirtualGridList.js
+++ b/packages/sampler/stories/moonstone-stories/VirtualGridList.js
@@ -24,6 +24,7 @@ const
 		direction: {'horizontal': 'horizontal', 'vertical': 'vertical'}
 	},
 	items = [],
+	dataSizeDefault = 1000,
 	// eslint-disable-next-line enact/prop-types
 	uiRenderItem = ({index, ...rest}) => {
 		const {text, subText, source} = items[index];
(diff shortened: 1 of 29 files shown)


packages/sampler/stories/moonstone-stories/VirtualGridList.js
```diff
@@ -51,16 +52,28 @@ const
...
 
-for (let i = 0; i < 1000; i++) {
+const updateDataSize = (dataSize) => {
 	const
-		count = ('00' + i).slice(-3),
-		text = `Item ${count}`,
-		subText = `SubItem ${count}`,
-		color = Math.floor((Math.random() * (0x1000000 - 0x101010)) + 0x101010).toString(16),
-		source = `http://placehold.it/300x300/${color}/ffffff&text=Image ${i}`;
+		itemNumberDigits = dataSize > 1 ? Math.floor(Math.log10(dataSize - 1)) + 1 : 0,
```
ybsung: How about (dataSize - 1 + '').length? In addition, we need to check if dataSize equals 0 or not.

packages/sampler/stories/moonstone-stories/VirtualGridList.js
```diff
@@ -24,6 +24,7 @@ const
 		direction: {'horizontal': 'horizontal', 'vertical': 'vertical'}
 	},
 	items = [],
+	dataSizeDefault = 1000,
```
ybsung: How about `defaultDataSize`?

packages/sampler/stories/qa-stories/VirtualList.js
```diff
@@ -34,9 +35,21 @@ const
...
-}
+const updateDataSize = (dataSize) => {
+	const
+		itemNumberDigits = dataSize > 0 ? ((dataSize - 1) + '').length : 0,
+		headingZeros = Array(itemNumberDigits).join('0');
+
+	items.length = 0;
+
+	for (let i = 0; i < dataSize; i++) {
+		items.push({item :'Item ' + (headingZeros + i).slice(-itemNumberDigits), selected: false});
```
ybsung: It seems that we need to initialize the `StatefullSwitchItem`'s `selected` state when updating `items` array.

ybsung: `items[this.props.index].selected` and `this.state.selected` are not matched each other in `StatefulSwitchItem `.

packages/sampler/stories/moonstone-stories/VirtualGridList.js
```diff
@@ -51,16 +52,28 @@ const
...
+			count = (headingZeros + i).slice(-itemNumberDigits),
+			text = `Item ${count}`,
+			subText = `SubItem ${count}`,
+			color = Math.floor((Math.random() * (0x1000000 - 0x101010)) + 0x101010).toString(16),
+			source = `http://placehold.it/300x300/${color}/ffffff&text=Image ${i}`;
+
+		items.push({text, subText, source});
+	}
+
+	return dataSize;
```
ybsung: I could see the following warnnings.
```
The specified value "NaN" is not a valid number. The value must match to the following regular expression: -?(\d+|\d+\.\d+|\.\d+)([eE][-+]?\d+)?
```
Do we need to return 0 if `dataSize` equals null?

0x64: I've tested all changed samples with various values including 'no value', but I could not see that error. Could you let me know how can I reproduce it?

packages/sampler/stories/qa-stories/VirtualList.js
0x64: Oh, you're right. Hmm. any better way to clear up states of all items?

0x64: Note: discussed with @MikyungKim and @ybsung, and concluded not to update qa-sampler. It is unclear to update all switch items when `dataSize` is changed, and qa-sampler is to test some special cases, not for general samples. I will revert changes of qa-sampler.

packages/sampler/stories/moonstone-stories/VirtualGridList.js
0x64: It turns out that it is not from our code but from a knob-addon.

ybsung:
:LGTM

@0x64 Please fix the conflict.

MikyungKim:
LGTM

PR #1647, opened 2018-05-14
ENYO-5112: Touchable End Gestures when Blurred

teckliew:
Add a blur handler to end gestures when the focused component is blurred.

### Links
ENYO-5112

### Comments

### DIFF
--- a/packages/ui/Touchable/Touchable.js
+++ b/packages/ui/Touchable/Touchable.js
@@ -527,7 +527,7 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		onBlur () {
-			return this.endGesture();
+			return this.suspendGesture();
 		}
 
 		enterGesture () {
@@ -548,6 +548,14 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
 			this.hold.end();
 			this.flick.end();
 			this.drag.end();
+
+			return true;
+		}
+
+		suspendGesture () {
+			this.hold.suspend();
+			this.flick.suspend();
+			this.drag.suspend();
 
 			return true;
 		}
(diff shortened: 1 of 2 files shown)


packages/ui/Touchable/Touchable.js
```diff
@@ -521,6 +526,13 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
 			return true;
 		}
 
+		onBlur () {
+			this.hold.suspend();
+			this.drag.end();
```
webOS101: Would it make sense to call the same handler as `onMouseLeave` calls here instead of trying to handle it differently?

webOS101: Also, keep in mind that 5-way isn't the only way that blur can be called.

ryanjduffy:
Let's add a refinement to this. Rather than indiscriminately canceling gestures on blur, let's restrict it to when the target had focus when the gesture began. That will, hopefully, guard against cases in which the gesture support is used independent of focus (e.g. scrollers).

Generally:
```js
const handleBlur = handle(
  forward('onBlur'),
  call('hadFocus'),
  call('endGesture')
);

// ...

startGesture () {
  // ...
  this.targetHadFocus = this.target === document.activeElement;
  // ...
}

// ...

hadFocus () {
  return this.targetHadFocus;
}
```

packages/ui/Touchable/Touchable.js
```diff
@@ -175,6 +175,11 @@ const handleGlobalMove = handle(
 	call('moveGesture')
 );
 
+const handleOnBlur = handle(
```
ryanjduffy: Please rename to `handleBlur` for consistency with the rest of the handlers.

PR #1650, opened 2018-05-15
ENYO-5253: Fix Spottable to prevent making component spottable when disabled

ryanjduffy:
The logic in `componentWillReceiveProps` was incorrectly setting the `spottableDisabled` state to `true` for `spotlightDisabled` components

### Resolution
* Always (correctly) update the value of `spottableDisabled` on prop change so the state is accurate
* Do not consider `spotlightDisabled` in the value of `spottableDisabled` because it is included in the `spottable` determination in `render()`.
* Fixed a regression from #1570 which inadvertently dropped assignment of `this.node`

### DIFF
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -182,13 +182,11 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		componentWillReceiveProps (nextProps) {
-			const becomingDisabled = nextProps.disabled && !this.props.disabled;
+			const spottableDisabled = this.isFocused && nextProps.disabled && !this.props.disabled;
 
-			if (!this.state.spottableDisabled && this.isFocused && becomingDisabled) {
-				this.setState({
-					spottableDisabled: true
-				});
-			}
+			this.setState({
+				spottableDisabled
+			});
 		}
 
 		componentDidUpdate (prevProps, prevState) {
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -182,10 +182,12 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		componentWillReceiveProps (nextProps) {
-			const spottableDisabled = this.isFocused && nextProps.disabled || nextProps.spotlightDisabled;
+			const becomingDisabled = nextProps.disabled && !this.props.disabled;
 
-			if (!this.state.spottableDisabled && spottableDisabled) {
-				this.setState({spottableDisabled});
+			if (!this.state.spottableDisabled && this.isFocused && becomingDisabled) {
+				this.setState({
+					spottableDisabled: true
+				});
 			}
 		}
 
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -10,6 +10,7 @@ import {forward, forwardWithPrevent, handle} from '@enact/core/handle';
 import hoc from '@enact/core/hoc';
 import {is} from '@enact/core/keymap';
 import React from 'react';
+import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 
 import {getContainersForNode} from '../src/container';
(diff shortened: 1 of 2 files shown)


packages/spotlight/Spottable/Spottable.js
```diff
@@ -182,11 +182,11 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		componentWillReceiveProps (nextProps) {
-			const spottableDisabled = this.isFocused && nextProps.disabled || nextProps.spotlightDisabled;
+			const spottableDisabled = this.isFocused && nextProps.disabled && !this.props.disabled;
```
webOS101: Does this prevent `spotlightDisabled` from working anymore?

ryanjduffy: Nope. spottableDisabled only tracks if a previously focused comp became disabled. It'll still prevent Spottable altogether when spotlightDisabled is set.

webOS101: My concern is we also use `spottableDisabled` for setting spottability on line 346.  Why not use the same mechanism and clean up the becoming enabled part on line 202, as well?

webOS101: The idea was that at line 202, we check to see if we've become enabled after being disabled, so we can potentially restore spotlight to the component.  If `spottableDisabled` is a tracker for if a component is enabled or not, this seems like a reasonable use of that state.  That said, this does appear to work although I still can't figure out how `spotlightDisabled` works for anything other than the `isActionable` handler.

ryanjduffy: I don't see `spottableDisabled` as tracking the overall spottability of the component which is the condition tested at L202. `spottableDisabled` would be more aptly named `spottedWhenBecameDisabled`. I'll take a cut at refactoring for readability and see if that clarifies things.

webOS101: Perhaps so.  Interestingly, if a component toggles off disabled while hovered the `onLeave` event does not fire so the component thinks it's still hovered.  If it's then restored by a component that is `spotlightDisabled` then it will gain focus even in pointer mode.  I think that's a pretty pathological set of circumstances, though.

webOS101: Hmm, I'm now not entirely sure of the mechanism for that...

packages/spotlight/Spottable/Spottable.js
```diff
@@ -177,34 +180,34 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
...
+		}
+
 		componentWillReceiveProps (nextProps) {
-			const spottableDisabled = this.isFocused && nextProps.disabled || nextProps.spotlightDisabled;
+			const focusedWhenDisabled = this.isFocused && nextProps.disabled && !this.props.disabled;
 
-			if (!this.state.spottableDisabled && spottableDisabled) {
-				this.setState({spottableDisabled});
-			}
+			this.setState({
```
germboy: Can you add the original guard back? Setting state to a value that may already exist will result in an unnecessary forced update. Seems like React could/should prevent against that ¯\_(ツ)_/¯

ryanjduffy: There _shouldn't_ be any "extra" updates caused by setting state in `componentWillReceiveProps`. It will already be in the middle of an update cycle due to new props.

webOS101: The latest React versions will not update if `setState` is called with `null`, but I don't think that setting it to an existing state does the same thing.

ryanjduffy: @webOS101 I don't believe that applies here. The handling of `setState` is different in terms of triggering re-renders when called within `componentWillReceiveProps` than elsewhere. Consider that its being migrated to a static method `getDerivedStateFromProps` (or whatever the name is). It's designed to generate state during a update cycle (initiated by a change of props).

Here's a quick example counting renders with a toggling `setState()` in `componentWillReceiveProps`.

https://codepen.io/ryanjduffy/pen/wjYmQw?editors=0010

webOS101: Sorry, it was a general comment, not specific to this case.  I understood that calling `setState` here causes no additional render.  I was speaking about calling it from elsewhere.

PR #1651, opened 2018-05-15
ENYO-5250: ContextualPopupDecorator to not steal focus on close

ryanjduffy:
`ContextualPopupDecorator` was indiscriminately setting focus to its activator on close regardless of where the focus may have been set under the assumption it had captured focus. If an app had programmatically set the focus elsewhere, the focus would be stolen. :sadpanda:

### Resolution

Since focusing the activator occurs in `componentDidUpdate` (good) but the container has already been unmounted then, I added a new state member, `shouldFocusActivator` which is calculated in `componentWillReceiveProps` to determine the focus behavior after update.

### DIFF
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -234,7 +234,8 @@ const ContextualPopupDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				arrowPosition: {top: 0, left: 0},
 				containerPosition: {top: 0, left: 0},
 				containerId: Spotlight.add(this.props.popupSpotlightId),
-				activator: null
+				activator: null,
+				shouldSpotActivator: true
 			};
 
 			this.overflow = {};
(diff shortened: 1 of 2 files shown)


packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -234,7 +234,8 @@ const ContextualPopupDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				arrowPosition: {top: 0, left: 0},
 				containerPosition: {top: 0, left: 0},
 				containerId: Spotlight.add(this.props.popupSpotlightId),
-				activator: null
+				activator: null,
+				shouldSpotActivator: true
```
viodragon2: It doesn't have to be a `state`, does it? If `shouldSpotActivator` is being assigned when `this.state.activator` gets reset, I don't see the point of making it a `state`.

ryanjduffy: I try to avoid members whenever possible. They run the risk of becoming out of sync with state since state updates are batched up.

PR #1652, opened 2018-05-15
ENYO-5271: Add debugging utilities to core/handle

ryanjduffy:
Debugging `core/handle` usage can be frustrating when everything is named `anonymous`.

### Resolution

* Added some names to internal functions that show up in the call stack
* Added `named()` to handlers to provide a means to name them when dev tools can't figure it out
* Added `bindAs()` to handlers to combine binding and naming when using `core/handle` with ES6 classes
* Adopting naming in `ui/Touchable` which is used everyone and could benefit from the naming the most

Was | Now
-----|-----
image | image

### Additional Considerations

See https://github.com/ariya/phantomjs/issues/14310 for why the `try`/`catch` block exists :( Its `__DEV__` only so a minor inconvenience.

### DIFF
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -90,11 +90,15 @@ const hasPropsAndContext = (obj) => {
 
 const named = (fn, name) => {
 	if (__DEV__) {
-		Object.defineProperty(fn, 'name', {
-			value: name,
-			writeable: false,
-			enumerable: false
-		});
+		try {
+			Object.defineProperty(fn, 'name', {
+				value: name,
+				writeable: false,
+				enumerable: false
+			});
+		} catch (err) {
+			// unable to set name of function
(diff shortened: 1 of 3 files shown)


packages/core/handle/handle.js
```diff
@@ -27,28 +27,31 @@
...
+ *   constructor () {
+ *     super();
+ *
+ *     // logEnter will be bound to `this` and set as this.handleKeyDown
+ *     //
+ *     // Equivalent to the following with the advantage of set the function name to be displayed in
+ *     // development tool call stacks
+ *     //
+ *     //   this.handleKeyDown = logEnter.bind(this)
+ *     logEnter.bindAs(this, 'handleKeyDown');
```
webOS101: Where does `logEnter` come from?

PR #1654, opened 2018-05-16
ENYO-5199: Fix ContextualPopupDecorator to properly stop keydown event propagation

viodragon2:
In react 16, [event  bubbling in `Portal`](https://reactjs.org/docs/portals.html#event-bubbling-through-portals) acts the same way as a normal React child, meaning that it will fire up the chain to the parent instead of the root node. This change has affected the event propagation path for `ContextualPopupDecorator` as it tries intercept keydown event to handle popup showing/closing. Since `keydown` event listener was attached to `document`, it no longer correctly stops event propagation. 

### Resolution
Proposed solution suggests to 
1. attach the `onKeyDown` handler directly to the popup container for handling the closing
2. only handle spotlight focus from the activator to the popup container at `document`

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -11,6 +11,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/IconButton` to allow external customization of the `Icon` vertical alignment (by setting `line-height`)
+- `moonstone/ContextualPopupDecorator` to properly stop propagating keydown event if fired from the popup container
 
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Added some minor refactoring mostly around optimizations and DRY-ness

* Moved the common event and spotlight logic into a new method, `handleDirectionalEvent`
* Used `direction` as an early bail-out in both keydown handlers to reduce code indent and avoid unnecessary processing
* Moved the spottable lookup after the `spotlightRestrict` prop check to avoid the expensive spottable determination when not necessary

PR #1659, opened 2018-05-17
ENYO-5255: Handle back key in MediaControls

viodragon2:
Added `onClose` prop to `moonstone/VideoPlayer.MediaControls` and updated `MediaControlsDecorator` to handle back key. 

When the back key is pressed, 
- It should close the media controls, if more components are NOT showing
- It should transition back to playback controls from more components, if more components are showing

Also updated the documentation to reflect recent changes in `VideoPlayer` and `MediaControls`

### Additional Consideration
In cases back keys are handled either by the `MediaControls` or `VideoPlayer`, the event will stop propagating. This was intentional for two reasons:
1. Since the states being managed for two cases are separated in MediaControls and VideoPlayer (i.e. visibility of MediaControls is managed by VideoPlayer and visibility of "more" components are managed by MediaControls), one cancel action can be handled by both components. We can create two separate callback props for each (e.g. `onCloseMore`, `onClose`), but  I wanted to only add one callback function prop for cancel.
2. Apps may add another `Cancelable` wrapped around the video player to hide the video player, and their condition might rely on `MediaControls`'s visibility . By the time `onCancel` event is handled in the wrapper, media controls would be hidden and apps would have no way to know if the back key was fired for hiding media controls or hiding video player. Another way is to add callback prop to `VideoPlayer` which only fires when back key is pressed and media controls aren't visible. This seems a bit out of reach for the general purpose of the video player, but I'm open for discussion.

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -37,14 +37,15 @@ const forwardToggleMore = forward('onToggleMore');
  * A set of components to control media playback and render additional components in designated
  * areas.
  *
- * @class MediaControls
+ * @class MediaControlsBase
  * @memberof moonstone/VideoPlayer
  * @ui
- * @public
+ * @private
  */
 const MediaControlsBase = kind({
 	name: 'MediaControls',
 
+	// intentionally assigning these props to MediaControls instead of Base (which is private)
 	propTypes: /** @lends moonstone/VideoPlayer.MediaControls.prototype */ {
 		/**
 		 * A string which is sent to the `backward` icon of the player controls. This can be
@@ -286,7 +287,7 @@ const MediaControlsBase = kind({
 		 * When `true`, more components are visible.
 		 *
 		 * @type {Boolean}
-		 * @public
+		 * @private
 		 */
 		showMoreComponents: PropTypes.bool,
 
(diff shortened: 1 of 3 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -652,6 +667,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 				if (selectedButton) {
 					selectedButton.blur();
 					selectedButton.focus();
+					Spotlight.focus(this.props.moreButtonSpotlightId);
```
viodragon2: @germboy: I may need your help on this. This seems a bit odd to me.
If the current spotted item is not the one that needs to be refocused, then I'd like to focus on `moreButton` determined by `moreButtonSpotlightId`. Hence this line.
The reason why I couldn't remove the `focus` call above is that if the previously focused component is the same, then highlight will not be restored when clicked with the pointer. (It's okay in 5-way)

Is there another way to properly set "(re)focus" to a component?

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -664,9 +664,11 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 
 				// Readout 'more' or 'back' button explicitly.
 				let selectedButton = Spotlight.getCurrent();
-				if (selectedButton) {
+				if (selectedButton === this.mediaControlsNode.querySelector(`.${css.moreButton}`)) {
```
viodragon2: Decided to check to whether more button is focused to separate the readout case and moving focus to more button.

Note that the focus will move to "more" button when back key is pressed no matter where it was focused, including `leftComponents`, `rightComponents` while showing more controls.

teckliew:
Should the spot be on the play/pause button when pressing back in more section?

viodragon2:
No. It should always focus back to "more" button when coming back from showing more section state.
cc: @xbinary

PR #1660, opened 2018-05-17
ENYO-5187: Fix VideoPlayer feedback timeouts

rundmt:
The video feedback wasn't working according to spec.

### Resolution
Some changes were made in the past to `FeedbackIcons` we're reverting some so `Feedback` and `MiniFeedback` work properly. Also added some logic so play should work properly now.

### Links
ENYO-5187

### DIFF
--- a/packages/moonstone/VideoPlayer/VideoPlayer.js
+++ b/packages/moonstone/VideoPlayer/VideoPlayer.js
@@ -902,6 +902,8 @@ const VideoPlayerBase = class extends React.Component {
 					miniFeedbackVisible: !(this.state.loading || !this.state.duration || this.state.error)
 				});
 			}
+		} else if (this.state.mediaControlsVisible && this.state.feedbackVisible && !this.state.feedbackIconVisible) {
+			this.setState({feedbackIconVisible: true});
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -902,6 +902,8 @@ const VideoPlayerBase = class extends React.Component {
 					miniFeedbackVisible: !(this.state.loading || !this.state.duration || this.state.error)
 				});
 			}
+		} else if (this.state.feedbackVisible && !this.state.feedbackIconVisible) {
```
viodragon2: Let's simplify this logic in one condition. I suggest something like:
```
    showFeedback = () => {
        if (this.state.mediaControlsVisible) {
            this.setState({
                feedbackVisible: true,
                feedbackIconVisible: true
            });
        } else {
```

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1534,8 +1536,9 @@ const VideoPlayerBase = class extends React.Component {
 	handleSliderBlur = () => {
 		this.sliderScrubbing = false;
 		this.startDelayedFeedbackHide();
+
 		this.setState({
-			feedbackIconVisible: true,
+			feedbackIconVisible: this.state.paused,
```
viodragon2: this is fine, but can we add inline comment why we use `this.state.paused`?

PR #1662, opened 2018-05-17
PLAT-53612: Add ability to select a range of a MediaSlider

ryanjduffy:

We have a use case for `VideoPlayer` in which the app wants to select a range of the `MediaSlider`

### Resolution

* Add `selection` and `onSeekOutsideSelection` props to `moonstone/VideoPlayer` to support selecting a range and notification of interactions outside of that range.
* Add `progressBarComponent` prop to `ui/Slider` as part of composition of progress bar by slider.
* Compose `ui/ProgressBar` within `moonstone/Slider`. Using `ui` version to minimize change impact due to current configuration of ProgressBar (Specifically its use of `percent` tooltip).
* Add `progressAnchor` prop to `ui/Slider` and `ui/ProgressBar`


packages/ui/Slider/Slider.js
```diff
@@ -214,19 +245,30 @@ const SliderBase = kind({
...
+		min,
+		orientation,
+		percent,
+		progressBarComponent,
+		progressAnchor,
+		tooltipComponent,
+		value = min,
+		...rest
+	}) => {
 		delete rest.backgroundProgress;
```
sugardave: Can remove this because it got destructured

packages/ui/Slider/Slider.less
```diff
@@ -2,27 +2,16 @@
 	// The "end" referenced below refers to the current "max value" of the proportion. If Slider
```
sugardave: Can you update this comment block (is it still necessary)?

PR #1663, opened 2018-05-17
ENYO-5276: ToggleButton padding fix

teckliew:
Set `ToggleButton` padding start to auto so it'll use the padding from Button.

### Links
ENYO-5276

### Comments

### DIFF
--- a/packages/moonstone/ToggleButton/ToggleButton.less
+++ b/packages/moonstone/ToggleButton/ToggleButton.less
@@ -8,7 +8,7 @@
 .toggleButton {
 	text-align: center;
 	position: relative;
-	.padding-start-end(@moon-toggle-button-text-switch-gutter, @moon-toggle-button-text-switch-gutter + @moon-button-h-padding + @moon-toggle-button-switch-width + 2*@moon-toggle-button-switch-border-width);
+	.padding-start-end(auto, @moon-toggle-button-text-switch-gutter + @moon-button-h-padding + @moon-toggle-button-switch-width + 2*@moon-toggle-button-switch-border-width);
 
 	&::after {
 		position: absolute;


ryanjduffy:
I think this is a workable solution.

@teckliew + @Djspaceg - For discussion, what do you think of adding an LTR class in I18nDecorator so we can safely distinguish these cases? We'd then be able to have a start or end mixin or the "null" parameter option we previously discussed. It does add another layer of specificity but there's some value in the consistent specificity that'd be applied.

```js
classes.push(base + isRtlLocale() ? 'right-to-left' : 'left-to-right');
```

```css
// Applies RTL-compatible start and end position to a selector
.padding-start-end (@start, @end) {
	:global(.enact-locale-left-to-right) & {
		padding-left: @start;
		padding-right: @end;
	}

	:global(.enact-locale-right-to-left) & {
		padding-left: @end;
		padding-right: @start;
	}
}
```

Djspaceg:
While this does work, I've been interested in avoiding using an LTR class for as long as is possible.

If given the choice, I'd lean more toward updating our `padding` and `margin` mixins to use `-webkit-padding-start:` and `-webkit-padding-end:` to keep things simpler and the class precedence stack smaller.

```
.padding-start-end (@start, @end) {
	-webkit-padding-start: @start;
	-webkit-padding-end: @end;
}
.padding-start (@start) {
	-webkit-padding-start: @start;
}
.padding-end (@end) {
	-webkit-padding-end: @end;
}
```

ryanjduffy:
I don't think we can include vendor-prefixed CSS in `ui`. They're tenable in moonstone which has some webkit affinity already but not elsewhere.

Djspaceg:
My example was terse. There are other vendor prefixed solutions for this scenario where the mixin would provide value: [`padding-block-start`](https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block-start). There may not be any edge support for this concept: https://caniuse.com/#search=padding-start

PR #1666, opened 2018-05-19
ENYO-4825: Add closeAllFloatingLayer to FloatingLayerDecorator

viodragon2:
`FloatingLayerDecorator` allows it to be used in a subtree of an app. In `moonstone/VideoPlayer` for example, `MediaControls` opens a video player's own floating layer and renders all popups in it so that it can be controlled by VideoPlayer. However, there's no way to close all the opened floating layer components.

### Resolution
Added `closeAllFloatingLayer`, `registerFloatingLayer`, `unregisterFloatingLayer` APIs as context in `FloatingLayerDecorator`. 


ryanjduffy:
Added an additional condition before calling `closeAllFloatingLayers` to validate that the current visible state as `false` in addition to validating that the visible state changed.

PR #1667, opened 2018-05-21
ENYO-4987: Set correct active Panel container id on render

germboy:
On certain platforms (chrome, ff), pointer events (`mouseover`, `mouseenter`, `mouseleave`) are emitted whether or not the pointer actually moves or if an element is moving/transitioning behind the non-moving pointer. Other platforms (TV, other browsers) do not emit these mouse events when elements transition behind a non-moving pointer.
This can cause a) de-synced spotlight states and b) varying spotlight behaviors for `SpotlightContainerDecorator` and `Spottable` components within `Panels` when they transition past the pointer.

### Resolution
The initial thought was: since these mouse events are being triggered during an animation/transition, we could likely prevent mouse actions while spotlight is paused - since we pause spotlight during `Panel` transitions. However, these mouse events get queued up throughout the transition and ultimately get emitted after `transitionend` - when spotlight is not longer paused. That isn't helpful.

There seems to be an inherent difference in how platforms handle this scenario, so I feel the best solution is to normalize how all platforms perform by preventing actions within the mouse events unless the pointer has actually moved (similarly to what we do in spotlight for `mousemove`). Then we can set the correct active container id on `Panel` render (similarly to how we do with `Popup` for the same reasons).

### Additional Considerations
1) Instead of requiring `Spottable` and `SpotlightContainerDecorator` to individually track pointer movement, I wonder if we'd want to expose spotlight's existing internal methods for this vs allowing each `hoc` to be more self-contained.
2) To go with the above item, the current fix causes our test cases to require additional work by attaching mouse positions to mouse events, in order to simulate movement.
3) The proposed fix in this PR sets the correct container id for `Panel` after transitioning. If a panel transitions past a non-moving pointer and the pointer ultimately hides (on TV), the `Panel` gains focus, as expected. However, if a `Spottable` component transitions and ultimately stops under a non-moving pointer, we are in an edge-case scenario where the component is not visibly focused and its internal state is neither spotted nor hovered. When the non-moving pointer hides in this scenario, the default panel component will be focused. I'm not sure of a great solution for this. It should be noted, this edge-case existed prior to this PR ... it's just something to be aware of.


packages/spotlight/Spottable/Spottable.js
```diff
@@ -334,12 +335,16 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 
 		handleEnter = (ev) => {
 			forward('onMouseEnter', ev, this.props);
-			this.isHovered = true;
+			if (hasPointerMoved(ev.clientX, ev.clientY)) {
```
webOS101: Can we guarantee the ordering of events such that the move will always happen -after- the enter/leave?

germboy: I can't find/link to a definitive spec, however in my experience across platforms (Chrome, FF, Edge, TV) the `mousemove` event always fires after the other related mouse events: `mouseover`, `mouseout`, `mouseenter`, and `mouseleave`

webOS101: OK!

PR #1670, opened 2018-05-21
ENYO-5280: Fix MarqueeController incorrectly canceling valid animations

ryanjduffy:

MarqueeController batches up cancel animation requests so that multiple synced marquees don't incur extra work when multiple cancel at the same time. However, if the cancellation was due to a change in prop and the synced marquees should start on render, they will likely be cancelled (when the job timer fires) before they can start (after the `marqueeOnRenderDelay`).

### Resolution

* Prevent starting the cancel job if there aren't any marquees running.
* (Incidentally) Avoid trying to start a marquee if `marqueeDisabled`

### DIFF
--- a/packages/ui/Marquee/MarqueeDecorator.js
+++ b/packages/ui/Marquee/MarqueeDecorator.js
@@ -396,8 +396,8 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		 * @returns {Boolean} - `true` if a possible marquee condition exists
 		 */
 		shouldStartMarquee () {
-			const {disabled, marqueeOn} = this.props;
-			return (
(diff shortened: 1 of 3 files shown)


packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -396,8 +396,8 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		 * @returns {Boolean} - `true` if a possible marquee condition exists
 		 */
 		shouldStartMarquee () {
-			const {disabled, marqueeOn} = this.props;
-			return (
+			const {disabled, marqueeDisabled, marqueeOn} = this.props;
+			return !marqueeDisabled && (
```
viodragon2: I'm assuming this isn't required for the fix, but it's nice to have.

PR #1672, opened 2018-05-23
ENYO-5153: Fix Button's sampler to toggle minWidth correctly

viodragon2:
`minWidth` prop's default value is set to `true`, and by `nullifying` the boolean value, it will always set to `true` even if the knob is set to `false`. Fixed by removing `nullifying` in the select knob.

### Additional Considerations
- Button's props weren't being properly merged in the info section and made an update to it. It seems like other components that uses `ui` as a base component show similar pattern and we need to update them all.
- Any props that were added by HOCs (e.g. `casing` prop from `i18n/Uppercase`) will not be fetched and need to be set with constants. Should we declare those `propTypes` in the base component or is there a way to fetch them and feed it to `propTables` in the story?

### DIFF
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -49,7 +49,7 @@ const ButtonBase = kind({
 		 * @type {String}
 		 * @public
 		 */
-		backgroundOpacity: PropTypes.oneOf([null, 'translucent', 'lightTranslucent', 'transparent']),
+		backgroundOpacity: PropTypes.oneOf(['translucent', 'lightTranslucent', 'transparent']),
 
 		/**
 		 * The color of the underline beneath button's content. Used for `IconButton`.
@@ -60,7 +60,7 @@ const ButtonBase = kind({
 		 * @type {String}
 		 * @public
 		 */
-		color: PropTypes.oneOf([null, 'red', 'green', 'yellow', 'blue']),
+		color: PropTypes.oneOf(['red', 'green', 'yellow', 'blue']),
 
 		/**
 		 * Customizes the component by mapping the supplied collection of CSS class names to the
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -86,6 +86,10 @@ const ButtonBase = kind({
 		css: PropTypes.object
 	},
 
+	defaultProps: {
+		backgroundOpacity: 'opaque'
+	},
+
 	styles: {
 		css: componentCss,
 		publicClassNames: ['button', 'bg', 'selected', 'small']
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -42,21 +42,14 @@ const ButtonBase = kind({
 		 * The background-color opacity of this button.
 		 *
 		 * Valid values are:
-		 * * `'opaque'`,
 		 * * `'translucent'`,
 		 * * `'lightTranslucent'`, and
 		 * * `'transparent'`.
 		 *
 		 * @type {String}
-		 * @default 'opaque'
 		 * @public
 		 */
-		backgroundOpacity: PropTypes.oneOf([
-			'opaque',
-			'translucent',
-			'lightTranslucent',
-			'transparent'
-		]),
+		backgroundOpacity: PropTypes.oneOf([null, 'translucent', 'lightTranslucent', 'transparent']),
 
 		/**
 		 * The color of the underline beneath button's content. Used for `IconButton`.
@@ -84,10 +77,6 @@ const ButtonBase = kind({
 		 * @public
 		 */
 		css: PropTypes.object
-	},
-
-	defaultProps: {
-		backgroundOpacity: 'opaque'
 	},
 
 	styles: {
--- a/packages/moonstone/IconButton/IconButton.js
+++ b/packages/moonstone/IconButton/IconButton.js
@@ -43,16 +43,14 @@ const IconButtonBase = kind({
 		 * The background-color opacity of this icon button
 		 *
 		 * Valid values are:
-		 * * `'opaque'`,
 		 * * `'translucent'`,
 		 * * `'lightTranslucent'`, and
 		 * * `'transparent'`.
 		 *
 		 * @type {String}
-		 * @default 'opaque'
 		 * @public
 		 */
-		backgroundOpacity: PropTypes.oneOf(['opaque', 'translucent', 'lightTranslucent', 'transparent']),
+		backgroundOpacity: PropTypes.oneOf([null, 'translucent', 'lightTranslucent', 'transparent']),
 
 		/**
 		 * The color of the underline beneath the icon.
--- a/packages/moonstone/IconButton/IconButton.js
+++ b/packages/moonstone/IconButton/IconButton.js
@@ -50,7 +50,7 @@ const IconButtonBase = kind({
 		 * @type {String}
 		 * @public
 		 */
-		backgroundOpacity: PropTypes.oneOf([null, 'translucent', 'lightTranslucent', 'transparent']),
+		backgroundOpacity: PropTypes.oneOf(['translucent', 'lightTranslucent', 'transparent']),
 
 		/**
 		 * The color of the underline beneath the icon.
@@ -61,7 +61,7 @@ const IconButtonBase = kind({
 		 * @type {String}
 		 * @public
 		 */
-		color: PropTypes.oneOf([null, 'red', 'green', 'yellow', 'blue']),
+		color: PropTypes.oneOf(['red', 'green', 'yellow', 'blue']),
 
 		/**
 		 * Customizes the component by mapping the supplied collection of CSS class names to the
--- a/packages/sampler/stories/moonstone-stories/Button.js
+++ b/packages/sampler/stories/moonstone-stories/Button.js
@@ -14,7 +14,8 @@ const Config = mergeComponentMetadata('Button', ButtonBase, Button, UIButton, UI
 
 // Set up some defaults for info and knobs
 const prop = {
-	backgroundOpacity: ['opaque', 'translucent', 'lightTranslucent', 'transparent'],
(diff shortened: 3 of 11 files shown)


packages/sampler/stories/moonstone-stories/Button.js
```diff
@@ -9,11 +10,11 @@ import {withInfo} from '@storybook/addon-info';
 import {mergeComponentMetadata} from '../../src/utils/propTables';
 import nullify from '../../src/utils/nullify.js';
 
-const Config = mergeComponentMetadata('Button', ButtonBase, Button);
+const Config = mergeComponentMetadata('Button', ButtonBase, Button, UIButton, UIButtonBase);
 
 // Set up some defaults for info and knobs
 const prop = {
-	backgroundOpacity: ['', 'translucent', 'lightTranslucent', 'transparent'],
+	backgroundOpacity: ['opaque', 'translucent', 'lightTranslucent', 'transparent'],
```
Djspaceg: Having no blank (`''`) option implies that the component must use this prop. `backgroundOpacity` is optional and the sample should represent that. I'm fine with _also_ having a blank/opaque option, but not ok with replacing the blank with 'opaque'.

packages/moonstone/Button/Button.js
```diff
@@ -86,6 +86,10 @@ const ButtonBase = kind({
 		css: PropTypes.object
 	},
 
+	defaultProps: {
+		backgroundOpacity: 'opaque'
```
Djspaceg: I'd almost prefer that opaque be removed from the available options, rather than have it be always applied as a default, so the default (and default applied class) can just the normal resting state of the Button. "Opaque" was originally added just for consistency if we needed it, but we don't need it yet at this point.

viodragon2: From documentation perspective, it is set to `@default 'opaque'`. But it seemed the property wasn't being set when rendered. Wanted to fill in the gap. I think it'd be okay to remove the `defaultProps` as long as it is aligned with the documentation.

packages/sampler/stories/moonstone-stories/Button.js
viodragon2: Similar to the comment below, it seemed awkward to me when Button's `backgroundOpacity` has `@default 'opaque` in the documentation, but not set anything in the sampler. Plus, it is documented as one of the options but the user cannot select it in the sampler. I thought that was inconsistent. Thoughts?

Djspaceg: I agree. Let's remove the `@default 'opaque'` from the docs. It was put there as a way to explain that the "normal" state of a Button's appearance is opaque, and not translucent, so PropTypes wouldn't error if someone attempted to use a logical word for an opacity property, even though it's just, and already, _the default_.

`''` in this case, due to `nullify`, below removes `backgroundOpacity` from the JSX entirely, so it appears as `<Button>` rather than implying there's a required property: `<Button backgroundOpacity="opaque">`.

We're getting into "developer experience" (DX) territory here, rather than discussing `minWidth`. A fundamental question like "should an empty prop value be enumerated in an array of valid options so it's clear to a developer that they can specify nothing and get the default on a non-required property". My preference is "no", because as a developer I want to see what I'm _practically_ able to set a prop to, not an exhaustive set of values. (We don't list `null` or `undefined` as options, even though those technically also work.) If I choose to not set it in my app, I'll just exclude that prop from my code and expect it to work. I definitely don't want developers to think they must assign an empty value: `<Button backgroundOpacity="">` to make a simple button work. That being said, we do have to sort of play nicely with Storybook stories, and massage them into accurately representing our options. There is no way to temporarily _remove_ a knob from the Knobs list, so we need a way to indicate, in our existing `select` knob, that not providing a prop is a valid option. This is where the `''` comes in (as the best, most representative option we've found so far). It's necessary to show that not choosing a value is an acceptable option.

"opaque" was not listed as a knob value because I didn't want to cause confusion about what the values mean. If `''` and "opaque" are both present, it could imply that they should visually represent two different states, i.e. whatever the default/blank is, and opaque, which must be different, because it's listed as a separate option.

If we **must** change the documentation, PropTypes, and Story, I'd personally prefer to remove the `@default 'opaque'` and leave the `''` in the knob values. The documentation will also need to be updated to describe that Button, by default, is visually opaque but that an "opaque" string value is not a valid option for the `backgroundOpacity` prop, so it's clear what is going on.

packages/moonstone/Button/Button.js
Djspaceg: I agree. Let's remove the `@default 'opaque'` from the docs. It was put there as a way to explain that the "normal" state of a Button's appearance is opaque, and not translucent, so PropTypes wouldn't error if someone attempted to use a logical word for an opacity property, even though it's just, and already, _the default_.

`''` in this case, due to `nullify`, below removes `backgroundOpacity` from the JSX entirely, so it appears as `<Button>` rather than implying there's a required property: `<Button backgroundOpacity="opaque">`.

We're getting into "developer experience" (DX) territory here, rather than discussing `minWidth`. A fundamental question like "should an empty prop value be enumerated in an array of valid options so it's clear to a developer that they can specify nothing and get the default on a non-required property". My preference is "no", because as a developer I want to see what I'm _practically_ able to set a prop to, not an exhaustive set of values. (We don't list `null` or `undefined` as options, even though those technically also work.) If I choose to not set it in my app, I'll just exclude that prop from my code and expect it to work. I definitely don't want developers to think they must assign an empty value: `<Button backgroundOpacity="">` to make a simple button work. That being said, we do have to sort of play nicely with Storybook stories, and massage them into accurately representing our options. There is no way to temporarily _remove_ a knob from the Knobs list, so we need a way to indicate, in our existing `select` knob, that not providing a prop is a valid option. This is where the `''` comes in (as the best, most representative option we've found so far). It's necessary to show that not choosing a value is an acceptable option.

"opaque" was not listed as a knob value because I didn't want to cause confusion about what the values mean. If `''` and "opaque" are both present, it could imply that they should visually represent two different states, i.e. whatever the default/blank is, and opaque, which must be different, because it's listed as a separate option.

If we **must** change the documentation, PropTypes, and Story, I'd personally prefer to remove the `@default 'opaque'` and leave the `''` in the knob values. The documentation will also need to be updated to describe that Button, by default, is visually opaque but that an "opaque" string value is not a valid option for the `backgroundOpacity` prop, so it's clear what is going on.

packages/sampler/stories/moonstone-stories/Button.js
Djspaceg: I agree. Let's remove the `@default 'opaque'` from the docs. It was put there as a way to explain that the "normal" state of a Button's appearance is opaque, and not translucent, so PropTypes wouldn't error if someone attempted to use a logical word for an opacity property, even though it's just, and already, _the default_.

`''` in this case, due to `nullify`, below removes `backgroundOpacity` from the JSX entirely, so it appears as `<Button>` rather than implying there's a required property: `<Button backgroundOpacity="opaque">`.

We're getting into "developer experience" (DX) territory here, rather than discussing `minWidth`. A fundamental question like "should an empty prop value be enumerated in an array of valid options so it's clear to a developer that they can specify nothing and get the default on a non-required property". My preference is "no", because as a developer I want to see what I'm _practically_ able to set a prop to, not an exhaustive set of values. (We don't list `null` or `undefined` as options, even though those technically also work.) If I choose to not set it in my app, I'll just exclude that prop from my code and expect it to work. I definitely don't want developers to think they must assign an empty value: `<Button backgroundOpacity="">` to make a simple button work. That being said, we do have to sort of play nicely with Storybook stories, and massage them into accurately representing our options. There is no way to temporarily _remove_ a knob from the Knobs list, so we need a way to indicate, in our existing `select` knob, that not providing a prop is a valid option. This is where the `''` comes in (as the best, most representative option we've found so far). It's necessary to show that not choosing a value is an acceptable option.

"opaque" was not listed as a knob value because I didn't want to cause confusion about what the values mean. If `''` and "opaque" are both present, it could imply that they should visually represent two different states, i.e. whatever the default/blank is, and opaque, which must be different, because it's listed as a separate option.

If we **must** change the documentation, PropTypes, and Story, I'd personally prefer to remove the `@default 'opaque'` and leave the `''` in the knob values. The documentation will also need to be updated to describe that Button, by default, is visually opaque but that an "opaque" string value is not a valid option for the `backgroundOpacity` prop, so it's clear what is going on.

packages/moonstone/Button/Button.js
```diff
@@ -42,21 +42,14 @@ const ButtonBase = kind({
...
-		 * @default 'opaque'
 		 * @public
 		 */
-		backgroundOpacity: PropTypes.oneOf([
-			'opaque',
-			'translucent',
-			'lightTranslucent',
-			'transparent'
-		]),
+		backgroundOpacity: PropTypes.oneOf([null, 'translucent', 'lightTranslucent', 'transparent']),
```
Djspaceg: The literal `null` can (should) be omitted entirely from this list, since a `null` is implied by simply not including the `backgroundOpacity` property. We're not saying that `<Button backgroundOpacity={null}>` should be used in practice, but instead saying that `<Button>` is a valid option. If this prop were "required", and it needed to support a non-value (for whatever reason), then I'd say we should to keep the `null` literal.

viodragon2: I had a bit of confusion with `color` having `null` value as an option. I'll go ahead and make that change as well.

FYI: I don't think there's a good way to include `null` as prop type when `isRequired`. See https://github.com/facebook/react/issues/3163

packages/moonstone/IconButton/IconButton.js
```diff
@@ -63,7 +61,7 @@ const IconButtonBase = kind({
 		 * @type {String}
 		 * @public
 		 */
-		color: PropTypes.oneOf([null, 'red', 'green', 'yellow', 'blue']),
+		color: PropTypes.oneOf(['red', 'green', 'yellow', 'blue']),
```
Djspaceg: 👍 Good catch. Didn't see this one initially.

PR #1674, opened 2018-05-23
ENYO-5070 - Enact Picker: Text Unexpectedly Continues to Marquee

rundmt:
Enact Picker would still marquee even when you 5way off picker if the picker was previously hovered.

### Resolution
Added some TV specific code to check for when the pointer hides. 

### Links
ENYO-5070

### DIFF
--- a/packages/ui/Marquee/MarqueeController.js
+++ b/packages/ui/Marquee/MarqueeController.js
@@ -3,6 +3,7 @@ import hoc from '@enact/core/hoc';
 import {Job} from '@enact/core/util';
 import React from 'react';
 import PropTypes from 'prop-types';
+import Spotlight from '@enact/spotlight';
 
 const STATE = {
 	inactive: 0,	// Marquee is not necessary (render or focus not happened)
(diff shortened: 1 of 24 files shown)


packages/ui/Marquee/MarqueeController.js
```diff
@@ -135,8 +137,23 @@ const MarqueeController = hoc(defaultConfig, (config, Wrapped) => {
 			};
 		}
 
+		componentDidMount () {
+			if (platform.webos) {
```
ryanjduffy: @teckliew + @rundmt wdyt of removing these checks?

Seems like it would unnecessarily restrict handling the event. It complicates testing in Chrome (though you can simulate the webos UA string) but more importantly would preclude this from working on another platform just to avoid setting a key handler. We could avoid that overhead by using `core/dispatcher` which will manage setting the single listener on `document`.

rundmt: Yeah I don't mind removing them.

teckliew: Sounds good. `pointerHide` should take care of the platform specific handling.

rundmt:
@ryanjduffy Updated! Should be good now.

PR #1683, opened 2018-05-29
ENYO-5151: Adjust Spotlight Partitions/Priorities

germboy:
Spotlight sometimes fails to select the expected "most natural" spottable element when using 5-way. We adjusted the spotlight partitions and priorities in #1495. That fix appears to be ever-so-slightly heavy-handed in its execution - as there are rare cases where 5-way does not perform as expected after that change.

### Resolution
When pressing a 5-way directional key, spotlight calculates where the "next" element is located on a 9-tile grid location, relative to the currently focused element (top-left, top-center, top-right, middle-left, etc). The correct grid location of "next" (when not an internal group) should be based on an additional factors we aren't taking into consideration at the time - 5-way direction pressed. The "next" element could actually live in multiple grid locations. We are able to isolate and identify which grid location to target based on 5-way direction pressed.

The solution is to revert the original calculation, unless navigating vertically. Spotlight internal group calculations are not relevant or affected by this change.

### Links
Issue caused by #1495

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -13,6 +13,10 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 No significant changes.
 
+### Fixed
+
+- `spotlight` to partition and prioritize next spottable elements for more natural 5-way behavior
+
 ## [2.0.0-beta.4] - 2018-05-21
 
 ### Fixed
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -6,16 +6,13 @@ The following is a curated list of changes in the Enact spotlight module, newest
  
 ### Fixed
 
+- `spotlight` to partition and prioritize next spottable elements for more natural 5-way behavior
 - `spotlight` to handle pointer events only when pointer has moved
 - `spotlight` to correctly set the active container id when unable to set focus to a container
 
 ## [2.0.0-beta.5] - 2018-05-29
 
 No significant changes.
-
-### Fixed
-
-- `spotlight` to partition and prioritize next spottable elements for more natural 5-way behavior
 
 ## [2.0.0-beta.4] - 2018-05-21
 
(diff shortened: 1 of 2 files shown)


ryanjduffy:
> The solution is to revert the original calculation, unless navigating vertically.

What is unique about vertical navigations?

germboy:
Nothing is necessarily unique about vertical navigations, other than how spotlight decides to designate a grid location for a "next" element that could be a target both vertically and horizontally. Take the following layout:

	      -------------
	        Button A
	      -------------
			-------------
			   Button B
			-------------
Prior to #1495, spotlight favored a more horizontal ruleset, placing "Button A" as left-only (not within/middle) from "Button B", meaning it wasn't a 1st choice element when navigating vertically "up" from "Button B".
```
// spotlight/src/navigate.js
if (center.x < destRect.left) {
	x = 0;
} else if (center.x <= destRect.right) {
	x = 1;
} else {
	x = 2;
}
```
So `x` was being set to `0`, while in reality it could also be `1` (based on which 5-way direction was pressed). For the majority of our general uses-cases, it's most likely that when an element is valid for multiple grid locations (where one location is within/middle), it's in a vertical orientation. Elements that are in a horizontal layout are generally less jumbled and less likely to be eligible targets of both straight *and* oblique.

#1495 adjusted this behavior, but did so a little too strongly. It fixed the issues with vertical navigation grid positioning, but it resulted in scenarios where it could cause issues (albeit, much fewer issues) when navigating horizontally. It attempted to use a singular modified calculation to determine which grid position to place the "next" element. Take the following layout:

	-----------------------------
	              Button C
	-----------------------------
	----------
	Button D
	----------
When pressing 5-way right from "Button D", the updated code was treating "Button C" as "within/middle" instead of right.
```
if (rect.right <= destRect.left) {
	x = 0;
} else if (rect.left < destRect.right) {
	x = 1;
} else {
	x = 2;
}
```
The solution is ultimately a mix of the 2 calculations. We can determine which to use based on whether we are navigating horizontally or vertically.

ryanjduffy:
@germboy Great explanation. Thanks!

PR #1684, opened 2018-05-30
Update resolution workspaceBounds to window dimensions during Resolut…

JayCanuck:
…ionDecorator's constructor to fix snapshot sizing.

* In v8 snapshot, the modular level initial workspaceBounds fallback to 1080p. So when the snapshot app loads into the window, it loads as 1080p even on larger/smaller screens.

### Resolution
* Use the `window` node during the `ResolutionDecorator` constructor to update the workspace bounds.


Djspaceg:
ok I'll merge it

PR #1686, opened 2018-05-30
ENYO-5231: Fix scrolling by page down at Scroller bounds

ryanjduffy:
`moonstone/Scroller` was not passing a container id to spotlight when attempting to determine the next spottable by position. As a result, spotlight was ignoring the spottables that were out of bounds and selecting a different, in bounds, target.

### Resolution
Update `moonstone/Scroller.scrollToNextPage` to access `spotlightId` instead of `containerId` which is the correct object member passed to the function.


0x64:
LGTM

PR #1687, opened 2018-05-31
ENYO-5268: Fix MediaControls to only handle directional events when player has focus

viodragon2:
Attaching key event handlers to `document` for media control is not desirable and we need to confine the scope of handling to the player. 
~Suggest passing down the player node to `MediaControls` and attach key event handlers to it.~
Suggest handling key down event when current focus is in `VideoPlayer` 

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -572,6 +572,14 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			paused: PropTypes.bool,
 
 			/**
+			 * A DOM Node of the player.
+			 *
+			 * @type {Node}
+			 * @public
+			 */
+			playerNode: PropTypes.node,
+
+			/**
 			 * Sets the `disabled` state on the media playback-rate control buttons; the inner pair.
 			 *
 			 * @type {Boolean}
@@ -637,8 +645,8 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 				countReactChildren(this.props.rightComponents),
 				countReactChildren(this.props.children)
 			);
-			on('keydown', this.handleKeyDown);
-			on('keyup', this.handleKeyUp);
+			on('keydown', this.handleKeyDown, this.props.playerNode);
+			on('keyup', this.handleKeyUp, this.props.playerNode);
 		}
 
 		componentWillReceiveProps (nextProps) {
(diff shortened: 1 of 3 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -571,6 +571,14 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			 */
 			paused: PropTypes.bool,
 
+			/**
+			 * A DOM Node of the player.
+			 *
+			 * @type {Node}
+			 * @public
+			 */
+			playerNode: PropTypes.node,
```
teckliew: This fails with DOM nodes. Maybe use `PropTypes.object`?

ryanjduffy:
By listening on the node rather than `document`, we're no longer able to jump forward/backward when in pointer mode. The first left/right press cancels pointer mode and then another keypress is required to jump. Further, I'm concerned this is introducing some awkward coupling between MediaControls and VideoPlayer by passing the node down.

Would an alternative be to inspect the currently focused item and only jump if nothing is focused or if the focused item is the placeholder?

```
		handleKeyDown = (ev) => {
			const {
				mediaDisabled,
				no5WayJump,
				visible
			} = this.props;

			const current = Spotlight.getCurrent();

			if (!no5WayJump &&
					!visible &&
					!mediaDisabled &&
					(!current || current.classList.contains(css.controlsHandleAbove)) &&
					(is('left', ev.keyCode) || is('right', ev.keyCode))) {
				this.paused.pause();
				this.startListeningForPulses(ev.keyCode);
			}
		}
```

viodragon2:
Hmm. I am not sure handling key events in pointer mode was intended. Most key events only deactivate the pointer mode and do not function other than the enter key. It's rare to handle pointer hide and the direction key at the same time (though we handle "down" from pointer mode to show controls in `VideoPlayer` 🤔). Given that left/right key does jump in pointer mode in 2.0.0-alpha.4, I am not worried too much about the change.

I generally like the idea of you your suggestion, but isn't `css.controlsHandleAbove` out of the scope of `MediaControls`? Should it be passed down from `VideoPlayer`? Or is there a way to detect whether spotlight is inside `MediaControls` (i.e. `<div data-media-controls>`)?

ryanjduffy:
> I generally like the idea of you your suggestion, but isn't css.controlsHandleAbove out of the scope of MediaControls? Should it be passed down from VideoPlayer?

It is still breaking encapsulation but I'm not terribly concerned about it within the same module. We do run the risk of an upstream change to VideoPlayer (e.g. renaming that class or some other change to the spotlight placeholder) causing a regression without some explicit integration that would have to be updated as well.

If we wanted to provide that integration, I'd like to aim for a more explicit prop indicating the current state (that the video player has focus) rather than the class name (or some other indirect data element).

> Or is there a way to detect whether spotlight is inside MediaControls (i.e. <div data-media-controls>)?

Focus can't be within MediaControls since this hander takes effect when they are hidden.

PR #1688, opened 2018-05-31
ENYO-5293: Fix FloatingLayer to render if already opened at mounting time

viodragon2:
If `FloatingLayer` is `open`ed at mounting time, it does not render correctly. More specifically, the `this.context.getFloatingLayer()` would not be available until it has been mounted, which in turn will fail to create a node to attach to the `floatingLayer` DOM tree.

I believe we haven't run into the issue as rerendering of the parent component usually resolves the problem. (e.g. `View`, animated `Popup`, etc.)

### Resolution
Created a state `nodeRendered` so that `FloatingLayer` only renders the children when it is inserted into DOM tree.

### DIFF
--- a/packages/ui/FloatingLayer/FloatingLayer.js
+++ b/packages/ui/FloatingLayer/FloatingLayer.js
@@ -10,10 +10,14 @@ import Cancelable from '../Cancelable';
 import {contextTypes} from './FloatingLayerDecorator';
 import Scrim from './Scrim';
 
-const forwardDismiss = adaptEvent(
-	() => ({type: 'onDismiss'}),
-	forward('onDismiss')
+const forwardWithType = type => adaptEvent(
+	() => ({type}),
+	forward(type)
 );
+
+const forwardDismiss = forwardWithType('onDismiss');
+const forwardClose = forwardWithType('onClose');
+const forwardOpen = forwardWithType('onOpen');
 
 /**
  * {@link ui/FloatingLayer.FloatingLayerBase} is a component that creates an entry point to the new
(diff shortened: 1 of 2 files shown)


packages/ui/FloatingLayer/FloatingLayer.js
```diff
@@ -193,27 +197,30 @@ class FloatingLayerBase extends React.Component {
 		const {floatLayerClassName} = this.props;
 		const floatingLayer = this.context.getFloatingLayer();
 
-		if (!this.node && floatingLayer && typeof document !== 'undefined') {
-			invariant(
-				this.context.getFloatingLayer,
-				'FloatingLayer cannot be used outside the subtree of a FloatingLayerDecorator'
-			);
+		if (this.node || !floatingLayer) return;
```
viodragon2: don't we need to check for `typeof document !== 'undefined'` for snapshot?

ryanjduffy: It's only necessary in module-scoped code and in the constructor (and componentWillMount). `componentDidMount` and `componentWillReceiveProps` aren't called during pre-render.

packages/ui/FloatingLayer/FloatingLayer.js
```diff
@@ -179,31 +194,33 @@ class FloatingLayerBase extends React.Component {
...
+		this.node = document.createElement('div');
+		this.node.className = floatLayerClassName;
+		this.node.style.zIndex = 100;
 
-		return this.node;
+		floatingLayer.appendChild(this.node);
+		on('scroll', this.handleScroll, this.node);
+
+		// render children when this.node is inserted in the DOM tree.
+		this.setState({nodeRendered: true});
```
Djspaceg: Should we be using the newer `setState` technique?

PR #1693, opened 2018-05-31
PLAT-54239: Storybook - Set correct Display Name and remove Prop Table

teckliew:
Set the correct component display names for JSX source in Storybook sampler for the components that are wrapped in HOC's. Also removed the prop table for simplicity because props from different layers of the component are sometimes not shown or are misrepresented. 

### Links
PLAT-54239

### Comments


Djspaceg:
I made a bunch of changes here and added some awesome new knobs. 

Basically, `enact-knobs` does a bunch of convenient stuff by accepting the knob/prop name, a `Config` which is just an object that has a `defaultProps` object in it with the defaults, and the displayName, and an optional default override value. I've included docs on each of the knobs. It then groups the knobs in storybook into tabs. You'll need to be sure to update storybook's knobs to v3.4 at least, by doing an `npm install` to get the new tabs feature.

Also of note, I switched the internal knob from `select` to `selectV2` (the future knob format) which reverses the key/value prop/label relationship. Now, it's `{'Human Readable Label': 'prop value'}`.

teckliew:
added `object` to `enact-knobs` and also `utils/emptify` for knobs that needed to output into empty string when given falsey values.

PR #1697, opened 2018-06-01
ENYO-5319: Fix ContextualPopupDecorator to allow focusing components under its popup

ryanjduffy:
The `keydown` handler in `ContextualPopupDecorator` was hijacking the 5-way event to focus the popup when it didn't have any target to focus.

### Resolution
Alter the condition to only attempt to focus the popup when it has spottable descendants.

### DIFF
--- a/docs/creating-components.md
+++ b/docs/creating-components.md
@@ -38,7 +38,7 @@ For the following sample, the `'Badge'` component accepts the `children` and `gr
 `greeting` having a default of `'Hello, my name is ...'`. It applies the `'badge'` `className` (combined
 with any passed-in `className`), it computes a new value for `children` and renders the result.
 
-```
+```javascript
 import kind from '@enact/core/kind';
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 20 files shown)


packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -490,10 +490,9 @@ const ContextualPopupDecorator = hoc(defaultConfig, (config, Wrapped) => {
 
 			if (!direction) return;
 
-			const spotlessSpotlightModal = spotlightRestrict === 'self-only' && (
-				Spotlight.getSpottableDescendants(containerId).length === 0
-			);
-			const shouldSpotPopup = current === activator && direction === this.adjustedDirection;
+			const hasSpottables = Spotlight.getSpottableDescendants(containerId).length > 0;
+			const spotlessSpotlightModal = spotlightRestrict === 'self-only' && !hasSpottables;
+			const shouldSpotPopup = current === activator && direction === this.adjustedDirection && hasSpottables;
```
webOS101: Technically line 502 below could be shortened to just `if (!spotlessSpotlightModal) {` because the conditional is redundant with 497.  not changing it, tho.

PR #1698, opened 2018-06-01
ENYO-5309 - Scroller added spotlightPaging prop

rundmt:
ExpandableList inside of a scroller has incorrect spotlight focus when paging up/down.

### Resolution
Added `spotlightPaging` prop to allow the app developer to choose how they want to handle paging. When set to the default of `scroller` it will work the way it currently works, but if set to `container` it will only spot within the container that the spotlight is currently in.

### Links
ENYO-5309

Enact-DCO-1.0-Signed-off-by: Derek Tor (derek.tor@lge.com)


ryanjduffy:
We've decided not to pursue this change at this time.

PR #1704, opened 2018-06-04
ENYO-5322: Fix floatingLayer unmount condition

viodragon2:
A recent change (#1688) in `FloatingLayer` has adopted the recommendation of react portals, and added `removeChild` in `componentWillUnmount`. However, `removeChild` should be guarded against `null`. 


ryanjduffy:
So if the `FloatingLayer` is mounted but never opened, `this.node` is `null` and therefore never attached and so it doesn't need to be removed?

viodragon2:
Correct

PR #1710, opened 2018-06-06
ENYO-5047: Add QA sample for ExpandableList with default selection

teckliew:
QA-sample add default selected ExpandableList

### Links
ENYO-5047

### Comments

### DIFF
--- a/packages/sampler/stories/qa-stories/ExpandableList.js
+++ b/packages/sampler/stories/qa-stories/ExpandableList.js
@@ -1,5 +1,6 @@
 import Button from '@enact/moonstone/Button';
 import ExpandableList from '@enact/moonstone/ExpandableList';
+import Scroller from '@enact/moonstone/Scroller';
 import {RadioControllerDecorator} from '@enact/ui/RadioDecorator';
 import React from 'react';
 import {storiesOf} from '@storybook/react';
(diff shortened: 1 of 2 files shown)


packages/sampler/stories/qa-stories/ExpandableList.js
```diff
@@ -73,24 +80,18 @@ storiesOf('ExpandableList', module)
 		)
 	)
 	.add(
-		'with multiples (to test "lockBottom" prop)',
+		'with default selected',
 		() => (
-			<div>
-				<ExpandableList title="First">
-					{['One', 'Two', 'Three']}
+			<Scroller style={{height: 600}}>
```
webOS101: Should we use ri or base on current window size?

ryanjduffy: Should probably use ri. It _should_ work without a height because both participate in flex layout but it was sizing too large so I gave it a fixed height to avoid the mess.

ryanjduffy: Found the cause. Sampler injects a DOM node in the Panel body that was interfering with the Scroller's height determination. Added a guard around that (which I'm not sure we still use anyway) to prevent rendering it when not valued.

Sadly, this will likely fail @Sangwook-Lee's TAS tests since all the samples will shift up some pixels but it's the right change to make.

viodragon2:
Added back the old sample that was replaced by the new one. Otherwise, LGTM!

PR #1711, opened 2018-06-07
PLAT-57992: Add support for scrolling via voice control

changgilee:
Change scroll status using by voice control event

### Resolution

### Additional Considerations

### Links


packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -560,6 +570,60 @@ class ScrollableBaseNative extends Component {
...
 
+	updateFocusAfterVoiceControl = () => {
+		const
+			{direction} = this.props,
+			viewportBounds = this.uiRef.containerRef.getBoundingClientRect(),
+			spotItemBounds = document.activeElement.getBoundingClientRect(),
+			nodes = this.uiRef.containerRef.querySelectorAll('[data-index]'),
+			isVertical = (direction === 'vertical' || direction === 'both'),
+			first = isVertical ? 'top' : 'left',
+			last = isVertical ? 'bottom' : 'right';
```
ybsung: Do you intend to scroll only vertically even though `direction` is equal to `both`? If yes, no problem. I just wanted to check if it was intended or not.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -560,6 +570,60 @@ class ScrollableBaseNative extends Component {
...
+
+		if (spotItemBounds[last] < viewportBounds[first] || spotItemBounds[first] > viewportBounds[last]) {
+			for (let i = 0; i < nodes.length; i++) {
+				const nodeBounds = nodes[i].getBoundingClientRect();
+				if (nodeBounds[first] > viewportBounds[first] && nodeBounds[last] < viewportBounds[last]) {
```
ybsung: I don't think that it is not the good solution to find a focusable node because nodes[i+1] could be positioned upper than nodes[i] with absolute or relative position.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -536,6 +542,9 @@ class ScrollableBaseNative extends Component {
 			childContainerRef.addEventListener('mouseover', this.onMouseOver, {capture: true});
 			childContainerRef.addEventListener('mousemove', this.onMouseMove, {capture: true});
 			childContainerRef.addEventListener('focusin', this.onFocus);
+			childContainerRef.addEventListener('webOSVoice', this.onVoice);
+			// FIXME In case of VirtualList, addEventListener target is not matched childContainerRef. So, scroll intent is added dynamically.
+			childContainerRef.setAttribute('data-webos-voice-intent', 'Scroll');
```
ybsung: It might be needed to add condition (e.g> platform.tv) for any TV specific features. For other devices, we don't have to handle them.

```
import {platform} from '@enact/webos/platform';

if (platform.tv) {
	childContainerRef.addEventListener('webOSVoice', this.onVoice);
	FIXME In case of VirtualList, addEventListener target is not matched childContainerRef. So, scroll intent is added dynamically.
	childContainerRef.setAttribute('data-webos-voice-intent', 'Scroll');
}

File: packages/moonstone/Scrollable/ScrollableNative.js [hunk 4]
```diff
@@ -560,6 +570,60 @@ class ScrollableBaseNative extends Component {
...
+					Spotlight.focus(nodes[i]);
+					break;
+				}
+			}
+		}
+	}
+
+	onVoice = (e) => {
+		const {scroll} = e.detail;
+		this.isVoiceControl = true;
```
ybsung: What if a browser passes a wrong e.detail value? We do not do anything in this function, but we tries to focus in scrollStopOnScroll () -> updateFocusAfterVoiceControl()

packages/moonstone/Scrollable/ScrollableNative.js
ybsung: If no, VirtualList supports `topleft` and `bottomright` option with `align` property.

packages/moonstone/Scrollable/ScrollableNative.js
0x64: I prefer setting attributes in `componentDidMount` and/or `componentDidUpdate' if possible, but it looks not bad here.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -536,6 +542,9 @@ class ScrollableBaseNative extends Component {
 			childContainerRef.addEventListener('mouseover', this.onMouseOver, {capture: true});
 			childContainerRef.addEventListener('mousemove', this.onMouseMove, {capture: true});
 			childContainerRef.addEventListener('focusin', this.onFocus);
+			childContainerRef.addEventListener('webOSVoice', this.onVoice);
+			// FIXME In case of VirtualList, addEventListener target is not matched childContainerRef. So, scroll intent is added dynamically.
```
0x64: I think the comment is unclear.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -473,6 +475,10 @@ class ScrollableBaseNative extends Component {
 		this.lastFocusedItem = null;
 		this.lastScrollPositionOnFocus = null;
 		this.isWheeling = false;
+		if (this.isVoiceControl) {
+			this.isVoiceControl = false;
+			this.updateFocusAfterVoiceControl();
```
0x64: It looks that the focus will not appear until scroll stops. Is this intended?

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -560,6 +570,60 @@ class ScrollableBaseNative extends Component {
 		}
 	}
 
+	updateFocusAfterVoiceControl = () => {
+		const
+			{direction} = this.props,
+			viewportBounds = this.uiRef.containerRef.getBoundingClientRect(),
+			spotItemBounds = document.activeElement.getBoundingClientRect(),
```
0x64: Is this safe to refer `activeElement` here? If this function is called when the focus is out of scrollable area, it may lead abnormal behaviors in some cases.

packages/moonstone/Scrollable/ScrollableNative.js
ybsung: As @0x64 mentioned, the item should not get focus after stopping scrolling. The item should get focus after creating and before showing it and stopping scrolling. Please check ENYO-4806 issue.

ybsung:
@changgilee 
@0x64 and I reviewed your PR. Please check our comments.

packages/moonstone/Scrollable/ScrollableNative.js
changgilee: In case of both, it operates only horizontally and vertically.(topleft and bottomright is not supported by voice control: http://collab.lge.com/main/pages/viewpage.action?pageId=840355974&focusedCommentId=853653402#comment-853653402 ) Relate to this, change to check the voice control direction using by seperate variable(voiceControlDirection)

packages/moonstone/Scrollable/ScrollableNative.js
changgilee: If there is a better solution, I will improve it.

packages/moonstone/Scrollable/ScrollableNative.js
changgilee: I did add condition(platform.tv) base on your guide. Thanks

packages/moonstone/Scrollable/ScrollableNative.js
changgilee: Add *default:* condition in switch statement.

packages/moonstone/Scrollable/ScrollableNative.js
changgilee: I did remove comment because it seems that setting attributes in this position can be allowed. Thanks

packages/moonstone/Scrollable/ScrollableNative.js
changgilee: Your comment is right. So, I did change to use *Spotlight.getCurrent()* and add to checking whether container contains spot item before comparing position.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -560,6 +576,71 @@ class ScrollableBaseNative extends Component {
 		}
 	}
 
+	updateFocusAfterVoiceControl = () => {
+		const spotItem = Spotlight.getCurrent();
+		if (spotItem && this.uiRef.containerRef.contains(spotItem)) {
+			const
+				viewportBounds = this.uiRef.containerRef.getBoundingClientRect(),
+				spotItemBounds = spotItem.getBoundingClientRect(),
+				nodes = this.uiRef.containerRef.querySelectorAll('[data-index]'),
```
ybsung: It would be better to use `spotlight.getSpottableDescendants(containerID)` instead of `this.uiRef.containerRef.querySelectorAll('[data-index]')`.

changgilee: I did change it base on your guide. Thanks~

ybsung:
Currently, we have no better ideas for the following issue.
1. Before stopping scrolling with voice control, we need to focus an item (Related comment: https://github.com/enactjs/enact/pull/1711/#discussion_r194590860)
2. If possible, we need the better way to find the item which will get focus

So we'd like to keep going to merge this PR as of now. **Looks good to me.**

In addition, I don't know why the Travis C build failed.
1. ERROR in ./Scrollable/ScrollableNative.js
lerna ERR! Module not found: Error: Can't resolve '@enact/webos/platform' in '/home/travis/build/enactjs/enact/packages/moonstone/Scrollable'
2. ERROR in ./Scrollable/Scrollable.js
lerna ERR! Module not found: Error: Can't resolve '@enact/webos/platform' in '/home/travis/build/enactjs/enact/packages/moonstone/Scrollable'

ybsung:
Good!!!

PR #1712, opened 2018-06-07
PLAT-60136: Remove TooltipDecorator Support From Button

germboy:
Improve mounting performance of `Button` and `IconButton` by removing the rarely-used built-in `TooltipDecorator` support.

### Resolution
We're simply removing `TooltipDecorator` from `Button` and `IconButton`. While this is simplistic, it does introduce a couple of required changes to other/related components.

Manually creating a `Button` with tooltip-support now changes the hoc order that was used. In order to retain `Uppercase` support for `Button`, `Uppercase` now needs to support formatting all eligible children via the supplied `casing` prop. Ryan and I could not think of why we shouldn't support this in lieu of only formatting the 1st child.

Maually creating a `moonstone/IconButton` with tooltip-support now changes the children within the `IconButton`. We need to provide an optional `icon` prop (in addition to the current `children` prop so that a user can provide an icon when the tooltip modifies the children (similar to our existing implementation in `ui/IconButton`). I don't love this alternate way of instantiating an `IconButton`, but don't see an obvious way around how to prevent the modification of children when using `TooltipDecorator`.


germboy:
I don't think it makes sense to keep a `tooltipNode` prop for a component we're removing built-in tooltip support for. And due to the new hoc ordering, the `children` of `moonstone/IconButton` will always be modified from a single child to an array. However, since a) `moonstone/IconButton` only ever expects a single child (to be used as the icon) and b) `ui/IconButtonBase` can accept both `icon` and `children`, I believe it's safe to check the children count in `moonstone/IconButton` and direct the 1st child to be used/parsed as the icon and the rest would be valid children. I've updated the PR to show this behavior. Thoughts?

germboy:
@viodragon2  In my previous change, I had reverted the `icon` prop implementation, so `children` was being used instead (as it is in our current implementation) - so specifying `icon` would result in no icon being rendered. I agree with the direction of removing `Tooltip` outside of `children` in `TooltipDecorator`. I've made additional adjustments to your proposed changes to factor in some legibility and performance help
- only rendering when `tooltipText` contains a value
- not separating `IconButton`'s `icon` and `children` props
- removing the `defaultConfig` option and `renderTooltip` method for `TooltipDecorator`
Let me know what you think.

packages/moonstone/IconButton/IconButton.js
```diff
@@ -106,9 +97,7 @@ const IconButtonBase = kind({
 				css={css}
 				icon={children}
```
viodragon2: remove this and pass `children` to `children`

PR #1713, opened 2018-06-07
PLAT-54548: Disable VideoPlayer playback controls until it's playable

viodragon2:
Apps want to disable playback controls until it's playable. Changed `this.state.sourceUnavailable`'s default value to `true` which will be updated to `false` only when there's enough data to play. Note that this is equivalent to `oncanplaythrough` event condition which means we can handle the event instead of managing the condition at the component level, but given that we allow custom media components, I chose not to expand the required events for `mediaComponents.

Some clarifications:
- `loading` state in `VideoPlayer` not only represents the initial loading of the video, but also anytime the video doesn't have enough content to play. IOW, buffering.
- The app only wants to disable playback controls during the initial loading, and not during buffering.
- Showing `MediaControls` should still work while loading because the app has a "go back to playlist" button in it. So `disabled` props will not satisfy the requirement.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
+- `moonstone/VideoPlayer` to disable bottom controls when loading until it's playable
+
 ## [2.0.0-beta.6] - 2018-06-04
 
 ### Removed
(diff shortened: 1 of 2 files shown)


ryanjduffy:
When the video is loading, calling `fastForward()` or `rewind()` will still set the playback rate. Is that the expected behavior or should those methods bail if `sourceUnavailable`?

viodragon2:
That's a good catch! I'll make a change.

viodragon2:
Added guards for all playback control imperative APIs

PR #1717, opened 2018-06-11
ENYO-5338: Add repositionProps to trigger reposition

viodragon2:
If the wrapper node or the container node changes the position, `ContextualPopup` needs to update its position, if necessary. 

### Resolution
Suggest adding a config option of a list of props that will trigger reposition, as
1. repositioning of the container is one of the rare use cases which we don't need to validate for every prop update
2. exposing as a prop to the application will require for developers to know some knowledge of how the component works

### Additional Considerations
- Since reposition isn't triggered by existing props (in normal use cases), `repositionProps` are expected to be custom props, which means they need to be cleaned up and make sure it does not get passed down. So the usage would look something like:
```
const ContextualPopup = ContextualPopupDecorator({repositionProps: ['customProp']}, (props) => {
	const prop = Object.assign({}, props);
	delete prop.customProp;
	return <IconButton {...prop} />
});
```
which is quite verbose :(

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `core/util.didPropChange` to support prop change check
+
 ## [2.0.0-beta.6] - 2018-06-04
 
 No significant changes.
--- a/packages/core/util/util.js
+++ b/packages/core/util/util.js
@@ -177,6 +177,17 @@ const mergeClassNameMaps = (baseMap, additiveMap, allowedClassNames) => {
 };
 
 /**
+ * Checks whether any props has changed or not for a given list of properties
+ *
+ * @param {Array} propsList An array of props to check
+ * @param {Object} prev Previous props
+ * @param {Object} next Next props
+ * @returns {Boolean} `true` if any of the props changed
+ * @private
+ */
+const didPropChange = (propsList, prev, next) => Array.isArray(propsList) && propsList.some((prop) => prev[prop] !== next[prop]);
+
+/**
  * Creates a function that memoizes the result of `fn`. Note that this function is a naive
  * implementation and only checks the first argument for memoization.
  *
@@ -203,6 +214,7 @@ export {
 	cap,
 	coerceArray,
 	coerceFunction,
+	didPropChange,
 	extractAriaProps,
 	isRenderable,
 	Job,
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -24,6 +24,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 - `moonstone/VideoPlayer.MediaControls` to handle left and right key to jump when `moonstone/VideoPlayer` is focused
 - `moonstone/Popup` to spot its content correctly when `open` by default
 
+### Added
+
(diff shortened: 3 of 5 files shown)


ryanjduffy:
This is a creative solution but I wonder if it is the simplest for the developer. Would an imperative API on `ContextualPopupDecorator` be any simpler? Is there a way we could position the contextual popup relative to the activator rather than in a global floating layer which would avoid developer intervention altogether?

viodragon2:
Exposing an imperative API could be easier, but I'd like to avoid implementing patterns we do not recommend.

Per your second question, the contextual popup does position itself relative to the activator (or the client node to be precise). This happens at mounting time. The only time we "reposition" is when `direction` property changes.

To recap the problem, we have two cases to account for repositioning:
1. activator/client node's dimension or position changes
2. contextual popup container's dimension changes

In either case, I'm not sure if there's a good signal for `ContextualPopupDecorator` to handle reposition itself. Both are states managed by the user. Other concern is that repositioning requires rerender. It should be prevented for developers to access such APIs IMO. Hence, added a config for prop list that app can pass.

We could remove `repositionProps` to handle unknown prop warning at component level to help developers not worry about it. But it can cause a problem if `repositionProps` contains non-custom props.

ryanjduffy:
> Exposing an imperative API could be easier, but I'd like to avoid implementing patterns we do not recommend when possible.

Agree that imperative APIs are not recommended but they are sometimes necessary when you start to stray from the typical use cases. In this case, I'm weighing it against the overhead of:
* Understanding how to use `repositionProps` and
* Wrapping the target component with an SFC to `delete` the repositioning props

While we don't recommend imperative APIs, I don't feel like we can recommend this prop gymnastics either.

> Per your second question, the contextual popup does position itself relative to the activator (or the client node to be precise). This happens at mounting time. The only time we "reposition" is when direction property changes.

That is the "global floating layer" part to which I was referring. If the contextual popup were positioned relative to the client node (= within the same DOM subtree and not in the global floating layer), it would automatically reposition. This approach seems like the best long term but it's a more significant undertaking, presuming it is even possible.

cc: @Djspaceg for ideas or known constraints

viodragon2:
My assumption was that this is not a normal use case, and wanted to make it harder for a developer to access because of the risks involved. But if you think that we have a further usage, I'm okay with exposing the API for now.

> This approach seems like the best long term but it's a more significant undertaking, presuming it is even possible.

Agreed. We do have a limitation with a current implementation where the container would not anchor to the activator if the activator's position changes by scrolling. I've been wanting to refactor both `ContextualPopupDecorator` and `TooltipDecorator` to use an abstract layer of HOC or functions to position the container in relation to an activator (something like how tether.js does), but didn't get a chance to do it. It involves tinkering with refs and observing the changes in DOM nodes which would be challenging, but maybe it's worth looking into now?

Djspaceg:
Personally, I don't feel that a props solution is the right approach to solving this problem. The way the popup is positioned onscreen should be refactored so its content and size can change and not interfere with the anchoring. A popup appearing to the left of the anchor should have its _right_ positioning value set, so as it grows or changes in width it extends to the left, leaving it anchored on the right side. A popup above should also be positioned only using the `bottom` prop, so the bottom remains fixed, and so forth.

Positioning _not_ using the floating layer does introduce other issues that will have to be up to the developer to decide (we'll need to offer both floating and non-floating options, like Enyo). For example, if a container is set to `overflow:hidden`, a contextualpopup or tooltip won't be visible if it draws off the edge.

I think we need further discussion about this.

viodragon2:
Closing in favor of #1756

PR #1721, opened 2018-06-14
ENYO-5352: Change Divider's marqueeOn default value to `render`

viodragon2:
Following up on #1271, change `Divider`'s `marqueeOn` default value to `render`


viodragon2:
Thanks for pointing it out. Yeah, it's weird that `casing` isn't warned. I made a change so that the decorator sets the default props by returning a `kind`. It's an extra layer so there is a performance concern with the change. I'm okay to revert and change `MarqueeDecorator` and `Uppercase` HOC applied to the base component, but it seemed nice to me. Let me know what you think.

viodragon2:
:mindblown:

viodragon2:
I like it! Not sure what the proper way to document this though. Any thoughts?

ryanjduffy:
I think the simplest solution right now would be to document them as properties on `Divider` even though they're really on `DividerDecorator`. Check out `DayPicker` for a couple examples of documenting props on the wrapped default export for an illustration on what I mean.

ryanjduffy:
:+1: Added `@memberof` to the "floating" prop docs to connect them to the class and adopted the `smartSelect` knob instead of the default storybook knob.

teckliew:
@ryanjduffy fyi, we will be using `select` from `enact-knob` #1693

https://github.com/enactjs/enact/pull/1693/files#diff-c6074a69c68e116026a86f248de1d709R6

packages/moonstone/Divider/Divider.js
```diff
@@ -84,21 +75,73 @@ const DividerBase = kind({
...
+ * @see i18n/Uppercase#casing
+ * @type {String}
+ * @default 'word'
+ * @memberof moonstone/Divider.Divider.prototype
+ * @public
+ */
+
+/**
+ * Determines what triggers the header content to start its animation. Valid values are
+ * `'hover'` and `'render'`. The default is `'render'`.
```
webOS101: The docs will render the default value from `@default`.  It's generally not needed to specify it in the text.

packages/moonstone/Divider/Divider.js
```diff
@@ -84,21 +75,73 @@ const DividerBase = kind({
...
+ * @memberof moonstone/Divider
+ * @mixes moonstone/Divider.DividerDecorator
+ * @ui
+ * @public
+ */
+
+/**
+ * Configures how the `children` string will be capitalized. By default, each word is capitalized.
+ *
+ * @see i18n/Uppercase#casing
```
webOS101: Without an `@name` field, there's no way for the documentation to know what to call this property.

PR #1725, opened 2018-06-14
ENYO-5339: Fix Scrollable to scroll properly after Scroller changes

teckliew:
Scrollable doesn't respond to Scroller updates

### Resolution
Because Scrollable doesn't update when Scroller changes, it needed something to hook on to when Scroller changes size. This solution uses a callback for Scrollable to update after Scroller changes.

### Links
ENYO-5339

### Comments


ryanjduffy:
Generally, I think that emitting an event when the component updates is not a practice we should promote. Events are most often emitted in response to user action or component state change. Tying them to lifecycle methods doesn't give appropriate context on what initiated the event.

That said, I don't see another way forward on this particularly issue so I'm inclined to move ahead with this change. Overall, the change is good and includes some nice refactoring that improves the code base.

@teckliew - do we need to adopt a similar change in the "native" versions as well or does this behavior only reproduce in the JS scrollers?

teckliew:
This also reproduces in native scrollable, so added the update to it as well.

PR #1728, opened 2018-06-15
Update documentation guidelines

webOS101:

Docs were not up-to-date with our standard

### Resolution
Updated and added a section on tone.

### Additional Considerations

### Links

### DIFF
--- a/docs/contributing/documentation.md
+++ b/docs/contributing/documentation.md
@@ -203,18 +203,21 @@ Tips:
 * Use the active voice when writing docs.
 * Descriptions should focus on what the componet or property provides.
 * Only document the `true` state for boolean propertues unless the `false` state's operation is unclear.
+* Properties should use a future tense (e.g. 'applies') and callbacks should use past tense (e.g. 'called')
 
 Good:
 
 * A Moonstone-styled button with built-in support for ...
 * The color of the underline beneath button's content.
-* When `true`, the control is in the expanded state with the contents visible.
+* Applies a disabled style and prevents interacting with the component.
+* Called when the internal input is focused.
 
 Could be better:
 
 * This component could be used to display a button ...
 * Set this to change the color of the underline beneath the button's content.
 * Controls whether the item is expanded or not.
+* A callback function invoked when the internal input is focused.
 
 ## Verifying Correct Documentation
 
--- a/docs/contributing/documentation.md
+++ b/docs/contributing/documentation.md
@@ -20,18 +20,25 @@ There are three general levels of documentation that will appear within the sour
 
 ## Module Documentation
 
-Each module within Enact should have one (and only one) file that includes the `@module` tag. The name of the module should be prefixed with the name of the package that contains it (e.g. `moonstone/Button`). Each module has a limited number of exports so they should be documented within the module block and the default export should be identified:
+Each module (and, consequently, directory) within Enact should have one (and only one) file that includes the `@module` tag. The name of the module should be prefixed with the name of the package that contains it (e.g. `moonstone/Button`). Each module has a limited number of exports so they should be documented within the module block using the `@exports` tag. The module description should not list the exports or repeat information that should appear at the component level.
 
 ```
 /**
- * Exports the {@link moonstone/Button.Button} and {@link moonstone/Button.ButtonBase}
- * components.  The default export is {@link moonstone/Button.Button}.
+ * Provides Moonstone-themed button components and behaviors.
+ *
+ * @example
+ * <Button small>Hello Enact!</Button>
  *
  * @module moonstone/Button
+ * @exports Button
+ * @exports ButtonBase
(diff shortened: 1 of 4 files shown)


docs/contributing/documentation.md
```diff
@@ -190,16 +194,38 @@ There are some special cases that appear within the Enact framework. One example
...
+
+* Use the active voice when writing docs.
+* Descriptions should focus on what the componet or property provides.
+* Only document the `true` state for boolean propertues unless the `false` state's operation is unclear.
+
+Good:
+
+* A Moonstone-styled button with built-in support for ...
+* The color of the underline beneath button's content.
+* When `true`, the control is in the expanded state with the contents visible.
```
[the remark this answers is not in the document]
Djspaceg: Your example doesn't include the words from your template. Based on the template it would read:

> The `open` prop opens the ExpandableItem to reveal its contents

I am not a fan of that template as I feel it's even less meaningful than "When true...". That being said, I do think the wording of the example is acceptable:

> Opens the `ExpandableItem` to reveal its contents

It describes the action as a sentence fragment stating by implying the relevant property name.

docs/contributing/documentation.md
```diff
@@ -190,16 +194,38 @@ There are some special cases that appear within the Enact framework. One example
...
+Tips:
+
+* Use the active voice when writing docs.
+* Descriptions should focus on what the componet or property provides.
+* Only document the `true` state for boolean propertues unless the `false` state's operation is unclear.
+
+Good:
+
+* A Moonstone-styled button with built-in support for ...
+* The color of the underline beneath button's content.
```
ryanjduffy: In more recent components, I've attempted to use a short (< 80 characters) description for a prop on the first line and then expand on it in subsequent lines if necessary. I think this forces us to succinctly describe the feature for the user to enable them to quickly decide if they need it without wading through the full body.

As a result, the short description (like this case) isn't a sentence and so a period at the end isn't appropriate.

* Should we follow the approach I've outline above (one line summary + longer body as needed)?
* Should we
  * always use an ending period,
  * omit ending punctuation altogether, or
  * restrict it to when using a complete sentence

Djspaceg: This is a considerate idea. I like it. Should the next line of the description (the longer part) be separated by a blank line to begin a new paragraph? If not, it will run-in as the next sentence of the docs, rather than a new paragraph.

ryanjduffy: Yes. I've always included a blank line between the short desc and body.

```
/**
 * Short description
 *
 * longer body
 *
 * @pragmas
 */
```

webOS101: The only problem with not adding the period is that some descriptions have it and some done.  I Think keeping it to a sentence is most useful.  Even a terse sentence deserves a period.

Opens the ExpandableItem to reveal its contents

That's still a sentence.  The subject is implied, but all the parts are there.

ryanjduffy: But the example isn't a sentence

> The color of the underline beneath button's content.

Should it be reworded to be one?

> Sets the color of the underline beneath button's content.

Seems like we'd have a lot of "Sets ..." just so we can add a period.

webOS101: I don't think we need to get too hung up on whether something is a proper sentence or not.  Just stick a period on it and make me happy. :P  Fragments are fine for the short description.  We definitely don't want 'sets', I think.  Because props don't really -do- things, they -are- things (in most contexts).

docs/contributing/documentation.md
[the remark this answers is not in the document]
webOS101: As commented below, 'Opens the...' implies some sort of imperative operation.  Which is why I had used the the 'when true,  ....' because it tells you the result when it's true.  It doesn't imply any sort of transition or action.  I am open to better wording, of course.  I was trying to avoid things like 'Controls the open status of the ExpandableItem', though it as descriptive.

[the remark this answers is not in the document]
ryanjduffy: @Djspaceg I meant that YYY would the text used for the description such that it would read like the template. Sorry I wasn't clear on that.

To close the loop on the two comment threads, should props that "do" things (e.g. `open`) begin with verbs (e.g. "Opens ...") and props that "are" things (e.g. `backgroundColor`) not being with verbs (e.g. "The color used in the back ..." ;) )?

In a declarative framework, changing a prop is the equivalent of invoking a method so I think the use of verbs can be appropriate in some cases.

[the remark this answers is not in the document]
webOS101: I don't entirely agree.  If I put this in my code:

```
<Popup open />
```

It does not render a popup and open it.  It renders an open popup.  Maybe the distinction is subtle (and, maybe, overstated?) but I think that saying props invoke methods is not accurate in many, many circumstances.

PR #1733, opened 2018-06-19
PLAT-61040: (release/1.x branch) Add disableBackHistoryAPI with `true` option to the appinfo.json of sampler

Baekwoo-s:
Add the disableBackHistoryAPI flag to the appinfo.json of Enact sampler
We expect popup close when pressing back key but App goes history back without this flag
### Links
PLAT-61040

### Comments


0x64:
LGTM

PR #1737, opened 2018-06-20
PLAT-49380: Add `initialScrollPosition` prop to scroll when initialized

ybsung:

Add `initialScrollPosition` prop in Scrollable.

### Resolution

I would like to scroll a list or a scroller first, then to render the items or the content in it when creating it. But at that time `bounds` information is incorrect. So I render the items or the content first, then scroll them.

### Additional Considerations

### Links

PLAT-49380

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -399,18 +399,17 @@ class ScrollableBase extends Component {
 			{hasDataSizeChanged} = this.childRef;
 
 		// Need to sync calculated client size if it is different from the real size
-		if (this.childRef.syncClientSize) {
-			// If we actually synced, we need to reset scroll position.
-			const isSync = this.childRef.syncClientSize();
+		const isSync = this.childRef.syncClientSize ? this.childRef.syncClientSize() : null;
 
-			if (this.props.initialScrollPosition) {
-				const {left, top} = this.props.initialScrollPosition;
-				this.scroll(left, top);
-			} else if (isSync) {
-				this.setScrollLeft(0);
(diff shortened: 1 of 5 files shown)


packages/ui/Scrollable/Scrollable.js
```diff
@@ -385,13 +399,17 @@ class ScrollableBase extends Component {
 			{hasDataSizeChanged} = this.childRef;
 
 		// Need to sync calculated client size if it is different from the real size
-		if (this.childRef.syncClientSize) {
+		const isSync = this.childRef.syncClientSize ? this.childRef.syncClientSize() : null;
+
+		if (!this.initialScrollTo && this.props.initialScrollPosition) {
+			const {left, top} = this.props.initialScrollPosition;
+			this.scroll(left, top);
```
0x64: If this routine does not forward 'onScrollStart', it looks better to use `start` instead of `scroll`.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -385,13 +399,17 @@ class ScrollableBase extends Component {
 			{hasDataSizeChanged} = this.childRef;
 
 		// Need to sync calculated client size if it is different from the real size
-		if (this.childRef.syncClientSize) {
+		const isSync = this.childRef.syncClientSize ? this.childRef.syncClientSize() : null;
```
0x64: If `syncClientSize()` MUST be called before handling the initial position prop, let's leave some comment about it for the future.

0x64:
LGTM ;)

ybsung:
We decided that we needed to find out better solution later. We won't merge this PR.

PR #1746, opened 2018-06-26
PLAT-48029: Scroller makes apps to know it's scrolled to the edge.

0x64:
We need to tell app developers when a scroller is scrolled to the top or the bottom so that app can add specified visual effect. We already provide `onScroll` event-type callback but this callback is called whenever a scroller is scrolled, so quite costly.
We have this requirement only for limited internal scenes for now.

### Resolution
The new prop `onReachEdge` is added as **private** which is called back when and only when a scroller reached to the edge or left from the edge.

### Additional Considerations

Note: If a scroller is shortened and not scrollable anymore, a scroller will notify that it is NOT reached to any edge since apps don't need to show a visual effect for short contents.

### Links
PLAT-48029

### Comments

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -183,7 +183,7 @@ class ScrollableBase extends Component {
 		onMouseDown: PropTypes.func,
 
 		/**
-		 * Called when the edges are reached.
+		 * Called when reached to an edge.
 		 * Passes `orientation`, `position`, and `reached`.
 		 *
 		 * Example:
@@ -409,6 +409,7 @@ class ScrollableBase extends Component {
 
 	componentDidUpdate (prevProps, prevState) {
 		const
+			{onReachEdge} = this.props,
 			{isHorizontalScrollbarVisible, isVerticalScrollbarVisible} = this.state,
 			{hasDataSizeChanged} = this.childRef;
 
@@ -423,8 +424,28 @@ class ScrollableBase extends Component {
 
 		this.clampScrollPosition();
 
-		this.forwardReachEdgeEvent('horizontal');
-		this.forwardReachEdgeEvent('vertical');
+		if (onReachEdge) {
+			const name = {horizontal: 'Left', vertical: 'Top'};
+
+			['horizontal', 'vertical'].forEach((orientation) => {
+				const
+					info = this.reachEdgeInfo[orientation],
+					currentPosition = this['scroll' + name[orientation]];
+				let condition = {begin: false, end: false};
+
+				if (this.getScrollabilities(orientation)) {
+					condition.begin = currentPosition <= 0;
+					condition.end = currentPosition >= this.getScrollBounds()['max' + name[orientation]];
+				}
+
+				['begin', 'end'].forEach((position) => {
+					if (info[position] !== condition[position]) {
+						info[position] = condition[position];
+						forward('onReachEdge', {orientation, position, reached: info[position]}, this.props);
+					}
+				});
+			});
+		}
 
 		this.addEventListeners();
 		if (
@@ -534,7 +555,7 @@ class ScrollableBase extends Component {
 	scrollLeft = 0
 	scrollTop = 0
 	scrollToInfo = null
-	touchingEdgeInfo = {
+	reachEdgeInfo = {
 		horizontal: {
 			begin: false,
 			end: false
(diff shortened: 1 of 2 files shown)


packages/ui/Scrollable/Scrollable.js
```diff
@@ -785,18 +826,57 @@ class ScrollableBase extends Component {
...
 	}
 
+	compareAndSet (obj, name, value) {
+		const changed = (obj[name] !== value);
+		obj[name] = value;
+		return changed;
+	}
+
+	forwardReachEdgeEvent (orientation) {
+		const {onReachEdge} = this.props;
```
ybsung: Add one empty line.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1210,6 +1292,7 @@ class ScrollableBase extends Component {
 		delete rest.onScroll;
 		delete rest.onScrollStart;
 		delete rest.onScrollStop;
+		delete rest.onReachEdge;
```
ybsung: `onReachEdge` should be placed before `onScroll`.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -182,6 +182,34 @@ class ScrollableBase extends Component {
 		 */
 		onMouseDown: PropTypes.func,
 
+		/**
+		 * Called when the edges are reached.
```
ybsung: How about not plural but singular for edge? It is called when one of the edges is reached.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -182,6 +182,34 @@ class ScrollableBase extends Component {
...
+		 * Passes `orientation`, `position`, and `reached`.
```
ybsung: Is it correct to wrap with ' in this case?
e.g> Passes `'orientation'`, `'position'`, and `'reached'`.

0x64: Nope. they are not string values but names of property.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -785,18 +826,57 @@ class ScrollableBase extends Component {
 		forward(type, {scrollLeft: this.scrollLeft, scrollTop: this.scrollTop, moreInfo: this.getMoreInfo()}, this.props);
 	}
 
+	compareAndSet (obj, name, value) {
+		const changed = (obj[name] !== value);
+		obj[name] = value;
+		return changed;
+	}
+
+	forwardReachEdgeEvent (orientation) {
```
ybsung: Could you suggest another function name because this function does not always forward an event? On the other hand, `forwardScrollEvent ` function always send an event.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -394,6 +422,9 @@ class ScrollableBaseNative extends Component {
 			}
 		}
 
+		this.forwardReachEdgeEvent('horizontal');
+		this.forwardReachEdgeEvent('vertical');
```
ybsung: If `clientSize` prop is passed, then `componentDidMount` will be called, but `componentDidUpdate` will not be called. Please check this edge case.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -916,12 +996,14 @@ class ScrollableBaseNative extends Component {
 		const
 			bounds = this.getScrollBounds(),
 			maxValue = bounds.maxTop,
-			{type, ratio} = this.getOverscrollStatus('vertical');
+			orientation = 'vertical',
+			{type, ratio} = this.getOverscrollStatus(orientation);
```
ybsung: ratio before type, please..

packages/ui/Scrollable/Scrollable.js
```diff
@@ -785,18 +826,57 @@ class ScrollableBase extends Component {
...
 	setScrollLeft (value) {
 		const
 			bounds = this.getScrollBounds(),
 			maxValue = bounds.maxLeft,
-			{type, ratio} = this.getOverscrollStatus('horizontal');
+			orientation = 'horizontal',
+			{type, ratio} = this.getOverscrollStatus(orientation);
 
 		this.scrollLeft = clamp(0, maxValue, value);
+		this.forwardReachEdgeEvent(orientation);
```
ybsung: As we discussed, it would be better to move the code in `forwardReachEdgeEvent` here after cleaning up like the following code.

```
		if (onReachEdge && this.getScrollabilities(orientation)) {
			const
				info = this.touchingEdgeInfo[orientation],
				condition = {begin: this.scrollLeft <= 0, end: this.scrollLeft >= this.getScrollBounds()[maxValue]};

			['begin', 'end'].forEach((edge) => {
				if (info[edge] !== condition[edge]) {
					info[edge] = condition[edge];
					onReachEdge({orientation, position: edge, reached: info[edge]});
				}
			});
		}
```

packages/ui/Scrollable/Scrollable.js
```diff
@@ -785,18 +826,57 @@ class ScrollableBase extends Component {
...
+				if (this.getScrollabilities(orientation)) {
+					const
+						name = (orientation === 'vertical') ? 'Top' : 'Left',
+						curPos = this['scroll' + name];
+					conditionBegin = (curPos <= 0);
+					conditionEnd = (curPos >= this.getScrollBounds()['max' + name]);
+				}
+
+				if (compareAndSet(info, 'begin', conditionBegin)) {
+					onReachEdge({orientation, position: 'begin', reached: info.begin});
```
ybsung: Let's use `forward` instead of `onReachEdge`.

packages/ui/Scrollable/Scrollable.js
0x64: 👌

packages/ui/Scrollable/Scrollable.js
```diff
@@ -182,6 +182,34 @@ class ScrollableBase extends Component {
 		 */
 		onMouseDown: PropTypes.func,
 
+		/**
+		 * Called when reached to an edge.
+		 * Passes `orientation`, `position`, and `reached`.
```
ybsung: * Passes `'orientation'`, `'position'`, and `'reached'`.

0x64: Nope. they are not string values but names of property.

ybsung: Oh, I missed. Good!!

packages/ui/Scrollable/Scrollable.js
```diff
@@ -395,6 +424,26 @@ class ScrollableBase extends Component {
 
 		this.clampScrollPosition();
 
+		if (onReachEdge) {
```
MikyungKim: If we put this for initial firing, we also need this logic in `componentDidMount` phase for specified scrollbar visibility.
I don't feel right about doing so in both lifecycles though.

0x64: Thanks, @MikyungKim. I missed a case which a scrollbar is intentionally hidden and `Scrollable` has a given `clientSize` prop value.

I agree that it looks not good to add the logic into both lifecycle methods. We need more discussion.

0x64:
Closing in favor of #1933

PR #1748, opened 2018-06-26
PLAT-62141: Add spotlight 5-way debugging

Djspaceg:
Duper tedious to verify where Spotlight will move to at any given focused element.

### Resolution
Added a way to get a sort of preview of which elements will be focused if any direction key is pressed.

### Additional Considerations
Extended the debugging classes to require refinement of what specifically you are attempting to debug: `.debug.spotlight`, `.debug.layout`, etc, rather than just a blanket `.debug` for every 'ol thing at once.


lishammel:
@Djspaceg  - does this have a jira to get it merged?

Djspaceg:
I don't think this actually has a JIRA ticket. We'd talked about it a little bit, but one just hadn't been made.

ryanjduffy:
Looks good, @Djspaceg! I think this will be really helpful in diagnosing spotlight behavior issues. We'll definitely want to write up some documentation on how these features can be used by applications for debugging.

One question: Is it worth handling cases where an element is spottable for multiple directions (e.g. left and down from the close button to an Item)? Does the value outweigh the additional complexity?

@JayCanuck - Do we have a way to exclude the CSS from production builds to minimize parsing unused rules? Maybe using the LESS variable injection support + [CSS Guards](http://lesscss.org/features/#css-guards-feature)?

Djspaceg:
@ryanjduffy, we can actually add this support in the future (stacking multiple directions) since it's just class names being applied. We can do `.spotlight-next-up.spotlight-next-left` and combine the colors. We'll just have to explicitly write the rules to combine them. It was something I'd considered while making this, and I don't think it's actually terribly difficult. It would be further improved by the ability to conditionally exclude the CSS for it in production.

ryanjduffy:
Tried the following on lesstester.com and it seems to have the desired affect.
```
@__DEV__: false;
:global when (@__DEV__ = true) {
  .abc {
    color: blue;
  }
}
```

So, if we could set the __DEV__ LESS var for development mode, we could omit any debug code from prod builds. I'll create a follow up ticket to address this.

JayCanuck:
Also a reminder that we already support `.__DEV__` classname keyword for debug code, which, like the javascript equivalent, is physically removed from the output css in production mode builds, but exists as-is in development mode. https://github.com/enactjs/cli/blob/master/docs/building-apps.md#__dev__-keyword

Djspaceg:
`.__DEV__` is news to me! Good news to know though. Let's go with this approach for simplicity. We can wrap all of the `.debug` classes in this as well, and add the combination classes to this as well.

ryanjduffy:
If we wanted to use `.__DEV__`, we may need to adjust how the debug classes are applied. Unlike a CSS Guard, the `.__DEV__` class would also need to be applied somewhere for the debug classes to take effect. So, instead of
```
<div class="debug aria">
```

we'd need

```
<div class="__DEV__ debug aria">
```
or, if we wanted to replace `debug` with `__DEV__` (fwiw, I don't think this is a good idea as it conflates "dev mode" with "debugging" which may not always be synonymous).

```
<div class="__DEV__ aria">
```

@Djspaceg @JayCanuck - Thoughts?

ryanjduffy:
Going to go ahead and merge this as is and reopen the JIRA to manage the CLI side. We can make any framework updates as part of that ticket as well.

Djspaceg:
My expectation was that the `__DEV__` class was simply automatically applied to the root decorator or body element or something so it was actually available. Having logic in our component(s) to add this class seems odd; it would just need to be removed along with the selectors themselves. This class could be added to the MoonstoneDecorator via the spotlight decorator behind guarded code but that's really specific for something so general as a dev-tools DEV mode.

JayCanuck:
If we want to go with the LESS variable route for a more 1-to-1 comparison between JS and CSS usage, I've setup a proof-of-concept of CLI changes on https://github.com/enactjs/cli/compare/develop...feature/PLAT-62880

PR #1750, opened 2018-06-26
ENYO-5378: Fix TooltipDecorator to prevent unnecessary renders

germboy:
`TooltipDecorator` always re-renders (to hide the tooltip) when losing focus, even when a tooltip is not yet visible.

### Resolution
We should check that the `showing` state of `TooltipDecorator` is `true` before setting it to `false`. 

### DIFF
--- a/packages/moonstone/TooltipDecorator/TooltipDecorator.js
+++ b/packages/moonstone/TooltipDecorator/TooltipDecorator.js
@@ -331,7 +331,9 @@ const TooltipDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				this.clientRef = null;
 				currentTooltip = null;
 				this.showTooltipJob.stop();
-				this.setState({showing: false});
+				if (this.state.showing) {
+					this.setState({showing: false});
+				}
 			}
 		}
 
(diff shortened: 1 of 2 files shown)


packages/moonstone/TooltipDecorator/TooltipDecorator.js
```diff
@@ -331,7 +331,9 @@ const TooltipDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				this.clientRef = null;
 				currentTooltip = null;
 				this.showTooltipJob.stop();
-				this.setState({showing: false});
+				if (this.state.showing) {
+					this.setState({showing: false});
```
ryanjduffy: Should we use functional `setState()` here and in `showTooltipJob()` above? Could generalize them with something like:

```
const toggleShowing = (showing) => (state) => state.showing !== showing ? {showing} : null;
// ...
this.setState(toggleShowing(false));
```

germboy: I like the readability of that and the idea of functional `setState`, but I've been avoiding it since it creates (or adds time to) a react tree reconciliation update simply when utilizing it - even when nothing should actually update.

ryanjduffy: I think we need resolve this issue (using function `setState`) in the longer term but the usage here is trivial so let's go ahead with the current impl.

PR #1752, opened 2018-06-26
PLAT-61706: Add moreButtonColor knob to VideoPlayer sample

viodragon2:
Add `moreButtonColor` knob to `VideoPlayer` sample

### Additional Consideration
Not excited about adding `color` prop in `onlyUpdateForKeys()` for `MediaButton`. In a normal use case, no developer would ever change that unless they want to confuse the hell out of the user. I wonder if there's a way we can configure that only for the sampler context?


ryanjduffy:
Since we expose a prop to configure the color, it's reasonable to allow it to be changed even though it'd be a poor UX decision to do so in most cases. :shipit:

PR #1754, opened 2018-06-28
PLAT-61006: QA-Sampler - Update Knobs

teckliew:
Update QA Sampler knobs and fix `Button`'s `minWidth` knob to work as expected.

`Button`'s `minWidth` is the only prop that is `true` by default so we want to hide `minWidth` in the Storybook JSX by default, but show `minWidth={false}`. As a result, we have this for `minWidth` knob: `storybookBoolean('minWidth', true) ? void 0 : false`, where `void 0` hides the prop when `true`.

### Links
PLAT-61006

### Comments

### DIFF
--- a/packages/sampler/stories/qa-stories/CheckboxItem.js
+++ b/packages/sampler/stories/qa-stories/CheckboxItem.js
@@ -1,9 +1,16 @@
 import CheckboxItem from '@enact/moonstone/CheckboxItem';
+import ToggleItem from '@enact/moonstone/ToggleItem';
+import UiToggleItem, {ToggleItemBase as UiToggleItemBase} from '@enact/ui/ToggleItem';
+import Item, {ItemBase} from '@enact/moonstone/Item';
 import Group from '@enact/ui/Group';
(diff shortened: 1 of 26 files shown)


packages/sampler/stories/qa-stories/CheckboxItem.js
```diff
@@ -70,7 +81,7 @@ storiesOf('CheckboxItem', module)
 				itemProps={{
 					inline: boolean('ItemProps-Inline', false)
```
Djspaceg: This `boolean` needs a config object.

Djspaceg:
There's a Travis warning on Button, but provided that gets fixed, I've conditionally approved:
```
lerna ERR! /home/travis/build/enactjs/enact/packages/sampler/stories/qa-stories/Button.js
lerna ERR!   32:59  warning  Using propTypes from another component is not safe because they may be removed in production builds  react/forbid-foreign-prop-types
```
I believe this is a recent addition to our linting rules.

PR #1755, opened 2018-06-28
ENYO-5387: Add disableFullscreen config to MoonstoneDecorator

haileyR:
The `.enact-fit` class is currently a required addition to `MoonstoneDecorator`. It means that if Moonstone is used in a normal DOM node, it takes up no space, so layout does not flow around it as one would expect.

### Resolution
This change will give consumers an option to exclude the `.enact-fit` class which, by default expects that MoonstoneDecorator will be used in a "full screen" context.

### Additional Considerations
This absolute position breaks some other places. e.g. MoonstonePreview in the docs.

### DIFF
--- a/packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.js
+++ b/packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.js
@@ -28,17 +28,17 @@ import {configure} from '@enact/ui/Touchable';
  * @hocconfig
  */
 const defaultConfig = {
-	i18n: true,
+	disableFullscreen: false,
 	float: true,
+	i18n: true,
 	noAutoFocus: false,
 	overlay: false,
 	ri: {
 		screenTypes
 	},
-	spotlight: true,
-	textSize: true,
 	skin: true,
-	disableFullscreen: false
+	spotlight: true,
+	textSize: true
 };
 
 /**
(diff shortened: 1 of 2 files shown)


packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.js
```diff
@@ -58,10 +59,14 @@ const defaultConfig = {
...
 const MoonstoneDecorator = hoc(defaultConfig, (config, Wrapped) => {
-	const {ri, i18n, spotlight, float, noAutoFocus, overlay, textSize, skin, highContrast} = config;
+	const {ri, i18n, spotlight, float, noAutoFocus, overlay,
+		textSize, skin, highContrast, disableFullscreen} = config;
 
 	// Apply classes depending on screen type (overlay / fullscreen)
-	const bgClassName = 'enact-fit' + (overlay ? '' : ` ${css.bg}`);
+	const bgClassName = classNames({
+		'enact-fit': !disableFullscreen,
+		[`${css.bg}`]: !overlay
```
Djspaceg: I believe it's possible to reduce this from `[\`${css.bg}\`]` to simply `bg`. I may be confusing this with features of our `kind()` though.

haileyR: tried, but failed...

packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.js
```diff
@@ -58,10 +59,14 @@ const defaultConfig = {
  * @public
  */
 const MoonstoneDecorator = hoc(defaultConfig, (config, Wrapped) => {
-	const {ri, i18n, spotlight, float, noAutoFocus, overlay, textSize, skin, highContrast} = config;
+	const {ri, i18n, spotlight, float, noAutoFocus, overlay,
+		textSize, skin, highContrast, disableFullscreen} = config;
```
ryanjduffy: We should add `disableFullscreen` to the `defaultConfig` object above as well. `defaultConfig` acts as the "interface" for the HOC configuration and is the home for documentation (even though it's lacking in this case) so we should keep it in sync with the config options used.

PR #1756, opened 2018-06-28
ENYO-5338: Add positionContextualPopup method

viodragon2:
`moonstone/ContextualPopup` may need to update its position if the position or dimensions of the activator or popup updates. Unfortunately, `moonstone/ContextualPopupDecorator` is not suited for monitoring those states. We can think about different ways of handling positioing so that it always positions relative to the activator and the window. Howver, given the complexity and time required for the suggested change, we choose to provide an imperative API for developers to call `positionContextualPopup` method to adjust positions if needed.

See also: #1717 

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
+### Added
+
+- `moonstone/ContextualPopup`'s new API `positionContextualPopup`
+
 ### Fixed
 
 - `moonstone/TooltipDecorator` to prevent unnecessary re-renders when losing focus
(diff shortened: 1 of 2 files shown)


ryanjduffy:
WDYT of ```
	const Decorator = hoc(...);
	return compose(
		ApiDecorator,
		Subscription({
			channels: ['i18n'],
			mapMessageToProps: (key, {rtl}) => ({rtl})
		}),
		Decorator
	);
```

**Comment by viodragon2** [comment]
I've thought about that. The only (minor) problem I had with that was the it needed to have a separate name other than `ContextualPopupDecorator`. IOW,
```
	const Decorator = hoc(...);
	const ContextualPopupDecorator = compose(
		ApiDecorator,
		Subscription({
			channels: ['i18n'],
			mapMessageToProps: (key, {rtl}) => ({rtl})
		}),
		Decorator
	);

export default ContextualPopupDecorator;
export {ContextualPopupDecorator}

```

And where does the doc belong to in this case? `Decorator` or `ContextualPopupDecorator`?

ryanjduffy:
I don't think it greatly matters on which you document it. My take would be to document on the member named after the export.

viodragon2:
Updated per discussion!

PR #1758, opened 2018-06-29
ENYO-5304: Fix ContextualPopup to refocus activator on close for a11y

ryanjduffy:
If a contextual popup doesn't contain spottable children, closing it should refocus the activator so a11y read out can occur.

### Resolution
Add another condition to `shouldSpotActivator` to refocus the activator if it had focus on close


ryanjduffy:
Boo. I'll put this back in the queue then since I can't safely test it.

viodragon2:
It turns out it was because I was testing it on an old version of gld4tv. I think your fix works great. Let's reopen this.

PR #1761, opened 2018-07-02
PLAT-61631 - Added Docs for Interoperability

rundmt:
Added Docs for Interoperability with third-party libraries

### Links
PLAT-61631

Enact-DCO-1.0-Signed-off-by: Derek Tor <derek.tor@lge.com>

### DIFF
--- a/docs/interoperability.md
+++ b/docs/interoperability.md
@@ -4,16 +4,16 @@ title: Using Enact with third-party libaries
 
 ## Overview
 
-Enact is a fanstastic way to build apps. By default we give you a template using our `Moonstone`
-library. `Moonstone` is a library designed for TVs. However, if you're not looking to create, TV 
-apps you can easily use Enact to build applications for Desktop/Mobile. You can mix and match 
-third-party libaries to use within your Enact very easily. Alternatively, if you just want to use
-`Enact` components in your own `React` app that is very straightforward.
+`Enact` is a fantastic way to build apps. By default we give you a template using our `Moonstone`
+library. `Moonstone` is a library designed for TVs. However, if you're not looking to create TV
+apps, you can easily use `Enact` to build applications for desktop and mobile. You can mix and match
+third-party libraries to use with `Enact` easily. Alternatively, it is also straightforward if you
+just want to use `Enact` components in your own `React` app.
 
 ## Using Third-Party Components inside of Enact
-When creating an app using `enact create`, we give you default boilerplate that is setup for a 
-`Moonstone` app. However if you want to use something outside of Enact like `material-ui`, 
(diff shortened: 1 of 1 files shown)


docs/interoperability.md
```diff
@@ -0,0 +1,73 @@
...
+## Overview
+
+`Enact` is a fantastic way to build apps. By default we give you a template using our `Moonstone`
+library. `Moonstone` is a library designed for TVs. However, if you're not looking to create TV
+apps, you can easily use `Enact` to build applications for desktop and mobile. You can mix and match
+third-party libraries to use with `Enact` easily. Alternatively, it is also straightforward if you
+just want to use `Enact` components in your own `React` app.
+
+## Using Third-Party Components inside of Enact
+When creating an app using `enact create`, it creates an `Enact` boilerplate that setups for a
```
webOS101: reword.

docs/interoperability.md
```diff
@@ -0,0 +1,73 @@
...
+`Moonstone` styled app. However if you want to use something outside of `Moonstone` like
+`material-ui`, `reactstrap`, or `react-router` just use `npm install`.
+
+You can include components just like you normally would by using import.
+
+```JavaScript
```
webOS101: Show command line for using a different `enact create` template?  I guess we don't have a 'core' or 'ui' template.  Wonder if we should whip one up as part of public release?

docs/interoperability.md
```diff
@@ -0,0 +1,73 @@
...
+import { Button as BSButton} from 'reactstrap';
```
webOS101: Is `BSButton` the recommended way for `reactstrap`?  It seems odd to me, but I don't know `reactstrap`.

docs/interoperability.md
```diff
@@ -0,0 +1,73 @@
...
+
+const MainPanel = kind({
+	name: 'MainPanel',
+
+	render: (props) => (
+		<Panel>
```
webOS101: Where did `Panel` come from?  Should we show the full imports for this?  Doesn't add anything?

docs/interoperability.md
rundmt: Yeah this is a good idea. I think we should have something that is just a `ui` template.

docs/interoperability.md
rundmt: Yeah I'll switch it back to button

docs/interoperability.md
rundmt: Yeah I'll remove this and just change it to a div

rundmt:
Yeah I think the changes you made are fine. I made some changes as well.

PR #1768, opened 2018-07-03
PLAT-62147: Limit voice control area of ContexturePopup

changgilee:
Limit voice control area using by data-webos-voice-exclusive attribute

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -72,6 +72,15 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			popupComponent: PropTypes.func.isRequired,
 
 			/**
+			 * When `true`, the range of voice control is limited to popup.
+			 *
+			 * @type {String}
+			 * @default true
+			 * @public
+			 */
+			'data-webos-voice-exclusive': PropTypes.bool,
+
+			/**
 			 * Direction of ContextualPopup
 			 *
 			 * @type {String}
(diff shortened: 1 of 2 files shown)


packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -71,6 +71,15 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			 */
 			popupComponent: PropTypes.func.isRequired,
 
+			/**
+			 * When `true`, the range of voice control is limited to popup.
+			 *
+			 * @type {Boolean}
```
webOS101: This needs an `@memberof` to doc correctly.  I'll fix it.

webOS101: Second point:  Are there use cases where would we not want to base this off the `spotlightRestrict` setting so it matches with 5-way and pointer mode restrictions?  If `self-only` set the restrict `true` and if not, set to `false`?  Or, would we want to make it nullable and allow it to follow the default above unless specifically set to `false` or `true`?

changgilee: In fact, the policy of voice control should be set separately from spotlightRestrict.
First, when popup is opened, the basic UX policy is to limit the voice control area to pop-in.
For example, if spotlightRestrict is set to self-first or none, the voice control range is limited to pop-up internals.
However, for ContexturePopup, there are scenarios where exception handling is required.
For example, ContexturePopup is used in the Settings menu during video playback in the Photo Video app.
In this case, if we limit the voice control area to pop-up, the Intent(PlayListControl) related to video control will not work.
(The web engine gathers intent information based on the dom node, and if the data-webos-voice-exclusive attribute is present, limits the scope of the queryselector to the subordinate attribute)

PR #1769, opened 2018-07-03
ENYO-5365: Fix Scroller overscroll effects

0x64:
- Fixed not to show an overscroll effect when a focus moves between paging control buttons right after a list reached to the edge.
- Fixed not to show an overscroll effect when a focus moves between a horizontal scrollbar and a vertical scrollbar.
- Fixed to show an overscroll effect by the first wheeling on native lists.
- Fixed to remove an overscroll effect by drag when the pointer moves the opposite direction.

### Resolution
The internal status variable is shared for effects on both edges (beginning and ending), so updating the variable by reaching one edge affected an effect for the other edge. In this PR, effects for both edges have their own status variable.

The existing code has a concept of all-in-one effect handling logic which is called from every user inputs, and it made a logic heavier than the original design. In this PR, the key effect handling logic handles for each edges basically and user input handlers make a decision which edge should be monitored.

### Links
ENYO-5364
ENYO-5365
ENYO-5366
ENYO-5408

### Comments


packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -759,79 +769,74 @@ class ScrollableBaseNative extends Component {
 		return clamp(0, 1, 2 * overDistance / baseSize);
```
ybsung: It would be better to add the following `else` cluase to reduce unnecessary calculation.
```
	if (position < 0) {
		overDistance = -position;
	} else if (position > maxPos) {
		overDistance = position - maxPos;
	} else {
		return 0;
	}
```

ybsung: Simply, use Math.min instead of clamp for this case.
```
Math.min(1, 2 * overDistance / baseSize);
```

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -841,8 +846,10 @@ class ScrollableBaseNative extends Component {
 			this.props.clearAllOverscrollEffects();
 		}
 
-		this.setOverscrollStatus('horizontal', overscrollTypes.none);
-		this.setOverscrollStatus('vertical', overscrollTypes.none);
+		this.setOverscrollStatus('horizontal', 'before', overscrollTypes.none);
+		this.setOverscrollStatus('horizontal', 'after', overscrollTypes.none);
+		this.setOverscrollStatus('vertical', 'before', overscrollTypes.none);
+		this.setOverscrollStatus('vertical', 'after', overscrollTypes.none);
```
ybsung: The 4th parameter for `ratio` was omitted. Is it OK or is it better to add it for this case?

packages/ui/Scrollable/ScrollableNative.js
0x64: Okay!

packages/ui/Scrollable/ScrollableNative.js
0x64: It does not make a trouble for now, but it's a bug actually. Thank you for pointing out.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -693,79 +699,74 @@ class ScrollableBase extends Component {
...
-				} else if (type === overscrollTypes.sticking) {
-					if (position === 'before' && curPos <= 0) { // Already on the beginning egde
-						applyOverscrollEffect(orientation, 'before', type, ratio);
-					} else if (position === 'after' && curPos >= maxPos) { // Already on the ending egde
-						applyOverscrollEffect(orientation, 'after', type, ratio);
-					}
+				maxPos = this.getScrollBounds()[isVertical ? 'maxTop' : 'maxLeft'],
+				{none, scrolling} = overscrollTypes;
+
+			if (ratio > 0 && type === scrolling) {
```
ybsung: Why don't you mix the following conditions in a line?

AS-IS
```
if (applyOverscrollEffect && this.getScrollabilities(orientation)) {
if (ratio > 0 && type === scrolling) {
```

TO-BE
```
if (applyOverscrollEffect && this.getScrollabilities(orientation) && ratio > 0 && type === scrolling) {
```

packages/ui/Scrollable/Scrollable.js
```diff
@@ -693,79 +699,74 @@ class ScrollableBase extends Component {
...
+				if (position === 'before' && curPos <= 0) { // Already on the beginning egde
```
ybsung: typo error: egde -> edge

packages/ui/Scrollable/Scrollable.js
```diff
@@ -693,79 +699,74 @@ class ScrollableBase extends Component {
...
-						newType = overscrollTypes.none;
-						newRatio = 0;
-					} else if (position >= maxPos && curPos >= maxPos) { // Already on the ending egde
-						applyOverscrollEffect(orientation, 'after', type, newRatio);
-						newType = overscrollTypes.none;
-						newRatio = 0;
+				showEffectBefore = false,
+				showEffectAfter = false;
+
+			ratio = (ratio === null) ? this.getOverscrollRatio(orientation, targetPos) : ratio;
```
ybsung: Please do not reassign the value to a parameter of a function.
Reference:
https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#31-reassigning-a-defined-parameter-while-also-mentioning-arguments-in-the-body-in-sloppy-mode-only-typical-example

packages/ui/Scrollable/Scrollable.js
0x64: Sure. Probably I forgot to clear this.

0x64:
Hmm. it's a just clearing code and is not *required*. I will change the name of the first function.

ybsung:
I found there was one issue with dragging.
If dragging vertically to show overscroll effect on the top or on the bottom while showing overscroll effect on the left or on the right, the vertical and horizontal overscroll effects are hidden at the same time even though the horizontal overscroll effect was shown first.
I think it is the reason to call clearAllOverscrollEffects() in https://github.com/enactjs/enact/blob/8104e49a5352f574126d472734b4fad4a9872031/packages/ui/Scrollable/Scrollable.js#L560, then all overscroll effect are hidden at the same time.

0x64:
Thanks for reporting. I have already one ticket to handle drag cases and I will check this also.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -791,13 +794,17 @@ class ScrollableBase extends Component {
...
 
 		this.scrollLeft = clamp(0, maxValue, value);
 
-		if (type === overscrollTypes.scrolling) {
-			this.updateOverscrollEffect('horizontal', this.scrollLeft, type, ratio);
-		}
+		['before', 'after'].forEach((position) => {
+			const {type, ratio} = this.getOverscrollStatus(orientation, position);
+
+			if (type === overscrollTypes.scrolling) {
```
ybsung: I don't think we need this condition because we already did with the following condition in `updateOverscrollEffectByDirection()`.
```
if (ratio > 0 && type === scrolling)
```

0x64: Yes, we need this due to cases through `updateOverscrollEffect()`.

ybsung: If we could omit the condition here, we could just call this.updateOverscrollEffectByDirection(orientation, position). Then type and ratio could be get from this.getOverscrollStatus(orientation, position) not out of but in this.updateOverscrollEffectByDirection().

packages/ui/Scrollable/Scrollable.js
```diff
@@ -693,79 +699,74 @@ class ScrollableBase extends Component {
...
-					} else if (position === 'after' && curPos >= maxPos) { // Already on the ending egde
-						applyOverscrollEffect(orientation, 'after', type, ratio);
-					}
+				maxPos = this.getScrollBounds()[isVertical ? 'maxTop' : 'maxLeft'],
+				{none, scrolling} = overscrollTypes;
+
+			if (ratio > 0 && type === scrolling) {
+				if (position === 'before' && curPos <= 0) { // Already on the beginning egde
+					applyOverscrollEffect(orientation, 'before', type, ratio);
+					this.setOverscrollStatus(orientation, position, none, 0);
```
ybsung: Why don't you use `'before'` string instead of position variable for code readability?

packages/ui/Scrollable/Scrollable.js
0x64: My intention is not to call another function since this routine is possibly called frequently especially on dragging.

packages/ui/Scrollable/Scrollable.js
0x64: Okay.

0x64:
@ybsung 
> I found there was one issue with dragging.
If dragging vertically to show overscroll effect on the top or on the bottom while showing overscroll effect on the left or on the right, the vertical and horizontal overscroll effects are hidden at the same time even though the horizontal overscroll effect was shown first.

I thought you pointed out some flicking, but it looks a dragging case. All effects by dragging should be gone at the same time. It is not a problem. If I misunderstood, please let me know.

0x64:
To make relations between internal functions clearer and to remove redundant check codes, I updated to check conditions (to show effects) within a requesting methods (input handlers, scrolling logics, etc) instead of effect handling methods. It is a simple change of a concept but most code for overscroll effect is affected.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -533,7 +541,6 @@ class ScrollableBase extends Component {
 			targetX = (direction === 'vertical') ? 0 : this.dragStartX - this.getRtlX(ev.x), // 'horizontal' or 'both'
 			targetY = (direction === 'horizontal') ? 0 : this.dragStartY - ev.y; // 'vertical' or 'both'
 
-		this.updateOverscrollEffectOnPosition(targetX, targetY, overscrollTypes.sticking);
 		this.start({targetX, targetY, animate: false});
```
ybsung: From now, targetX, targetY variables are used once.

0x64: ✔️

packages/ui/Scrollable/Scrollable.js
```diff
@@ -823,22 +804,44 @@ class ScrollableBase extends Component {
...
 		}
-		if (Math.abs(bounds.maxTop - targetY) < epsilon) {
-			targetY = bounds.maxTop;
+
+		if (overscrollEffect) {
+			if (scrollLeft !== targetX && this.canScrollHorizontally(bounds)) {
+				const edge = this.isOverEdge(targetX, maxLeft);
+
+				if (isDragging) {
+					this.applyOverscrollEffectOnDrag('horizontal', edge, targetX, overscrollTypeHold);
```
ybsung: Do we need to call `applyOverscrollEffectOnDrag` when `edge` is equals to `null`?

packages/ui/Scrollable/Scrollable.js
```diff
@@ -648,35 +653,34 @@ class ScrollableBase extends Component {
...
-		});
+		this.start({targetX: this.accumulatedTargetX, targetY: this.accumulatedTargetY});
 	}
 
 	// overscroll effect
 
-	getScrollabilities = (orientation) => {
-		const bounds = this.getScrollBounds();
-		return (orientation === 'vertical') ? this.canScrollVertically(bounds) : this.canScrollHorizontally(bounds);
+	isOverEdge = (position, maxPosition) => {
```
ybsung: `get` function would be better than `is`.

0x64: What about `getEdgeFromPosition`?

ybsung: 👍

packages/ui/Scrollable/Scrollable.js
```diff
@@ -348,6 +350,8 @@ class ScrollableBase extends Component {
 			vertical: true
 		};
 
+		this.overscrollEnabled = !!(props.applyOverscrollEffect);
```
ybsung: All member variables should be initialized before using it from the point of view of V8 engine optimization

ybsung:
LGTM

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -391,7 +383,9 @@ class ScrollableBase extends Component {
 	scrollByPage = (keyCode) => {
 		// Only scroll by page when the vertical scrollbar is visible. Otherwise, treat the
 		// scroller as a plain container
-		if (!this.uiRef.state.isVerticalScrollbarVisible) return;
+		if (!this.uiRef.state.isVerticalScrollbarVisible) {
+			return true;
```
MikyungKim: I think we need to change the return value in this function. true <-> false

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -469,27 +460,48 @@ class ScrollableBase extends Component {
...
-						element = Spotlight.getCurrent(),
-						nextSpottable = element ? getTargetByDirectionFromElement(direction, element) : null;
-
-					if (!nextSpottable) { /* if the spotlight focus will not move */
-						const
-							isRtl = this.uiRef.state.rtl,
-							orientation = (direction === 'up' || direction === 'down') ? 'vertical' : 'horizontal',
-							position = (direction === 'up' || !isRtl && direction === 'left' || isRtl && direction === 'right') ? 'before' : 'after';
-						this.uiRef.updateOverscrollEffectByDirection(orientation, position, overscrollTypes.scrolling, 1);
+				direction = getDirection(keyCode);
```
MikyungKim: What about `else if (getDirection(keyCode)` ?

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -469,27 +460,48 @@ class ScrollableBase extends Component {
...
+				if (direction) {
```
MikyungKim: I think below is redundant.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -469,27 +460,48 @@ class ScrollableBase extends Component {
...
+				const
+					orientation = (direction === 'up' || direction === 'down') ? 'vertical' : 'horizontal',
+					bounds = this.uiRef.getScrollBounds(),
+					scrollability = orientation === 'vertical' ? this.uiRef.canScrollVertically(bounds) : this.uiRef.canScrollHorizontally(bounds);
+
+				if (scrollability) {
+					const
+						isRtl = this.uiRef.state.rtl,
+						edge = (direction === 'up' || !isRtl && direction === 'left' || isRtl && direction === 'right') ? 'before' : 'after';
+					this.uiRef.checkAndApplyOverscrollEffect(orientation, edge, overscrollTypeOnce, 1);
```
MikyungKim: ratio parameter can have default value '1'

packages/ui/Scrollable/Scrollable.js
```diff
@@ -688,95 +698,81 @@ class ScrollableBase extends Component {
...
-						newRatio = 0;
-					}
-				} else if (type === overscrollTypes.sticking) {
-					if (position === 'before' && curPos <= 0) { // Already on the beginning egde
-						applyOverscrollEffect(orientation, 'before', type, ratio);
-					} else if (position === 'after' && curPos >= maxPos) { // Already on the ending egde
-						applyOverscrollEffect(orientation, 'after', type, ratio);
-					}
-				}
+		if (edge === 'before' && curPos <= 0) { // On the beginning edge
```
MikyungKim: We could reduce the code here.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -688,95 +698,81 @@ class ScrollableBase extends Component {
...
+		} else {
+			this.setOverscrollStatus(orientation, edge, type, ratio);
+		}
+	}
 
-				this.setOverscrollStatus(orientation, newType, newRatio);
-			}
+	clearOverscrollEffect = (orientation, edge) => {
+		if (this.props.clearOverscrollEffect) {
+			this.props.clearOverscrollEffect(orientation, edge);
```
MikyungKim: I think we can call this function only if the type is not 'none' from UI side not in moonstone.

PR #1770, opened 2018-07-03
ENYO-5399: Use specific caret range of Storybook dependencies.

JayCanuck:
* It appears by using `^3` version for `@storybook` dependencies in a recent change https://github.com/enactjs/enact/commit/2f546210163a39b6d6d9f92a121d749c2317248d there is some internal conflict being created, such that sideloading onto webOS TV results in a blank white sample area.

### Resolution
* Using the specific `^3.4.8` release range value seems to get around the issue.


JayCanuck:
Closing out; does not fix issue

PR #1772, opened 2018-07-04
PLAT-61242: VirtualList/Scroller: Handle a page up or down key in pointer mode

ybsung:

If pressing a page up or down key in pointer mode in VirtualList or Scroller, the key is ignored

### Resolution

We misunderstood the Spotlight requirement, then we ignored key events in pointer mode. So we removed the guard ignoring it and focus an item after changing from pointer mode to 5 way key mode.

### Additional Consideration

There is the known issue that Scroller shuttles when pressing page up/down key. We'll fix it in PLAT-62228

### Links

PLAT-61242

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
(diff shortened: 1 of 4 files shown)


0x64:
LGTM

PR #1776, opened 2018-07-06
PLAT-61241: Handle direction keys properly over ScrollButtons when `focusableScrollbar` is false

ybsung:

Even though pressing up/down/left/right keys in the ScrollButton when `focusableScrollbar` is false, the ScrollButton keep get focus.

### Resolution

If there is any spottable component to the direction of the key, give it focus. If not, give a content focus.

The expected result with my PR is...

Step
1. Launch http://nebula.lgsvl.com/enyojs/enact-sampler/develop/
2. Select VirtualList
3. Wheel a little bit

- Hover the previous ScrollButton, then press a 'right' key.
Expected result is that the VirtualList item gets focus.
- Hover the previous ScrollButton, then press a 'left' key.
Expected result is that the VirtualList item gets focus.
- Hover the previous ScrollButton, then press a 'down' key.
Expected result is that the VirtualList item gets focus.
- Hover the previous ScrollButton, then press a 'up' key.
Expected result is that the 'close' button gets focus.
- Hover the previous ScrollButton, then press a 'page down' key.
Expected result is that the VirtualList scrolls.

### Links

PLAT-61241

### Comments

### DIFF
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -275,8 +275,7 @@ class ScrollButtons extends Component {
 		// We don't need to navigate manually if `focusableScrollbar` is `false`
 		if (focusableScrollbar) {
 			const
-				{keyCode} = ev,
-				direction = getDirection(keyCode),
+				direction = getDirection(ev.keyCode),
 				fromNextToPrev = (vertical && direction === 'up') || (!vertical && direction === (rtl ? 'right' : 'left')),
 				fromPrevToNext = (vertical && direction === 'down') || (!vertical && direction === (rtl ? 'left' : 'right'));
 
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -67,6 +67,16 @@ class ScrollButtons extends Component {
 		disabled: PropTypes.bool,
 
 		/**
+		 * When `true`, allows 5-way navigation to the scrollbar controls. By default, 5-way will
+		 * not move focus to the scrollbar controls.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		focusableScrollbar: PropTypes.bool,
+
+		/**
 		* Sets the hint string read when focusing the next button in the scroll bar.
 		*
 		* @type {String}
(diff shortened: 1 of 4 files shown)


packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -259,16 +269,34 @@ class ScrollButtons extends Component {
...
+				fromNextToPrev = (vertical && direction === 'up') || (!vertical && direction === (rtl ? 'right' : 'left')),
+				fromPrevToNext = (vertical && direction === 'down') || (!vertical && direction === (rtl ? 'left' : 'right'));
+
+			// manually focus the opposite scroll button when 5way pressed
+			if ((fromNextToPrev && target === this.nextButtonNodeRef) ||
+				(fromPrevToNext && target === this.prevButtonNodeRef)) {
+				this.focusOnOppositeScrollButton(ev, direction);
+			}
+		} else {
+			const direction = !vertical && 'up' || rtl && 'right' || 'left';
```
0x64: We need a comment about this code for the future works.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -259,16 +269,34 @@ class ScrollButtons extends Component {
...
+
+			if (Spotlight.getPointerMode()) {
+				setTimeout(() => {
```
0x64: Let's leave some comment like other async cases.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -279,23 +307,33 @@ class ScrollButtons extends Component {
...
 			{prevButtonDisabled} = this.state,
 			{keyCode} = ev;
 
-		if (isPageUp(keyCode) && !prevButtonDisabled) {
-			Spotlight.focus(this.prevButtonNodeRef);
+		if (isPageUp(keyCode)) {
+			if (focusableScrollbar && !prevButtonDisabled) {
+				Spotlight.focus(this.prevButtonNodeRef);
+			} else if (!focusableScrollbar) {
+				this.onClickPrev(ev);
```
0x64: Do we have a requirement to scroll up in this case?

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -279,23 +307,33 @@ class ScrollButtons extends Component {
...
 			{nextButtonDisabled} = this.state,
 			{keyCode} = ev;
 
-		if (isPageDown(keyCode) && !nextButtonDisabled) {
-			Spotlight.focus(this.nextButtonNodeRef);
+		if (isPageDown(keyCode)) {
+			if (focusableScrollbar && !nextButtonDisabled) {
+				Spotlight.focus(this.nextButtonNodeRef);
+			} else if (!focusableScrollbar) {
+				this.onClickNext(ev);
```
0x64: Do we have a requirement to scroll down in this case?

packages/moonstone/Scrollable/ScrollButtons.js
ybsung: I just implemented it as Enyo VirtualList and Scroll do.

packages/moonstone/Scrollable/ScrollButtons.js
ybsung: Same above.

0x64:
LGTM

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -259,16 +269,38 @@ class ScrollButtons extends Component {
...
+				// When changing from "pointer" mode to "5way key" mode,
+				// a pointer is hidden and a last focused item get focused after 30ms.
+				// To make sure the content in `VirtualList` or `Scroller` to be focused after that, we used 50ms.
+				setTimeout(() => {
+					if (Spotlight.getCurrent() === target) {
+						Spotlight.move(direction);
+					}
+				}, 50);
+			} else {
+				Spotlight.move(direction);
```
MikyungKim: I think we need `if (Spotlight.getCurrent() === target)` this condition.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -66,6 +66,16 @@ class ScrollButtons extends Component {
...
 
+		/**
+		 * When `true`, allows 5-way navigation to the scrollbar controls. By default, 5-way will
+		 * not move focus to the scrollbar controls.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		focusableScrollbar: PropTypes.bool,
```
MikyungKim: How about just `focusable`?

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -279,23 +311,33 @@ class ScrollButtons extends Component {
...
 	onKeyDownPrev = (ev) => {
 		const
+			{focusableScrollbar} = this.props,
 			{nextButtonDisabled} = this.state,
 			{keyCode} = ev;
 
-		if (isPageDown(keyCode) && !nextButtonDisabled) {
-			Spotlight.focus(this.nextButtonNodeRef);
+		if (isPageDown(keyCode)) {
+			if (focusableScrollbar && !nextButtonDisabled) {
```
MikyungKim: I think `!nextButtonDisabled` should go upper if.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -279,23 +311,33 @@ class ScrollButtons extends Component {
...
 	onKeyDownNext = (ev) => {
 		const
+			{focusableScrollbar} = this.props,
 			{prevButtonDisabled} = this.state,
 			{keyCode} = ev;
 
-		if (isPageUp(keyCode) && !prevButtonDisabled) {
-			Spotlight.focus(this.prevButtonNodeRef);
+		if (isPageUp(keyCode)) {
+			if (focusableScrollbar && !prevButtonDisabled) {
```
MikyungKim: Same here

packages/moonstone/Scrollable/ScrollButtons.js
ybsung: I'll change it to `focusableScrollButtons`.

PR #1778, opened 2018-07-06
PLAT-58640: Fix VirtualList focus behavior when entering 5way mode

germboy:
When in pointer mode, pressing a 5-way directional key (switching from pointer-mode to 5-way mode) while a `VirtualList` a) is using an `isItemDisabled` function prop and b) contains a currently focused item, focus will:
- not change on Chrome and is unable to switch into 5-way mode, even with repeated 5-way directional presses
- not change initially on TV, but will only change with successive 5-way directional presses

### Resolution
There is some logic in the `VirtualList` internal method `jumpToSpottableItem` that is not called or used unless the instance is using a `isItemDisabled` prop. That logic will request focus to be changed to the "next" list item. However, since this method is being called via an internal `keydown` event (and propagation will be stopped anyway), spotlight will be unable to turn off pointer mode, so a focus change cannot occur.

This is remedied by turning off pointer mode in response to the keydown event that initiates the focus change call. We don't usually want to explicitly change the pointer mode status, however this scenario fits the use-case (see `moonstone/Popup.Popup`) where it should be safe to do so as:
- it's called in response to a `keydown` event
- is guarded by `getDirection(keyCode)`
- a focus change is required that would otherwise be prevented by being in pointer mode

### Comments
I didn't notice any unexpected side-effects to focus behavior when applying this fix and testing the various focus behaviors contained in `VirtualList` (switching between pointer & 5-way modes, page up/down behavior from items and/or scroll buttons, scroll wheel focus behavior, etc.).

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -662,6 +662,7 @@ const VirtualListBaseFactory = (type) => {
 			if (getDirection(keyCode)) {
 				ev.preventDefault();
 				this.setSpotlightContainerRestrict(keyCode, target);
+				Spotlight.setPointerMode(false);
 				if (this.jumpToSpottableItem(keyCode, repeat, target)) {
 					ev.stopPropagation();
 				}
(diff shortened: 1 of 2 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -662,6 +662,7 @@ const VirtualListBaseFactory = (type) => {
 			if (getDirection(keyCode)) {
 				ev.preventDefault();
 				this.setSpotlightContainerRestrict(keyCode, target);
+				Spotlight.setPointerMode(false);
```
ybsung: @webOS101 
Do you agree to call `Spotlight.setPointerMode(false)` in `VirtualList? You did not recommend to do it before. We did not consider it as our solution.

ryanjduffy: Generally, that's true. We should avoid manually setting pointer mode to allow Spotlight to manage it according to its rules. However in this case, we _may_ be preventing Spotlight's management by stopping event propagation. There's likely some API refinement we should do to better handle this but, in the mean time, this approach within the framework components is usually acceptable.

@germboy - In that light, should we move the call to `setPointerMode()` into the `if` block below to only set it in the case that we'd be preventing Spotlight management?

ybsung: @ryanjduffy
Thanks for clarifying it.

PR #1779, opened 2018-07-06
ENYO-5406: Fix spotlight navigation within Popup without onClose handler

rundmt:
When popup is set with no children and with no `onClose` prop spotlight will still move throughout the background Panel.

### Resolution
Moved the code to check for `onClose` after the `stopPropagation`. This allows for the same behavior for Spotlight whether you have `onClose` or not.

### Additional Considerations
The `onClose` check did check a bigger block of code than my change. I think it's unneeded, but there could be something I'm missing.

### Links
ENYO-5406


lishammel:
closing out as jira associated with this is closed

PR #1786, opened 2018-07-09
ENYO-5377: Dialog: 'open' prop is applied as opposed to the actual popup state

haileyR:
- ENYO-5377
- The `floatLayerOpen` state now reflects the `open` prop of popup. 

### Resolution
This change will open for close the floating layer that contains the popup according to the popup's `open` prop.

### Additional Considerations
- When a popup should open, there should be a floating layer open. So having  handleFloatingLayerOpen(), handlePopupHide() and handlePopupShow() to handle open/close state changes in addition to state updates in componentWillReceiveProps() seems overcomplicate things and duplicates jobs. We should consider refactoring this.
- Also, why the `popupOpen` state is set to `noAnimation` prop?

Enact-DCO-1.0-Signed-off-by: HanGyeol Ryu hangyeol.ryu@lge.com


webOS101:
Please add Enact DCO

lishammel:
closing out as jira associated with this is closed

PR #1788, opened 2018-07-11
ENYO-5367: Fix spotlight pause in ExpandableItem with no children

viodragon2:
Opening `ExpandableItem` with no children in 5-way results in indefinite spotlight pause.

### Resolution
Handle `onOpen` and `onClose` instead of `open` prop change to pause.

See #1634


packages/moonstone/ExpandableItem/ExpandableSpotlightDecorator.js
```diff
@@ -183,6 +176,18 @@ const ExpandableSpotlightDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+			forward('onClose'),
+			this.pause
+		)
+
+		handleOpen = this.handle(
+			forward('onOpen'),
+			this.pause
+		)
+
 		handleShow = () => {
```
teckliew: Maybe we could use `this.handle` here for consistency?

viodragon2: The change would look something like
```
		handleShow = this.handle(
			() => {
				this.resume();
				return true;
			},
			() => this.highlight(this.highlightContents)
		)
```
which I don't think it's necessary.

ryanjduffy:
This looks good! I was concerned about regressions but everything appears to be working as expected. 👏 👏

PR #1789, opened 2018-07-11
PLAT-58297: Add voice control to Pickers

changgilee:
- This PR will be support picker value setting with voice
(Picker, RangePicker, DatePicker, TimePicker are added this feature)
- "Select" intent is used for implement this feature
- When user speak like "Group Name" + "value", matched value will be set in Picker

### Resolution

### Links
PLAT-58297

### Comments

### DIFF
--- a/packages/moonstone/DatePicker/DatePickerBase.js
+++ b/packages/moonstone/DatePicker/DatePickerBase.js
@@ -96,6 +96,14 @@ const DatePickerBase = kind({
 		year: PropTypes.number.isRequired,
 
 		/**
+		 * Disable voice control feature.
+		 *
+		 * @type {Boolean}
+		 * @public
+		 */
+		'data-webos-voice-disabled': PropTypes.bool,
+
+		/**
 		 * The "aria-label" for the day picker.
 		 *
 		 * @type {String}
@@ -280,6 +288,7 @@ const DatePickerBase = kind({
 	},
 
 	render: ({
+		'data-webos-voice-disabled': voiceDisabled,
 		day,
 		dayAriaLabel,
 		dayLabel = $L('day'),
@@ -312,6 +321,7 @@ const DatePickerBase = kind({
 				{...rest}
 				showLabel="always"
 				autoClose={false}
+				data-webos-voice-disabled={voiceDisabled}
 				lockBottom={false}
 				onSpotlightDisappear={onSpotlightDisappear}
 				onSpotlightLeft={onSpotlightLeft}
@@ -332,6 +342,8 @@ const DatePickerBase = kind({
 										accessibilityHint={dayLabel}
 										aria-label={dayAriaLabel}
 										className={css.day}
+										data-webos-voice-disabled={voiceDisabled}
+										data-webos-voice-group-label={dayLabel}
 										key="day-picker"
 										label={noLabels ? null : dayLabel}
 										max={maxDays}
@@ -352,6 +364,8 @@ const DatePickerBase = kind({
 										accessibilityHint={monthLabel}
 										aria-label={monthAriaLabel}
 										className={css.month}
+										data-webos-voice-disabled={voiceDisabled}
+										data-webos-voice-group-label={monthLabel}
 										key="month-picker"
 										label={noLabels ? null : monthLabel}
 										max={maxMonths}
@@ -372,6 +386,8 @@ const DatePickerBase = kind({
 										accessibilityHint={yearLabel}
 										aria-label={yearAriaLabel}
 										className={css.year}
+										data-webos-voice-disabled={voiceDisabled}
+										data-webos-voice-group-label={yearLabel}
 										key="year-picker"
 										label={noLabels ? null : yearLabel}
 										max={maxYear}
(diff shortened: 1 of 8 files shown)


packages/moonstone/Picker/Picker.js
```diff
@@ -187,14 +187,17 @@ const PickerBase = kind({
 				validateRange(value, 0, max, 'Picker', '"value"', 'min', 'max index');
 			}
 			return clamp(0, max, value);
+		},
+		voiceLabel: ({children}) => {
+			return JSON.stringify(children.map((child) => (typeof child === 'object' ? child.props.children : child)));
```
webOS101: I don't think this is a safe way to extract the values from `children`.  Any arbitrary node could be passed and calling `stringify` on an unknown object is not a very good idea.  Perhaps only handle the most common cases and don't supply a voice label for others?

packages/moonstone/DatePicker/DatePickerBase.js
```diff
@@ -95,6 +95,14 @@ const DatePickerBase = kind({
 		 */
 		year: PropTypes.number.isRequired,
 
+		/**
+		 * Disable voice control feature.
+		 *
+		 * @type {Boolean}
+		 * @public
```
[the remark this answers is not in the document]
changgilee: I will add it~

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -69,9 +69,36 @@ const ExpandableItemBase = kind({
 		 */
 		title: PropTypes.string.isRequired,
 
-		// TODO: Document voice control props and make public
+		/**
+		 * Disable voice control feature.
+		 *
+		 * @type {Boolean}
+		 * @public
```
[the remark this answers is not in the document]
changgilee: I will add it~

packages/moonstone/internal/Picker/Picker.js
```diff
@@ -669,6 +676,17 @@ const PickerBase = class extends React.Component {
 		}
 	}
 
+	handleVoice = (ev) => {
+		const {max, min, onChange, value} = this.props;
+		const voiceIndex = ev && ev.detail && ev.detail.matchedIndex && Number(ev.detail.matchedIndex);
```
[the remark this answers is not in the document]
changgilee: - I did insert Number.isInteger() before using it.
- In current state, Web Engine is delegate matchedIndex value as string type. But it seems more better that check matchedIndex refer to your concern. So, I did change to check whether it is undefined or not.

packages/moonstone/Picker/Picker.js
changgilee: - As you mentioned, error can be occur when unexpected children is used in Picker(Thanks for comment about it)
- I did check that how to picker is used in application side. In most case, they declare array at children. But there is still possibility that div(object) can be used.
- Basically, I try to set data-webos-voice-labels-ext automatically refer to children if the item's type is string or number.
- Exceptional case, we can't gain proper children label value only component side. In this case, app developer has to declare 'data-webos-voice-labels-ext' attribute and it will be work as expected.

packages/moonstone/Picker/Picker.js
```diff
@@ -48,6 +48,16 @@ const PickerBase = kind({
 		 */
 		'aria-valuetext': PropTypes.string,
 
+		/**
+		 * By default, `data-webos-voice-labels-ext` is generated reference to children.
+		 * However, if the children is not an array of number or string, label value should be explicitly decalred.
+		 *
+		 * @type {Number[], String[]}
```
webOS101: I believe this needs to be `{Number[]|String[]}`

packages/moonstone/Picker/Picker.js
```diff
@@ -187,14 +197,23 @@ const PickerBase = kind({
 				validateRange(value, 0, max, 'Picker', '"value"', 'min', 'max index');
 			}
 			return clamp(0, max, value);
+		},
+		voiceLabel: ({children, 'data-webos-voice-labels-ext': voiceLabelsExt}) => {
+			let voiceLabel;
+			if (voiceLabelsExt) {
+				voiceLabel = voiceLabelsExt;
+			} else {
+				voiceLabel = children.map((child) => ((typeof child === 'number' || typeof child === 'string') ? child : ''));
```
webOS101: I might be good to verify that `children` does exist and is an array (or array-like object) before using.  Even better would be to use `React.Children.map()`

packages/moonstone/Picker/Picker.js
changgilee: I did fix it~

packages/moonstone/Picker/Picker.js
changgilee: I did add array check condition of children and change to React.Children.map(). Thanks~

packages/moonstone/Picker/Picker.js
```diff
@@ -203,7 +203,7 @@ const PickerBase = kind({
 			if (voiceLabelsExt) {
 				voiceLabel = voiceLabelsExt;
 			} else {
-				voiceLabel = children.map((child) => ((typeof child === 'number' || typeof child === 'string') ? child : ''));
+				voiceLabel = children && Array.isArray(children) && React.Children.map(children, (child) => ((typeof child === 'number' || typeof child === 'string') ? child : ''));
```
webOS101: It's not necessary to do both checks `React.Children.map`, I believe.  I'll fix this up.

PR #1790, opened 2018-07-11
PLAT-62334:  Add data-spotlight-container-disabled knob in Scroller qa-sampler

ybsung:

We need the way to check if a `Scroller` supports `data-spotlight-container-disabled` prop well.

### Resolution

I added `data-spotlight-container-disabled` knob in Scroller qa-sampler

### Additional Considerations

### Links

PLAT-62334

### Comments

### DIFF
--- a/packages/sampler/stories/qa-stories/Scroller.js
+++ b/packages/sampler/stories/qa-stories/Scroller.js
@@ -1,5 +1,6 @@
 import Button from '@enact/moonstone/Button';
 import ExpandableList from '@enact/moonstone/ExpandableList';
+import {ScrollableBase} from '@enact/moonstone/Scrollable';
 import Scroller from '@enact/moonstone/Scroller';
 import Item from '@enact/moonstone/Item';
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa-stories/Scroller.js
```diff
@@ -28,6 +29,7 @@ storiesOf('Scroller', module)
 		'List of things',
 		() => (
 			<Scroller
+				data-spotlight-container-disabled={boolean('data-spotlight-container-disabled', ScrollableBase, false)}
```
0x64: Do we need `ScrollableBase` instead of `Scroller` here?

ybsung: No. I missed it. Thanks.

0x64:
LGTM

MikyungKim:
LGTM, but I'm worried about the behavior we saw that "data-spotlight-container-disabled" attribute is working on the node which does not have "data-spotlight-container". Please investigate this further more.

PR #1792, opened 2018-07-11
ENYO-5405: Fix Picker 5-way focus behavior

germboy:
When using 5-way to navigate between buttons in a non-joined `Picker` component, certain layouts can cause focus to change to other outside elements, rather than the intended alternate `Picker` button. 

Spotlight attempts to set focus to the natural "next" target based on the direction pressed. The targets gaining focus here are generally expected, however when pressing a 5-way directional key towards an opposing `Picker` button, it is always assumed that the other button should gain focus.

### Resolution
Since we handle the `Picker`'s internal button `keydown` events and have access to the spottables, we can decide during `keydown` when to prevent default spotlight behavior and set focus elsewhere, if needed.

### DIFF
--- a/packages/moonstone/internal/Picker/Picker.js
+++ b/packages/moonstone/internal/Picker/Picker.js
@@ -1,8 +1,8 @@
-import {forward} from '@enact/core/handle';
+import {forward, stopImmediate} from '@enact/core/handle';
 import clamp from 'ramda/src/clamp';
 import equals from 'ramda/src/equals';
 import {is} from '@enact/core/keymap';
-import {Job} from '@enact/core/util';
+import {cap, Job} from '@enact/core/util';
 import React from 'react';
 import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


packages/moonstone/internal/Picker/Picker.js
```diff
@@ -627,45 +627,55 @@ const PickerBase = class extends React.Component {
...
-		} else if (isLeft(keyCode) && onPickerSpotlightLeft && (orientation === 'vertical' || this.hasReachedBound(step * -1))) {
-			onPickerSpotlightLeft(ev);
-		} else if (isRight(keyCode) && onPickerSpotlightRight) {
-			onPickerSpotlightRight(ev);
-		} else if (isUp(keyCode) && onPickerSpotlightUp) {
-			onPickerSpotlightUp(ev);
+		const direction = getDirection(keyCode);
+
+		if (direction) {
+			const {orientation, step, ...rest} = this.props;
```
ryanjduffy: Somewhat minor optimization but there's no need to use `...rest` here which causes (in babel, at least) an extra object allocation and an iteration over all the keys to drop the specified values.

Instead, it could be something like:
```
const {orientation, step} = this.props;

if (...) {
  // existing code
} else {
  forward(`onPickerSpotlight${cap(direction)}`, ev, this.props);
}
```

haileyR:
This works as expected, but we found another bug from a **joined vertical picker** within a item list.
5-way into joined vertical picker itself works as expected, but we have an issue with using 5-way left and right.

packages/moonstone/internal/Picker/Picker.js
germboy: Ah, that's definitely cleaner - good suggestion 👍

germboy:
@haileyR That appears to be a consequence of a vertical `Picker` being equal distance from 2 `Item` components, so the 1st match will receive focus. As I commented on ENYO-5421, I wouldn't classify this as a bug - it's a layout issue. If a developer wished to direct focus elsewhere, they should use directional spotlight function props (`onSpotlightRight`) on the `Picker` to do so.

PR #1793, opened 2018-07-11
PLAT-62474: Fix spotlight in Scrollable after flick

germboy:
… event

Flick events emitted in `Scrollable` can cause the spotlight container within to become stuck in a "disabled" state, making it impossible to set focus to components within the `Scroller`.

### Resolution
Flick events in `Scrollable` immediately disable the container, relying on its own scroller `stop` method to re-enable the container when the scrolling (or "flicked") animation comes to an end. When disabling the container, `Scrollable` was not taking into account whether or not its contents can actually scroll or not. If the contents cannot scroll, then the `stop` method is not called, thus can never re-enable the container.

Before disabling the container, we should check to see the `direction` of the flick event and whether or not we can scroll in that direction.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/ProgressBar.ProgressBarTooltip` not to display unknown props warning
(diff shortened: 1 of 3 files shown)


0x64:
@germboy I'm not 100% sure but `focusedItem.blur()` may not be needed either if `Scrollable` cannot actually scroll.

PR #1795, opened 2018-07-12
ENYO-5424: VirtualListNative: scroll animation occurs when calling scrollTo with animate false (1 QE)

yeram:
Fix unexpected scrolling animation when calling scrollTo API

### Resolution
Move settings some values inside the condition that starts scroll

### Additional Considerations

### Links
ENYO-5424


MikyungKim:
LGTM

PR #1797, opened 2018-07-12
ENYO-5355: Show Title/TitleBelow only when title is available.

haileyR:

Resolves error to show only titleBelow when title is not provided.

### Resolution
Only shows title block when we have title provided.

### Additional Considerations
Buttons are wrapped in the `titleWrapper` class. We might need to rename it or move buttons out from the wrapper.

### Links
ENYO-5355


haileyR:
Updated as per change request. @webOS101

PR #1798, opened 2018-07-13
ENYO-5356: Fix Feedback tooltip showing condition

viodragon2:
Feedback tooltip should display only time when hovering out from slider while playing. Currently, it hides the tooltip.

### Resolution
We do not have a boolean condition for showing feedback icon as "play" relies on "idle", "focus", and "blur" conditions to display either both icon and time or just time in the tooltip.

Replaced `noFeedback` prop with `action` prop in `FeedbackTooltip` to handle the state.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/ProgressBar.ProgressBarTooltip` not to display unknown props warning
+- `moonstone/VideoPlayer` to display feedback tooltip when hover out from slider while playing
 
 ## [2.0.0-rc.1] - 2018-07-09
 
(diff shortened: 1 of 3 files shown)


ryanjduffy:
I wonder if the difficulty of naming comes from misplaced responsibilities. Is the data passed in `action` something that the tooltip needs to understand or should those rules be owned by the consumer and the result passed instead?

viodragon2:
`FeedbackTootlip` determines the visibility of icons by itself to a certain degree already. It considers `playBackState` and the "visibility" of the tooltip. This was an oversight. The "visibility" of the tooltip is determined by "how it was invoked to show tooltip". Hence, the name "action". I agree that the word "action" gives you an active element type of feel which sounds like it should be taken care by the consumer, so something more passive would suit better? (e.g. 'invokedBy', 'shownCondition', 'visibleState', etc.) All the examples I can think of doesn't really fit tho 🤷‍♂️ 

On a sidenote, I would personally **love** to drop this requirement and simply display the play icon on hover out TBH. It's a complexity for almost no UX benefits.

ryanjduffy:
Thanks for the explanation, @viodragon2. Given that it's a private API, I'm not too concerned with it right now. Let's take the fix as is and we can revisit later if necessary. :+1: for incremental improvements!

PR #1804, opened 2018-07-16
ENYO-5431: Set initial value for boolean knobs to avoid getting warning messages.

haileyR:
No more warning message on knob setting for boolean values.

### Resolution
Set initial values for boolean values.

related to https://github.com/twisty/formsy-react-components/issues/66


teckliew:
Weird how `DatePicker`'s `noLabel` prop doesn't throw an error without the initial value 🤔

Djspaceg:
With this set of changes, had we considered just assigning `false` as the default return for our `boolean()` function if there is no default set by `defaultProps`? This would insulate us from having to make 196 line changes and future knobs.

Djspaceg:
Closed in favor of #1819

PR #1661, opened 2018-05-17
ENYO-5273: VideoPlayer show MediaControls on mount

teckliew:
### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
VideoPlayer should show `MediaControls` on mount and also whenever it switches to another video.
Set the default state to show the title and `MediaControls` and show controls if the `source` is changed.

### Links
[//]: # (Related issues, references)
ENYO-5273

### Comments
Enact-DCO-1.0-Signed-off-by: Teck Liew teck.liew@lge.com

PR #1805, opened 2018-07-17
ENYO-5434: VideoPlayer bottom control renders

viodragon2:
Some apps control the visibility of bottom controls via imperative APIs (`showControls` and `hideControls`) because there are cases where app wants to prevent showing controls upon newly loaded/played event. For example, the app does not want to display controls at everytime video changes when in store demo mode.
Because apps control the visibility of bottom controls in a similar way as frameworks do, it could render unwanted results.

### Resolution
Apps shouldn't need to control the visibilities via calling imperative APIs if we allow them to prevent the default action via event.

### Additional Considerations
The general idea behind this is that apps want to display bottom controls when initially playing a video in most cases. So instead of letting apps to take care of, we fixed the framework to support that in #1661. 
Proposed change can still break the app as long as apps try to handle the same requirement on their end. We _can_ remove supporting "auto showing controls" and let app control everything entirely, too.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
+### Changed
+
+- `moonstone/VideoPlayer` to allow preventing bottom control render via play event handler
+
 ### Fixed
 
 - `moonstone/Input` to not focus by *tab* key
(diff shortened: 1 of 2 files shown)


viodragon2:
My initial implementation of allowing to prevent "play" event wouldn't work as play button pressed on a remote should only display the mini feedback. So I've decided to handle all visibility control in `loadstart` event. However, preventing `loadstart` doesn't seem intuitive as it needs to reset internal states no matter what. I've created an alternative approach to add additional prop in #1829.

ryanjduffy:
Closing in favor of #1829

PR #1807, opened 2018-07-17
PLAT-62418: Fixed to scroll properly when holding down paging control buttons

0x64:
Fixed a bug that a list or a scroller does not scroll until a user releases the button if a user holds down a paging control button right after pressing the opposite paging control button.

### Resolution
After #1570, a focus does not move to the opposite paired button but it stays on the current button even if the button is disabled by scrolling. This bug occurs by old code to prevent scrolling from `holdpulse` events when a focus moved to the opposite button during holding down.
So, removed the old code.

### Links
PLAT-62418

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


ybsung:
LGTM, but please update CHANGELOG.md.

0x64:
@ybsung I will update CHANGELOG.md later. As you know, it can make a conflict frequently if we update it too early, so intentionally I defer updating normally.

MikyungKim:
Please add CHANGELOG.

0x64:
Updated CHANGELOG.md!

MikyungKim:
LGTM

PR #1811, opened 2018-07-19
ENYO-5455: Fix audio guidance issue about more button in video player

xbinary:
1) When the `more button` is blurred and focused for audio guidance, it  must be `visible === true`
2) `blur()` and `focus()` of the button requires a brief delay for readout `infoComponents` first

### Resolution
1) Add condition to `this.props.visible`
2) Add `setTimeout` for delay

### Links
ENYO-5455

### Comments
Enact-DCO-1.0-Signed-off-by: Hyeok Jo hyeok.jo@lge.com

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -674,8 +674,12 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 				// Readout 'more' or 'back' button explicitly.
 				let selectedButton = Spotlight.getCurrent();
 				if (selectedButton === this.mediaControlsNode.querySelector(`.${css.moreButton}`)) {
-					selectedButton.blur();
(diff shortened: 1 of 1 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -674,8 +674,12 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 				// Readout 'more' or 'back' button explicitly.
 				let selectedButton = Spotlight.getCurrent();
 				if (selectedButton === this.mediaControlsNode.querySelector(`.${css.moreButton}`)) {
-					selectedButton.blur();
-					selectedButton.focus();
+					if (this.props.visible) {
+						setTimeout(() => {
```
teckliew: Not sure if it's the right approach to defer focus here, but if it is, you could let `core/Job.start()` handle it.

PR #1814, opened 2018-07-19
PLAT-62358: Updated the RadioItem stroke to ~"2px"

Djspaceg:
The circle thickness is just too nice looking and convenient.

### Resolution
The stroke has been updated to be thinner, and use the one value we simply can't incorporate into resolution independence.

### Additional
There may be a really terrible solution involving `calc()`. Something like `calc(2apx * 1rem / 24px)`


webOS101:
https://www.youtube.com/watch?v=69fPof-ZTnU

ryanjduffy:
They alignment of the selected radio button is a bit off in Chrome for FHD. It appears correct for HD and 4k, however. Does it render correctly on the TV?

image

Djspaceg:
The effect you're seeing, @ryanjduffy, is the consequence of not using a multiple of 3, then scaling by 2/3. When previewing this on a HiDPI display, you must view it at 100% scale emulation. The following is what I see when at 100% scale for each of the resolutions. We'll need to properly verify on the TV, due to this awful rounding issue.
combinedimages

My idea from above (`calc(2apx * 1rem / 24px)`) will need some tuning due to the fact that evidently, one cannot multiply values in CSS that have different units, un-unit-ed or addition/subtraction works just fine... Alternatively, if we use `apx` (absolute pixels, an invention of ours) the result will be HD being too bold and 4K UHD being too thin, with FHD at the Goldilocks sweet spot of 2px. :facepalm:

webOS101:
It is off-center on those, too.  Just less noticeably.

Djspaceg:
It's only off-center in 4k. The others are perfectly centered.

If you have a better suggestion of how to get a 2px border in a system that doesn't allow 2px, I'm happy to hear it.

ryanjduffy:
I think it's the best we can do with `ri`. Would this be an appropriate occasion to use a media query to achieve the exact requirement rather than approximating it with `ri`?

Djspaceg:
The most straight-forward approach would be to hard-code (via LESS math) `apx` values for each of the key RI classes (not media queries). I'll submit an update to this ticket using that approach.

Djspaceg:
PR Updated

Djspaceg:
Good change. I'm going to tinker a little more with this to see if I can snap the standard value to `16px`  instead of `16.031px`.

PR #1815, opened 2018-07-19
ENYO-5454 - Fix closeOnSelect spotlight action

webOS101:

When switching from pointer mode to 5-way mode after an expandable
with `closeOnSelect` collapsed it was possible for spotlight to remain
on the hidden contents of the expandable.

### Resolution
Blur the hidden item when in pointer mode.

### Additional Considerations
I removed an old test that would unconditionally try to spotlight the header
when in pointer mode and closing.  I believe the current solution works
better than what the previous code was trying (and failing) to do.

### Links
ENYO-5454

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ExpandableItem` and other expandable components to spotlight correctly when switching from
+	pointer mode to 5-way with `closeOnSelect`
+
 ## [2.0.0-rc.2] - 2018-07-16
 
 ### Added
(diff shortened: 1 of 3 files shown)


packages/spotlight/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ExpandableItem` and other expandable components to spotlight correctly when switching from
```
teckliew: Shouldn't this be in the `moonstone` changelog instead?

webOS101: :disappear:

ryanjduffy:
@webOS101 - I noticed that when closing an Expandable with the pointer, focus would be lost even though the pointer did not leave the label. This was caused by the `blur()` call in `highlightLabeledItem`. To address this, I added an early bail out to that method when `current === label` + some refactoring to support it.

PR #1816, opened 2018-07-20
ENYO-5449: Fix Slider to prevent 5-way actions when activated

teckliew:
`Slider` incrementing/decrementing handlers to use `stopPropagation` instead of `stopImmediatePropagation` 🤔

### Links
ENYO-5449

### Comments


ryanjduffy:
:lgtm: The reason this works is that spotlight is listening for events on `window` whereas both React and `core/dispatcher` are listening on `document` (`document` is the default for `dispatcher` but can be configured per listener). As a result, calling `stopPropagation` prevents the event both from from bubbling up the virtual dom _as well as_ the real DOM (from `document` -> `window`).

I'm not sure `stopImmediate` was ever necessary for this case but I do know it was needed in other places previously. We'll need to keep our eyes out for those and fix them if they are causing issues.

teckliew:
so `stopImmediate` doesn't stop event bubbling in the virtual DOM?

ryanjduffy:
correct. It _could_ ... and maybe should for functional parity with the native events ... but currently it only calls `ev.nativeEvent.stopImmediatePropagation` which doesn't affect the virtual dom.

PR #1817, opened 2018-07-20
ENYO-5442: Defer to reset Scrollable restore flag

ybsung:

ui/Scrollable.componentDidUpdate tried to restore scroll position and focus, then this.scrollToInfo is equal to `null`.
But moonstone/Scrollable.componentDidUpdate also tried to sync scroll position if this.scrollToInfo is equal to `null`.
So two logics conflicted, then the VirtualList worked incorrectly.

### Resolution

For the case to restore scroll position, we don't need to sync scroll position in moonstone/Scrollable. To do it, I deferred to reset this.scrollToInfo to `null`.

### Additional Considerations

I don't want to add any flags more for this issue. So I used `setTimeout`.

### Links

ENYO-5442

### Comments

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -1000,7 +1000,7 @@ class ScrollableBase extends Component {
 			if (this.props.scrollTo) {
 				this.props.scrollTo(opt);
 			}
-			this.scrollToInfo = null;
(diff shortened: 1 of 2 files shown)


ybsung:

0x64:
LGTM

ybsung:
@ryanjduffy 
As you mentioned, I'll keep in mind to find better ideas. Thanks.

PR #1819, opened 2018-07-23
ENYO-5431 Sampler: VideoPlayer: Selecting 'noAutoPlay' knob causes console 'Uncontrolled input' Warning

haileyR:
Sampler: VideoPlayer: Selecting 'noAutoPlay' knob causes console 'Uncontrolled input' Warning

### Resolution
Set defaultValue for boolean props that are not defined to false.

### DIFF
--- a/packages/sampler/src/enact-knobs/boolean.js
+++ b/packages/sampler/src/enact-knobs/boolean.js
@@ -30,7 +30,10 @@ const boolean = (name, Config, preferredValue) => {
 		Config.groupId = Config.displayName;
 	}
 
-	return nullify(booleanKnob(name, (preferredValue != null ? preferredValue : Config.defaultProps[name]), Config.groupId));
+	// Set false for default boolean props that are not defined.
+	const defaultValue = Config.defaultProps[name] != null ? Config.defaultProps[name] : false;
+	// console.log(name, Config, preferredValue, preferredValue != null, Config.defaultProps);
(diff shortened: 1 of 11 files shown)


packages/sampler/src/enact-knobs/boolean.js
```diff
@@ -30,7 +30,10 @@ const boolean = (name, Config, preferredValue) => {
 		Config.groupId = Config.displayName;
 	}
 
-	return nullify(booleanKnob(name, (preferredValue != null ? preferredValue : Config.defaultProps[name]), Config.groupId));
+	// Set false for default boolean props that are not defined.
+	const defaultValue = (Config.defaultProps[name] != null) ? Config.defaultProps[name] : false;
+	// console.log(name, Config, preferredValue, preferredValue != null, Config.defaultProps);
```
teckliew: should we remove this log?

teckliew:
Looks good, but there are a few samples that needed `mergeComponentMetadata` to work. For example, the prop `disabled` for `RadioItem` in QA-sample: `disabled={boolean('disabled', RadioItem, false)}` needs to be `disabled={boolean('disabled', Config, false)}` where `const Config = mergeComponentMetadata('RadioItem', RadioItem);` to work.

Maybe we should just have a `mergeComponentMetadata` built into the `enact-knobs`? It's getting a bit too repetitive to have to `mergeComponentMetadata` even if we only needed one component metadata.

PR #1820, opened 2018-07-23
ENYO-5417: Changed overscroll color more recognizable on the focused element

0x64:
Overscroll effects are not recognizable by eye due to the color of the effect is as same as our focus effect color.

### Resolution
Changed the overscroll effect color to bright gray (almost white) for a dark theme and dark gray for a light theme.

### Links
ENYO-5417

### Comments

### DIFF
--- a/packages/moonstone/styles/colors-light.less
+++ b/packages/moonstone/styles/colors-light.less
@@ -224,4 +224,4 @@
 
 // Scrollable
 // ---------------------------------------
-@moon-scrollable-overscroll-color: @moon-accent;
+@moon-scrollable-overscroll-color: #454545;
--- a/packages/moonstone/styles/colors.less
+++ b/packages/moonstone/styles/colors.less
@@ -230,4 +230,4 @@
 
 // Scrollable
 // ---------------------------------------
-@moon-scrollable-overscroll-color: @moon-accent;
+@moon-scrollable-overscroll-color: #ccc;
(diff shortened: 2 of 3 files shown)


packages/moonstone/styles/colors-light.less
```diff
@@ -224,4 +224,4 @@
 
 // Scrollable
 // ---------------------------------------
-@moon-scrollable-overscroll-color: @moon-accent;
+@moon-scrollable-overscroll-color: #454545;
```
seunghoh: How about to assign variable rather than hardcoded value?

packages/moonstone/styles/colors.less
```diff
@@ -230,4 +230,4 @@
 
 // Scrollable
 // ---------------------------------------
-@moon-scrollable-overscroll-color: @moon-accent;
+@moon-scrollable-overscroll-color: #ccc;
```
seunghoh: Same as above. How about to use a variable?

packages/moonstone/styles/colors-light.less
0x64: The color is unique and it is similar to `@moon-dark-gray` but not exactly the same. So it is not a universal color but a specific color only for overscroll effect, in my opinion.

packages/moonstone/styles/colors.less
0x64: Same here as above. :) `#ccc` is used for other cases also, but it is not a universal color.

PR #1821, opened 2018-07-24
PLAT-62228: Set a 5way key mode when pressing a key to focus an item

ybsung:

If pressing a page up or down key over the item in a VirtualList with a pointer mode, VirtualList could give focus to the item to move because Spotlight.focus() don't work in a pointer mode.

### Resolution

As we did in a VirtualList with the #1778, I forced to call `Spotlight.setPointerMode(false) ` when pressing a key.

### Additional Considerations

I found we also tried to move focus between `ScrollButtons` when pressing a key. So I also fixed it for this case.

### Links

PLAT-62228

### Comments

### DIFF
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -120,6 +120,20 @@ const ButtonBase = kind({
 });
 
 /**
+ * Enforces a minimum width on the Button.
+ *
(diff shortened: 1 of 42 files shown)


ybsung:
I created this PR incorrectly. I'd like to create it again.

PR #1822, opened 2018-07-24
PLAT-62228: Set a 5way key mode when pressing a key to focus an item

ybsung:

When pressing a page up or down key over the item in a VirtualList with a pointer mode, VirtualList could not move focus properly.

### Resolution

VirtualList called Spotlight.focus(). But the VirtualList could not move focus because it was in a pointer mode.
As we did in a VirtualList with the #1778, I forced to call `Spotlight.setPointerMode(false) ` when pressing a key.

### Additional Considerations

When pressing a 4 direction key over the page control in a VirtualList with a pointer mode to move another page control, VirtualList could not move focus properly. So I fixed it in a same way.

### Links

PLAT-62228

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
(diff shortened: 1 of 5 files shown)


0x64:
LGTM

PR #1826, opened 2018-07-25
ENYO-5481: Spotlight: Blur is not occur when exit window (5 QE)

haileyR:
ENYO-5481: Spotlight: Blur is not occur when exit window (5 QE)

### Resolution
Added webOSMouseHandler so that spotlight blurs when the mouse leaves.

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
+### Added
+
+- `spotlight` webOS mouse event handler.
+
 ### Changed
 
 - `spotlight` to default to 5-way mode on initialization
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -354,6 +354,12 @@ const Spotlight = (function () {
 		_pointerMoveDuringKeyPress = false;
 	}
 
+	function webOSMouseHandler (ev) {
+		if (ev && ev.detail && ev.detail.type === 'Leave') {
+			onBlur();
+		}
+	}
+
 	function onFocus () {
 		// Normally, there isn't focus here unless the window has been blurred above. On webOS, the
 		// platform may focus the window after the app has already focused a component so we prevent
@@ -482,6 +488,7 @@ const Spotlight = (function () {
 				window.addEventListener('keyup', onKeyUp);
 				window.addEventListener('mouseover', onMouseOver);
 				window.addEventListener('mousemove', onMouseMove);
+				document.addEventListener('webOSMouse', webOSMouseHandler);
 				setLastContainer(rootContainerId);
 				configureDefaults(containerDefaults);
 				configureContainer(rootContainerId);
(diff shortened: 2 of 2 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -482,6 +488,7 @@ const Spotlight = (function () {
 				window.addEventListener('keyup', onKeyUp);
 				window.addEventListener('mouseover', onMouseOver);
 				window.addEventListener('mousemove', onMouseMove);
+				document.addEventListener('webOSMouse', webOSMouseHandler);
```
teckliew: Let's add a check for `webOS` so that we don't have to add the `webOSMouse` handler needlessly.

You can do this by importing `import platform from '@enact/core/platform'`, then check `platform.webos`, and if that's true, attach the `webOSMouse` handler.

packages/spotlight/src/spotlight.js
```diff
@@ -354,6 +354,12 @@ const Spotlight = (function () {
 		_pointerMoveDuringKeyPress = false;
 	}
 
+	function webOSMouseHandler (ev) {
```
teckliew: What do you think about renaming this `onWebOSMouse`?

packages/spotlight/src/spotlight.js
```diff
@@ -504,6 +511,7 @@ const Spotlight = (function () {
 			window.removeEventListener('keyup', onKeyUp);
 			window.removeEventListener('mouseover', onMouseOver);
 			window.removeEventListener('mousemove', onMouseMove);
+			document.removeEventListener('webOSMouse', webOSMouseHandler);
```
teckliew: Add the platform check here as well.

packages/spotlight/src/spotlight.js
webOS101: Handlers should be `handleXxxx` not `onXxxxx`.

haileyR: it's handleWebOSMouseEvent now.

packages/spotlight/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
+### Added
+
+- `spotlight` webOS mouse event handler.
```
webOS101: This should be a bug fix, not an additon.  There's no new public API.

packages/spotlight/src/spotlight.js
```diff
@@ -354,6 +355,12 @@ const Spotlight = (function () {
 		_pointerMoveDuringKeyPress = false;
 	}
 
+	function handleWebOSMouseEvent (ev) {
+		if (ev && ev.detail && ev.detail.type === 'Leave') {
```
webOS101: Do we need to detect an 'Enter' event to reset the pointer mode?  Is it possible for pointer mode to change after the 'leave' and still get an 'enter' event?  I don't know!

haileyR: no we don't need the enter event handler. We tested.

ryanjduffy:
@haileyR + @teckliew - Good work on this!

A future consideration might be to refactor the logic out of `onBlur` into a method which can be called from either handler. We run a risk by reusing event handlers that we will decide to use the event payload in the called method (`onBlur` in this case) to make some decision but that payload isn't available in the calling method (`handleWebOSMouseEvent`) so it either breaks or we have to manufacture a synthetic event to pass to the called method.

Not a critical issue so I've merged this as is.

PR #1828, opened 2018-07-26
PLAT-62487: Add inline support to kinds

ryanjduffy:
On low powered devices, the overhead of component layers add up and result in a material performance degradation.

### Resolution
While we can eliminate layers in places, it's beneficial to be able to "inline" functional components, particularly in the `ui` layer, in the themes to reduce layers. This PR adds a currently private static `inline()` method to components created with `core/kind` which can be called to inline the contents of the component into the host rather than introducing another component boundary within React.

### DIFF
--- a/packages/ui/Button/Button.js
+++ b/packages/ui/Button/Button.js
@@ -50,10 +50,10 @@ const ButtonBase = kind({
 		css: PropTypes.object,
 
 		/**
-		 * Disables the [ButtonBase]{@link ui/Button.ButtonBase}
+		 * Applies the `disabled` class.
 		 *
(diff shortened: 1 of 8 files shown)


packages/ui/Button/Button.js
```diff
@@ -156,13 +155,6 @@ const ButtonBase = kind({
 		}
 	},
 
-	handlers: {
```
webOS101: Need to update the docs for `disabled` to indicate that it's up to `Touchable` or the controlling component to prevent clicks from firing.  That for `ButtonBase` it's just styling.

PR #1833, opened 2018-07-27
ENYO-5488: ui/Image to require either src or placeholder props

germboy:
`ui/Image` cannot display placeholder images when `src` is not set.

### Resolution
The `src` prop was marked as required before the implementation of `placeholder`. Since then, we were unable to display placeholders in cases where the source is not yet known. This fix provides a fallback to `placeholder` in those cases, which aids developers when dealing with delayed images. A `warning` is applied, requiring either of the 2 props to be specified, to make up for removing `.isRequired` on the `src` prop-type.

### DIFF
--- a/packages/ui/Image/Image.js
+++ b/packages/ui/Image/Image.js
@@ -8,6 +8,8 @@
 import kind from '@enact/core/kind';
 import PropTypes from 'prop-types';
 import React from 'react';
+import warning from 'warning';
+
 import {selectSrc} from '../resolution';
 
 import componentCss from './Image.less';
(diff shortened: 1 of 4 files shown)


packages/ui/Image/Image.js
```diff
@@ -132,12 +124,22 @@ const ImageBase = kind({
...
+		 * @public
+		 */
+		src: PropTypes.oneOfType([PropTypes.string, PropTypes.object])
 	},
 
 	defaultProps: {
 		placeholder: '',
-		sizing: 'fill'
+		sizing: 'fill',
+		src: ''
```
webOS101: Do we need to specify a default here?  And, if so, should we document it in the prop?

packages/ui/Image/Image.js
```diff
@@ -132,12 +124,22 @@ const ImageBase = kind({
 		 * @default 'fill'
 		 * @public
 		 */
-		sizing: PropTypes.oneOf(['fit', 'fill', 'none'])
+		sizing: PropTypes.oneOf(['fit', 'fill', 'none']),
+
+		/**
+		 * String value or Object of values used to determine which image will appear on
+		 * a specific screenSize.
+		 *
```
webOS101: Should we update the description of the `placeholder` prop to say that it will be used if `src` is not specified?

PR #1837, opened 2018-07-30
ENYO-5493: Fix Dialog read order

Baekwoo-s:
Fixed the reading order of Dialog's prop `buttons` and `children`
- AS-IS
  - Title + TitleBelow + **Buttons + Children** + Focused button
- TO-BE
  - Title + TitleBelow + **Children + Buttons** + Focused button

### Resolution
- Set a unique id(Using IdProvider) for each props(title, titleBelow, buttons, and children)
- Set the id to aria-labelledby according to reading order

### Links
ENYO-5493

### Comments

### DIFF
--- a/packages/moonstone/Dialog/Dialog.js
+++ b/packages/moonstone/Dialog/Dialog.js
@@ -185,28 +185,35 @@ const DialogBase = kind({
 
 	computed: {
 		className: ({noDivider, styler}) => styler.append({showDivider: !noDivider}),
+		generateId: () => () => Math.random().toString(36).substr(2, 8),
 		titleBelow: ({title, titleBelow}) => title ? titleBelow : ''
 	},
 
-	render: ({buttons, casing, css, children, title, titleBelow, ...rest}) => {
+	render: ({buttons, casing, css, children, generateId, title, titleBelow, ...rest}) => {
 		delete rest.noDivider;
 
+		const
+			buttonsId = generateId(),
+			childrenId = generateId(),
+			titleBelowId = generateId(),
+			titleId = generateId();
+
 		return (
-			<Popup {...rest} css={css}>
+			<Popup {...rest} aria-labelledby={`${titleId} ${titleBelowId} ${childrenId} ${buttonsId}`} css={css}>
 				<div className={css.titleWrapper}>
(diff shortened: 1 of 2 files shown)


packages/moonstone/Dialog/Dialog.js
```diff
@@ -185,28 +185,35 @@ const DialogBase = kind({
...
+		generateId: () => () => Math.random().toString(36).substr(2, 8),
 		titleBelow: ({title, titleBelow}) => title ? titleBelow : ''
 	},
 
-	render: ({buttons, casing, css, children, title, titleBelow, ...rest}) => {
+	render: ({buttons, casing, css, children, generateId, title, titleBelow, ...rest}) => {
 		delete rest.noDivider;
 
+		const
+			buttonsId = generateId(),
```
viodragon2: Do we have to regenerate on every render? If the purpose of having an ID is to sort the order of aria-label, can't we create ids per instances?

Baekwoo-s: @viodragon2 
I modified it not to generate every render.
But there is one thing to worry about.
It creates an id outside of kind, which is common to all dialogs.
If two or more dialogs appear on one screen, it may cause unexpected behavior.j

viodragon2: This would be module-scoped and not instance-scoped. I suggest using `moonstone/internal/IdProvider` to resolve this. It's a HOC responsible for creating a unique id per component. You may extend the id to subcomponents by giving suffixes as `${id}_body`, `${id}_buttons`, etc.

Baekwoo-s: Thanks for your comment.
I updated using the IdProvider.

packages/moonstone/Dialog/Dialog.js
```diff
@@ -78,6 +79,14 @@ const DialogBase = kind({
 		 */
 		css: PropTypes.object,
 
+		/**
+		 * The dialog id reference for setting aria-labelledby.
```
ybsung: How about `referenced` instead of `reference`?

ybsung:
LGTM

PR #1841, opened 2018-07-31
ENYO-5511: Add VideoPlayer prop to disable play/pause button

viodragon2:
### Checklist

- [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
- [x] I have included the [Enact DCO](http://enactjs.com/docs/developer-guide/contributing/dco/) sign-off on each commit or in this Pull Request
- [x] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
- [ ] At least one test case is included for this feature or bug fix
- [x] Documentation was added or is not needed

- [ ] This is an API breaking change

There's no way to only disable play/pause button. Added `playPauseButtonDisabled` property in `MediaControlsBase`.

Note that if play/pause button is disabled, spotlight needs to fallback to focus on other spottable item in playback controls. Changed the outer wrapper of `MediaControls` to `SpotlightContainer` with the default spotlight behavior configured to the spec.

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -277,7 +277,7 @@ const MediaControlsBase = kind({
 		playIcon: PropTypes.string,
 
 		/**
-		 * Disables state on the media "play"/"pause" button
+		 * Disables the media "play"/"pause" button.
 		 *
 		 * @type {Boolean}
 		 * @public
@@ -317,7 +317,7 @@ const MediaControlsBase = kind({
 		spotlightDisabled: PropTypes.bool,
 
 		/**
-		 * A custom media controls container ID to use with Spotlight.
+		 * The spotlight ID for the media controls container.
 		 *
 		 * @type {String}
 		 * @public
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -584,6 +584,14 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			paused: PropTypes.bool,
 
 			/**
+			 * Disables state on the media "play"/"pause" button
+			 *
+			 * @type {Boolean}
+			 * @public
+			 */
+			playPauseButtonDisabled: PropTypes.bool,
+
+			/**
 			 * Disables the media playback-rate control buttons; the inner pair.
 			 *
 			 * @type {Boolean}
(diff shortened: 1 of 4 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -22,6 +22,15 @@ import {countReactChildren} from './util';
 
 import css from './VideoPlayer.less';
 
+const OuterContainer = SpotlightContainerDecorator({
+	enterTo: 'default-element',
+	defaultElement: [
+		`.${css.leftComponents} .${spotlightDefaultClass}`,
+		`.${css.rightComponents} .${spotlightDefaultClass}`,
+		`.${spotlightDefaultClass}`,
+		`.${css.mediaControls} > .spottable`
```
germboy: Instead of hard-coding the `spottable` class name, I'd suggest importing it from `spotlight/Spottable`. Better yet, you could update this line to be `.${css.mediaControls} *` since we know each contained control is a spottable component.

viodragon2:
I believe the symptom you're seeing is because the media playback controls (e.g. rew/ff/play/pause/etc.) are disabled from when video fires `loadstart` event until there's enough data to be played. So if the initial video loading time is longer and rendering of media controls happens before, then it will try to spot whatever is available in `MediaControls`, which in this case `leftComponents`. I think that's okay.

germboy:
Yes, that sounds right. I agree that focusing on the 1st available control when all default directives have failed should be fine. Plus with this simplified "default selection" logic, it'd be much easier to dictate where to focus in this case, if needed in the future. 👍

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -267,6 +276,14 @@ const MediaControlsBase = kind({
 		 */
 		playIcon: PropTypes.string,
 
+		/**
+		 * Disables the media "play"/"pause" button.
+		 *
+		 * @type {Boolean}
+		 * @public
+		 */
+		playPauseButtonDisabled: PropTypes.bool,
```
ryanjduffy: Since this disables handling the key event for play/pause as well, what do you think of renaming to `playPauseDisabled`?

ryanjduffy:
Also, let's add a knob for this in the sampler for testability.

cc: @LucieRoy - we should add test cases to validate that when this prop is set, that play/pause is disabled for both the button and the remote control button press.

packages/moonstone/VideoPlayer/MediaControls.js
ryanjduffy: Nah, we can't really do that at this point, sadly. We could deprecate them but let's stick with what you have. Thanks for pointing that out!

viodragon2:
Added knob to the sampler

PR #1842, opened 2018-07-31
ENYO-5480: Prevent unexpected scrolling when updated by app

ybsung:

`onUpdate` prop in moonstone/Scroller.componentDidUpdate() is designed to be called after ui/Scrollable.enqueueForceUpdate() is called. But `this.setState()` from app can also execute this function and lead Scroller to scroll unexpectedly.

### Resolution

I prevent calling `onUpdate` in `ui/Scrollable.componentDidUpdate()` by adding the `shouldUpdateScroller` flag. The `shouldUpdateScroller` flag could be set to `true` when `ui/Scrollable.enqueueForceUpdate()` is called.

### Additional Considerations

If `onUpdate` is called from Scroller, the `onUpdate` could be called twice.
The first time when calling `this.forceUpdate()` in `ui/Scrollable.enqueueForceUpdate()`.
The second time when calling `this.setState()` in `ui/Scrollable.componentDidUpdate()` to update scrollbar visibility.

So I moved `onUpdate()` from moonstone/Scroller to ui/Scrollable and called it depending on the visibility change of the scrollbar

In addition, I removed the Scroller test case related with `onUpdate`.

### Links

ENYO-5480

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller` to prevent scrolling when hovering the truncated spottable item
+- `moonstone/Scroller` to prevent unexpected scrolling when updated by app
 - `moonstone/Scroller` to go to next page properly via page up/down keys
 
 ## [2.0.0] - 2018-07-30
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller` to prevent scrolling based on focused item when hovering the item
+- `moonstone/Scroller` to prevent scrolling hovering the truncated spottable item
 
 ## [2.0.0] - 2018-07-30
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,11 +2,16 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
-## [2.0.1] - 2018-08-01
+## [unreleased]
 
 ### Fixed
 
 - `moonstone/Scroller` to prevent unexpected scrolling when updated by app
+
+## [2.0.1] - 2018-08-01
+
+### Fixed
+
 - `moonstone/Dialog` read order of dialog contents
 - `moonstone/Scroller` to go to next page properly via page up/down keys
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller` to prevent scrolling hovering the truncated spottable item
+- `moonstone/Scroller` to prevent scrolling when hovering the truncated spottable item
 
 ## [2.0.0] - 2018-07-30
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/Scroller` to prevent scrolling based on focused item when hovering the item
+
 ## [2.0.0] - 2018-07-30
 
 ### Added
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -311,7 +311,6 @@ class ScrollableBase extends Component {
 
 			if (pos && (pos.left !== this.uiRef.scrollLeft || pos.top !== this.uiRef.scrollTop)) {
 				this.startScrollOnFocus(pos);
-				this.uiRef.isInvalidated = false;
 			}
 
 			// update `scrollHeight`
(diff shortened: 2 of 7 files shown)


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -548,8 +548,9 @@ class ScrollableBase extends Component {
 
 	// Callback for scroller updates; calculate and, if needed, scroll to new position based on focused item.
 	handleScrollerUpdate = () => {
-		if (this.uiRef.scrollToInfo === null && Spotlight.getPointerMode()) {
+		if (this.uiRef.scrollOnFocusInPointerMode && this.uiRef.scrollToInfo === null && Spotlight.getPointerMode()) {
```
MikyungKim: Can't we move these conditions to ui/Scrollable?

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -548,8 +548,9 @@ class ScrollableBase extends Component {
...
 			this.calculateAndScrollTo();
+			this.uiRef.scrollOnFocusInPointerMode = false;
```
MikyungKim: This one also.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -259,6 +259,14 @@ class ScrollableBase extends Component {
 		 */
 		onScrollStop: PropTypes.func,
 
+		/**
+		 * Called when invalidating [Scroller]{@link moonstone/Scroller.Scroller}'s bounds
```
MikyungKim: Shouldn't it be ui/Scroller?

packages/ui/Scrollable/Scrollable.js
```diff
@@ -406,8 +415,8 @@ class ScrollableBase extends Component {
 		) {
 			this.deferScrollTo = false;
 			this.isUpdatedScrollThumb = this.updateScrollThumbSize();
-		} else {
-			this.updateScrollbars();
+		} else if (!this.updateScrollbars() && onUpdate && !this.childRef.hasOwnProperty('hasDataSizeChanged')) {
```
MikyungKim: Let's refactor this part based on more trimmed logic.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -446,6 +455,7 @@ class ScrollableBase extends Component {
 	// TODO: consider replacing forceUpdate() by storing bounds in state rather than a non-
 	// state member.
 	enqueueForceUpdate = () => {
+		this.scrollOnFocusInPointerMode = true;
```
MikyungKim: Let's rename it to... `shouldUpdateScroller`?

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -252,6 +252,14 @@ class ScrollableBaseNative extends Component {
 		 */
 		onScrollStop: PropTypes.func,
 
+		/**
+		 * Called when invalidating [Scroller]{@link ui/Scroller.Scroller}'s bounds
```
MikyungKim: ScrollerNative?

packages/moonstone/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Scroller` to prevent scrolling when hovering the truncated spottable item
```
MikyungKim: please update the log

packages/ui/Scrollable/Scrollable.js
```diff
@@ -404,10 +414,15 @@ class ScrollableBase extends Component {
 			hasDataSizeChanged === false &&
 			(isHorizontalScrollbarVisible && !prevState.isHorizontalScrollbarVisible || isVerticalScrollbarVisible && !prevState.isVerticalScrollbarVisible)
 		) {
+			// For VirtualList
 			this.deferScrollTo = false;
 			this.isUpdatedScrollThumb = this.updateScrollThumbSize();
-		} else {
-			this.updateScrollbars();
+		} else if (!this.updateScrollbars()) {
+			if (this.shouldUpdateScroller === true && !this.childRef.hasOwnProperty('hasDataSizeChanged')) {
```
MikyungKim: let's remove `=== true`

MikyungKim:
Looks okay. I think we need to revisit this later after refactoring `enqueueForceUpdate` and logic in `componentDidUpdate`.

teckliew:
:lgtm:

ryanjduffy:
I'd like to better understand the necessity of this change before moving ahead. I've done some analysis (which I've included below) but recognize that the others on this PR have a deeper knowledge of this codebase than I

### Analysis
I noticed a disconnect between the original requirement for adding `onUpdate` and the way in which it was implemented. The requirement surfaced due to a need to scroll a Component (in that case, the entire ExpandableList) into view when that component caused a change in scroll bounds. It appears in the original implementation that we asserted that nearly any update of the component should cause this behavior but discovered that updates introduced upstream (as was described in this ticket) would inadvertently cause scrolling.

The gap in my view is that we didn't verify that the scroll bounds changed before invoking this logic. I adjusted the callback with the following and it seems to resolve the issue but I'm not sure it's a complete or appropriate solution.

```
	// Callback for scroller updates; calculate and, if needed, scroll to new position based on focused item.
	handleScrollerUpdate = () => {
		if (this.uiRef.scrollToInfo === null && Spotlight.getPointerMode()) {
			const scrollHeight = this.uiRef.getScrollBounds().scrollHeight;
			if (scrollHeight !== this.uiRef.bounds.scrollHeight) {
				this.calculateAndScrollTo();
			}
		}
	}
```

ryanjduffy:
Closing in favor of #1873

PR #1847, opened 2018-07-31
ENYO-5502: Fix Image to hide broken image icon when image loading fails

viodragon2:
### Checklist

Broken image icon appears even if height and width are set to 0 in webOS.

### Resolution
Set `ui/Image`'s `` visibility hidden

### Additional Considerations
Setting `display: none` does the same trick but it changes the vertical alignment.

See https://codepen.io/stephenhchoi/pen/wxXjOq?editors=1100 for the sample. The red box mimics the `Button` and blue mimics `Image`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [2.0.1] - 2018-08-01
+
+### Fixed
+
+- `moonstone/Dialog` read order of dialog contents
+- `moonstone/Scroller` to go to next page properly via page up/down keys
+- `spotlight` to not blur when pointer leaves floating webOS app while paused
+
 ## [2.0.0] - 2018-07-30
 
 ### Removed
--- a/lerna.json
+++ b/lerna.json
@@ -1,6 +1,6 @@
 {
   "lerna": "2.8.0",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "command": {
     "bootstrap": {
       "npmClientArgs": [
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "enact",
-  "version": "2.0.0",
+  "version": "2.0.1",
   "description": "Monorepo for all Enact front end libraries.",
   "private": true,
   "scripts": {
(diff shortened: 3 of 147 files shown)


ryanjduffy:
I'm curious why we're "hiding" the `` using zero width/height vs a more explicit means (display/visibility/opacity). As far as I can tell, there's no case in which we need it to be visible and Chrome (latest, at least) doesn't appear to report `error` or `load` differently for an `` with `display: none`.

https://codepen.io/anon/pen/rrKvyK?editors=0010

viodragon2:
The broken image icon only appears in webOS TV even with width/height 0. It doesn't reproduce the same issue in Chrome.

ryanjduffy:
I understand but my question is still if we can fully hide the `` all the time and thereby avoid that behavior.

viodragon2:
Thanks for the input.  Updated to set `visibility: hidden` at all times for ``. I'm not sure why `display: none` changes the layout when aligned with other `inline-block` elements. So I decided to go with `visibility: hidden` for now. If anybody can explain what causes the misalignment, I'd appreciate it.

Djspaceg:
Buckle up, here's why the alignment changes. _(Remember, you asked for this.)_

Inline-blocks work like text. By default, they align on their baseline. When a div has no content, its baseline is the same as the "bottom". By setting `display: none` you're telling the `div` that it no longer has any contents, and it should align according to its own dimensions (ignoring children). When you set the size of the `img` inside the top set of divs, the blue box still has content, and that content (the 0x0 `img` also has a baseline – 1em). It's actually not in the upper left corner but about 1em down from the top left because it is vertically aligned with the baseline of the "text" it is pretending to be (since images are inline-block). This position inside the blue box is then middle aligned with the red box, because the red box was told to be middle aligned. The vertical-middle of the red box is where following siblings will anchor to, and in this case, it's the baseline of the child of the blue box, which is 1em down from the top left corner of the blue box. Meanwhile, back to the bottom pair, the blue box has no children (because of `display: none`) so it is anchoring its bottom (baseline) to the vertical middle of the red box.

The attached image shows a little orange square. This is the 0x0 image. Note it aligns perfectly with the middle of the red box, and 1em down from the top of the blue box.
Box alignment issue

Djspaceg:
My opinion is that `display: none`, if it doesn't cause any loading or event-firing issues, is the way to go. The blue box will behave more intuitively with no (text/inline) content inside.

PR #1852, opened 2018-08-01
Feature/enyo 5512 hailey

haileyR:
ProgressTooltip: there's gap between body and arrow (2 QE)

### Resolution
Move tooltipArrow by -1px to avoid having a gap.


Djspaceg:
I can't reproduce this. Can you verify that the `1px` rule works proportionally on HD (720) FHD (1080) and 4k?

haileyR:
This solution doesn't work. I will close this.

PR #1853, opened 2018-08-02
ENYO-5512: Remove space between tooltipLabel and tooltipArrow

Sangwook-Lee:
When `tooltip` is displayed with `ProgressBar` or `Slider`, There's a gap between `tooltipLabel` and `tooltipArrow`. (about 1 to 2 px)
To reproduce, progress value should be more than 50%. And problem arises with the `leftArrow` type.
So, we have to remove space between tooltipLabel and tooltipArrow.

### Resolution
Currently, arrow svg moves to the upper right corner or the bottom right by `translate()` function. But, as the width of the TooltipLabel changes in decimal units, a problem sometimes arises.
If the width value is an integer, there is no problem. But it is impossible to do so because the width varies according to the length of the text.

### Links
ENYO-5512

### Comments
Enact-DCO-1.0-Signed-off-by: Sangwook Lee (sangwook1203.lee@lge.com)

### DIFF
--- a/packages/moonstone/TooltipDecorator/Tooltip.less
+++ b/packages/moonstone/TooltipDecorator/Tooltip.less
@@ -68,7 +68,7 @@
 	&.above {
 		&.rightArrow {
 			.tooltipArrow {
-				transform: translateY(100%) translateY(-1px) rotate(-90deg) scaleX(-1);
+				transform: translateY(99%) rotate(-90deg) scaleX(-1);
 				left: 0;
 				bottom: 0;
 			}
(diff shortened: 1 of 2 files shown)


packages/moonstone/TooltipDecorator/Tooltip.less
```diff
@@ -78,9 +80,10 @@
 
 		&.leftArrow {
 			.tooltipArrow {
-				transform: translate(100%, 99%) rotate(90deg);
-				bottom: 0;
-				right: 0;
+				transform-origin: 100% 100%;
+				transform: rotate(270deg) rotateX(180deg) rotateY(180deg);
```
Djspaceg: This is a really convoluted set of transforms. :(

Djspaceg:
After some deeper testing, I found that this approach does not resolve the issue. The tail is offset by more pixels in some cases, rather than fewer. We'll need a different approach.

PR #1854, opened 2018-08-02
ENYO-5435: Updated `moonstone` lists and a scroller not to be scrolled by drag

0x64:
### Checklist

We've got a request for `moonstone` not to scroll lists or a scroller by drag.

### Resolution
Simply changed not to use `Touchable('div')` for `moonstone`. We are still supporting scroll by drag in `ui`.

### Additional Considerations

### Links
ENYO-5435

### DIFF
--- a/packages/moonstone/Scrollable/ScrollableNative.js
+++ b/packages/moonstone/Scrollable/ScrollableNative.js
@@ -9,7 +9,6 @@ import React, {Component} from 'react';
 import ReactDOM from 'react-dom';
 import Spotlight from '@enact/spotlight';
 import SpotlightContainerDecorator from '@enact/spotlight/SpotlightContainerDecorator';
-import Touchable from '@enact/ui/Touchable';
 
 import $L from '../internal/$L';
 
@@ -38,8 +37,6 @@ const
 		right: 'left',
 		up: 'down'
 	};
-
-const TouchableDiv = Touchable('div');
 
 const navigableFilter = (elem) => {
 	if (
@@ -805,6 +802,7 @@ class ScrollableBaseNative extends Component {
 				addEventListeners={this.addEventListeners}
 				applyOverscrollEffect={this.applyOverscrollEffect}
 				clearOverscrollEffect={this.clearOverscrollEffect}
+				noScrollByDrag
 				onFlick={this.onFlick}
 				onKeyDown={this.onKeyDown}
 				onMouseDown={this.onMouseDown}
@@ -818,6 +816,8 @@ class ScrollableBaseNative extends Component {
 					childComponentProps,
 					className,
 					componentCss,
+					contentComponent: ContentComponent,
+					contentComponentProps: {className: contentClassName, ...restContentProps},
 					horizontalScrollbarProps,
 					initChildRef: initUiChildRef,
 					initContainerRef: initUiContainerRef,
@@ -826,7 +826,6 @@ class ScrollableBaseNative extends Component {
 					rtl,
 					scrollTo,
 					style,
-					touchableProps: {className: touchableClassName, ...restTouchableProps},
 					verticalScrollbarProps
 				}) => (
 					<div
@@ -838,7 +837,7 @@ class ScrollableBaseNative extends Component {
 						style={style}
 					>
 						<div className={classNames(componentCss.container, overscrollCss.overscrollFrame, overscrollCss.vertical, isHorizontalScrollbarVisible ? overscrollCss.horizontalScrollbarVisible : null)} ref={this.initVerticalOverscrollRef}>
-							<TouchableDiv className={classNames(touchableClassName, overscrollCss.overscrollFrame, overscrollCss.horizontal)} ref={this.initHorizontalOverscrollRef} {...restTouchableProps}>
+							<ContentComponent className={classNames(contentClassName, overscrollCss.overscrollFrame, overscrollCss.horizontal)} ref={this.initHorizontalOverscrollRef} {...restContentProps}>
 								{childRenderer({
 									...childComponentProps,
 									cbScrollTo: scrollTo,
@@ -850,7 +849,7 @@ class ScrollableBaseNative extends Component {
 									rtl,
 									spotlightId
 								})}
-							</TouchableDiv>
+							</ContentComponent>
 							{isVerticalScrollbarVisible ?
 								<Scrollbar
 									{...verticalScrollbarProps}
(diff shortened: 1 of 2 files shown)


0x64:
I will update CHANGELOG after the first review. ;)

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -764,7 +763,7 @@ class ScrollableBase extends Component {
 						style={style}
 					>
 						<div className={classNames(componentCss.container, overscrollCss.overscrollFrame, overscrollCss.vertical, isHorizontalScrollbarVisible ? overscrollCss.horizontalScrollbarVisible : null)} ref={this.initVerticalOverscrollRef}>
-							<TouchableDiv className={classNames(touchableClassName, overscrollCss.overscrollFrame, overscrollCss.horizontal)} ref={this.initHorizontalOverscrollRef} {...restTouchableProps}>
+							<ContentComponent className={classNames(contentClassName, overscrollCss.overscrollFrame, overscrollCss.horizontal)} ref={this.initHorizontalOverscrollRef} {...restContentProps}>
```
ybsung: How about `ChildContainer` instead of `ContentComponent`?

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -732,6 +729,7 @@ class ScrollableBase extends Component {
 				addEventListeners={this.addEventListeners}
 				applyOverscrollEffect={this.applyOverscrollEffect}
 				clearOverscrollEffect={this.clearOverscrollEffect}
+				noScrollByDrag
 				onFlick={this.onFlick}
```
ybsung: Do we still need `onFlick` even though we don't support dragging and flicking?

packages/moonstone/Scrollable/Scrollable.js
0x64: Hmm. Similar to me. Is that better?

0x64: The term `container` is used already for the top DOM element. Finding another name...

0x64:
I've added `noScrollByDrag` prop to `ui` lists and a scroller. (Technically added to `ui/Scrollable`). So, by default, dragging is supported.
And, regarding future signage products, I want to add the prop also into `moonstone` components.

@webOS101 , @ryanjduffy 
I'm wondering whether we should disable dragging by default in moonstone.
If we should do, the prop name will be `scrollByDrag` not `noScrollByDrag` to make its default value to _false_. (Inconsistent prop names between ui and moonstone)
If not, we may consider enabling drag by default and _add `noScrollByDrag` to existing apps_.
I gathered opinions here and there is no clear conclusion.
Could you please let me know the proper way?

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -130,6 +127,15 @@ class ScrollableBaseNative extends Component {
 		 */
 		focusableScrollbar: PropTypes.bool,
 
+		/**
+		 * Prevents scroll by dragging or flicking on the list or the scroller.
```
ybsung: How about `scrolling`?

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -130,6 +127,15 @@ class ScrollableBaseNative extends Component {
...
+		 *
+		 * @type {Boolean}
+		 * @default false
```
ybsung: The default value for `Moonstone` is not `false` but `true`.

packages/moonstone/Scrollable/ScrollableNative.js
0x64: `scroll` can be used as a noun and pretty clear here what it means, IMO.

packages/moonstone/Scrollable/ScrollableNative.js
0x64: It is the origin of my question above. Please see my comment.

ybsung:
LGTM now. But depending on the decision for the comment - https://github.com/enactjs/enact/pull/1854#issuecomment-410174535, please let me know if needed to update the PR. I'll review it again.

0x64:
@ybsung Thanks. I will fix the doc issue as soon as the way decided.

0x64:
I will change the prop name of `moonstone/Scrollable`.

0x64:
@ybsung I updated the PR.

ybsung:
LGTM, but please update CHANGELOG.md and create a TC if needed.

0x64:
@ybsung Sure!

packages/moonstone/Scrollable/ScrollableNative.js
0x64: I updated the name and its default value is now `false`.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -1328,17 +1344,18 @@ class ScrollableNative extends Component {
 				{...rest}
 				containerRenderer={({ // eslint-disable-line react/jsx-no-bind
 					childComponentProps,
+					childWrapper : ChildWrapper,
```
MikyungKim: Remove the space between `r` and `:`.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1283,6 +1300,8 @@ class Scrollable extends Component {
 				{...rest}
 				containerRenderer={({ // eslint-disable-line react/jsx-no-bind
 					childComponentProps,
+					childWrapper : ChildWrapper,
```
MikyungKim: Remove the space between `r` and `:`.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -732,7 +740,8 @@ class ScrollableBase extends Component {
 				addEventListeners={this.addEventListeners}
 				applyOverscrollEffect={this.applyOverscrollEffect}
 				clearOverscrollEffect={this.clearOverscrollEffect}
-				onFlick={this.onFlick}
+				noScrollByDrag={!scrollByDrag}
```
MikyungKim: Do we really need this....? Why don't we have the same name for this prop?

MikyungKim:
Please resolve the merge conflict.

0x64:
@ryanjduffy `touch` looks good but I have a concern that we don't block actually touch input for paging control buttons and items. If an app is running on a touchable device, a user can still press paging control buttons with his/her fingers. For app devs, `touch` may make some misunderstandings.

ryanjduffy:
That's a fair point! Thanks for bringing it up.

0x64:
Note: @MikyungKim and I decided to remove `moonstone`'s prop. A new prop of `moonstone` may give confusing to app developers. App developers can use `ui`'s prop without adding a new one to `moostone`.

I will update the PR.

0x64:
I did not resolve a conflict on CHANGELOG.md. It should be resolved right before merging.

PR #1856, opened 2018-08-03
ENYO-5518: Fix year picker when min year and max year don't exist

viodragon2:
### Checklist

If `minYear` or `maxYear` were provided, then `DatePicker` disables the year picker. Recent change expects `minYear` and `maxYear` prop will be transformed into a local calendar year without a default value provided.

### Resolution
Added `yearFormatter` prop to `DateTimePickerBase` which will transform year value to local calendar year if it exists. 

### Additional Considerations
Adding defaultProps to `DateTimeDecorator` would've been an easier way to fix it, but `minYear` and `maxYear` is irrelevant to `TimePicker` so it's not the right place to handle this.

Would like to add a unit test case, but wasn't sure what the right test would be. Any thoughts?

### DIFF
--- a/packages/moonstone/DatePicker/DatePicker.js
+++ b/packages/moonstone/DatePicker/DatePicker.js
@@ -22,7 +22,7 @@ import Skinnable from '../Skinnable';
 import DatePickerBase from './DatePickerBase';
 
 const dateTimeConfig = {
-	customProps: function (i18n, value, props) {
+	customProps: function (i18n, value) {
 		const values = {
 			maxMonths: 12,
 			maxDays: 31,
@@ -37,8 +37,7 @@ const dateTimeConfig = {
 			values.day = value.getDays();
 			values.maxMonths = i18n.formatter.cal.getNumMonths(values.year);
 			values.maxDays = i18n.formatter.cal.getMonLength(values.month, values.year);
-			values.maxYear = i18n.toLocalYear(props.maxYear);
-			values.minYear = i18n.toLocalYear(props.minYear);
+			values.yearFormatter = i18n.toLocalYear;
 		}
 
 		return values;
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
+### Added
+
+- `moonstone/DatePicker` property `yearFormatter` to accept a function to handle different types of calendar years
+
 ### Fixed
 
 - `moonstone/Slider` to forward `onActivate` event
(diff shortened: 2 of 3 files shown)


packages/moonstone/DatePicker/DatePicker.js
```diff
@@ -37,8 +37,7 @@ const dateTimeConfig = {
 			values.day = value.getDays();
 			values.maxMonths = i18n.formatter.cal.getNumMonths(values.year);
 			values.maxDays = i18n.formatter.cal.getMonLength(values.month, values.year);
-			values.maxYear = i18n.toLocalYear(props.maxYear);
```
ryanjduffy: What do you think of plucking `defaultProps` from `DatePickerBase` here?

```
values.maxYear = i18n.toLocalYear(props.maxYear || DatePickerBase.defaultProps.maxYear);
values.minYear = i18n.toLocalYear(props.minYear || DatePickerBase.defaultProps.minYear);
```

I'm trying to decide which is the right overlap of concerns. Does it make more sense to:
* inject the default values (either statically or via ref to `defaultProps`) in this config object
* delegate i18n formatting down to the component.

Injection duplicates default prop handling from `DatePickerBase` in the common case of i18n support but leaves it with the base component in prerendering. Using `defaultProps` mitigates that overlap a bit by avoiding using different values.

Delegation shifts _one_ piece of i18n formatting to the base component while keeping the rest in `DateTimeDecorator`. I worry that mixed responsibilities creates unclear APIs and can lead to bugs with multiple layers implementing different parts of the same capability. It does, however, keep prop management more clean between the layers.

viodragon2: I think your first suggestion is okay in a way that it simplifies the concern. The only reason I didn't follow this way was that I imagine `DateTimeDecorator` (and most of our `*able` HOCs) to be a "behavior-setter" which does not associate with default values.

That being said, I also agree that the latter approach is mixing responsibilities. `DateTimeDecorator` handles i18n adjustments, and `DateTimePickerBase` is a simple render component. And it seems like a "formatter"  lies in between. I'd thought making it an opt-in function would mitigate somewhat, but I don't think that's true.

ryanjduffy: I tend to agree with your definition of our `*able` HOCs. In other cases (`disabled` being an example), we have set a default for props that are required to be valued by the HOC in order for it to operate. `DateTimeDecorator` is an anomaly in its design in that it supports author-land functions that operate on its props. Because of that, I think it's okay to do what is proposed here.

Another option that isn't feasible right now would be to support hoisting statics (`defaultProps` in this case) up the HOC stack to ensure that the right defaults are set on the generated component. That could introduce other issues if we start to assume values have been set due to hoisting rather than explicitly handling them.

PR #1857, opened 2018-08-03
PLAT-63942: ScrollerNative: items are rendered slowly in TV

yeram:
### Checklist

Optimize rendering performance for `Scrollable` and `Marquee`

### Resolution
`position: relative` makes bad rendering performance during scrolling. To fix it, adding above style when it's really needed
And use GPU acceleration on rendering children of `Scroller`.

### Additional Considerations

### Links
PLAT-63942

### Comments

Enact-DCO-1.0-Signed-off-by: Yeram Choi <yeram.choi@lge.com>

### DIFF
--- a/packages/ui/Marquee/Marquee.less
+++ b/packages/ui/Marquee/Marquee.less
@@ -21,13 +21,13 @@
 		overflow: hidden;
 		width: 100%;
 		white-space: pre !important;
-		position: relative;
-		will-change: left, right;
 		transition-property: left, right;
 		transition-timing-function: linear;
 
 		&.animate {
 			overflow: visible;
+			position: relative;
+			transform: translateZ(0);
 		}
 	}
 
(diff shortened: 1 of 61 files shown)


packages/ui/Scrollable/Scrollable.less
```diff
@@ -3,6 +3,7 @@
 	position: relative;
 	width: 100%;
 	height: 100%;
+	transform: translateZ(0);
```
Djspaceg: Why doesn't this have `will-change: transform;` on it instead?

packages/ui/Marquee/Marquee.less
```diff
@@ -21,13 +21,12 @@
 		overflow: hidden;
 		width: 100%;
 		white-space: pre !important;
-		position: relative;
-		will-change: left, right;
```
Djspaceg: Do the left and right properties no longer change?

PR #1858, opened 2018-08-03
ENYO-5513: A11y: Not read out infoComponent when selecting more button on VideoPlayer

Baekwoo-s:
### Checklist

- Fixed not read out infoComponent when select more button on VideoPlayer.
- This issue also includes the following issues:
  - Fixed an issue where InfoComponent was not read when Blue key was pressed while MediaControls panel was floating
  - Fixed an issue where infoComponent was not read when moreButton was pressed with 5way key
  - Fixed an issue that gives focus when the role, aria-* property are not applying yet in the real DOM
  - Fixed conflicts when `title` and `info` roles are region role

### Resolution
- Issues that did not read infoComponent
  - The behavior was different when handling more button and blue key in the scenario of reading infoComponent. If we press the more button, the toggleMoreComponents function changes the state of the MoreComponents and gives focus, but the Blue key gives focus before changing the state. I unified this behavior so that the state changes first and gives focus.
- `Title` and `info` role conflicts with region role
  - Since `title` and `info` give focus to the same children(buttons in MediaControls), they will not be read if they contain the same region role. It is OK to remove the button focus once. However, I applied the alert role because the alert role is applied to the first focus in the VideoPlayer scenario, so there is no change in the region role and the reading order.
- Issues that role and aria-* property did not apply to real DOM
  - Added an `isRenderedAriaProps` variable to VideoPlayer.js and added prop to MediaControls.js
  - The render cyle of VideoPlayer and MediaControls are handled at the same time, but the role and aria- * properties have not yet been applied to real DOM at the first render time. This caused an issue because the `this.state.announce` was changed so that the objects of controlsAriaProps were immediately applied at the time of rendering. 
  - I have modified MediacControls to pass `this.isRenderedAriaProps` is true after setState which changes` this.state.announce` is rendered.

### Links
ENYO-5513

### Comments

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -443,7 +443,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			initialJumpDelay: PropTypes.number,
 
 			/**
-			 * The status of infoComponent's id is set to aria-labelledby and is ready to be read.
+			 * The status of video player when aria-labelledby and region are rendered.
 			 *
 			 * @type {Boolean}
 			 * @public
@@ -633,7 +633,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			this.paused = new Pause('VideoPlayer');
 
 			this.state = {
-				isFirstReadInfoComponents: true,
+				hasReadInfoComponents: false,
 				showMoreComponents: false
 			};
 
@@ -680,7 +680,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 				forwardToggleMore({showMoreComponents: this.state.showMoreComponents}, this.props);
 			}
 
-			if (this.state.showMoreComponents && this.props.isRenderedAriaProps && this.state.isFirstReadInfoComponents) {
+			if (this.state.showMoreComponents && this.props.isRenderedAriaProps && !this.state.hasReadInfoComponents) {
 				// Readout the infoComponents for the first time.
 				this.focusMoreButton();
 			}
@@ -815,7 +815,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 		}
 
 		toggleMoreComponents () {
-			if (!this.state.isFirstReadInfoComponents) {
+			if (this.state.hasReadInfoComponents) {
 				this.focusMoreButton();
 			}
 
@@ -832,17 +832,18 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 				return;
 			}
 
-			// Readout 'more' or 'back' button explicitly.
-			let selectedButton = Spotlight.getCurrent();
+			const selectedButton = Spotlight.getCurrent();
 			if (selectedButton === this.mediaControlsNode.querySelector(`.${css.moreButton}`)) {
+				// Press enter of 5way key
 				selectedButton.blur();
 				selectedButton.focus();
 			} else {
+				// Press color key
 				Spotlight.focus(this.props.moreButtonSpotlightId);
 			}
 
 			this.setState({
-				isFirstReadInfoComponents: false
+				hasReadInfoComponents: true
 			});
 		}
 
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -851,7 +851,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			if (selectedButton === this.moreButtonNode) {
 				// Press enter of 5way key
 				selectedButton.blur();
-				selectedButton.focus();
+				Spotlight.focus(selectedButton);
 			} else {
 				// Press color key
 				Spotlight.focus(this.props.moreButtonSpotlightId);
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -155,6 +155,14 @@ const MediaControlsBase = kind({
 		moreButtonLabel: PropTypes.string,
 
 		/**
+		 * The method to run when the more button rendered, giving a reference to the DOM.
+		 *
+		 * @type {Function}
+		 * @public
+		 */
+		moreButtonRef: PropTypes.func,
+
+		/**
 		 * A custom more button ID to use with Spotlight.
 		 *
 		 * @type {String}
@@ -349,6 +357,7 @@ const MediaControlsBase = kind({
 		moreButtonClassName,
 		moreButtonColor,
 		moreButtonDisabled,
+		moreButtonRef,
 		moreButtonSpotlightId,
 		moreIcon,
 		moreIconLabel,
@@ -402,10 +411,11 @@ const MediaControlsBase = kind({
 							color={moreButtonColor}
 							disabled={moreButtonDisabled}
 							onClick={onMoreClick}
-							tooltipProps={{role: 'dialog'}}
-							tooltipText={moreIconLabel}
+							ref={moreButtonRef}
 							spotlightId={moreButtonSpotlightId}
 							spotlightDisabled={spotlightDisabled}
+							tooltipProps={{role: 'dialog'}}
+							tooltipText={moreIconLabel}
 						>
 							{moreIcon}
 						</MediaButton>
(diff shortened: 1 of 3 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -815,11 +815,35 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
...
+
+		}
+
+		focusMoreButton = () => {
+			if (!this.props.visible) {
+				return;
+			}
+
+			// Readout 'more' or 'back' button explicitly.
+			let selectedButton = Spotlight.getCurrent();
```
ybsung: Can we replace `let` with `const`?

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -815,11 +815,35 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
...
+			if (selectedButton === this.mediaControlsNode.querySelector(`.${css.moreButton}`)) {
+				selectedButton.blur();
+				selectedButton.focus();
```
ybsung: Why don't you use Spotlight.blur() and Spotlight.focus() here?

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -442,6 +442,14 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			 */
 			initialJumpDelay: PropTypes.number,
 
+			/**
+			 * The status of infoComponent's id is set to aria-labelledby and is ready to be read.
+			 *
+			 * @type {Boolean}
+			 * @public
+			 */
+			isRenderedAriaProps: PropTypes.bool,
```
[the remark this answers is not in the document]
Baekwoo-s: Yes, Since this prop is a variable holding the initial value in the parent VideoPlayer, I don't think the value of defalutProps is needed.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1695,7 +1700,8 @@ const VideoPlayerBase = class extends React.Component {
 	getControlsAriaProps () {
 		if (this.state.announce === AnnounceState.TITLE) {
 			return {
-				role: 'region',
+				role: 'alert',
+				'aria-live': 'off',
 				'aria-labelledby': `${this.id}_title`
```
ybsung: Please order the properties in the object above.

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -625,6 +633,7 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			this.paused = new Pause('VideoPlayer');
 
 			this.state = {
+				isFirstReadInfoComponents: true,
```
ybsung: How about `isFirstToReadInfoComponents`?

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -815,11 +815,35 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
...
+		}
+
+		focusMoreButton = () => {
+			if (!this.props.visible) {
+				return;
+			}
+
+			// Readout 'more' or 'back' button explicitly.
+			let selectedButton = Spotlight.getCurrent();
+			if (selectedButton === this.mediaControlsNode.querySelector(`.${css.moreButton}`)) {
```
ybsung: Why did you use `querySelector` instead of `React Ref`?

Baekwoo-s: This is the code that @ryanjduffy , @viodragon2 wrote.
I will change this code to use React ref instead of querySelector. 
Please let me know if you have any other comments.

packages/moonstone/VideoPlayer/MediaControls.js
Baekwoo-s: I changed the variable name based on the discussion result.
isFirstToReadInfoComponents -> hasReadInfoComponents.

packages/moonstone/VideoPlayer/VideoPlayer.js
Baekwoo-s: Done.

packages/moonstone/VideoPlayer/MediaControls.js
Baekwoo-s: This should be Focused for a specific button.
Spotlight.getCurrent(). Focus() is correct, not Spotlight.focus().

packages/moonstone/VideoPlayer/MediaControls.js
Baekwoo-s: Done.

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -442,6 +442,14 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			 */
 			initialJumpDelay: PropTypes.number,
 
+			/**
+			 * The status of infoComponent's id is set to aria-labelledby and is ready to be read.
```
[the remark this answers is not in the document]
Baekwoo-s: Done.

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -802,6 +813,10 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
 			this.mediaControlsNode = ReactDOM.findDOMNode(node); // eslint-disable-line react/no-find-dom-node
 		}
 
+		getMoreButtonRef = (node) => {
+			this.moreButtonNode = ReactDOM.findDOMNode(node); // eslint-disable-line react/no-find-dom-node
```
ybsung: I am not sure but, ...do we need to use `ReactDOM.findDOMNode()` here?
And don't we need to check if node is equals to `null`?

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -154,6 +154,14 @@ const MediaControlsBase = kind({
 		 */
 		moreButtonLabel: PropTypes.string,
 
+		/**
+		 * The method to run when the more button rendered, giving a reference to the DOM.
+		 *
+		 * @type {Function}
+		 * @public
```
ybsung: Should this prop be `public`? If yes, please update CHANGELOG.md `added`.

Baekwoo-s: I changed prop to private. Thanks.

packages/moonstone/VideoPlayer/MediaControls.js
Baekwoo-s: > I am not sure but, ...do we need to use ReactDOM.findDOMNode() here?
Yes, we compare the button Node and the Spotlight.getCurrent() Node in the MediaControlsDecorator, not in the MediaControlsBase.

> And don't we need to check if node is equals to null?
I didn't see the node returning null where refs function.

ybsung: Got it.

ybsung:
LGTM

packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -815,11 +830,36 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {
...
+		focusMoreButton = () => {
+			if (!this.props.visible) {
+				return;
+			}
+
+			const selectedButton = Spotlight.getCurrent();
+			if (selectedButton === this.moreButtonNode) {
+				// Press enter of 5way key
+				selectedButton.blur();
+				Spotlight.focus(selectedButton);
```
viodragon2: any reason why we use `Spotlight.focus` instead of `selectedButton.focus()`? Isn't the purpose of this block only to reset the focus for a11y?

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1669,7 +1672,9 @@ const VideoPlayerBase = class extends React.Component {
 			infoVisible: showMoreComponents,
 			titleVisible: true,
 			announce: announce < AnnounceState.INFO ? AnnounceState.INFO : AnnounceState.DONE
-		}));
+		}), () => {
+			this.isRenderedAriaProps = true;
```
viodragon2: Using `setState` callbacks are generally discouraged, especially when you can achieve the same thing in `componentDidUpdate`. Is there any reason why we can't do this in `cDU`?

viodragon2:
closing in favor of #1874

PR #1860, opened 2018-08-03
ENYO-5523: Update pointer mode after hiding VKB

germboy:
### Checklist

Spotlight's pointer mode isn't updated when attempting to switch modes while the VKB is visible (ex: when an input has focus).

### Resolution
While the VKB is visible, spotlight or currently-focused components (ex: `Input`) are unable to receive key events (`keydown`) events. During this time a user can change from pointer to 5-way mode. This prevents spotlight from restoring focus in some cases after dismissing the VKB, since spotlight falsely thinks it is still in pointer mode - which prevents programmatic focus changes.

Since we know pointer-mode can change while the VKB is open, we can remedy this issue by setting the platform pointer mode as a response to the webOS event `keyboardStateChange` when the VKB closes.

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -361,6 +361,12 @@ const Spotlight = (function () {
 		}
 	}
 
+	function handleKeyboardStateChangeEvent ({visibility}) {
+		if (!visibility) {
+			setPlatformPointerMode();
+		}
+	}
+
 	function onFocus () {
 		// Normally, there isn't focus here unless the window has been blurred above. On webOS, the
 		// platform may focus the window after the app has already focused a component so we prevent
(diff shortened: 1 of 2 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -489,7 +495,10 @@ const Spotlight = (function () {
 				window.addEventListener('keyup', onKeyUp);
 				window.addEventListener('mouseover', onMouseOver);
 				window.addEventListener('mousemove', onMouseMove);
-				if (platform.webos) document.addEventListener('webOSMouse', handleWebOSMouseEvent);
+				if (platform.webos) {
+					document.addEventListener('webOSMouse', handleWebOSMouseEvent);
```
webOS101: This will need to attach to the top-level document for use with iframes.

germboy: of course :facepalm:

packages/spotlight/src/spotlight.js
```diff
@@ -361,6 +361,12 @@ const Spotlight = (function () {
 		}
 	}
 
+	function handleKeyboardStateChangeEvent ({visibility}) {
```
teckliew: Is `visibility` the state of VKB?

germboy: Yes. When the state of the VKB changes (shows/hides), we receive a custom `keyboardStateChange` event that contains a `visibility` property that reflects the state of the VKB.

ryanjduffy:
@teckliew Failed at first because of a CLI issue. Failed now because of a test failure surfaced because of the CLI update. Fortunately, it's fixed in `develop` so the push CI failure isn't meaningful.

PR #1862, opened 2018-08-06
PLAT-64158: Fix Popup to prevent duplicated 5-way navigation

jeonghee27:
### Checklist

`moonstone/Popup` components handle focus movement redundantly. 
In `moonstone/Popup` case with two or more buttons, focus is moved twice at one 5way-key input.

### Resolution
In `moonstone/Popup` keydown event handler, it called `Spotlight.move(dir)` without stopPropagation() . So `Spotlight` handle the key event redundantly.

`moonstone/Popup` is fixed to prevent additional key event handling after focus changed using stopPropagation() and preventDefault(). 
And added more button to verify the spotlight operation in Popup story of qa-sampler .

### Additional Considerations

Since the focus has been moved, the event could be regarded as consumed. Thus I added preventDefault() and stopPropagation(). But I don't have confidence it need preventDefault() because when I checked several additional cases(include scroller case), there was no difference in the behavior.

### Links
PLAT-64158

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Popup` to prevent moving twice at one 5-way key input if there is two or more button.
+- `moonstone/Popup` to prevent moving twice at one 5-way key input when the `Popup` has `spotlightRestrict='self-first'` option and there is two or more button
 - `moonstone/Slider` to forward `onActivate` event
 - `moonstone/GridListImageItem` to properly vertically align when the content varies in size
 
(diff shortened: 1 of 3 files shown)


0x64:
LGTM

packages/moonstone/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Popup` to prevent moving twice at one 5-way key input when the `Popup` has `spotlightRestrict='self-first'` option and there is two or more button
```
ybsung: there is two or more button
=> there are two or more buttons

ybsung:
@jeonghee27 
Please fix merge conflict.

ybsung:
LGTM

PR #1864, opened 2018-08-07
ENYO-5426: Make the images visually lined up

ryanjduffy:
Resubmitting #1810 for post 2.0.2

### DIFF
--- a/packages/ui/GridListImageItem/GridListImageItem.less
+++ b/packages/ui/GridListImageItem/GridListImageItem.less
@@ -25,4 +25,9 @@
 		pointer-events: none;
 		height: 100%;
 	}
+
+	.caption,
+	.subCaption {
+		/* Public Class Names */
+	}
 }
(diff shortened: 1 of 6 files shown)


packages/ui/GridListImageItem/GridListImageItem.less
```diff
@@ -3,9 +3,7 @@
 
 	.icon {
 		position: relative;
-		display: block;
-		margin: 0;
-		opacity: 0.5;
+		display: none;
```
ryanjduffy: Why did we change this to `display: none` when the icon is already conditionally rendered?

ryanjduffy: I guess my observation was that by setting `display: none` here, we require a theme set the `display` in their CSS for `.icon` to be usable. That doesn't seem like the right default to me.

Djspaceg: I'll take it out.

PR #1871, opened 2018-08-13
ENYO-5426: Fix the GridListImageItem's image height

ybsung:
### Checklist

- If using only a caption or a subCaption, there is the gap between its text and the GridListImageItem border.
- `.caption` and `.subCaption` css classes are not needed any more.

### Resolution

- Increase the image height if using only a caption or a subCaption.
- Remove useless css classes. 

### Additional Considerations

- Add CSS styles for subCaption that we used before. (I don't know the exact history for the styles, but we defined the styles for subCaption before and it is not applied to the subCaption now.)

### Links

ENYO-5426

### Comments

The target branch for this PR is not `develop` but `feature/ENYO-5426-same-height`. So I did not update `CHANGELOG.md` and no test case is needed.

### DIFF
--- a/packages/moonstone/GridListImageItem/GridListImageItem.less
+++ b/packages/moonstone/GridListImageItem/GridListImageItem.less
@@ -25,18 +25,14 @@
 		height: ~"calc(100% - " @moon-gridlist-item-two-captions-padding-bottom ~")";
 	}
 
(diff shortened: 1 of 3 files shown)


ybsung:
I created the new PR #1872 instead of this PR.

PR #1872, opened 2018-08-13
ENYO-5426: Fix GridListImageItem image, caption, and subcaption styles.

ybsung:
### Checklist

- `moonstone` caption and subcaption styles did not override `ui` ones.
- If using only a caption or a subCaption, there is the gap between its text and the `GridListImageItem` border.

### Resolution

- Add the comments in the `ui` caption and subcaption classes so that `moonstone` ones could override.
- Adjust the `GridListImageItem`'s height depending on the caption and subcaption height.

### Additional Considerations

### Links

ENYO-5426

### Comments

The target branch for this PR is not `develop` but `feature/ENYO-5426-same-height`. So I did not update `CHANGELOG.md` and no test case is needed.

### DIFF
--- a/packages/moonstone/GridListImageItem/GridListImageItem.less
+++ b/packages/moonstone/GridListImageItem/GridListImageItem.less
@@ -30,13 +30,11 @@
 	.caption {
 		.moon-sub-header-text();
 		padding-top: @moon-gridlist-item-caption-padding-top;
-		line-height: initial;
 	}
 
 	.subCaption {
 		.moon-body-text();
 		padding-top: @moon-gridlist-item-caption-padding-top;
-		line-height: initial;
 	}
 
 	// Skin colors
--- a/packages/moonstone/GridListImageItem/GridListImageItem.less
+++ b/packages/moonstone/GridListImageItem/GridListImageItem.less
@@ -4,6 +4,9 @@
 @import '../styles/skin.less';
 
 .gridListImageItem {
+	display:flex;
+	flex-direction: column;
+
 	box-sizing: border-box;
 	border-width: @moon-gridlist-item-border-width;
 	border-style: solid;
@@ -21,10 +24,6 @@
 		opacity: 0.5;
 	}
 
-	.image {
-		height: ~"calc(100% - " @moon-gridlist-item-two-captions-padding-bottom ~")";
-	}
-
 	&.selected {
 		.icon {
 			opacity: 1;
@@ -33,11 +32,13 @@
 
 	.caption {
 		.moon-sub-header-text();
+		height: @moon-gridlist-item-caption-height;
 		padding-top: @moon-gridlist-item-caption-padding-top;
 	}
 
 	.subCaption {
 		.moon-body-text();
+		height: @moon-gridlist-item-subcaption-height;
 		padding-top: @moon-gridlist-item-caption-padding-top;
 	}
 
--- a/packages/moonstone/GridListImageItem/GridListImageItem.less
+++ b/packages/moonstone/GridListImageItem/GridListImageItem.less
@@ -30,11 +30,13 @@
 	.caption {
 		.moon-sub-header-text();
 		padding-top: @moon-gridlist-item-caption-padding-top;
+		line-height: initial;
 	}
 
 	.subCaption {
 		.moon-body-text();
 		padding-top: @moon-gridlist-item-caption-padding-top;
+		line-height: initial;
 	}
 
 	// Skin colors
--- a/packages/moonstone/GridListImageItem/GridListImageItem.less
+++ b/packages/moonstone/GridListImageItem/GridListImageItem.less
@@ -29,13 +29,11 @@
 
 	.caption {
 		.moon-sub-header-text();
-		height: @moon-gridlist-item-caption-height;
 		padding-top: @moon-gridlist-item-caption-padding-top;
 	}
 
 	.subCaption {
 		.moon-body-text();
-		height: @moon-gridlist-item-subcaption-height;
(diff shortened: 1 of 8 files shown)


packages/moonstone/GridListImageItem/GridListImageItem.less
```diff
@@ -4,6 +4,9 @@
 @import '../styles/skin.less';
 
 .gridListImageItem {
+	display:flex;
```
Djspaceg: I think you may want `display: inline-flex` here. Let me know if that works for your scenarios. Some of the alignment issues of this ticket initially were due to the component being used in an inline context. If it needs to be both, we may want to add an `inline` prop, like `Item`.

Maybe consider using `Layout` and `Cell`? It may offer some nice built-in benefits at low cost. It got a perf optimization recently.

packages/ui/GridListImageItem/GridListImageItem.less
```diff
@@ -25,4 +25,12 @@
 		pointer-events: none;
 		height: 100%;
 	}
+
+	.caption {
+		/* Public Class Names */
+	}
+
+	.subCaption {
```
Djspaceg: We can (should) combine these into one set of rules:
```
	.caption,
	.subCaption {
		/* Public Class Names */
	}
```
They just need to exist in the file, with _some_ content in them, so they will be available in the hash when imported and won't be pruned out by webpack. Currently, the CSS comment is the trick middle-spot between being pruned and available.

packages/moonstone/GridListImageItem/GridListImageItem.less
ybsung: If applying `display: inline-flex`, it also worked. I had to apply `Layout` to `GridListImageItem`. I have done with `inline` and `orientation` props.

ybsung:
Hi @Djspaceg 
I fixed the issues that you mentioned.

The Travis CI build failed. I seems to be related with not my PR but our Enact build.

Djspaceg:
I've made a few changes:
1. Switched from `Layout` to `Column` (trivial, it's just an alias to `Layout orientation="vertical"`)
2. Removed the (now) unnecessary LESS variables. With the captions being `Cell shrink`, they'll automatically size to the dimensions of their content. (36 and 39 px respectively)
3. BONUS: Updated the VL stories to use 100% height instead of a pixel value since now Storybook allows content to fill the content space reliably.

@ybsung, if these changes are ok with you, you can go ahead and merge this into the other ticket. 👍

ryanjduffy:
@LucieRoy - Updating the height of the VirtualList stories might impact some test cases.

ybsung:
Oh we could remove the unnecessary LESS variables. Cool!!!. I'll verify the PR tomorrow.

ybsung:
Hi @Djspaceg 

> 1. Switched from Layout to Column (trivial, it's just an alias to Layout orientation="vertical")

: I verified all samples worked well. Thanks.

> 2. Removed the (now) unnecessary LESS variables. With the captions being Cell shrink, they'll automatically size to the dimensions of their content. (36 and 39 px respectively)

: By doing it, the code in my PR is  much better now.

I found that if removing `line-height` styles in the caption and the subcaption, it would be much flexible. If app developers override the `font-size` of the `caption` or the `sub-caption`, the `caption` or the `subcaption` could be clipped because of the `line-height`. By removing it, we could prevent clipping it. Please let me know your opinion about my [commit](https://github.com/enactjs/enact/pull/1872/commits/19312e27f5a8038bf58d4f0a6a33c425a6750d8d).

> 3. BONUS: Updated the VL stories to use 100% height instead of a pixel value since now Storybook allows content to fill the content space reliably.

: `100% height` is the one that we expected to use when creating VirtualList samplers. But as @ryanjduffy mentioned, we need to check its TCs. At the first time, I tried to add `height: 100%; min-height: 552px(or 549px);` styles not to break any VirtualList and VirtualGrdiList TCs. But I found that I needed to update the Scroller samples, too. So it would be better to separate it with this PR. I created PLAT-64761 to apply `100% height` to all VirtualList, VirtualGridList, and Scroller samplers.

CC: @LucieRoy

Djspaceg:
Separating the height change into its own PR makes sense for trackability/history.

Resetting the line-height to `initial` means that we lose the ability to control how tall-glyphs are handled in locales that need that feature. Additionally, we don't want any developer to effortlessly override the UX defined font-size for these components; that would remove consistency (perceived visual quality) and limit our ability to make framework-level design updates in the future. If a developer is going to undertake such an action, they need to approach it with the level of care and consideration that it involves, e.g. consideration for RTL, Marquee, dynamic Item sizing, tall-glyphs in every locale, which fonts are used, sub-pixel font rendering, font hinting sizes, and multiple screen sizes/scales. While it does make it easier for the styles to be overridden, by an external user, it's my opinion that that isn't the function of this component.

As an alternate perspective, and reminder, if the Moonstone defined styles are not meeting the needs of the app developers using it, *we* need to update those internally. Requiring every app to correctly define all of their own custom rules consistently along with every other app not only wastes their time but defeats the purpose of a well designed, well maintained, framework. We have to change if they need us to be different.

I'm going to recommend moving forward with this PR and reverting the [line-height change](https://github.com/enactjs/enact/pull/1872/commits/19312e27f5a8038bf58d4f0a6a33c425a6750d8d). What I'll do is I'll revert that change, merge this into the other PR, and if we change our mind about the line-height addition, we can cherrypick that back into the other PR before it's merged.

ybsung:
@Djspaceg
Thanks for your explanation. I agree with your opinion. I had to think about it more deeply.

PR #1873, opened 2018-08-13
ENYO-5480: Fix prevent scrolling on focus in pointer mode

ryanjduffy:
### Checklist

I noticed a disconnect between the original requirement for adding onUpdate and the way in which it was implemented. The requirement surfaced due to a need to scroll a Component (in that case, the entire ExpandableList) into view when that component caused a change in scroll bounds. It appears in the original implementation that we asserted that nearly any update of the component should cause this behavior but discovered that updates introduced upstream (as was described in this ticket) would inadvertently cause scrolling.

### Resolution
The gap in my view is that we didn't verify that the scroll bounds changed before invoking this logic. I adjusted the callback to verify the bounds have changed as well as updating the Scrollable-managed scrollHeight value.

### Additional Considerations

I also made an update to Scroller to only update the scroll top when the scroll size decreases. This fixed a bug in which repeatedly opening and closing an ExpandableList in a Scroller would alternately scroll it into view on open and then fail to scroll on open. We need to validate this change with the work in ENYO-5539 (PR yet to be created).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ The following is a curated list of changes in the Enact project, newest changes 
 - `moonstone/LabeledIcon` and `moonstone/LabeledIconButton` to have proper spacing and label-alignment with all label positions
 - `moonstone/Popup` to prevent duplicate 5-way navigation when `spotlightRestrict="self-first"`
 - `moonstone/Scroller` not to scroll to wrong position via 5way navigation in RTL languages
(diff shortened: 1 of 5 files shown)


ryanjduffy:
Yes, I believe so. It's solving a couple different problems. First, it's preventing scrolling when the focused item is partially out of view when the Scroller updates while in pointer mode (a rare, but possible case). Second, it fixes scrolling Expandables correctly into view when opening or closing them.

PR #1876, opened 2018-08-15
ENYO-5354: Delay PickerButton hold event

germboy:
### Checklist

On TV-only, after a hold event `Picker` can get into a state where selecting a `PickerButton` once can result in the picker value changing multiple times.

### Resolution
This issue appears to be the result of a performance issue on TV (whether through Storybook or specifically-created app) where the `keyup` isn't able to register in time before a `holdPulse` event is able to fire (causing the additional value change). The solution for the time being appears to be delaying the hold event for the `PickerButton` so that `holdPulse` isn't fired before the actual hold event is registered.

### Additional Considerations
This is more of a remedy, as an actual fix would require preventing the TV from getting into the odd state where this issue is observed. However, it isn't obvious to us how this is happening or how to prevent this.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -4,12 +4,13 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
-### Changed
-
-- `moonstone/Picker` key down hold threshold to 800ms before firing the `onChange` event
 ### Added
 
 - `moonstone/VideoPlayer` property `noMediaSliderFeedback`
+
+### Changed
+
+- `moonstone/Picker` key down hold threshold to 800ms before firing the `onChange` event
 
 ### Fixed
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/VideoPlayer` to round the time displayed down to nearest second
+- `moonstone/Picker` to correctly update its value once when button is selected
 
 ## [2.0.2] - 2018-13-01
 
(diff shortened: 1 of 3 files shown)


packages/moonstone/internal/Picker/PickerButton.js
```diff
@@ -88,7 +88,15 @@ const PickerButtonBase = kind({
 			);
 		} else {
 			return (
-				<IconButton {...rest} backgroundOpacity="transparent" disabled={disabled} small>
+				<IconButton
+					{...rest}
+					backgroundOpacity="transparent"
+					disabled={disabled}
+					holdConfig={{events: [
```
ryanjduffy: Should we externalize this to a module-scoped variable to avoid generating a new instance every render? Should it also be applied to the `joined` version as well?

packages/moonstone/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/VideoPlayer` to round the time displayed down to nearest second
+- `moonstone/Picker` to correctly update its value once when button is selected
```
ryanjduffy: This might be too vague to document the change. Perhaps we should mention the change in hold threshold since it could impact apps or testing.

packages/moonstone/internal/Picker/PickerButton.js
webOS101: Seems reasonable to apply to `joined` as well so the mouse or arrow key behavior is similar in both configurations.

germboy: The performance issue does not replicate with the `joined` version of `Picker`, so a fix there isn't required. However, I'm ok with keeping the hold behavior consistent across `Picker` variations, even though it's inconsistent across components. I don't have strong feelings either way.

germboy:
Added the `holdConfig` to `internal/Picker` (passing props down to `PickerButton)`, since the timing there will ultimately be used when normalizing behavior between joined/un-joined variants (future work).

PR #1878, opened 2018-08-15
PLAT: 64797: Add support for async i18n resource retrieval

ryanjduffy:
### Checklist

Taking a first cut at additive changes to support asynchronous retrieval of i18n resources without requiring its adoption in Moonstone.

### Additional Considerations

* [ ] Minimal testing has been done on impacts to Moonstone and still needs to happen.
* [ ] Would like to add tests for sync and async loading
* [ ] Would like to add tests for the new `i18n/Text` component
* Should `sync` be `true` or `false` by default for `i18n/I18nDecorator`? `false` is the direction but `true` is the current reality for Moonstone. Is setting it to `true` by default for `moonstone/MoonstoneDecorator` enough?


packages/core/util/Job.js
```diff
@@ -206,6 +206,19 @@ class Job {
 			this.fn(...args);
 		}
 	}
+
+	promise = (promise) => {
```
viodragon2: docs plz

viodragon2: Don't want to start exploring `Observable` for this PR, but it might be worth using it for canceling.

packages/i18n/I18nDecorator/getI18nClasses.js
```diff
@@ -1,60 +1,93 @@
...
+
+		options.onLoad(classes.filter(Boolean).join(' '));
+	});
+}
+
+/*
+ * A function that returns locale in class name.
+ *
+ * @memberof i18n/I18nDecorator
+ * @returns {String} Class
```
viodragon2: This needs an update

viodragon2: This needs an update

packages/i18n/src/Loader.js
```diff
@@ -201,12 +214,14 @@ EnyoLoader.prototype.loadFiles = function (paths, sync, params, callback) {
...
-					get(this._pathjoin(_root, path), handler);
+				this.loadManifestsSync(_root);
+
+				if (this.isAvailable(_root, path, true)) {
+					getSync(this._pathjoin(_root, path), handler);
 				}
 
-				if (!found && this.isAvailable(locdata, path)) {
-					get(this._pathjoin(locdata, path), handler);
+				if (!found && this.isAvailable(locdata, path, true)) {
```
viodragon2: what does this thrid parameter do?

packages/i18n/src/resBundle.js
```diff
@@ -57,8 +81,19 @@ function clearResBundle () {
 	resBundle = null;
 }
 
+function getIStringFromBundle (str, rb) {
```
viodragon2: Can we update `$L.toIString()` instead of creating a semi-duplicate code? I think we can simply add a check for `rb` value and create bundle when not provided.

packages/i18n/I18nDecorator/I18nDecorator.js
```diff
@@ -10,21 +10,44 @@
...
+const defaultConfig = {
+	/**
+	 * Retrieve i18n resource files synchronously
+	 *
+	 * @type {Boolean}
+	 * @default false
+	 * @public
+	 * @memberof i18n/I18nDecorator.I18nDecorator.defaultConfig
+	 */
+	sync: false
```
viodragon2: In response to your question in the description
> Should `sync` be `true` or `false` by default for `i18n/I18nDecorator`? `false` is the direction but `true` is the current reality for Moonstone. Is setting it to `true` by default for `moonstone/MoonstoneDecorator` enough?

IMHO, I think it depends on the status of `ilib`'s async support. If the version of `ilib` used in `enact` does not support 100%, then setting `sync` to `true` would be better. If not, async is probably a preferable way to do so.

packages/i18n/src/resBundle.js
ryanjduffy: I created this to be used by the new `Text` component to consolidate the key => IString logic. Good catch on the duplication, though! I've refactored `$L` to use this method.

packages/i18n/src/Loader.js
ryanjduffy: I think at one point it was a `sync` flag but now it's just cruft. :/ 🗑

packages/i18n/I18nDecorator/getI18nClasses.js
ryanjduffy: Well, it _eventually_ returns a string ... sort of ...

... Okay, I'll update it!

packages/core/util/tests/Job-specs.js
```diff
@@ -100,4 +100,28 @@ describe('Job', function () {
...
+
+		it('should not start job when stopped before promise resolves', function (done) {
+			const j = new Job(() => {
+				done(new Error('Job ran for stopped promise'));
+			});
+			j.promise(new Promise(resolve => setTimeout(resolve, 20)));
+			setTimeout(() => j.stop(), 10);
+			setTimeout(done, 30);
+		});
+	});
```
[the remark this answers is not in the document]
ryanjduffy: This test passes but not for the reason you might think.

The `promise()` method doesn't control the execution of the promise but only the execution of the "side effect" passed as the callback to `Job`.  In the test above, the "Previous promise ran" promise still runs (and calls `done()`) but:

* doesn't end the test because `done()` was already called
* doesn't invoke the job because another promise was registered with the job after it

The last test, `it('should not start job when another is started'`, is verifying the "cancelling" feature by passing a new promise to `promise()` thereby ignoring the resolution of the first. If you debug it, you'll notice it still resolves and still calls the `then` but bails when it tests if `this.id === promise`.

packages/core/util/tests/Job-specs.js
```diff
@@ -100,4 +100,28 @@ describe('Job', function () {
...
+		it('should start job for a resolved promise', function (done) {
+			const j = new Job(() => done());
+			j.promise(Promise.resolve(true));
+		});
+
+		it('should not start job for a rejected promise', function (done) {
+			const j = new Job(() => {
+				done(new Error('Job ran for rejected promise'));
+			});
+			j.promise(Promise.reject(true));
```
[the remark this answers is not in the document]
ryanjduffy: Took a little different approach which I think makes sense but would like your feedback on it.

I changed `promise()` to return a Promise (surprising, I know) which can then be handled via `catch` or `then`. The returned promise will resolve with the result from the job which is a bit different than the other Job functions which don't return anything but I think makes sense in the context of promises.

packages/core/util/Job.js
```diff
@@ -206,6 +206,19 @@ class Job {
...
+		if (!(promise instanceof Promise)) return;
```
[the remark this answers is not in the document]
ryanjduffy: Good call! I've adopted `invariant` and added a couple tests (including generalizing support for thenables rather than explicitly Promise).

packages/i18n/I18nDecorator/I18nDecorator.js
```diff
@@ -67,49 +92,91 @@ const I18nDecorator = hoc((config, Wrapped) => {
...
+				rtl
+			};
+		}
+
+		loadResources (locale) {
+			const options = {sync, locale};
+			const resources = Promise.all([
+				wrapIlibCallback(isRtlLocale, options),
+				wrapIlibCallback(getI18nClasses, options),
+				setResBundleLocale(locale, sync)
```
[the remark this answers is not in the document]
ryanjduffy: :+1: I've removed `setResBundleLocale` and added a simpler `setResBundle` and then separated the create and set phases as discussed.

packages/core/util/tests/Job-specs.js
```diff
@@ -131,4 +131,74 @@ describe('Job', function () {
 			j.idle('second');
 		});
 	});
+
+	describe('#promise', function () {
+		it('should throw when passed a non-thenable argument', function (done) {
+			const j = new Job(() => done('Unexpected job execution'));
```
webOS101: Should this have `new Error(...) in the done?

packages/core/util/tests/Job-specs.js
```diff
@@ -131,4 +131,74 @@ describe('Job', function () {
...
+				expect(value).to.equal(2);
+				done();
+			});
+			j.promise(new Promise(resolve => resolve(1)));
+			j.promise(new Promise(resolve => resolve(2)));
+		});
+
+		it('should return the value from the job to the resolved promise', function (done) {
+			const j = new Job(() => 'job value');
+			j.promise(Promise.resolve(true)).then(value => {
```
webOS101: Do we need to bother to add a catch on any of these for testing?

packages/core/util/tests/Job-specs.js
```diff
@@ -131,4 +131,74 @@ describe('Job', function () {
...
+				expect(value).to.equal('job value');
+				done();
+			});
+		});
+
+		it('should not return the value from the job to the resolved promise', function (done) {
```
webOS101: this test title is just the inverse of the title above, which seems to imply they can't both be true.  This should reference 'when job replaced'?

packages/i18n/I18nDecorator/I18nDecorator.js
```diff
@@ -66,49 +91,97 @@ const I18nDecorator = hoc((config, Wrapped) => {	// eslint-disable-line no-unuse
...
+			this.loadResourceJob.stop();
 			if (typeof window === 'object') {
 				off('languagechange', this.handleLocaleChange, window);
 			}
 		}
 
+		getDerivedStateForLocale (spec) {
+			const locale = updateLocale(spec);
+
+			let rtl = false;
```
webOS101: Is there any reason to default to `false` while waiting for a resource load?  Could this cause thrashing when switching from one rtl to another rtl locale?

packages/i18n/locale/locale.js
```diff
@@ -21,45 +20,70 @@ import {setResBundleLocale} from '../src/resBundle';
...
-	);
+	// eslint-disable-next-line no-new
+	new LocaleInfo(spec, {
+		...options,
+		onLoad: (li) => {
+			const locale = li.getLocale();
+
+			// We use the non-latin fonts for these languages (even though their scripts are technically
+			// considered latin)
+			const nonLatinLanguageOverrides =  ['vi', 'en-JP'];
```
webOS101: It occurs to me we should document this (if we didn't?) and, possibly, move this override to `moonstone` to separate our non-standard stuff.

packages/core/util/tests/Job-specs.js
ryanjduffy: It doesn't have to but probably should for consistency.

From https://mochajs.org/#asynchronous-code:
> This callback accepts both an Error instance (or subclass thereof) or a falsy value; anything else will cause a failed test.

packages/core/util/Job.js
```diff
@@ -206,6 +208,42 @@ class Job {
 			this.run(args);
 		}
 	}
+
+	/**
+	 * Starts the job when `promise` resolves.
+	 *
+	 * The job execution is tied to the resolution of the last promise passed. Like other methods on
+	 * `Job`, calling `promise()` again with a new Promise will block execution of the job until
+	 * that new Promise resolves. Any previous promises will still resolve normally but will not
```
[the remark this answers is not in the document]
ryanjduffy: Fixed ... I pRomise

packages/i18n/I18nDecorator/I18nDecorator.js
ryanjduffy: Good call. I'll rework this to only update that state member when in sync.

packages/i18n/locale/locale.js
ryanjduffy: Yeah. This should live in moonstone. I'll do a bit of refactoring to expose those as well.

packages/i18n/locale/locale.js
```diff
@@ -24,33 +24,33 @@ import {initCaseMappers} from '../src/case';
...
+					) ||
 					// the combination of language and region is treated as non-latin
-					nonLatinLanguageOverrides.indexOf(locale.toString()) !== -1
+					(
+						Array.isArray(nonLatinLanguageOverrides) &&
+						nonLatinLanguageOverrides.indexOf(locale.toString()) !== -1
+					)
 				) && (
 					// the non-latin language should be treated as latin
+					!Array.isArray(latinLanguageOverrides) ||
```
webOS101: I feel like this should be a direct mirror of the `nonLatinLanguageOverrides` so I will duplicate the locale check as well.

PR #1879, opened 2018-08-16
ENYO-5395: Change icon props to be required

teckliew:
### Checklist

Mark `children` or `icon` props as required for `Icon` and for components that use `Icon`.
The following will throw a `propType` required warning:

- `<ToggleIcon />`
- `<Icon />`
- `<IconBase />`
- `<IconButton />`
- `<LabeledIcon>plus</LabeledIcon>`
- `<LabeledIconButton>plus</LabeledIconButton>`

### Additional Considerations
One thing to consider is `RadioItem`; because it uses `ToggleIcon` in a special way, `icon` prop was set as an empty string, so it can bypass the `Proptype` warning.
We might want to consider using an actual `Icon` for `RadioItem` in the future.

### Links
ENYO-5395

### Comments


packages/ui/LabeledIcon/LabeledIcon.js
```diff
@@ -39,7 +39,7 @@ const LabeledIconBase = kind({
 		 * @type {Node}
 		 * @public
 		 */
-		children: PropTypes.node,
+		children: PropTypes.node.isRequired,
```
webOS101: Would it make more sense to make the `icon` property required rather than the `children` property?

packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
+- `moonstone/Icon` and `moonstone/labeledIcon` to require children
```
webOS101: Can you update to `ui` instead of `moonstone`?

haileyR:
hmm I doubt that `children` is a right term to use.. I know there are a lot of components using children prop, but like this one, it is an icon component(named so). so to be more clear with it, it should be just type or something that actually describes the prop.

haileyR:
Yea having a “type” prop should work since it is referring what type of icon it is.

Djspaceg:
Technically the content of what is displayed _in_ an icon is not the "type" of the icon. A type would be like a classification of the icon, like where it is used, rather than what the picture inside it is. `src` is the most appropriate I've seen on here so far, but it would a little overloaded to be used for the glyph name that the Icon component uses, and would only really be appropriate for the URI of the icon picture file. `name` is an HTML attribute which we wouldn't want to conflate here.

`icon` is totally redundant (combined with `src` for URIs), but at least lines up with the naming convention that `` tags use. And independent of all of this, that's a major breaking API change which we wouldn't be able to _enact_ until v3.0.0 anyway. We'll have plenty of time to discuss between now and then.

Djspaceg:
For some history, when Icon was originally created, `children` was appropriate for the icon content, because the line of thinking was we have a general rule that says "all visible primary content goes outside the JSX tags", (`children`) and Icon's icon names and URIs fit into this category. I've more recently been thinking about Icon in the terms of an HTML `img` tag, which has no content and relies solely on attributes. This approach becomes more appropriate too when the Icon component is nested _inside_ of other more complex components, like LabeledIcon or Button, which has its primary text content making sense in the children prop; meaning that the Icon children content needs to be moved into a different prop name (`icon`), which is bad for consistency and predictability.

PR #1880, opened 2018-08-17
ENYO-5551: When comparing Slider value against min, ensure value is set

germboy:
### Checklist

`Slider` can emit an `onChange` event when `value` has not actually changed. The issue is a result of `value` being unset and `Slider` incorrectly checking `value` against `min` before firing the `onChange` event.

### Resolution
Since `value` (when unset) should default to `min`, we need to ensure that is the case when we compare these values in the `isNotMin` method of `Slider/utils`.

### DIFF
--- a/packages/moonstone/Slider/utils.js
+++ b/packages/moonstone/Slider/utils.js
@@ -29,7 +29,7 @@ const isNotMax = (ev, {value, max}) => {
 	return value !== max;
 };
 
-const isNotMin = (ev, {value, min}) => {
+const isNotMin = (ev, {min, value = min}) => {
 	return value !== min;
 };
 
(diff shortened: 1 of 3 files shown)


packages/moonstone/Slider/utils.js
```diff
@@ -29,7 +29,7 @@ const isNotMax = (ev, {value, max}) => {
 	return value !== max;
 };
 
-const isNotMin = (ev, {value, min}) => {
+const isNotMin = (ev, {min, value = min}) => {
```
webOS101: Is this not a failure that a unit test would catch?

ryanjduffy: Good call! I added a couple tests for an undefined value.

PR #1888, opened 2018-08-20
PLAT-64864: Remove the unexpected Scrollbar button composite layer

ybsung:
### Checklist

There is the unexpected Scrollbar button composite layer.

### Resolution

The ScrollButton size is 40px X 40px. But the "arrow" icon size of the ScrollButton is 24px X 48px, which makes the new composite layer for it. So I tried to clip the icon with 40px X 40px. But the icon was not clipped visually.

### Links

PLAT-64864


ryanjduffy:
Targeted this PR against `develop`. We have concerns about using the `nth-child` selector and would like more time to investigate.

PR #1891, opened 2018-08-21
ENYO-5569: Handle less in onScroll event

viodragon2:
### Checklist

`onScroll` event handler is responsible for updating scroller's scroll thumb position and scroller button disabled state.

Each event handling takes up about ~16ms on TV.

### Resolution
Refactored `onScroll` handler to do less amount of work and used `requestAnimationFrame` to handle UI updates.
The `onScroll` event handler now takes up about ~3ms, and each UI update (which takes about ~10ms) occurs less.

### DIFF
--- a/packages/ui/Scrollable/ScrollableNative.js
+++ b/packages/ui/Scrollable/ScrollableNative.js
@@ -1210,7 +1210,7 @@ class ScrollableBaseNative extends Component {
 
 		if (childRef && childRef.containerRef) {
 			if (childRef.containerRef.addEventListener) {
-				childRef.containerRef.addEventListener('scroll', this.onScroll, {capture: true});
+				childRef.containerRef.addEventListener('scroll', this.onScroll, {capture: true, passive: true});
 				childRef.containerRef.addEventListener('mousedown', this.onMouseDown);
 			}
 			this.childRef.containerRef.style.scrollBehavior = 'smooth';
@@ -1230,7 +1230,7 @@ class ScrollableBaseNative extends Component {
 		}
 
 		if (childRef && childRef.containerRef && childRef.containerRef.removeEventListener) {
-			childRef.containerRef.removeEventListener('scroll', this.onScroll, {capture: true});
+			childRef.containerRef.removeEventListener('scroll', this.onScroll, {capture: true, passive: true});
 			childRef.containerRef.removeEventListener('mousedown', this.onMouseDown);
 		}
 
--- a/packages/ui/Scrollable/ScrollableNative.js
+++ b/packages/ui/Scrollable/ScrollableNative.js
@@ -442,6 +442,7 @@ class ScrollableBaseNative extends Component {
 			this.forwardScrollEvent('onScrollStop');
 		}
 		this.scrollStopJob.stop();
+		this.startScrollJob.stop();
 
 		this.removeEventListeners();
 		off('keydown', this.onKeyDown);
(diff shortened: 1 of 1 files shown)


packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -902,9 +887,23 @@ class ScrollableBaseNative extends Component {
 		this.isScrollAnimationTargetAccumulated = false;
 		this.scrolling = false;
 		this.forwardScrollEvent('onScrollStop');
+		this.startHidingThumb();
```
0x64: If the timing of hiding is fine, it could be a good move. Please _feel_ it for sure. :)

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -922,9 +921,12 @@ class ScrollableBaseNative extends Component {
 			this.childRef.didScroll(this.scrollLeft, this.scrollTop, dirHorizontal, dirVertical);
 		}
 		this.forwardScrollEvent('onScroll');
+		this.scrollStopJob.start();
 	}
 
-	scrollStopJob = new Job(this.scrollStopOnScroll.bind(this), scrollStopWaiting);
+	startScrollJob = new Job(this.scrollOnScroll, 16);
+
+	scrollStopJob = new Job(this.scrollStopOnScroll, scrollStopWaiting);
```
0x64: Note: We don't need to call _bind_ for arrow functions in general, but we've experienced _this binding issue_ due to this line before. I'm not sure what was the cause. It could be an issue related to transpiling configuration. If it works well now, it's time to remove _bind_. :)

packages/ui/Scrollable/ScrollableNative.js
0x64: Do this change need to be applied to JS version, too?

viodragon2: No, I do not think so. It's already doing it at `stop` https://github.com/enactjs/enact/blob/master/packages/ui/Scrollable/Scrollable.js#L936

0x64: Okay!

packages/ui/Scrollable/ScrollableNative.js
ryanjduffy: Binding shouldn't be necessary because babel transpilation stores `this` in a closure. In the case of class members, the first transpilation step is to move it into the constructor and the second is to cache an ref to `this` and rewrite the usage of `this` in the function to the ref.

Here's an example from babel's REPL with a couple common variations of arrow function and class member transpilation to see it all in action.

http://babeljs.io/repl#?babili=false&browsers=&build=&builtIns=false&spec=false&loose=false&code_lz=MYGwhgzhAEYEbGgbwFDWsMITQBQEpk10MB7AOwgBdoBbAUyoAtSATaAXj0I4D4iSJZgEsIAOgikGI8gHMA3MXQBfRSuIzZnaACZiDWnHoAnbQU79UgkeMnSmwuWujLixUsxPaAZgFdywFTCFNwCQg62UowOTsSqKMQQ9MAU7FzmfGHoNhJRms6urkA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=false&fileSize=false&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2&prettier=false&targets=&version=6.26.0&envVersion=

viodragon2:
Passive event listener does not affect in this scenario because `onScroll` event is fired from explicit `HTMLElement.scrollTo` function rather than a wheel or a touch. Though our wheel performance wasn't bad, it wouldn't hurt to add.

PR #1893, opened 2018-08-21
ENYO-5555: Fix Picker hold behavior

germboy:
### Checklist

`Picker` fires `onChange` events (due to a hold) on an inconsistent timing-based pattern, depending on how the hold originated
- 5-way directionally using `joined`: uses custom `Job` with timing
- Selection-key based on unjoined: uses Spottable/Touchable `onHold` behavior
- Pointer hold: used no timing, rather used the repeated "down" events to handle changes

### Resolution
- Added `selectionKey` prop support to `Spottable`, enabling the use of alternate keys to initiate selection. This allows components like a joined horizontal `Picker` to fire selection and hold events via 5-way left-right keys.
- Refactored `Picker` selection and hold event management a bit.


packages/moonstone/internal/Picker/tests/Picker-specs.js
```diff
@@ -215,6 +215,7 @@ describe('Picker Specs', function () {
 
 		const expected = -1;
 		picker.simulate('keyDown', {keyCode: 37});
+		picker.simulate('mousedown');
```
germboy: Note for reviewer: The proposed change to `Spottable` will now allow for the `mousedown` event to be forwarded/simulated when specified keys (like in this case, the left arrow key) are pressed. In practice, there is no change in behavior from the previous implementation. Strangely, in these tests it appears we must now explicitly call `picker.simulate('mousedown')` even though I feel this should be included via the `keydown` event in the prior line.

The `keydown` event still sets the internal flag in `Picker` that designates how the value should change based on which direction was pressed, so the test is still valid. That being said, I'm wondering if there is a reason we must explicitly call this line now and if there is a way to avoid this. Thoughts?

ryanjduffy:
@germboy and I discussed this a bit late last week. I had reservations on adding API to configure selection keys because of the mismatch of additional key events triggering mouse events. However, I think we can consider mitigating that in the longer term by migrating away from `onClick` as the "selection" event in favor of a custom event that better represents how Spottable behaves.

This could all be moot if a native focus management solution lands before we get to that point.

PR #1894, opened 2018-08-22
PLAT-64761: Full-height Scroller Samples

ybsung:
### Checklist

Samples for Scroller use an awkward static height.

### Resolution

Update the samples to use the full height available in the window.

### Additional Considerations

By default, VirtualList, VirtualGridList, and Scroller stretches to fit to a window. So we don't need `width: 100%; height: 100%` style in them. I removed it.

### Links

PLAT-64761


Djspaceg:
These look like great additions. I forgot Scroller scroller expanded to the full size of its container.

PR #1895, opened 2018-08-22
PLAT-64864: Remove page control's composite layers

ybsung:
### Checklist

We overrided the Button's CSS styles by using CSS element selector to remove page control's composite layers and to enhance scroll performance. But the Button's DOM structure changed and some composite layers were not removed because we used the CSS element selector.

### Resolution

I reverted the previous patch to remove page control's composite layers.
I override the Button's CSS styles by using CSS class selector.

### Additional Considerations

1. The ScrollButton's CSS classes are defined in not ScrollButton.less but Scrollbar.less. So I moved them to the ScrollButton.less.
2. I forgot to remove `direction` prop in ScrollButton.js in  PLAT-39161. I'll do it in PLAT-64865.
3. To override Button's `client` class, I add `client` string to the `publicClassNames` of the Button and the Icon Button. If it is not proper, I could override it with the following code.

AS-IS
```
	.client {
		overflow: hidden;
	}
```

TO-BE
```
	.bg ~ div {
		overflow: hidden;
	}
```

4. To override the Button's CSS styles, I added `ccc={css}` in the ScrollButton's render function. I think we could remove it because `rest` includes `css` property. But it did not work if removing it.
==> As @0x64 mentioned, `css` is NOT enumarable.  So `rest` could not include `css` property with `render: ({children, disabled, ...rest})` code.

Before removing `css={css}`
```
	render: ({css, children, disabled, ...rest}) => {
		delete rest.active;
		delete rest.direction;

		return (
			<IconButton
				{...rest}
				backgroundOpacity="transparent"
				css={css}
				disabled={disabled}
				small
			>
				{children}
			</IconButton>
		);
	}
```

After removing `css={css}`
```
	render: ({/*css, */children, disabled, ...rest}) => {
		delete rest.active;
		delete rest.direction;

		return (
			<IconButton
				{...rest}
				backgroundOpacity="transparent"
				/*css={css}*/
				disabled={disabled}
				small
			>
				{children}
			</IconButton>
		);
	}
```

### Links

PLAT-64864

### DIFF
--- a/packages/moonstone/Scrollable/ScrollButton.js
+++ b/packages/moonstone/Scrollable/ScrollButton.js
@@ -4,7 +4,7 @@ import React from 'react';
 
 import IconButton from '../IconButton';
 
-import css from './ScrollButton.less';
+import componentCss from './ScrollButton.less';
 
 const classNameMap = {
 	down: css.scrollbarBottomButton,
@@ -88,7 +88,7 @@ const ScrollButton = kind({
 	},
 
 	styles: {
-		css,
+		css: componentCss,
 		className: 'scrollButton'
 	},
 
--- a/packages/moonstone/Scrollable/ScrollButton.js
+++ b/packages/moonstone/Scrollable/ScrollButton.js
@@ -7,10 +7,10 @@ import IconButton from '../IconButton';
 import componentCss from './ScrollButton.less';
 
 const classNameMap = {
-	down: css.scrollbarBottomButton,
-	left: css.scrollbarLeftButton,
-	right: css.scrollbarRightButton,
-	up: css.scrollbarUpButton
+	down: componentCss.scrollbarBottomButton,
+	left: componentCss.scrollbarLeftButton,
+	right: componentCss.scrollbarRightButton,
+	up: componentCss.scrollbarUpButton
 };
 
 /**
--- a/packages/moonstone/Scrollable/ScrollButton.js
+++ b/packages/moonstone/Scrollable/ScrollButton.js
@@ -4,13 +4,13 @@ import React from 'react';
 
 import IconButton from '../IconButton';
 
-import componentCss from './ScrollButton.less';
+import css from './ScrollButton.less';
 
 const classNameMap = {
-	down: componentCss.scrollbarBottomButton,
-	left: componentCss.scrollbarLeftButton,
-	right: componentCss.scrollbarRightButton,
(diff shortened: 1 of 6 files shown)


packages/moonstone/Scrollable/ScrollButton.js
```diff
@@ -88,14 +97,15 @@ const ScrollButton = kind({
 		className: ({direction, styler}) => styler.append(classNameMap[direction])
 	},
 
-	render: ({children, disabled, ...rest}) => {
+	render: ({css, children, disabled, ...rest}) => {
```
0x64: Let `css` be in `rest` in peace.

0x64: Hmm... It looks that this code works with desctructuring `css` because `css` is NOT _enumarable_. And, IMO, we don't need to define `css` prop for `ScrollButton` due to this, but I'm not sure what is the better way to implement this.

ybsung: I passed `css` variable from not the arguments of the render function but the variable imported from './ScrollButton.less' directly.

0x64: 💨

0x64:
LGTM

packages/moonstone/Scrollable/ScrollButton.less
```diff
@@ -0,0 +1,17 @@
...
+	margin: 0;
+
+	&.small:before {
+		content: none;
+	}
+
+	.bg {
+		will-change: auto;
+		-webkit-transform: initial;
+		transform: initial;
```
Djspaceg: This removes the transformation and acceleration of the button press animation but doesn't address any of the animation rules. This changes the way the button responds to an out-of-design style.

ybsung: Hi @Djspaceg 

There were some composite layers due to the page controls. I tried to remove them before to enhance list scroll performance.

This PR includes
* the revert of the previous patch to remove the layers and 
* the new patch to remove the layers again.

Without the `transform: initial;` change, there is no way to remove the layers. If my PR breaks the animation rules, I'd like
* to go ahead only to revert of the previous patch and
* not to include the new patch
in this PR. 

Could you let me know your opinion?
Thanks.

lishammel:
closing out as associated jira has been closed

PR #1896, opened 2018-08-22
ENYO-5529: Limited overscroll effects to wheel input only in `moonstone`

0x64:
### Checklist

We have a requirement to show overscroll effect only on wheel input in `moonstone` so far and current our components support various types of inputs including wheel, drag, arrow key, and so on.

So, limited overscroll effects to wheel input only to meet the requirement.

### Resolution
We only need effects on wheel input for `moonstone` for now, but it does not mean that we don't need to support other types of input for future requirements. We also need to be ready for other theme libraries beside `moonstone`.
In this PR, a way to select types of input for overscroll effects is added and `moonstone` supports only wheel input by default. If we need to expand effects to other inputs, it will be easily accomplished by setting the default selection.

### Additional Considerations
Even though `ui` does not have any overscroll effect, `ui` has a logic to support it for other theme libraries which are rely on the logic of `ui`. So most of changes by this PR is about `ui` library.

### Links
ENYO-5529

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -468,6 +468,7 @@ class ScrollableBase extends Component {
 		this.animateOnFocus = true;
 
 		if (!repeat && this.hasFocus()) {
+			const {overscrollEffectOn} = this.props;
 			let
 				overscrollEffectRequired = false,
 				direction = null;
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -147,6 +147,21 @@ class ScrollableBase extends Component {
 		focusableScrollbar: PropTypes.bool,
 
 		/**
+		 * Specifies overscroll effects shows on which type of inputs.
+		 *
+		 * @type {String[]}
+		 * @default null
+		 * @private
+		 */
+		overscrollEffectOn: PropTypes.shape({
+			arrowKey: PropTypes.bool,
+			drag: PropTypes.bool,
+			pageKey: PropTypes.bool,
+			scrollbarButton: PropTypes.bool,
+			wheel: PropTypes.bool
+		}),
+
+		/**
 		 * Sets the hint string read when focusing the next button in the vertical scroll bar.
 		 *
 		 * @type {String}
@@ -185,7 +200,14 @@ class ScrollableBase extends Component {
 
 	static defaultProps = {
 		'data-spotlight-container-disabled': false,
-		focusableScrollbar: false
+		focusableScrollbar: false,
+		overscrollEffectOn: {
+			arrowKey: false,
+			drag: false,
+			pageKey: false,
+			scrollbarButton: false,
+			wheel: true
+		}
 	}
 
 	constructor (props) {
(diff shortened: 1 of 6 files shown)


0x64:
Note)

1. I added a new prop as _private_. If anyone thinks it should be _public_, please let me know.
2. _CHANGELOG_ should be updated before merging. I will add after 1st review.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -146,6 +146,21 @@ class ScrollableBase extends Component {
 		 */
 		focusableScrollbar: PropTypes.bool,
 
+		/**
+		 * Specifies overscroll effects shows on which type of inputs.
+		 *
+		 * @type {String[]}
```
ybsung: The type might be Object or null, right?

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -269,6 +269,19 @@ class ScrollableBaseNative extends Component {
 		 */
 		onWheel: PropTypes.func,
 
+		/**
+		 * Specifies overscroll effects shows on which type of inputs.
+		 *
+		 * @type {String[]}
```
ybsung: Same above.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -146,6 +146,21 @@ class ScrollableBase extends Component {
...
+		 * @default null
```
ybsung: Oh, default is not null.

packages/moonstone/Scrollable/Scrollable.js
0x64: Oops, I missed to change it! Thanks.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -146,6 +146,21 @@ class ScrollableBase extends Component {
...
+		 * @private
+		 */
+		overscrollEffectOn: PropTypes.shape({
+			arrowKey: PropTypes.bool,
+			drag: PropTypes.bool,
+			pageKey: PropTypes.bool,
+			scrollbarButton: PropTypes.bool,
+			wheel: PropTypes.bool
```
ybsung: How about omitting `drag` and `wheel` here because those properties are not used in moonstone/Scrolllable? And we could merge `overscrollEffectOn` in moonstone into `overscrollEffectOn` default object in ui.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -146,6 +146,21 @@ class ScrollableBase extends Component {
 		 */
 		focusableScrollbar: PropTypes.bool,
 
+		/**
+		 * Specifies overscroll effects shows on which type of inputs.
+		 *
+		 * @type {Object}
+		 * @default {arrowKey: false, drag: false, pageKey: false, scrollbarButton: false, wheel: true}
```
ybsung: Could you define the default value in multiple lines?
e.g>
VideoPlayer.js
```
		 * @default {
		 *	fastForward: ['2', '4', '8', '16'],
		 *	rewind: ['-2', '-4', '-8', '-16'],
		 *	slowForward: ['1/4', '1/2'],
		 *	slowRewind: ['-1/2', '-1']
		 * }
		 * @public
		 */
		playbackRateHash: PropTypes.shape({
```

packages/moonstone/Scrollable/Scrollable.js
0x64: As you pointed out, `drag` and `wheel` (in JS version) values are not used in `moonstone`. My intention is to expose all possible features explicitly. Also, if we define all values in `moonstone`, we don't need to merge `prop` objects and `ui/Scrollable` can simply read prop value without any modification or assigning to the local(this) variable. Could you share your opinion?

packages/moonstone/Scrollable/Scrollable.js
0x64: Okay. It looks better to read. :)

packages/moonstone/Scrollable/Scrollable.js
ybsung: I got your point.

ybsung:
LGTM

PR #1900, opened 2018-08-22
ENYO-5574: Prevent browser scroll behavior when 5way navigating Video…

germboy:
…Player

### Checklist

When attempting to 5-way navigate in the `VideoPlayer`, the browser default scroll behavior is not prevented when attempting to:
- 5-way down when controls are hidden
- 5-way up from the slider
- 5-way down from the slider (when media controls are available)
- 5-way left/right from the slider (although you'll have to adjust CSS in the VideoPlayer sample to allow for panel overflow and scale the VP from 1 to at least 1.2 in order to see the effect)

### Resolution
We should call `preventDefault()` on events where we prevent default spotlight behavior by stopping propagation.


ryanjduffy:
Interestingly, this is the same class of problem we fixed in #1914.

I verified this in Chrome by loading up a custom video player sample and setting a static size for the player and placing it inside a smaller container with `overflow: scroll`. Then, scroll the controls into view and test 5-way interaction. With the fix applied, the unexpected scrolling doesn't occur. :+1:

PR #1904, opened 2018-08-27
ENYO-5579: Changed to call focus() function of the target element with `preventScroll` option

0x64:
### Checklist

Not only native versions but also JS version of lists has an issue that the container node is scrolled to show up the focused element is located in the center of the container's viewport when a child element's `focus` API is called.

### Resolution
So far, we applied some device-specific way to prevent scrolling by focus for native lists, but the legacy solution does not work on JS versions.
We have a few ways to try and it looks that the best way is to implement the standard way to prevent scroll by focus. (See the second link at the bottom.)

### Additional Considerations
Beside `spotlight`, there are several `focus()` calls in `ui` and `moonstone` and it looks okay to change only `spotlight` regarding purposes of each code blocks.

### Links
ENYO-5579
https://html.spec.whatwg.org/multipage/interaction.html#dom-focusoptions-preventscroll

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
+### Changed
+
+- `spotlight` to call focus() function of the target element with `{preventScroll: true}` option
+
 ### Fixed
 
 - `spotlight/SpotlightContainerDecorator` to unmount config instead of remove when spotlightId is changed if it preserves id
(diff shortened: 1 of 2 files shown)


ybsung:
LGTM

0x64:
@viodragon2 
I thought that we actually don't need to make it configurable. Scroll position by focus is not standardized so Webkit family including Blink scrolls to make a focused element to the center, and Firefox scrolls to the close edge. So UI devs need to handle the position anyway.
But, in general, I agree to make it configurable and also agree with you to close this one after the original your working branch is finished up.

0x64:
Note: @viodragon2's PR is #1906.

ryanjduffy:
Closing in favor of #1918

PR #1906, opened 2018-08-27
ENYO-5579: Add `preventScroll` option to SpotlightContainer

viodragon2:
### Checklist

By default browser repositions the container when a focus goes out of bounds so that focused element is centered. `Scrollable` and `ScrollableNative` tries to adjust the position on `focusin` event to anchor focused item to the edge of the container in the direction of the key. However, these operations are costly and there's a chance where a user can see a focused element in the center in a slower system (e.g. TV).

### Resolution
Add `preventScroll` option to be configured in `SpotlightContainer` and set default value to `true` in `moonstone/Scrollable` and `moonstoneScrollableNative`

### Additional Considerations
In this fix, only `moonstone/Scrollable*` are taking advantage of `preventScroll`, but we could expand this to be applied to `MoonstoneDecorator` if needed. But it seemed rather complicated to apply and decided only apply where it's needed.


viodragon2:
Added `preventScroll` option to `ExpandableItem`. This, however, is not a universal fix and will cause the issue if a custom component with `SpotlightContainer` without the config is inside the scroller.

I have created another branch https://github.com/enactjs/enact/tree/feature/ENYO-5532-stephen-context which uses the new context API which would resolve this problem, but I'm not a super big fan of this approach.

ryanjduffy:
Closing in favor of #1918

PR #1913, opened 2018-08-28
ENYO-5580 - It is impossible to handle up/down keycode when using isItemDisabled in virtualList

rundmt:
### Checklist

`VirtualList` stops `onKeyDown` from propagating so other components can't receive the event. This is done to prevent spotlight from calling focus again on its `onKeyDown`.

### Resolution
Instead of using stopPropagation we can pause spotlight temporarily so we don't get two `focus` calls, but we can still have the event propagate. We pause Spotlight `onKeyDown` and remove it `onKeyUp`.

### Links
ENYO-5580

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -604,7 +604,7 @@ const VirtualListBaseFactory = (type) => {
 				if (isWrapped) {
 					this.isWrappedBy5way = true;
 				}
-
+				
 				if (firstIndex <= nextIndex && nextIndex < firstIndex + numOfItems) {
 					this.focusOnItem(nextIndex);
 				} else {
@@ -639,8 +639,12 @@ const VirtualListBaseFactory = (type) => {
 				ev.preventDefault();
 				this.setSpotlightContainerRestrict(keyCode, target);
 				Spotlight.setPointerMode(false);
+				this.jumpToSpottableItem(keyCode, repeat, target)
 				if (this.jumpToSpottableItem(keyCode, repeat, target)) {
-					ev.stopPropagation();
+					// Pause Spotlight so we don't focus twice
+					// Resume after a short delay so Spotlight will work as intended
+					Spotlight.pause();
+					setTimeout(Spotlight.resume, 0);
 				}
 			}
 		}
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -274,7 +274,7 @@ const VirtualListBaseFactory = (type) => {
 
 			if (containerNode && containerNode.removeEventListener) {
 				containerNode.removeEventListener('keydown', this.onKeyDown);
-				containerNode.removeEventListener('keyup', this.revertSpotlightState);
+				containerNode.removeEventListener('keyup', this.resumeSpotlight);
 			}
 
 			this.setContainerDisabled(false);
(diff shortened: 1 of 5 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -268,12 +269,12 @@ const VirtualListBaseFactory = (type) => {
 				// remove a function for preventing native scrolling by Spotlight
 				if (containerNode && containerNode.removeEventListener) {
 					containerNode.removeEventListener('scroll', this.preventScroll);
-					containerNode.removeEventListener('keydown', this.onKeyDown);
 				}
 			}
 
 			if (containerNode && containerNode.removeEventListener) {
 				containerNode.removeEventListener('keydown', this.onKeyDown);
+				containerNode.removeEventListener('keyup', this.revertSpotlightState);
```
ybsung: Hi @rundmt 
I could not find where `revertSpotlightState` function is. Am I missing something?

ybsung: `revertSpotlightState` might be `resumeSpotlight`?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -274,7 +274,9 @@ const VirtualListBaseFactory = (type) => {
 
 			if (containerNode && containerNode.removeEventListener) {
 				containerNode.removeEventListener('keydown', this.onKeyDown);
-				containerNode.removeEventListener('keyup', this.resumeSpotlight);
+			}
+			if (document) {
```
ryanjduffy: Suggest we update this to use `this.resumeSpotlight()` as well so we don't accidentally leave spotlight paused when unmounting. Also gets DRYness bonus points. :)

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -645,19 +647,18 @@ const VirtualListBaseFactory = (type) => {
 					this.virtualListPause = new Pause('VirtualList');
 					this.virtualListPause.pause();
 
-					const containerNode = this.uiRef.containerRef;
-
-					if (containerNode && containerNode.addEventListener) {
-						containerNode.addEventListener('keyup', this.resumeSpotlight);
+					if (document) {
```
ryanjduffy: I think we're safe to access `document` in this code and in `resumeSpotlight`. IIRC, the only lifecycle methods that run during prerendering where `window` and `document` are undefined are the constructor and `componentWillMount` (which is going away). @JayCanuck can correct me if I've misspoke.

JayCanuck: For prerender support, `window`/`document` should be avoided in `constructor`, `getDerivedStateFromProps`, `componentWillMount` and `render`, yep.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -268,14 +269,15 @@ const VirtualListBaseFactory = (type) => {
...
 					containerNode.removeEventListener('scroll', this.preventScroll);
-					containerNode.removeEventListener('keydown', this.onKeyDown);
 				}
 			}
 
 			if (containerNode && containerNode.removeEventListener) {
 				containerNode.removeEventListener('keydown', this.onKeyDown);
 			}
 
+			this.resumeSpotlight();
```
ybsung: If `isItemDisabled` prop is not defined in `VirtualList`, `this.virtualListPause.resume();` in `resumeSpotlight()` generates an error when `this.resumeSpotlight()` is called because this.virtualListPause is `null`.

ryanjduffy:
Looks good, @rundmt!

PR #1914, opened 2018-08-29
PLAT-64196: Fix unexpected scrolling by 5-way key from inside an ExpandableItem

0x64:
### Checklist

When `ExpandableItem` has a `true` value of `lockBottom` prop and `down` key is pressed on the last item of `ExpandableItem`, a scroller which contains `ExpandableItem` scrolls unexpectedly.
Fixed not to make a scrolling in this case. Since `lockBottom` is `true`, there will be no action.

Edit:
Thanks to @MikyungKim, I found that the issue also happens on `up` key when `ExpandableItem` is located on the top of a viewport of a scroller (regardless `lockBottom` is set or not).
Added the fix for this, too.

### Resolution
A scrolling is a default behavior of web engine by `up` or `down` key. So added `preventDefault()`.

### Links
PLAT-64196

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -10,6 +10,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/ExpandableItem` to not make scrolling of a parent scroller when `down` key is pressed on the last item
 - `moonstone/VideoPlayer` so that activity is detected and the `autoCloseTimeout` timer is reset when using 5-way to navigate from the media slider
 
 ## [2.1.0] - 2018-08-20
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/ExpandableItem` to not make scrolling of a parent scroller when `down` key is pressed on the last item
+- `moonstone/ExpandableItem` to not make scrolling of a parent scroller when 5-way key is pressed on the first item or the last item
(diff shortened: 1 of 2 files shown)


packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -305,6 +305,7 @@ const ExpandableItemBase = kind({
 				} else if (isDown(keyCode) && wouldDirectionLeaveContainer('down', target)) {
 					if (lockBottom) {
 						ev.nativeEvent.stopImmediatePropagation();
+						ev.nativeEvent.preventDefault();
```
MikyungKim: ev.preventDefault?

packages/moonstone/CHANGELOG.md
```diff
@@ -10,6 +10,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/ExpandableItem` to not make scrolling of a parent scroller when `down` key is pressed on the last item
```
MikyungKim: Wrong position

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -305,6 +305,7 @@ const ExpandableItemBase = kind({
 				} else if (isDown(keyCode) && wouldDirectionLeaveContainer('down', target)) {
 					if (lockBottom) {
 						ev.nativeEvent.stopImmediatePropagation();
+						ev.preventDefault();
```
MikyungKim: do we need this for 'up' key too?

0x64: Good point! I just updated to fix `up` key too. The issue happens `ExpandableItem` is located on the top of a viewport of a scroller and the currently focused item is the first one.

MikyungKim:
:ok:

PR #1918, opened 2018-08-30
ENYO-5579: Prevent browser scrolling components into view on focus

ryanjduffy:
### Checklist

webOS applications using VirtualList were experiencing focus problems when using `isItemDisabled` event when native scrolling wasn't enabled for the app.

### Resolution
Support using the [focusOptions param of the `focus()` API](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus) to suppress scrolling focused elements into view. This PR "connects" the usage of the API with the `overflow` spotlight container config which indicates that the container supports focusable components out of view (e.g. from a scroller).

### Additional Considerations
* A future additive API should support passing the focus options to `Spotlight.focus()` to override this default behavior.
* This PR also removes the `data-preventscrollonfocus` key which was a stopgap until the new focus options API was supported. @0x64 - can you confirm this is appropriate here?

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -38,6 +38,7 @@ import {
 	isContainer,
 	isContainer5WayHoldable,
 	isNavigable,
+	isWithinOverflowContainer,
 	mayActivateContainer,
 	removeAllContainers,
 	removeContainer,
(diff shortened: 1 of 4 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -199,11 +200,13 @@ const Spotlight = (function () {
...
 		}
 
+		const focusOptions = isWithinOverflowContainer(elem, containerIds) ? {preventScroll: true} : null;
+
 		let silentFocus = function () {
 			if (currentFocusedElement) {
 				currentFocusedElement.blur();
 			}
-			elem.focus();
+			elem.focus(focusOptions);
```
webOS101: I wasn't able to determine how widespread support is for `focusOptions`.  But, given we're already replacing a proprietary solution with this experimental one, I think it's 👍 .

PR #1925, opened 2018-09-04
ENYO-5604: Tooltip is not repositioned when text changed

Sangwook-Lee:
### Checklist

Even if length of tooltipText, the position of tooltip is not recalculated.

### Resolution
I added  function to calculate position when text changed.

### Links
ENYO-5604


changgilee:
LGTM. At first time, I think setTooltipLayout() function can be called another lifecycle. (ex. componentWillReceiveProps). But it has to be called after DOM updated for recalculate base on new tooltip size. So, it seems that componentDidUpdate is proper lifecycle in this case.

PR #1927, opened 2018-09-04
ENYO-5605: Fix ScrollableNative to update scroll buttons

viodragon2:
### Checklist

`ui/ScrollableNative` needs to update scroll buttons when it scrolls. Using `requestAnimationFrame` helped us to increase performance (#1891), but setting `timeout` (i.e. 16ms) will throttle the event, and will not properly use the last input from the event.

### Resolution
Removed `timeout` parameter from the job.

### Additional Considerations
We could retain the throttling part of it to reduce the amount of work to be done while scrolling, and update the scroll buttons after scroll stop event. Note that since the wait time for scroll stop event is 200ms, there's a slight delay in a state change. But maybe it's okay as JS version scroller changes the state after scroll animation completes.

Another idea I had was to debounce job to handle the scroll event to correctly set the last position for calculation. But I haven't tried this method.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
+- `ui/ScrolllableNative` to update scroll buttons correctly on wheel
 - `ui/Marquee` positioning bug when used with CSS flexbox layouts
 
 ## [2.1.1] - 2018-08-27
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,6 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/ScrolllableNative` to update scroll buttons correctly on wheel
 - `ui/Marquee` positioning bug when used with CSS flexbox layouts
 
 ## [2.1.1] - 2018-08-27
--- a/packages/ui/Scrollable/ScrollableNative.js
+++ b/packages/ui/Scrollable/ScrollableNative.js
@@ -383,6 +383,7 @@ class ScrollableBaseNative extends Component {
 		};
 
 		this.overscrollEnabled = !!(props.applyOverscrollEffect);
+		this.scrollTicking = false;
 
 		props.cbScrollTo(this.scrollTo);
 	}
@@ -463,7 +464,6 @@ class ScrollableBaseNative extends Component {
 			this.forwardScrollEvent('onScrollStop');
 		}
 		this.scrollStopJob.stop();
-		this.startScrollJob.stop();
 
 		this.removeEventListeners();
 		off('keydown', this.onKeyDown);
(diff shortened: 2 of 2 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
+- `ui/ScrolllableNative` to update scroll buttons correctly on wheel
```
ryanjduffy: ScrollableNative isn't public so let's omit this entry.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -958,7 +958,7 @@ class ScrollableBaseNative extends Component {
 		this.scrollStopJob.start();
 	}
 
-	startScrollJob = new Job(this.scrollOnScroll, 16);
+	startScrollJob = new Job(this.scrollOnScroll);
```
ryanjduffy: I think this is a good change but perhaps not for the reason you intended it. When testing a VirtualListNative sample, I noticed that including a timeout along with my proposal to include `stop()` in `startRafAfter` resulted in long delays for rendering items because each scroll event cancelled the previous RAF so items weren't rendered until scrolling stopped.

viodragon2:
I do not believe `startRaf` was meant to be used to handle repeating calls like scroll events. The `timeout` is to merely "delay" the raf call, and nothing more. I thought it was throttling because it would get called each time after the timeout, but it's not correct behavior.

Instead of using `Job.startRaf()`, I've decided to use `window.requestAnimationFrame` directly.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -728,7 +728,13 @@ class ScrollableBaseNative extends Component {
 	onScroll = (ev) => {
 		const {scrollLeft, scrollTop} = ev.target;
 
-		this.startScrollJob.startRaf(scrollLeft, scrollTop);
+		if (!this.scrollTicking) {
+			window.requestAnimationFrame(() => {
```
ryanjduffy: This would be hard to reproduce but, for safety, should we cache the RAF id so we can cancel during `componentDidUnmount`? If so, we could probably repurpose `this.scrollTicking` to be the id holder.

viodragon2: sure

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -728,7 +732,12 @@ class ScrollableBaseNative extends Component {
 	onScroll = (ev) => {
 		const {scrollLeft, scrollTop} = ev.target;
 
-		this.startScrollJob.startRaf(scrollLeft, scrollTop);
+		if (!this.scrollRaFId) {
+			this.scrollRaFId = window.requestAnimationFrame(() => {
```
0x64: Is there any possibility that `scroll` event occurs twice or more times between rAF callback? If so, we may have still the same problem of ignoring the later scroll positions.

PR #1928, opened 2018-09-05
ENYO-5603: Clean up spotlight container disabled prop propagation

viodragon2:
### Checklist

`SpotlightContainer` is declared in `Scrollable`s and yet it's propagating `data-spotlight-disabled` property down to its children. I believe this is a cruft left over from #1553

### Resolution
Removed all `setContainerDisabled` from `moonstone/Scrollable`, `moonstone/Scroller` and `moonstone/ScrollableNative`

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -264,13 +264,6 @@ class ScrollableBase extends Component {
 		if (focusedItem) {
 			focusedItem.blur();
 		}
-
-		if (
-			direction === 'vertical' && this.uiRef.canScrollVertically(bounds) ||
-			direction === 'horizontal' && this.uiRef.canScrollHorizontally(bounds)
-		) {
-			this.childRef.setContainerDisabled(true);
-		}
 	}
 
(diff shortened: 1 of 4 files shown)


ryanjduffy:
I'm going to look into this more. I think there are some parts of this change that we should take but I'm concerned about the removal of the code that sets `data-spotlight-container-disabled`. That change allows the pointer to focus contents while scrolling which should be prevented (if I recall the requirements correctly).

ryanjduffy:
I don't think we can safely remove `spotlightId` from Scrollable either. VirtualList relies on it to configure the spotlight container.

It does seem like we have some confused responsibilities from the refactoring. `moonstone/Scrollable` "owns" the spotlight container but calls into its child for `setContainerDisabled` which then finds the parent node to set `data-spotlight-container-disabled`. Seems like there would be an opportunity to optimize this interaction and reduce the coupling.

What do you think, @viodragon2?

viodragon2:
The more I look into this, the more complicated it becomes to untangle spotlight from `Scrollable`. I don't think there's a clean or an easy way to address this issue, but the benefits of cleaning this up at this point seem tiny. I'll close this issue for now and reconsider this later as a bigger part of `Scroller`, `VirtualList`, and `Scrollable` refactor work.

PR #1931, opened 2018-09-06
ENYO-5612: Fix VideoPlayer a11y for infoComponents (Option 1)

Baekwoo-s:
### Checklist

### This PR is option 1 of ENYO-5612.
handle blur() before setTimeout and live with spotlight blinking issue
Do not merge with PR below. (option 2)
https://github.com/enactjs/enact/pull/1937

Fixed the timing to read out `infoComponents` accessibility value when `moreButton` or `moreButtonColor` is pressed.

### Resolution
I have fixed the timing of reading `infoComponents` temporarily using setTimeout due to an unknown cause.
(The unknown cause is being tracked in the web engine, and a framework-side workaround until web engine fix is identified and implemented.)

### Links
ENYO-5612

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+- `moonstone/VideoPlayer` timing to read out `infoComponents` accessibility value when `moreButton` or `moreButtonColor` is pressed
+
 ## [2.1.2] - 2018-09-04
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ybsung:
LGTM

Baekwoo-s:
@viodragon2 
> Is there any reason you chose 50ms for the delay? 

I wanted to reasonably delay the reading of the region role and aria-labelledby for unknown reasons.

> but I'm concerned for the case where user presses the more button color key (e.g. blue color key) when the focus is not in more button. 50ms delay seems a bit too long for the case.

I test setTimeout to 1ms, but the issue is not reproduced. But I think it is necessary to have a reasonable delay because I am worried about issues with low probability.
Could you adjust the setTimeout value? If it seems you need more delay than 1ms.

PR #1932, opened 2018-09-06
ENYO-5616: Added exception for cases where refs don't exist

Sangwook-Lee:
### Checklist

There's the problem when tooltipText is changed to an empty string or changed from an empty string to a string.

### Resolution
I added exception handling for cases where refs don't exist. If there's no ref, that returns.

### Links
[ENYO-5616]
[GT-25843]

### DIFF
--- a/packages/sampler/stories/qa/Tooltip.js
+++ b/packages/sampler/stories/qa/Tooltip.js
@@ -57,6 +57,8 @@ class ChangeableTooltip extends React.Component {
 			this.setState({text: 'long text'});
 		} else if (text === 'long text') {
 			this.setState({text: 'very loooooooooooong text'});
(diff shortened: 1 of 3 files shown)


Sangwook-Lee:
@ryanjduffy , I agree with you. Thank you for review.

packages/sampler/stories/qa/Tooltip.js
```diff
@@ -57,6 +57,8 @@ class ChangeableTooltip extends React.Component {
 			this.setState({text: 'long text'});
 		} else if (text === 'long text') {
 			this.setState({text: 'very loooooooooooong text'});
+		} else if (text === 'very loooooooooooong text') {
```
webOS101: It does seem like we could have added a unit test for this case, too.

PR #1933, opened 2018-09-06
PLAT-48029: Added reachedEdgeInfo within `onScrollStop` callback

0x64:
### Checklist

We have a requirement to notify a list or a scroller reached the scroll edge. With the existing APIs, an app should keep an eye on the info provided by _onScroll_ callback function and it is not good on the performance aspect.
In this PR, we provide (maybe little limited) alternative solution to provide reaching status by _onScrollStop_ callback to help apps to choose more light method if possible.

### Resolution
_onScrollStop_ callback provides `reachedEdgeInfo` which contains reaching status of `top`, `bottom`, `left`, and `right` edge.

### Additional Considerations
Note that if a list or a scroller is shortened and not scrollable anymore, no edge will be marked as _true_(reached)_ since apps don't need to show a visual effect for short contents.

### Links
PLAT-48029

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -868,7 +868,7 @@ class ScrollableBase extends Component {
 				edge = this.getEdgeFromPosition(this.scrollLeft, bounds.maxLeft);
 
 			if (edge) { // if edge is null, no need to check which edge is reached.
-				if ((edge === 'before' && !rtl) || ((edge === 'after') && rtl)) {
+				if ((edge === 'before' && !rtl) || (edge === 'after' && rtl)) {
 					reachedEdgeInfo.left = true;
 				} else {
 					reachedEdgeInfo.right = true;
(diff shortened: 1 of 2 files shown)


0x64:
Closing due to changing approach for the original issue.

0x64:
Reopened; closed by accident.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -857,6 +857,38 @@ class ScrollableBase extends Component {
...
+			bounds = this.getScrollBounds(),
+			reachedEdgeInfo = {bottom: false, left: false, right: false, top: false};
+
+		if (this.canScrollHorizontally(bounds)) {
+			const
+				rtl = this.state.rtl,
+				edge = this.getEdgeFromPosition(this.scrollLeft, bounds.maxLeft);
+
+			if (edge) { // if edge is null, no need to check which edge is reached.
+				if ((edge === 'before') === rtl) {
```
ybsung: I like the following code instead of the code above.
`rtl && edge !== 'before' || !rtl && edge === 'before'`

PR #1934, opened 2018-09-07
ENYO-5614: Fix Scrollable to notify user when scrolling is not possible via voice command

changgilee:
### Checklist

Base on UX requirement of Voice Control, proper message has to shown when scrolling is not possible

### Resolution
In enact side, we need to check scroll is possible or not before control scroll position with voice.

### Links
ENYO-5614

### Comments
Enact-DCO-1.0-Signed-off-by: Changgi Lee changgi.lee@lge.com

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -667,9 +667,8 @@ class ScrollableBase extends Component {
 	}
 
 	notifyVoiceException = (e, type) => {
-		const param = {'voiceUi': {'exception': type}};
 		if (window.webOSVoiceReportActionResult) {
-			window.webOSVoiceReportActionResult(param);
+			window.webOSVoiceReportActionResult({'voiceUi': {'exception': type}});
 			e.preventDefault();
 		}
 	}
@@ -688,7 +687,7 @@ class ScrollableBase extends Component {
 				return false;
 			}
 		} else if (scroll === 'down' || scroll === 'bottom') {
-			if (scrollTop === scrollBounds.maxTop) {
+			if (scrollTop >= scrollBounds.maxTop - 1) {
 				notifyAlreadyCompleted();
 				return false;
 			}
@@ -719,39 +718,46 @@ class ScrollableBase extends Component {
 				case 'up':
 					this.voiceControlDirection = 'vertical';
 					this.onScrollbarButtonClick({isPreviousScrollButton: true, isVerticalScrollBar: true});
+					e.preventDefault();
 					break;
 				case 'down':
 					this.voiceControlDirection = 'vertical';
 					this.onScrollbarButtonClick({isPreviousScrollButton: false, isVerticalScrollBar: true});
+					e.preventDefault();
 					break;
 				case 'left':
 					this.voiceControlDirection = 'horizontal';
 					this.onScrollbarButtonClick({isPreviousScrollButton: !isRtl, isVerticalScrollBar: false});
+					e.preventDefault();
 					break;
 				case 'right':
 					this.voiceControlDirection = 'horizontal';
 					this.onScrollbarButtonClick({isPreviousScrollButton: isRtl, isVerticalScrollBar: false});
+					e.preventDefault();
 					break;
 				case 'top':
 					this.voiceControlDirection = 'vertical';
 					this.uiRef.scrollTo({align: 'top'});
+					e.preventDefault();
 					break;
 				case 'bottom':
 					this.voiceControlDirection = 'vertical';
 					this.uiRef.scrollTo({align: 'bottom'});
+					e.preventDefault();
 					break;
 				case 'leftmost':
 					this.voiceControlDirection = 'horizontal';
 					this.uiRef.scrollTo({align: isRtl ? 'right' : 'left'});
+					e.preventDefault();
 					break;
 				case 'rightmost':
 					this.voiceControlDirection = 'horizontal';
 					this.uiRef.scrollTo({align: isRtl ? 'left' : 'right'});
+					e.preventDefault();
 					break;
 				default:
 					this.isVoiceControl = false;
 			}
-			e.preventDefault();
 		}
 	}
 
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -666,49 +666,71 @@ class ScrollableBase extends Component {
 		}
 	}
 
+	canVoiceScroll = (scroll) => {
+		const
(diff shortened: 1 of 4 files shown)


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -666,49 +666,93 @@ class ScrollableBase extends Component {
 		}
 	}
 
+	notifyVoiceException = (e, type) => {
+		const param = {'voiceUi': {'exception': type}};
+		if (window.webOSVoiceReportActionResult) {
+			window.webOSVoiceReportActionResult(param);
```
ybsung: I don't think we need to define the `param` variable. It would be better to pass `{voiceUi: {exception: type}}` object directly as the first parameter of the `webOSVoiceReportActionResult` function.

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -744,49 +744,93 @@ class ScrollableBaseNative extends Component {
...
+		const scrollBounds = this.uiRef.getScrollBounds();
+		const notifyAlreadyCompleted = () => this.notifyVoiceException(e, 'alreadyCompleted');
+
+		if (scroll === 'up' || scroll === 'top') {
+			if (scrollTop === 0) {
+				notifyAlreadyCompleted();
+				return false;
+			}
+		} else if (scroll === 'down' || scroll === 'bottom') {
+			if (scrollTop === scrollBounds.maxTop) {
```
ybsung: The condition might be `scrollTop >= scrollBounds.maxTop -1` because sometimes scrollTop is between scrollBounds.maxTop -1 and scrollBounds.maxTop when it reaches to the bottom in VirtualListNative or ScrollerNative (not VirtualList, Scroller).

ybsung:
@changgilee 
Please
- Check `A CHANGELOG entry is included` in the PR description
- Refresh your PR branch with the latest develop code due to the conflict
- Create a TC

packages/moonstone/Scrollable/Scrollable.js
changgilee: It seems more clean. I will fix it.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -666,49 +666,93 @@ class ScrollableBase extends Component {
...
+				return false;
+			}
+		} else if (scroll === 'right' || scroll === 'rightmost') {
+			const limit = isRtl ? 0 : scrollBounds.maxLeft;
+			if (scrollLeft === limit) {
+				notifyAlreadyCompleted();
+				return false;
+			}
+		}
+		return true;
```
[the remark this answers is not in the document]
changgilee: In this case, we can reduce code as you mentioned, but it seems that eslint warning occur(disallow nested ternary expressions: https://eslint.org/docs/rules/no-nested-ternary). Additionally, if we consider more specific warning message of each direction case(not now, but can be classification in the future), legacy code seems more acceptable.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -666,49 +666,93 @@ class ScrollableBase extends Component {
...
+					break;
+				case 'rightmost':
+					this.voiceControlDirection = 'horizontal';
+					this.uiRef.scrollTo({align: isRtl ? 'left' : 'right'});
+					break;
+				default:
+					this.isVoiceControl = false;
+			}
+			e.preventDefault();
+		}
```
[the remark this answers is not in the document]
changgilee: Basically, legacy code is something wrong because e.preventDefalut() has to called when voice control performed. So, I will change code for it will not called when default case of switch statement.

packages/moonstone/Scrollable/Scrollable.js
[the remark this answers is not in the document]
ybsung: OK.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -666,49 +666,99 @@ class ScrollableBase extends Component {
...
+				return false;
+			}
+		} else if (scroll === 'down' || scroll === 'bottom') {
+			if (scrollTop >= scrollBounds.maxTop - 1) {
+				notifyAlreadyCompleted();
+				return false;
+			}
+		} else if (scroll === 'left' || scroll === 'leftmost') {
+			const limit = isRtl ? scrollBounds.maxLeft : 0;
+			if (scrollLeft === limit) {
```
ybsung: As you did, it is needed to check `scrollLeft >= scrollBounds.maxLeft -1` in here as well as the case when `scroll === 'right' || scroll === 'rightmost'`.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -666,49 +666,99 @@ class ScrollableBase extends Component {
 		}
 	}
 
+	notifyVoiceException = (e, type) => {
+		if (window.webOSVoiceReportActionResult) {
+			window.webOSVoiceReportActionResult({'voiceUi': {'exception': type}});
```
ybsung: It would be better to use `'alreadyCompleted'` instead of type variable.

ybsung:
I hope we could reduce the code related with `if` and `switch` conditions. Could you check my commit - https://github.com/enactjs/enact/commit/cf885f180b8ef7390013c8a51cabef53597baa70 in `ENYO-5614-ybsung` branch and share your opinions?

ybsung:
Hi @changgilee 
I have one question for e.preventDefault(). Which default behavior in a browser do you want to prevent? According to my understanding, `onvoice` is our customized event. So there is no default behavior in a browser, then it seems that we don't need to call e.preventDefault(). Did I miss something?

changgilee:
Hi @ybsung 
Actually, e.preventDefault() is used specific purpose in Web Engine side.
In current status, scroll control with voice can be used different case.
Case 1
: App developer did not attach onVoice handler.
: When user speak, Web Engine find DOM node and control it directly.
: In this case, Web Engine can determine scroll status because they control it directly.
Case 2
: App developer attach onVoice event handler.
: When scroll related voice input occur, onVoice handler is called. (Web Engine does not control DOM node. Another words, app should change scroll status)
: In this case, Web Engine should distinguish app perform something or not in event handler. (It does not mean success or fail) e.preventDefault() is used for interface to Web Engine during voice control process and Web Engine Developer request to us call it after perform in voice handler.

changgilee:
@ybsung 
I did update PR base on yours.

ybsung:
@changgilee 
Thanks for your explanation about the default browser behavior for `onvoice`.
This PR looks good to me.

PR #1935, opened 2018-09-07
ENYO-5623: Fix ui/Marquee to stop when blurred during restart timer

ryanjduffy:
### Checklist

In #1301, we added a `Job` to manage cancelling marquee instances in `MarqueeController` to allow synchronized marquees to continue when a stop event (e.g. blurring) was immediately followed by a start event (focusing, pointer enter). As a result, it became possible (either through impeccable timing or sluggishness on lower powered hardware) to blur a synchronized marquee during the restart timer causing the job to be scheduled but then reset by the "complete" trigger before it fires.

### Resolution

* Remove the call to stop the cancel job on "complete"

### Additional Considerations

I couldn't come up with a reason that the "complete" trigger would stop the cancel job. The purpose of the complete trigger is to inform the controller that one instance is done in order to coordinate restarting all instances at the same time. It isn't a trigger to initiate marquee and so shouldn't prevent cancelling (as far as i can tell).

### DIFF
--- a/packages/ui/Marquee/MarqueeController.js
+++ b/packages/ui/Marquee/MarqueeController.js
@@ -230,7 +230,6 @@ const MarqueeController = hoc(defaultConfig, (config, Wrapped) => {
 		handleComplete = (component) => {
 			const complete = this.markReady(component);
 			if (complete) {
-				this.cancelJob.stop();
 				this.markAll(STATE.ready);
 				this.dispatch('start');
 			}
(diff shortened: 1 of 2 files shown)


packages/ui/Marquee/MarqueeController.js
```diff
@@ -230,7 +230,6 @@ const MarqueeController = hoc(defaultConfig, (config, Wrapped) => {
 		handleComplete = (component) => {
 			const complete = this.markReady(component);
 			if (complete) {
-				this.cancelJob.stop();
```
webOS101: Seems like this is safe enough.  I think the thought process was that any time we were going to issue a start, we should prevent cancel from stopping it.  But, I think in this instance, it makes sense that if we have a cancel pending, we would want to allow it to prevent the restart.

PR #1936, opened 2018-09-10
ENYO-5504: Scroll page by page via Channel Up / Down keys without focused item

ybsung:
### Checklist

We did not support that VirtualList scrolls page by page when any item and any paging control in the VirtualList did not get focus.

### Resolution

If pressing Channel Up or Down keys when any item and any paging control did not get focus, VirtualList scrolls page by page.

### Additional Considerations

I also fixed the following bug.

*Condition*
- the list is at the top position
- the Up Paging Control is disabled
- the cursor is on the gap between items (pointer mode)
- a Channel Down is pressed

*Expected result*: Spotlight still hides
*Actual result*: Spotlight shows because VirtualList converts to the 5-way key mode, then `Spottable` `componentDidUpdate` calls `Spotlight.focus(containerId);`.

So I converted to pointer mode before `Spottable` `componentDidUpdate` is called.

### Links

ENYO-5504

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -450,13 +450,19 @@ class ScrollableBase extends Component {
 
 			if (isPageUp(keyCode) || isPageDown(keyCode)) {
 				if (this.props.direction === 'vertical' || this.props.direction === 'both') {
-					// We need to convert to 5-way key mode to move Spot to another item manually.
-					Spotlight.setPointerMode(false);
+					const isPointerMode = Spotlight.getPointerMode();
+
+					if (isPointerMode) {
+						// We need to convert to 5-way key mode to move Spot to another item manually.
+						Spotlight.setPointerMode(false);
+					}
 					direction = isPageUp(keyCode) ? 'up' : 'down';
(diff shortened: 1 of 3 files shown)


0x64:
LGTM

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -450,9 +450,19 @@ class ScrollableBase extends Component {
 
 			if (isPageUp(keyCode) || isPageDown(keyCode)) {
 				if (this.props.direction === 'vertical' || this.props.direction === 'both') {
-					Spotlight.setPointerMode(false);
+					const isPointerMode = Spotlight.getPointerMode();
+
+					if (isPointerMode) {
```
MikyungKim: Let's confine this behavior only for focused item existing case.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -417,11 +423,17 @@ class ScrollableBase extends Component {
 				this.childRef.scrollToNextItem({direction, focusedItem, reverseDirection: rDirection, spotlightId});
 			}
 
-			// Need to check whether an overscroll effect is needed
-			return true;
+			if (isPointerMode) {
```
MikyungKim: This can be a little ambiguous. Let's call setPointerMode here also.

PR #1938, opened 2018-09-11
ENYO-5626: Fix Spottable to respect pause state when it becomes enabled

Sangwook-Lee:
### Checklist

The knob is blurred while dragging. So, it can focus to other components while dragging.

### Resolution
I fixed to pass `dragging` props from `SliderBehaviorDecorator` to `IncrementSlider`.
So, when `dragging` prop is `true`, it's impossible to spotlight to the increment/decrement buttons.

### Links
[ENYO-5626]

### DIFF
--- a/packages/sampler/stories/qa/IncrementSlider.js
+++ b/packages/sampler/stories/qa/IncrementSlider.js
@@ -1,6 +1,37 @@
 import IncrementSliderDelayValue from './components/IncrementSliderDelayValue';
 import React from 'react';
+import IncrementSlider from '@enact/moonstone/IncrementSlider';
+import Button from '@enact/moonstone/Button';
+import ri from '@enact/ui/resolution';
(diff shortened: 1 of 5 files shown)


packages/sampler/stories/qa/IncrementSlider.js
```diff
@@ -11,4 +42,13 @@ storiesOf('IncrementSlider', module)
 				<IncrementSliderDelayValue />
 			</div>
 		)
+	)
+	.add(
+		'spotlight behavior while dragging',
+		() => (
+			<div>
+				While holding down the knob (dragging), move the cursor quickly between knob and SliderButtons.
```
webOS101: Should this message indicate what should happen.  I suppose that either the test case has the instructions and the result or the sample should have both.  It doesn't make as much sense to have only half the info in two places.

PR #1940, opened 2018-09-11
ENYO-5630: focus to default element when moving to next panel

yeram:
### Checklist

Set `autoFocus` prop as `default-element` to `children` when panels' index increases

### Resolution
Based on legacy spec, it doesn't restore last focused when pushing panel

### Additional Considerations

### Links
ENYO-5630

### Comments


packages/moonstone/Panels/Viewport.js
```diff
@@ -117,10 +128,17 @@ const ViewportBase = class extends React.Component {
...
-		return child ? React.cloneElement(child, {
-			spotlightId: child.props.spotlightId || generateId(index, 'panel-container', Spotlight.remove),
+		const {spotlightId = generateId(index, 'panel-container', Spotlight.remove)} = child.props;
+		const props = {
+			spotlightId,
 			'data-index': index
-		}) : null;
+		};
+
+		if (this.state.direction === 'forward') {
```
webOS101: Should we only apply this if no other value is set?

PR #1942, opened 2018-09-11
ENYO-5586: Fix 5-way navigation into VideoPlayer

viodragon2:
### Checklist

If `VideoPlayer` were focused from the outside, the default spotlight logic is ignored and spots to the closest element.

### Resolution
Removed unnecessary config option `enterTo: default-element` from outer container of `MediaControls`. This is okay as `VideoPlayer` explicitly handles the "down" arrow key to focus to `this.mediaControlsSpotlightId` which will eventually focus to `defaultElement` set by the container config.

### Additional Considerations
We may need to look into `SpotlightContainer` behavior more deeply and sort out how `enterTo` options should be handled to find the next spottable item.

I did not include CHANGELOG since this does not affect the regular `VideoPlayer` use case which is supposed to be used in fullscreen.

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -23,7 +23,6 @@ import {countReactChildren} from './util';
 import css from './VideoPlayer.less';
 
 const OuterContainer = SpotlightContainerDecorator({
-	enterTo: 'default-element',
 	defaultElement: [
 		`.${css.leftComponents} .${spotlightDefaultClass}`,
 		`.${css.rightComponents} .${spotlightDefaultClass}`,
--- a/packages/moonstone/VideoPlayer/VideoPlayer.js
+++ b/packages/moonstone/VideoPlayer/VideoPlayer.js
@@ -50,7 +50,14 @@ import Video from './Video';
 import css from './VideoPlayer.less';
 
 const SpottableDiv = Touchable(Spottable('div'));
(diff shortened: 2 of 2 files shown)


ryanjduffy:
@viodragon2 Your analysis is correct for pointer mode (this is intentional) but maybe not for 5-way.

When I tested in 5-way, I noticed that:
1. Spotlight attempts to find a target first in the list (`container-1`) and then in the root (`spotlightRootDecorator`).
2. Within the root, it finds the Panel which has `enterTo: 'last-focused'`
3. Last focused fails because nothing has been focused yet so it falls back to default element which is any spottable element in the body.
4. When the slider and media controls are visible, both are considered candidates but the slider comes first so it is selected

I think another solution to this problem might be to configure VideoPlayer with `enterTo: 'default-element'` and add default element selectors for the media control frame and the spottable div. This preserves always landing on the right control in the media controls as well since Panel would now store the VP as the last focused instead of its children.

Something like:

```
const RootContainer = SpotlightContainerDecorator({
	enterTo: 'default-element',
	defaultElement: [`.${css.controlsHandleAbove}`, `.${css.controlsFrame}`]
}, 'div');
```

What do you think?

ryanjduffy:
We could decouple from that class with another selector by either passing another `className` or a custom attribute (e.g. `data-we-want-to-find-this-later`) to `ComponentOverride`. I'd consider that an API change, though, since it would affect consumers using the override so we might want to defer that sort of change until later.

PR #1943, opened 2018-09-12
ENYO-5634: Fix title of expandable is invisible when expandable opens

yeram:
### Checklist

Do not adjust `scrollTop` when nested item is focused

### Resolution
When expandable opens, `calculateScrollTop` is called two times. (expandable -> nested item)
I tried to skip second adjustment because it makes title invisible.

### Additional Considerations
Not sure it's right approach but I'm trying to modify as small as possible

### Links
ENYO-5634

### Comments


viodragon2:
LGTM

PR #1944, opened 2018-09-12
ENYO-5626: Fix Spottable to respect pause state when it becomes enabled

ryanjduffy:
### Checklist

When spotlight is paused and a spottable component becomes enabled while the pointer is over it, it will focus itself effectively ignoring the paused state.

### Resolution
Move the paused condition up to guard against focusing when enabled for both pointer and 5-way.

### Additional Considerations
Added a qa sample (Spotlight > Disabled with Pause) which can be used to test this sort of scenario.


ryanjduffy:
Closing in favor of #1938

PR #1946, opened 2018-09-13
ENYO-5638: Fix VideoPlayer to show correct playback rate feedback on play or pause

xbinary:
### Checklist

Fix playback rate value is shown on the `FeedbackTooltip` in the middle of changing playback

`FeedbackTooltip` has `playbackRate` and `playbackState` props
1) `play`() is called during 4x(or etc) playback
2) `playbackRate` is changed to 2x and then `FeedbackTooltip` is rendered.
3) `playbackState` is changed to 'play' and then `FeedbackTooltip` is rendered.
Despite the presence of `setState` in the `play()` , it is performed synchronously.
as a result, render occur twice

### Resolution
`playbackRate` and `playbackState` must be changed at once for rendering of `FeedbackTooltip`
I think that the simplest way at this time is to do `prevCommand` variable allocation first.

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/VideoPlayer` to show correct playback rate feedback on play or pause
+
 ## [2.1.3] - 2018-09-10
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Thanks for the fix, @xbinary!

This is good example of why we should avoid non-`state` stateful members. In this case, React is executing the state change from `showFeedback` synchronously because it's in executed while in the render phase (see https://github.com/facebook/react/blob/03ab1efeb4dde9e07b93ff038c87cc4fa443abf4/packages/react-reconciler/src/ReactFiberScheduler.js#L1761). As a result, the feedback component is re-rendered immediately using the recently-out-of-date `this.prevCommand` rather than after the play and pause functions complete when an async `setState` would have ran.

webOS101:
Given the (somewhat) obscure source of this issue, should we add an inline comment so this doesn't get refactored back into a bug?  Or, just a leave a comment about fixing this properly for future refactorers?

ryanjduffy:
Done!

PR #1948, opened 2018-09-13
ENYO-5639: Fix Button and IconButton styling of image-based icons

ryanjduffy:
### Checklist

When using an image-based icon for `Button` or `IconButton`, the focused and disabled state isn't apparent because the image's opacity isn't adjusted.

### Resolution
Update styling rules for `Button` and `IconButton` to restore color and set opacity for the internal `Icon` to give it the same visual appearance for both font icons and image icons when focused and disabled.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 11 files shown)


viodragon2:
How about a QA sampler for the case?

ryanjduffy:
Fair point! I added a somewhat generalized one to Spotlight to test this scenario but can be extended later if we want.

PR #1950, opened 2018-09-19
ENYO-5642: Add voice control to GridListImageItem

changgilee:
### Checklist

GridListImageItem should be support voice control feature

### Additional Considerations
data-webos-voice-intent='Select' is used as default.
But depending on the app, data-webos-voice-intent can be used another value.

### Links
ENYO-5642

### Comments

### DIFF
--- a/packages/moonstone/GridListImageItem/GridListImageItem.js
+++ b/packages/moonstone/GridListImageItem/GridListImageItem.js
@@ -120,6 +120,7 @@ const GridListImageItemBase = kind({
 	},
 
 	defaultProps: {
+		'data-webos-voice-intent': 'Select',
 		placeholder: defaultPlaceholder,
 		selected: false
 	},
@@ -140,7 +141,7 @@ const GridListImageItemBase = kind({
 				{...rest}
 				captionComponent={captionComponent}
 				css={css}
-				data-webos-voice-intent={voiceIntent || 'Select'}
+				data-webos-voice-intent={voiceIntent}
 				iconComponent={Icon}
 				imageComponent={Image}
 				selectionOverlay={selectionOverlay}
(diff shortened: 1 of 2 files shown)


packages/moonstone/GridListImageItem/GridListImageItem.js
```diff
@@ -132,6 +140,7 @@ const GridListImageItemBase = kind({
 				{...rest}
 				captionComponent={captionComponent}
 				css={css}
+				data-webos-voice-intent={voiceIntent || 'Select'}
```
ryanjduffy: What is the reason to apply the default value here rather than via `defaultProps`?

changgilee: It seems that using defaultProps is better way and I will change it. Thanks~

packages/moonstone/GridListImageItem/GridListImageItem.js
```diff
@@ -121,7 +131,7 @@ const GridListImageItemBase = kind({
 		publicClassNames: ['gridListImageItem', 'icon', 'image', 'selected', 'caption', 'subCaption']
 	},
 
-	render: ({css, selectionOverlay, ...rest}) => {
+	render: ({css, 'data-webos-voice-intent': voiceIntent, selectionOverlay, ...rest}) => {
```
webOS101: It is not necessary to destructure this since it's just passed in with `{...rest}` below.

packages/moonstone/GridListImageItem/GridListImageItem.js
```diff
@@ -51,6 +51,15 @@ const GridListImageItemBase = kind({
...
 
 	propTypes: /** @lends moonstone/GridListImageItem.GridListImageItemBase.prototype */ {
+		/**
+		 * The voice control intent.
+		 *
+		 * @type {String}
+		 * @default 'Select'
+		 * @public
+		 */
+		'data-webos-voice-intent': PropTypes.string,
```
webOS101: Normally we would need `@memberof` here but I just upgraded our doc parsing tool so this is no longer required!  :partyparrot:

PR #1955, opened 2018-09-22
ENYO-5648: Blur active element when container becomes spotlightDisabled

viodragon2:
### Checklist

When a spotlight container component manages, `spotlightDisabled` are normally propagated to its children to prevent from getting focused.
`moonstone/VideoPlayer` sets `MediaControls`'s `spotlightDisabled` prop to `true`
This state is all spread across the known controls (e.g. MediaButtons like play, rew/ff, etc.)
However, spottable components specified by the users (e.g. `leftComponents`, `rightComponents`, `children`) do not.

If the last focused item was one of those user specified spottable components and `MediaControls` is hidden, you can still interact with the component even though it’s hidden.

A real world example would be in P&V video player. Let’s say you have a focus on the icon button that opens up a contextual popup (e.g. subtitle) in the `rightComponents`, wait for MediaControls to disappear after timeout, press “enter”, then you would have a contextual popup being opened.

### Resolution
Blur active element from the container when `SpotlightContainerDecorator` receives `spotlightDisabled` changed to true.

### Additional Considerations

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -405,7 +405,7 @@ const MediaControlsBase = kind({
 
 		return (
 			<OuterContainer {...rest} spotlightId={spotlightId}>
-				<div className={css.leftComponents}>{leftComponents}</div>
+				<Container className={css.leftComponents} spotlightDisabled={spotlightDisabled}>{leftComponents}</Container>
 				<div className={css.centerComponentsContainer}>
 					<div className={centerClassName}>
 						<Container className={css.mediaControls} spotlightDisabled={showMoreComponents || spotlightDisabled}>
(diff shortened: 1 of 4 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -405,7 +405,7 @@ const MediaControlsBase = kind({
 
 		return (
 			<OuterContainer {...rest} spotlightId={spotlightId}>
```
ryanjduffy: Could we have assigned `spotlightDisabled` to this container rather than wrapping left/right components with a new container?

viodragon2:
After some investigation, I'm closing this ticket and leaning towards making a VideoPlayer specific change. `ExpandableTransitionContainer` uses `spotlightDisabled` only for preventing navigation while retaining focus, which conflicts with the proposed idea. `ExpandableSpotlightDecorator`'` `highlightLabeledItem` function may be updated according to the new rule, but I don't think it's worth the risk at this point given all related bugs reproduced from this logic (#1179, #1634, #1815 to name a few).

PR #1956, opened 2018-09-25
ENYO-5643: Fix spotlight focus on window focus

germboy:
### Checklist

Spotlight can sometimes set focus to an unintended target when the application window gains focus

### Resolution
If the last-focused element is unavailable, we should try to set focus to the next logical target within the same container as the last-focused element. Spotlight also should take into account overflow containers when selecting a container's focus target.

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -11,7 +11,7 @@ import {coerceArray} from '@enact/core/util';
 import intersection from 'ramda/src/intersection';
 import last from 'ramda/src/last';
 
-import {matchSelector} from './utils';
+import {contains, matchSelector, getContainerRect, getRect} from './utils';
 
 const containerAttribute = 'data-spotlight-id';
 const containerConfigs   = new Map();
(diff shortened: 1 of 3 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -377,13 +377,22 @@ const Spotlight = (function () {
...
+			while (isContainer(lastFocusedElement)) {
+				({lastFocusedElement} = getContainerConfig(lastFocusedElement));
+			}
+
+			if (!Spotlight.focus(lastFocusedElement)) {
 				// If the last focused element was previously also disabled (or no longer exists), we
-				// need to set focus somewhere
-				Spotlight.focus();
+				// need to set focus to the next target element within the same container
+				if (!Spotlight.focus(getTargetByContainer(last(getContainersForNode(lastFocusedElement))))) {
```
ryanjduffy: I think `getTargetByContainer` here is redundant as `Spotlight.focus()` will resolve that on its own.

packages/spotlight/src/container.js
```diff
@@ -740,7 +740,8 @@ function getContainerFocusTarget (containerId) {
 			return getContainerFocusTarget(nextId);
 		}
 
-		return element;
+		const {overflow} = getContainerConfig(containerId);
+		return !overflow ? element : contains(getContainerRect(containerId), getRect(element)) && element;
```
ryanjduffy: This ternary is a bit hard to read with the additional `&&` at the end. WDYT of breaking it up to:

```
if (!overflow || contains(getContainerRect(containerId), getRect(element))) {
  return element;
}
```

germboy:
I think those changes are fair. Looks good - thanks!

PR #1957, opened 2018-09-26
ENYO-5524: Add container events for focus changes

ryanjduffy:
### Checklist

There is a significant performance cost of determining if focus will leave the scroller on every key down.

### Resolution
In order to improve scroll on focus performance of scrollers with many spottable children, this PR adds new (private, unsafe for public use!) spotlight container events to notify containers when focus enters, leaves, or fails to leave them.

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -61,6 +61,8 @@ let GlobalConfig = {
 	leaveFor: null,         // {left: <extSelector>, right: <extSelector>, up: <extSelector>, down: <extSelector>}
 	navigableFilter: null,
 	obliqueMultiplier: 5,
+	onBlurContainer: null,
+	onFocusContainer: null,
 	overflow: false,
 	rememberSource: false,
 	restrict: 'self-first', // 'self-first', 'self-only', 'none'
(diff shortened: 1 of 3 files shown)


packages/spotlight/src/container.js
```diff
@@ -948,6 +951,82 @@ function isWithinOverflowContainer (target, containerIds = getContainersForNode(
...
+
+/**
+ * Notifies any containers that focus attempted to move but failed to find a target
+ *
+ * @param {String} direction up/down/left/right
+ * @param {Node} current currently focused element
+ * @param {String[]} currentContainerIds Containers for current
+ * @private
+ */
+function notifyBlurFailContainer (direction, current, currentContainerIds) {
```
germboy: Should we rename this to `notifyBlurContainerFail` in order to match the accompanying `onBlurContainerFail` naming convention used in the config?

packages/spotlight/src/container.js
```diff
@@ -948,6 +951,82 @@ function isWithinOverflowContainer (target, containerIds = getContainersForNode(
...
+				type: 'onBlurContainerFail',
+				direction,
+				target: current
+			});
+		}
+	});
+}
+
+/**
+ * Notifies any affected containers that they one of their children has received focus.
```
germboy: Minor: need to clean up the docs. Looks like an unintentional use of the word "they" above.

germboy:
Can we add a changelog as well?

germboy:
k, that's fair. 👍

Djspaceg:
The only issue so far I have with the ScrollerNative PRs is the naming of the new callbacks. `onFocusContainer`, `onBlurContainer`, and `onBlurContainerFail`. It's my opinion that "Enter" and "Leave" would be more appropriate for these cases. _(Even the comments use these alternate terms.)_ Proposing new terms, respectively: `onEnterContainer`, `onLeaveContainer`, and `onLeaveContainerFail`.

I web terminology, "Focus" and "Blur" are primarily for inputs, while "Enter"/"Leave" is used for containers. To put it another way, you'd "enter" a room and "focus" on the cup on the table. The room is the container, the cup is the object in the container; "you" in this case is Spotlight and it's attention.

I raise this because it's new API that we won't be able to change for quite some time, so I want to make sure it's as future-proof as possible.

ryanjduffy:
Those are good points, @Djspaceg. For background, these events fire when the focus moves into a container or out of a container (but not within a container).

I chose focus/blur because I was concerned that enter/leave imply pointer movement (from `mouseenter`/`mouseleave`) which isn't the trigger for the event. Pointer enter and leave have other consequences in spotlight (setting the active container).

Also, these APIs are intentionally private at this point and are therefore subject to change if necessary. However, I'd much rather have consensus naming now.

webOS101:
I think that makes sense to me.  If the notification is about an _element_ getting focus inside the container, then `onContainerElementFocus` might make sense.  Or maybe, `onFocusInContainer`?

sugardave:
If we want to not step on other meanings for enter/leave, we could use something like `onUnsealContainer`/`onSealContainer`, but I think @Djspaceg's names are fine.

germboy:
I think enter/leave makes sense. Initially it makes me think of our pointer enter/leave methods (like Ryan mentioned), but those are mouse events and are private so developers should not be confused between the two.

Djspaceg:
Thanks, @ryanjduffy!

PR #1962, opened 2018-09-28
Fix marquee measure before CSS load

webOS101:

### Checklist

Marquee would not start in sampler when in 5-way mode (now the default). The QA
sampler shows this off best as it starts on a button that has long text.  Reloading the
page showed off the error.

### Resolution
This removes a measurement forced when focus occurs.  There's no need to remeasure
at this time and we can safely defer it until later as we already remeasure.

### Additional Considerations
The only additional downside is that now when we go to start a marquee we may unecessarily
start a timer, then measure and decided we didn't need to animate at all.

### Links
ENYO-5654

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -16,6 +16,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Touchable` to correctly handle a hold cancelled from an onHold handler
+- `ui/Marquee.MarqueeDecorator` to handle situations where lazily loaded CSS could cause marquee to not start correctly
 
 ## [2.1.4] - 2018-09-17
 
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Pointing out for clarity that in the current code, the calculation doesn't occur on focus when in a scroller _because_ we calculated on resize which occurs when the scroller decides it needs a scroll bar. So, the combination of both changes in this ticket is necessary to achieve the dual benefit of improved render and scrolling perf.

ryanjduffy:
Also, did we verify that marquee start perf is acceptable on the target devices with this change since we’ve moved a calculation to when the animation starts? If it is causing any initial stuttering, we could optionally attach it to the `promoted` state change to decouple it a bit.

viodragon2:
From my investigation, the initial measurement of `calculateMetrics()` runs ~13ms on average. With the proposed change described above, this will be reduced to ~4ms on average. If `setState` is not batched, the total cost of job can go up to 200ms at times, but reducing extra `setState` can improve this. And from my observation, I did not see any stuttering at the start of marquee.

ryanjduffy:
Nice find, @webOS101! Thanks for the PR!

PR #1964, opened 2018-10-01
ENYO-5656: Correctly focus last-focused elements when entering restricted containers

germboy:
### Checklist

#1956 introduced a regression where you could not enter a restricted (overflow/last-focused) container when the last-focused element was outside the view of the overflow container.

### Resolution
The validation we were using was in the wrong location. By returning this value in `getContainerNavigableElements`, we are able to more conveniently/accurately return which elements should gain focus next depending on their container configs. In this case, we check for what priority should be used when entering another container: we should focus last-focused > default > first in-view.

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -641,13 +641,16 @@ function getContainerLastFocusedElement (containerId) {
 	}
 
 	// lastFocusedElement may be a container ID so try to convert it to a node to test navigability
-	const {lastFocusedElement} = config;
+	const {lastFocusedElement, overflow} = config;
 	let node = lastFocusedElement;
 	if (typeof node === 'string') {
 		node = getContainerNode(lastFocusedElement);
 	}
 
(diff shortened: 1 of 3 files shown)


packages/spotlight/src/container.js
```diff
@@ -693,22 +693,42 @@ function getContainerNavigableElements (containerId) {
...
 	if (!next) {
-		next = (!enterDefault && getContainerDefaultElement(containerId)) ||
-				getSpottableDescendants(containerId);
+		const spottables = getSpottableDescendants(containerId);
+
+		// if there isn't a preferred entry on an overflow container, try to find the first element
+		// in view
+		if (overflow) {
+			const containerRect = getContainerRect(containerId);
+			next = spottables.reduce((result, element) => {
```
webOS101: Why not use `array.find()`?  There doesn't seem to be a need to iterate over all elements once you've found the first match.

ryanjduffy: That would work!

PR #1965, opened 2018-10-02
ENYO-5664: Add onHoldEnd event to Touchable

ryanjduffy:
### Checklist

No way to know when a hold ends to support longpress + drag interactions

### Resolution
* Add `onHoldEnd` event to `ui/Touchable`
* Add `holdConfig.global` prop to opt-in to maintain hold when leaving node
* Update Hold and Drag to use `global` to indicate that a blur should not end the gesture


webOS101:
Regarding 2, if we wanted to do that, we'd need to support both for now and deprecate the original event.

ryanjduffy:
@germboy I see your point about the incongruence of `onHold` and `onHoldEnd`. Maybe `onReleaseHold` or `onHoldRelease` would soften that a bit? I don't think `onHoldStart` would convey the same message as `onHold`.

webOS101:
`onUnhold`  :Shipit:

webOS101:
More serious, `Start/End` mirror the touch events so it would be a useful analogy.

ryanjduffy:
It's not a direct parallel though as it stands. I think we may have to rework the hold events to get the semantics the same:

 Now | Then
------|-----
| `onHold` | `onHoldStart` |
| `onHoldPulse` | `onHold` |
| --- | `onHoldEnd` |

germboy:
@ryanjduffy  I think that makes sense, seeing it broken down like that. Ultimately we'd need `onHoldPulse` to become `onHold`, which we can't really do. :shipit:  for now and we can plan on changing the naming convention in the next major release. Approving 1st review, despite the changelog conflict.

PR #1969, opened 2018-10-04
ENYO-5673: Add Top Most Visible Item in the Scroller sample to QA sampler

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/qa/Scroller.js
+++ b/packages/sampler/stories/qa/Scroller.js
@@ -87,7 +87,7 @@ class ScrollerLastFocusedItem extends React.Component {
 			<Scroller
 				focusableScrollbar
 				verticalScrollbar="visible"
-				style={{height: 200}}
+				style={{height: ri.scale(200)}}
 				onScrollStop={this.handleScrollStop}
 			>
 				<Item>Item</Item>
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Scroller.js
```diff
@@ -69,6 +69,36 @@ class ScrollerWithResizable extends React.Component {
 	}
 }
 
+class ScrollerLastFocusedItem extends React.Component {
```
webOS101: This sample doesn't focus last focused, but 'top most visible item in the scroller'.  We should update names and titles to reflect that.

packages/sampler/stories/qa/Scroller.js
```diff
@@ -69,6 +69,36 @@ class ScrollerWithResizable extends React.Component {
...
+		return (
+			<Scroller
+				focusableScrollbar
+				verticalScrollbar="visible"
+				style={{height: ri.scale(200)}}
+				onScrollStop={this.handleScrollStop}
+			>
+				<Item>Item</Item>
+				<Item>Item</Item>
+				<Item>Focus me and press right</Item>
```
webOS101: Suggest moving this item to 2nd place and change text to 'focus me, press right, then select on down arrow'
Change 3rd item to say 'Focus should return here'.  This will clarify that the focused item has scrolled off the viewport and that the new focused item should be the topmost.

PR #1970, opened 2018-10-04
ENYO-5671: Fix TimePicker to not read meridem label when changing value

ryanjduffy:
### Checklist

When changing the value on webOS with audio guidance enabled, the picker would announce the current value plus the meridiem label. This is consistent with the other pickers (hour and minute) but doesn't make much sense given the meridem label.

### Resolution
Remove the `accessibilityHint` property from the `DateComponentPicker` to suppress the extra text read out.


ybsung:
The code looks good to me.
The checklist in the PR description is needed to update.

PR #1971, opened 2018-10-05
ENYO-5667: Scroll to the boundary when the first or last item is focused

viodragon2:
### Checklist

When focusing on the first or the last item in `Scroller`, the margins do not get accounted for the new scroll position. 

### Resolution
When the focused item's position is in somewhat close to `scrollHeight`, then the new position accounts for the `scrollHeight` instead of item's position.

### Additional Considerations
The threshold is set to `24px` which is half the height of Icon - the smallest moonstone component.

### DIFF
--- a/packages/core/util/Job.js
+++ b/packages/core/util/Job.js
@@ -114,6 +114,7 @@ class Job {
 	 */
 	throttleUntil = (timeout, ...args) => {
 		if (!this.id) {
+			this.type = 'timeout';
 			this.run(args);
 			this.id = setTimeout(this.stop, timeout);
 		}
@@ -202,7 +203,7 @@ class Job {
 			this.id = window.requestAnimationFrame(callback);
 		} else {
 			// If requestAnimationFrame is not supported just run the function immediately
-			this.fn(...args);
+			this.run(args);
 		}
 	}
 }
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,8 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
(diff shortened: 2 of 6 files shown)


packages/moonstone/Scroller/Scroller.js
```diff
@@ -82,6 +83,12 @@ class ScrollerBase extends Component {
 		spotlightId: PropTypes.string
 	}
 
+	constructor (props) {
+		super(props);
+
+		this.heightThreshold = ri.scale(24);
```
webOS101: Does this have issues with resizing and V8 snapshot?

viodragon2: I don't think so? And even if so, shouldn't `ui/resolution` handle that case? 
FYI: `moonstone/ContextualPopupDecorator` already uses this pattern. https://github.com/enactjs/enact/blob/master/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js#L246-L248

webOS101: Pretty sure that's wrong, too. :/  Maybe @JayCanuck or @Djspaceg could say?

JayCanuck: I believe `ri` in the constructor will result in an incorrect value in the prerender on non-default resolutions (anything non-1080p), resulting in a markup mismatch error on page load and additional re-rendering.

ryanjduffy: The run-time cost of `ri.scale()` is pretty small; it's only multiplication except for the first invocation after a resolution change. Let's move it into `calculateScrollTop` to avoid any possible issues with resolution changes.

viodragon2:
Does this mean we need to fix `ContextualPopupDecorator` as well?

ryanjduffy:
We should at some point

PR #1977, opened 2018-10-08
PLAT-50585: Fix ui/Marquee to marquee text after locale change

ryanjduffy:
### Checklist

When a locale change causes a visual change (e.g. a different font is used) that introduces the need to marquee, `ui/Marquee` would not start if already triggered.

### Resolution
Adopt `i18n/I18nDecorator.I18nContextDecorator` to invalidate metrics on locale change

### Additional Considerations
Doesn't appear to be any notably run-time performance different but still needs mount performance analysis due to the extra layer added


ryanjduffy:
Travis just reminded me that we had previously avoided the `i18n` dependency but this would require it. We need to make a decision on that as well before moving this ahead.

ryanjduffy:
Fixed up the i18n issue by shifting it to moonstone where it already exists and we were already providing a custom `MarqueeDecorator` to allow for the special `isRtlText` rule for text directionality.

PR #1980, opened 2018-10-10
ENYO-5681: Prevent scrolling on 5-way hold if the last item spotted

viodragon2:
### Checklist

Holding down 5-way directional key towards the end of the list in `moonstone/Scroller`, the last part of scroll animation is jaggy. This only happens if there's no spottable element in the direction. This is due to spotlight container firing `onLeaveContainerFail` for every `keydown` event which will  invoke `moonstone/Scroller.scrollToBoundary()`

### Resolution
Add a flag to mark next target to prevent notifying `onLeaveContainerFail` on hold.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/Scroller` to prevent scroll animation on 5-way hold when it reaches the boundary
+
 ## [2.2.1] - 2018-10-09
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ryanjduffy:
I think the spotlight is behaving as expected in this case but that scroller could be adjusted. For example, if I were to:
* 5-way down to the last item,
* 5-way once more to trigger leave failure,
* mousewheel up just a bit, and
* 5-way down again

My guess is that it wouldn't trigger onContainerBlurFail as designed by this PR but as a result wouldn't scroll to boundary. I think there are a couple other edge cases too.

Could we bail on scroller's scroll attempt if the destination position doesn't change?

ryanjduffy:
Here's a quick cut at that option. What do you think?

https://github.com/enactjs/enact/compare/develop...feature/ENYO-5681-ryan?expand=1

PR #1981, opened 2018-10-11
ENYO-5687: Enact core Job.idle() should stop an existing job before scheduling a new job

sugardave:
### Checklist

It is possible for a race condition to be exposed when multiple calls to a `Job`'s `idle` method are executed when the main thread is overloaded and finally released.  Multiple jobs can end up with the same `id` and if the component that the `Job` is operating on is unmounted at the time the thread is released, a console warning can occur:

```
Warning: Can't call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```

### Resolution
`Job`'s `idle` method is updated to stop any existing jobs before scheduling a new one.

### Links
ENYO-5687

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
+## [Unreleased]
+
+### Changed
+
+- `core/util.Job.idle` to cancel existing scheduled jobs before scheduling a new job and to use `startAfter` when there is no `window` object or no `window.requestIdleCallback` method
+
 ## [2.2.1] - 2018-10-09
 
 No significant changes.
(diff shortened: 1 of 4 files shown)


ryanjduffy:
I discovered that the change to properly fallback to `startAfter` prevented us from testing the fix to call `stop()` so I added a `requestIdleCallback` polyfill to the test so we can verify that the `type === 'idle'` code paths run correctly.

ryanjduffy:
@rundmt + @sugardave - Good reminder to ensure that Travis clears before moving it along. I understand that sometimes we hand off PRs before Travis clears (I've done it too!) but we need to keep an eye out and avoid moving tickets to integration with failing CI.

In this case, the change to fall back to `startAfter` rather than running the job immediately (a good change) caused another test (ExpandableList) which made an incorrect assumption to fail. All fixed up now and ready to go! :shipit:

PR #1982, opened 2018-10-12
Spotlight: Clean up unused spotlight code

germboy:
### Checklist

I removed a couple areas of old/unused code from spotlight.
1) The `SpotlightRootDecorator`'s `navigableFilter` method was removed from the container in #703 but the method remained.

2) The container's `tabIndexIgnoreList` property was a remnant of spatial navigation and was used to ignore specific element types when `makeFocusable()` was called (which would modify the dom to add `tabindex` to elements). We don't explicitly edit the dom to make things focusable in enact (as that's react's job via `Spottable)` and we didn't carry over that method anyway, so it's safe to remove this reference.


Djspaceg:
Is this directly related to Agate work?

germboy:
No, it's just some old code that still remains from an earlier spotlight implementation. We don't need to attach this to agate work @lishammel . I could/should open a new JIRA to attach this to instead.

PR #1985, opened 2018-10-12
ENYO-5683: Fix handling pointer events in Scroller when spotlight disabled

ybsung:
### Checklist

Even though the `data-spotlight-container-disabled` prop in Scroller is true, the Scroller Spotlight container becomes enabled when clicking the item in the Scroller.

### Resolution

When clicking the item in the Scroller, the `stop` in the Scroller was called, then it always made the Scroller Spotlight container enabled by calling `this.childRef.setContainerDisabled(false)`. Before calling it, I just checked if `data-spotlight-container-disabled` is false.

### Additional Considerations

### Links

ENYO-5683

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller` to not enable Spotlight by click when 'data-spotlight-container-disabled' prop is true
+- `moonstone/Scroller` to not enable Spotlight by click, flick, and wheel when 'data-spotlight-container-disabled' prop is true
 - `moonstone/Scroller` stuttering when page up/down key is pressed
 
 ## [2.2.1] - 2018-10-09
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller` to not enable Spotlight by click, flick, and wheel when 'data-spotlight-container-disabled' prop is true
+- `moonstone/Scroller` to not enable Spotlight by click and wheel when 'data-spotlight-container-disabled' prop is true
 - `moonstone/Scroller` stuttering when page up/down key is pressed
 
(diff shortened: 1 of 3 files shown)


germboy:
As I look over this issue, I'm curious if there is any particular reason why `Scrollable`/`ScrollableNative` and the qa-sample knob are using and directly referencing the private `data-spotlight-container-disabled` prop of the container instead of the public container companion prop `spotlightDisabled`? The private props in this case shouldn't generally be designated as a public proptype in a component. Is this something we can update?

The same can also be said for the `data-spotlight-id` proptype of `Scrollable` which should normally be referenced via `spotlightId` - however, that's not specifically an issue in this ticket/PR.

As for the qa-sample: I wonder if this was simply an oversight, seeing as how there's no need to use `data-spotlight-container-disabled` as a knob label. Or perhaps there's a need to clarify the container has been `spotlightDisabled` rather than the individual items being `spotlightDisabled` :shrug:

germboy:
@ybsung Ah, now that I look into it I see why `data-*` props were used instead. `SpotlightContainerDecorator` wraps the `ScrollableBase` before it is exported, so the data props are the only ones accessible at the time of use. That's probably something I'd change later, it's not worth changing for now.

I think the general approach in this PR is ok for now. I do see that ENYO-5688 will affect this in a minor way as well, but since this ticket preserves the `spotlightDisabled` status of the scroller, I would say this PR satisfies the ticket.

germboy:
@ybsung I've added `mousedown` event prevention in `Scrollable`, which should fix the issue that @MikyungKim pointed out. That was happening because the item was gaining focus via the mousedown/click even though the container was `spotlightDisabled`. Then when you pressed 5way up, focus was leaving the item/container, causing a scroll behavior. Can you see if this adequately fixes your issue?

The only issue I see now is a sampler-only issue. If you select the knob for "spotlightDisabled", then try to click on a scroller item, the `ev.preventDefault()` will actually prevent the window from gaining focus at that point. So then if you attempt to 5way to the app-close button, it doesn't gain focus because focus is still in the knobs pane. You need to click somewhere else in the window first, in order to give the window focus. This won't be an issue in the real-world, since this is a sampler-only scenario. I thought you should be aware, as it will affect how you test this.

ybsung:
Hi @germboy
Thank you so much for your help and all information.

PR #1990, opened 2018-10-16
ENYO-5691: Fix rendering blurry text issue

yeram:
### Checklist

Fix rendering blurry text issue when `Notification` is rendered on a non-integer boundary

### Resolution
Using `flex` layout instead of 3d `transform`

### Additional Considerations
It inserts `div` element between `Popup` and `body`

### Links
ENYO-5691
ENYO-3473
https://stackoverflow.com/questions/31109299/css-transform-translate-50-50-makes-texts-blurry

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Notification` to use flex layout for alignment instead of 3d transform
+- `moonstone/Notification` rendering bug that caused text to be blurry
 
 ## [2.2.2] - 2018-10-15
 
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/Notification` to use flex layout for alignment instead of 3d transform
+
 ## [2.2.2] - 2018-10-15
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


ryanjduffy:
@yeram This is an interesting bug and I appreciate the fix. When I was testing it, I noticed that it introduced the possibility for the top of the notification to be clipped. In `develop`, it will overflow out the bottom of the view but the top remains visible. Neither of these are particularly good but perhaps lie outside the bounds (:rimshot:) of the use cases for the component.

I'd also like to propose a possible alternative that introduces a smaller diff which is useful at this point in the lifecycle as it reduces the opportunities for bugs. I found that promoting the `.body` section to the GPU also eliminated the text issue on Chrome. Wondering if you see the same on the TVs.

```less
	.body {
		.moon-notification-content();
		// promote this layer to ensure that text is rendered by the GPU as well which avoids some
		// odd antialiasing when the notification is positioned in subpixel land
		.enact-composite();

		:global(.enact-locale-ja) &,
		:global(.enact-locale-zh) & {
			overflow-wrap: normal;
			word-break: normal;
		}
	}
```

yeram:
Hi @ryanjduffy , thanks for the review!
Unfortunately, promoting `body` section to GPU seems doesn't work on TV.

For the overflow issue, lines of text should be under 7 per spec, so I think there should not be the use cases. Legacy component also can overflow but I think it's good timing to prevent overflow on component level. I'm going to update PR for it.

ryanjduffy (approved):
Thanks, @yeram! :+1: Looks good to me!

Reminder to keep the change log entries consumer-centric vs implementation-centric. In other words, tell the consumer _what_ was fixed versus _how_ we fixed it. I updated the change log here in that vein.

yeram:
Got it! :+1:

Djspaceg:
I'm merging this, but only with reluctance. It uses a fixed-size invisible container to center the notification and adds one additional layer of DOM. Due to the dimensions of this container, the optional animation has reduced in visual "quality" by having a greater distance to travel during the same transition duration.
invisible centering container

That being said, it does a pretty good job of meeting the variable requirements of Notification, maintaining its expand-from-the-center characteristic, both horizontally and vertically. The transition animation is secondary to the blur issue, in that it's optional and not widely known about.

Good work, and hopefully this doesn't cause any regressions!

PR #1991, opened 2018-10-17
ENYO-5692: Fix VideoPlayer to clamp knob position

xbinary:
### Checklist

When pointer is located on the left side of the progress bar,
Knob is located to minimum X position

### Resolution
clamp `clientX` of pointer


packages/moonstone/VideoPlayer/MediaSliderDecorator.js
```diff
@@ -125,11 +126,10 @@ const MediaSliderDecorator = hoc((config, Wrapped) => {	// eslint-disable-line n
 
 		move (clientX) {
 			this.setState((state) => {
-				if (clientX >= state.minX && clientX <= state.maxX) {
-					return {
-						x: calcProportion(state.minX, state.maxX, clientX)
-					};
-				}
+				const value = clamp(state.minX, state.maxX, clientX);
+				return {
```
ryanjduffy: I think we _could_ optimize this to return `null` when `x` isn't changing but I doubt it'll make a material difference given that the slider will already be updating frequently during playback. Noting for future consideration only.

PR #1992, opened 2018-10-17
ENYO-5688: Prevent Spottable focus when it should not be focusable

germboy:
### Checklist

`Spottable` components within a `spotlightDisabled` `SpotlightContainerDecorator` will still gain spotlight highlights when directly clicking them via mouse/pointer. This is because Spottables are gaining focus when they should not have focus. This can only happen in pointer-mode, as 5-way correctly prevents these controls from gaining focus.

### Resolution
We can prevent unexpected focus to Spottables in these cases by calling `ev.preventDefault()` on the `mousedown` event when the spottable should not be focusable. This requires the additional step of updating the public `Spotlight.isSpottable()` method, which returns an incorrect value (simply matching a selector string against the element). 

### Additional Considerations
1) I've removed an additional piece of logic in the `handleFocus` method of `Spottable` originally added in #1607 . This can safely be removed in this PR, since its purpose was to prevent focus and forwarding the `onClick` event on disabled components. This check and prevention would now be handled by the `preventDefault` action in `onMouseDown`.

2) I've updated the qa-sample for spotlight > kitchen sink to handle the 2 different `spotlightDisabled` options (container vs individual spottables) so that we can test for the 2 behaviors. In doing so, I noticed some components that need to adjust their styles based on being `spotlightDisabled` in either way or `spotlightMuted` (`Slider`, `IncrementSlider`). This is probably better done in a separate ticket.


ryanjduffy:
Thanks for the good analysis! I want to clarify one point that might cause us to adjust the proposed solution. Should the purpose of `spotlightDisabled` be to prevent interactivity or to prevent focus? It's a thin line, I acknowledge. I think that we've historically used `disabled` to prevent interactivity and `spotlightDisabled` to prevent focus.

If we want to limit `spotlightDisabled` to only preventing focus, perhaps trapping `focus` in spotlight container during the capture phase would be an alternative. `stopPropagation()` would prevent `Spottable` from receiving the focus event and `ev.target.blur()` could clear the focus.

ryanjduffy:
Here's a commit you can cherry pick if you want to go that route: 470d045c42446c9bfc36eefff3f52c14369b98d1

germboy:
That's actually inline with the initial approach I took when fixing this. However in that solution (like in your commit), I grew a little iffy on how the spottables capture focus (even though they don't handle the focus event) only to immediately be explicitly blurred elsewhere 😐 , which is more preventing the effect/lifetime of focus rather than preventing focus itself. So that's why I opted for handling this in `Spottable`. I agree, it's a thin line - and may make sense handling in container despite my concern above . Do you have thoughts or concerns on the focus/blur vs focus prevention behavior aspect?

Additional thoughts:
1) At the least, I do think we need to normalize the behavior between a `spotlightDisabled` `Spottable` and a `Spottable` in a `spotlightDisabled` `SpotlightContainerDecorator`. After this fix (as seen in either solution) I have noticed a difference in how a `spotlightDisabled` `Spottable` will still emit an `onFocus` event when being clicked. In fact, this even causes 5-way to be trapped (and hidden) within the `Spottable` if there are no other valid `Spottable` components in the direction pressed. With that in mind, we should be preventing focus for both methods of using `spotlightDisabled` (container vs spottable).

2) Even though it is not used in your commit, the `isSpottable()` method probably still needs to be updated. It's fairly useless in its current state, as it just matches a selector to tell if the element is spottable without taking containers into consideration.

germboy:
This ticket is revealing more inconsistencies between `spotlightDisabled` components and components in `spotlightDisabled` containers that I want to clarify for the review process.

a) Focused components that become `spotlightDisabled`: The focused styling is removed, giving the appearance that the component no longer has focus. In reality, it still has focus and 5way is sort of "trapped" in that at this point any 5way directional press should set focus somewhere. However, if there's no eligible next target in the direction pressed, spotlight will not change focus from its now-hidden location - giving the appearance of focus being lost.

b) Focused components in a container that becomes `spotlightDisabled`: The focus styling remains and does not change (as opposed to becoming `disabled` where the spotlight highlight reduces opacity).

I think (b) is likely what we want across the board. We probably don't want to blur or explicitly change focus in that case. This was discovered when determining how/when we should prevent the blur event. If we're explicitly calling blur on focus attempts on `spotlightDisabled` components and containers, we need to prevent the blur event as well - but blur prevention would need to be conditional on if the blur attempt was done via an explicit call due to a `spotlightDisabled` focus attempt.

Also, it appears that `spotlightDisabled` marquee-able controls (in either configuration) do not marquee when hovered with the pointer, unlike `disabled` controls. Is this expected? They will, however, continue to marquee if they become `spotlightDisabled` while focused.

I will go the route of (b) above, as that seems to make the most sense - unless there are conflicting opinions which require more discussion. I would like more opinions on the marquee behavior though.

germboy:
The latest commits fix the major pain points of this issue. The proposed fix normalizes the behavior between `spotlightDisabled` components and components in `spotlightDisabled` containers.

Focused components that become `spotlightDisabled` will retain their focus until focus is set elsewhere (like we do with `disabled` components). I feel we shouldn't forceably remove focus in these cases (or most cases in general, including completely disabling), but this does have the consequence of there being no visual changes (unlike disabling) informing the user that it has become `spotlightDisabled`.

Would you expect a marquee-able component (`Item`, `Button`) to marquee on hover if it were `spotlightDisabled` (like it does when `disabled`)? I'd think so, but I'm not certain the best way to implement that. `ui/MarqueeDecorator` dictates the behavior, so outside of allowing a very specific custom prop config or creating a `moonstone/MarqueeDecorator` (both of which are :yuck:), I'm not sure how to account for `spotlightDisabled`. Thoughts?

PR #1993, opened 2018-10-17
ENYTO-5665: Job needs cleanup

sugardave:
### Checklist

This patch cleans up `Job` a little to add a type for throttled jobs and to ensure that all instances of the job execution are performed by `run`.

### Links
ENYO-5665

### DIFF
--- a/packages/core/util/Job.js
+++ b/packages/core/util/Job.js
@@ -114,7 +114,7 @@ class Job {
 	 */
 	throttleUntil = (timeout, ...args) => {
 		if (!this.id) {
-			this.type = 'throttle';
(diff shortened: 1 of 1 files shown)


packages/core/util/Job.js
```diff
@@ -114,6 +114,7 @@ class Job {
 	 */
 	throttleUntil = (timeout, ...args) => {
 		if (!this.id) {
+			this.type = 'throttle';
```
ryanjduffy: Let's use `timeout` instead of `throttle` since the primary purpose is to know what kind of "stop" to use. I'll make that change and move this along.

PR #1994, opened 2018-10-18
ENYO-5691: Implement alternate centering to avoid GPU layer "blurring"

Djspaceg:
### Checklist

Fix rendering blurry text issue when `Notification` is rendered on a non-integer boundary

### Resolution
Using `flex` layout instead of 3d `transform`

### Links
ENYO-5691
ENYO-3473
https://stackoverflow.com/questions/31109299/css-transform-translate-50-50-makes-texts-blurry

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/VideoPlayer` to position the slider knob correctly when beyond the left or right edge of the slider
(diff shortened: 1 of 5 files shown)


ryanjduffy:
Noting that this will be an API change due to the addition of `publicClassNames`

yeram:
- These lines should be removed :
 https://github.com/enactjs/enact/blob/5557f4c3f186d1331a4179760588cffe27d01a45/packages/moonstone/Notification/Notification.less#L77-L80
- Prevent overflow as suggested from previous PR : https://github.com/enactjs/enact/pull/1990/commits/41d31616a1fbbfe84ed9a2067b2f360b62a95fcd

Otherwise Looks good to me! :+1:

PR #1995, opened 2018-10-18
ENYO-5695: Fix spotlight target selection of clipped elements

ryanjduffy:
### Checklist

The center calculation was incorrect when an element was partially visible in an overflow container causing them to "appear" to be closer to spotlight in certain circumstances.

### Resolution
Fix the calculation to offset the center from the the left/top position


sugardave:
The code change :LGTM:

However, I cannot get the sample in the linked issue to run properly in my browser or on my TV.  It would be good to have that verification as well.

PR #2002, opened 2018-10-25
ENYO-5666: Cache client size, scroll size, and scroll thumb size ratio in VirtualList and Scroller

ybsung:
### Checklist

When moving scroll thumb, `client size`, `scroll size`, and `scroll thumb size ratio` were recalculated even thought they did not change.

### Resolution

- Cache `client size`, `scroll size`, and `scroll thumb size ratio` values.
- Recalculate them when updated.

### Additional Considerations

The scroll button disabled state was changed by `setState` which enforces react to rerender. This was a costly operation. @viodragon2 fixed this issue by calling `setState` when the IconButton status changed instead of calling it every time when updating scroll thumb position.

### Links

ENYO-5666

### DIFF
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -159,17 +159,17 @@ class ScrollButtons extends Component {
 			shouldDisableNextButton = maxPos - currentPos <= 1,
 			updatePrevButton = (prevButtonDisabled !== shouldDisablePrevButton),
 			updateNextButton = (nextButtonDisabled !== shouldDisableNextButton);
-		let nextState = null;
-
-		if (updatePrevButton && updateNextButton) {
-			nextState = {prevButtonDisabled: shouldDisablePrevButton, nextButtonDisabled: shouldDisableNextButton};
-		} else if (updatePrevButton) {
-			nextState = {prevButtonDisabled: shouldDisablePrevButton};
-		} else if (updateNextButton) {
-			nextState = {nextButtonDisabled: shouldDisableNextButton};
-		}
 
 		if (updatePrevButton || updateNextButton) {
+			let nextState = {};
+
+			if (updatePrevButton) {
+				nextState.prevButtonDisabled = shouldDisablePrevButton;
+			}
+			if (updateNextButton) {
+				nextState.nextButtonDisabled = shouldDisableNextButton;
+			}
+
 			this.setState(nextState);
 		}
 	}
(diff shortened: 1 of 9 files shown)


viodragon2:
The scroll thumb update isn't actually a bottleneck when scrolling. The `update` function in `ui/ScrollBar` and `setCSSVariable` takes about 1~3 ms which isn't a problem. Even with the proposed change, there's no improvement in performance. 

However, the perf loss comes from the render function ran in `ScrollButtons`.  Nomally this operation takes up ~5ms
screen shot 2018-10-25 at 6 19 59 pm

But sometimes it takes long time as well.
screen shot 2018-10-25 at 6 22 35 pm


If this `setState` call is solely for setting the `disabled` state which we can calculate in advanced, I was thinking we could prevent. See https://github.com/enactjs/enact/tree/feature/ENYO-5666-stephen

screen shot 2018-10-25 at 6 21 17 pm

This is essentially checking prop change in `shouldComponentUpdate` and probably be done at `moonstone/ScrollBar`, but you get the idea. Let me know what you think.

ybsung:
Hi @viodragon2 
Good catch!! I did not know the `setState` ( which returns nothing  not to re-render ) 's cost is high. Thank you so much for your effect. I'd like to cherry-pick your commit.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -150,26 +150,28 @@ class ScrollButtons extends Component {
...
 			/* If a scroll size or a client size is not integer,
 			   browsers's max scroll position could be smaller than maxPos by 1 pixel.*/
-			shouldDisableNextButton = maxPos - currentPos <= 1;
+			shouldDisableNextButton = maxPos - currentPos <= 1,
+			updatePrevButton = (prevButtonDisabled !== shouldDisablePrevButton),
+			updateNextButton = (nextButtonDisabled !== shouldDisableNextButton);
+		let nextState = null;
+
+		if (updatePrevButton && updateNextButton) {
+			nextState = {prevButtonDisabled: shouldDisablePrevButton, nextButtonDisabled: shouldDisableNextButton};
```
0x64: If we _let_ _nextState_ a simple empty object _{}_, it looks to possible to reduce _if_ clauses.

0x64:
LGTM!

ybsung:
Hi @viodragon2 
I agree with your opinions. At the first time, I expected that I could find any ways to improve VirtualList performance with a few code changes. But as you mentioned, this PR has no perf impact and increases complexity. Let's re-visit this issue later after restructuring components related with VirtualList and Scroller. I am sorry that you might spend your time much to help me. I'd like to close this PR.

I have not have a chance to use `memorize`. Next time I'll consider it when applicable. Thanks for the tip.

Thanks a lot.

PR #2003, opened 2018-10-29
ENYO-5703: Fix Marquee component to render font based on locale correctly

yeram:
### Checklist

Make `Marquee` as a styled component.

### Resolution
Fix `Marquee` to use `moon-font()` mixin.

### Additional Considerations
Some of moonstone components are using `Marquee` and handling font styling by themselves.
To minimize impact of change, I modified them to use `MarqueeDecorator('div')` instead of `Marquee`.

Here is the list of those components
- `moonstone/GridListImageItem`
- `moonstone/LabeledItem`
- `moonstone/Spinner`
- `moonstone/VideoPlayer/MediaTitle`
- `moonstone/internal/Picker/PickerItem`

Also note that `moonstone/MediaOverlay`  is using `Marquee` but has no font styling for it.
I leave it as-is to be fixed by styled `Marquee`.

### Links
ENYO-5703

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Marquee`, `moonstone/MediaOverlay` to display localed based font
 - `moonstone/DayPicker` separator character used between selected days in the label in fa-IR locale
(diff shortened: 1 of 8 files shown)


webOS101:
This seems to be very likely to cause regressions for applications.  Is there a narrower approach we could apply?

yeram:
IMO, it's a most narrow approach on framework side.
To fix it by application side, framework provides generic less mixin or new styled `Marquee` component and applications use it case by case.

Djspaceg:
The framework _does_ provide a styled `Marquee`, it's called `Item`.

ryanjduffy:
Closing in favor of #2007

PR #2004, opened 2018-11-01
PLAT-67660: Pass childProps through VirtualList

ybsung:
### Checklist

There is no way to pass custom parameters from app to VirtualList's itemRenderer.

### Resolution

Define `childProps` prop in VirtualList and spread it as arguments of the `itemRenderer`.

### Links

PLAT-67660

### DIFF
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -159,7 +159,7 @@ const VirtualListBaseFactory = (type) => {
 			getComponentProps: PropTypes.func,
 
 			/**
-			 * It is the object which is passed as arguments of `itemsRenderer` prop function.
+			 * Additional props included in the object passed to the `itemsRenderer` callback.
 			 *
 			 * @type {Object}
 			 * @private
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -27,6 +27,32 @@ const gridListItemSizeShape = PropTypes.shape({
 	minWidth: PropTypes.number.isRequired
 });
 
+const ListContext = React.createContext();
+
+class ListProvider extends Component {
+	static propTypes = /** @lends ui/VirtualList.ListProvider.prototype */ {
+		/**
+		 * It is the object which is passed to items.
+		 *
+		 * @type {Object}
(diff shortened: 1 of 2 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -269,7 +277,7 @@ const VirtualListBaseFactory = (type) => {
 		}
 
 		componentWillUpdate (nextProps, nextState) {
-			if (this.state.firstIndex === nextState.firstIndex) {
+			if (this.state.firstIndex === nextState.firstIndex || this.props.itemProps && this.props.itemProps !== nextProps.itemProps) {
```
0x64: Umm. Do we need to consider the case that the prop is _undefined_ at first then assigned later for updating?

ybsung: No. I don't think so.

0x64:
LGTM

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -158,6 +158,14 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			getComponentProps: PropTypes.func,
 
+			/**
+			 * It is the object which is passed as arguments of `itemsRenderer` prop function.
+			 *
+			 * @type {Object}
+			 * @private
+			 */
+			itemProps: PropTypes.object,
```
ryanjduffy: I'd like to use `childProps` rather than `itemProps`. It's a change I've wanted to make to `Repeater` and `Group` for while after we decided to standardize on `childComponent`. It avoids using `child` and `item` to refer to the same thing.

Also, I don't think you need to mark this as private. It's a stable concept that we've already demonstrated in `Repeater`.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -158,6 +158,14 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			getComponentProps: PropTypes.func,
 
+			/**
+			 * It is the object which is passed as arguments of `itemsRenderer` prop function.
```
ryanjduffy:
```suggestion
			 * Additional props included in the object passed to the `itemsRenderer` callback.
```

PR #2008, opened 2018-11-06
Feature/enyo 5715 lucie

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links


LucieRoy:
There are 2 issues with this PR:
1. we had to add style={{height: '100%'}} to ExpandableList else the list would not expand.
2. The first time we click on button '10', the list will not expand. Need to click twice.
In the original sample loaded alone, these issues do not happen.

PR #2009, opened 2018-11-06
PLAT-67808: Async ilib workaround for moonstone

viodragon2:
### Checklist

`moonstone` needs to load a resource bundle from a path that cannot be shared with `resBundle` that `i18n/I18nDecorator` uses. This lead us to create an internal `$L` function but it will only load resources synchronously.

### Resolution
~Added a `MoonstoneI18NDecorator` and `internal/i18n/resBudnle` to handle asynchronous resource fetching and translating.~

EDIT: Update description (Feb/5/2019)
The original proposal is reverted and https://github.com/enactjs/enact/pull/2014 was merged

> An alternative based on [#2009 (comment)](https://github.com/enactjs/enact/pull/2009#issuecomment-436819337) that adds API to hook into the i18n resource loading process. I also added a config prop to the I18nContextDecorator to get access to the `loaded` context value which was already passed but unavailable via the HOC.

> It also reverts some of the decomposition of res bundle code. That might not make sense as it couples $L and the moonstone string res bundle which we probably will want separated out later when we fully support async in Moonstone.

### Additional Considerations
~- `$L` lives outside the react component lifecycle and will not be updated upon resource loading. This can result in failure in translation if the component renders before the resources load.~
~- Note the capitalization in `N` which differs from `i18n/I18nDecorator`~


packages/moonstone/internal/$L/$L.js
```diff
@@ -1,58 +1,26 @@
...
-	}
-	return isObject ? rb.getString(str.value, str.key).toString() : rb.getString(str).toString();
+	return String(toIString(str));
 }
 
 export default $L;
 export {
-	$L,
-	getResBundle,
-	clearResBundle
```
JayCanuck: Just one note that moving/renaming `clearResBundle` from the module will require a PR to `@enact/dev-utils`/`cli` too (see https://github.com/enactjs/dev-utils/blob/master/plugins/SnapshotPlugin/snapshot-helper.js#L41)

ryanjduffy:
@viodragon2 - Do you think there a need for a standardized hook in `i18n` for userspace resource loading? For example, an app could make locale-dependent external service calls which are synchronized with the global locale change infrastructure in `I18nDecorator`. This same infrastructure could support preloading ilib resources as is required to shim async support into moonstone.

packages/moonstone/internal/$L/$L.js
viodragon2: Added `clearResBundle` in `internal/$L`. FYI: 
```	delete ResBundle.strings;
	delete ResBundle.sysres;```
seems unnecessary, but added to keep the parity with how `i18n/src/resBundle.clearResBundle` does clearing.

**Comment by viodragon2** [comment]
Doesn't the app already handle `onlocalechange` event to run locale-dependent logics? I can't think of a good reason why an app would need to know when the "resources" are available as the apps don't really manage resources on their source codes. It does sound like a good idea to support hooks in asynchronous loading, but I am not 100% sure whether `<Suspense>` already solves that problem or not.

**Comment by ryanjduffy** [comment]
I think the primary possibility this enables is a consistent means to preload ilib resources before the app renders. So as I look ahead to how to balance customer demands for synchronous retrieval for native-like experiences on a web platform that favors asynchronous methods, one option is to retrieve some/all external resources in batch before render so that all subsequent requests need not wait.

One concern I have with the proposed solution here is that it only addresses string resolution and, as you noted, doesn't provide a means for `$L` usage to be refreshed upon update, particularly when blocked by `shouldComponentUpdate`.

By providing the hook, the app (browser, in this case) can inject the string resolution code (and any other ilib resources they discover they need) and Moonstone and i18n can remain ignorant of those requirements.

Does that line of thinking make sense?

**Comment by ryanjduffy** [comment]
I'm not educated enough on `Suspense` yet to say! You're right that apps wishing to using async ilib right now should use `Text` over `$L`. The hope with this PR is to provide a workaround to support the existing `$L` usage in moonstone which we can't easily change at this point.

Primarily, I want to minimize the change we make to moonstone to support this under the assumption that we'll make broader changes later when async is fully supported. In that light, we could use `Suspense` when `sync: false` in `MoonstoneDecorator` but I'd also be okay holding off on that and allowing the app to manage that deferred rendering for the time being.

File: packages/i18n/I18nDecorator/I18nDecorator.js [hunk 2]
```diff
@@ -50,11 +50,44 @@ const defaultConfig = {
...
+	 * if loading a resource has side effects that should only be applied once all loading has
+	 * completed.
+	 *
+	 * ```
+	 * resources: [
+	 *   {resource: (options) => { ... fetch ... }, onLoad: (res) => { ... apply side effect ... }}
+	 * ]
+	 * ```
+	 *
+	 * @type {Array<Function|Object>}
```
webOS101: I'm not sure the docs deal well with this.  Could we use `{Function[]|Object[]}`?

packages/i18n/I18nDecorator/I18nDecorator.js
```diff
@@ -81,6 +114,18 @@ const defaultConfig = {
...
+	const extResources = Array.isArray(config.resources) ? config.resources.map(res => {
+		if (!res) return;
+
+		const fn = res.resource || res;
+		const onLoad = res.onLoad;
+
+		if (typeof fn !== 'function') return;
+
+		return {resource: fn, onLoad};
+	}).filter(Boolean) : [];
```
webOS101: Food for thought question on these cases where we `map().filter(Boolean)`.  Would these be clearer/easier with `reduce()` instead?  I don't know but it seems like they're pretty similar.

packages/i18n/I18nDecorator/I18nDecorator.js
```diff
@@ -259,7 +313,7 @@ const contextDefaultConfig = {
 };
 
 const I18nContextDecorator = hoc(contextDefaultConfig, (config, Wrapped) => {
-	const {localeProp, rtlProp, updateLocaleProp} = config;
+	const {loadedProp, localeProp, rtlProp, updateLocaleProp} = config;
```
webOS101: Should we add this to `contextDefaultConfig`?  This doesn't appear to be used anywhere (is it?)... is it something we should document for future use or is it not needed for this ticket?

packages/moonstone/internal/$L/$L.js
```diff
@@ -1,27 +1,48 @@
...
-		resBundle = new ResBundle({
+	return resBundle;
+}
+
+/**
+ * Creates a new ilib.ResBundle for string translation
+ *
+ * @param  {ilib.Locale} locale Locale for ResBundle
+ *
+ * @returns {Promise|ResBundle} Resolves with a new ilib.ResBundle
```
webOS101: Need to validate this will be valid with the types generator.

webOS101:
Is the description on this ticket still accurate?  Should we remove the reference to `MoonstoneI18NDecorator`?

packages/i18n/I18nDecorator/I18nDecorator.js
webOS101: @ryanjduffy Pointed out this would be wrong to make that substitution.

packages/moonstone/internal/$L/$L.js
webOS101: This is private, so don't need to, actually.

PR #2010, opened 2018-11-07
ENYO-5697: Added .moon-font-size mixin

Djspaceg:
Added for app convenience, not necessarily for framework use.

### Checklist

Occasionally app developers have a need to set the font size of *their* components, and it's awkward to set both latin and non-latin manually.

### Resolution
Added a mixin to facilitate this need.

### Note
The changelog was intentionally _not_ updated since this is more of a "special request" rather than a new official API feature.

### DIFF
--- a/packages/moonstone/styles/text.less
+++ b/packages/moonstone/styles/text.less
@@ -41,6 +41,22 @@
 	}, @target);
 }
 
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Note: Rebased this on `release/2.2.6` as well.

PR #2011, opened 2018-11-08
ENYO-5720: Storybook: cannot set local languages

Sangwook-Lee:
Fixed MoonstoneEnvironment.js to use window locale when selected local language.

### Checklist

We should check on TV that all the text inside Enact framework has been translated. The QE team uses Enact Sampler to check the strings of the input component.
However, after they change the language, they can not see the translated string in the Enact sampler.

### Resolution
When `local` is selected, Enact sampler passes locale value to MoonstoneDecorator/I18nDecorator as empty string `''`.But, string is not updated to current locale language.
So, I fixed some codes to use `window.navigator.language` when locale value is `''`.

### Links
ENYO-5720

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -156,7 +156,7 @@ const StorybookDecorator = (story, config) => {
 			className={classnames(classes)}
 			title={`${config.kind} ${config.story}`.trim()}
 			description={config.description}
-			locale={select('locale', locales, Config, getPropFromURL('locale'))}
+			locale={select('locale', locales, Config, getPropFromURL('locale') || window.navigator.language)}
 			textSize={boolean('large text', Config, (getPropFromURL('large text') === 'true')) ? 'large' : 'normal'}
 			highContrast={boolean('high contrast', Config, (getPropFromURL('high contrast') === 'true'))}
 			style={backgroundLabelMap[select('background', backgroundLabels, Config, getPropFromURL('background'))]}
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -55,7 +55,7 @@ const MoonstoneFullscreen = MoonstoneDecorator({overlay: false}, FullscreenBase)
 
 // NOTE: Locales taken from strawman. Might need to add more in the future.
 const locales = {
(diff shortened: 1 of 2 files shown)


packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -156,7 +156,7 @@ const StorybookDecorator = (story, config) => {
 			className={classnames(classes)}
 			title={`${config.kind} ${config.story}`.trim()}
 			description={config.description}
-			locale={select('locale', locales, Config, getPropFromURL('locale'))}
+			locale={select('locale', locales, Config, getPropFromURL('locale') || window.navigator.language)}
```
Djspaceg: We may need to guard against the use of `window` in case it doesn't exist. I'm not sure though.

ryanjduffy:
@Sangwook-Lee Thanks for the report and the proposed fix! I've looked into this a bit and discovered a deeper bug that is actually (just recently) fixed in `develop` but would like to include in 2.2.x as well to avoid other possible issues.

The root cause appears to be incorrect logic in `componentWillReceiveProps`. Instead of checking for a prop change, we were just checking for a truthy value which failed in the case of using `'local'` in sampler. In practice, we need not guard against a falsey value which is handled correctly by `updateLocale` currently.

https://github.com/enactjs/enact/blob/release/2.2.6/packages/i18n/I18nDecorator/I18nDecorator.js#L101

What do you think of updating this PR to perform and equality check on `locale` and verify it fixes the issue on the TV?

Sangwook-Lee:
@ryanjduffy , Thanks for the review.
I thought that modifying sampler was safer than modifying `I18nDecorator`.
So, If we have to fix `I18nDecorator`, we can pend this fix.

As you said, there's root cause in I18nDecorator. Also, when we select `local` knob, sampler passes default locale `en-US` to I18nDecorator. So, I think we should specify `local` value not empty string `''`.

Moreover, we should call `ilib.getLocale()` to update locale to `PalmSystem.locale`. And then it can be updated through the `updateLocale()`.

ryanjduffy:
Sampler passes the empty string which invokes the default locale logic in ilib which produces the same effect as passing the navigator language value.

Sangwook-Lee:
It seems to pass en-US when local is selected by below.

https://github.com/enactjs/enact/blob/e970a9e27fa6fbd3919154b2d07d19b8d355895d/packages/sampler/src/enact-knobs/select.js#L36

Sangwook-Lee:
@ryanjduffy , I've updated PR according to your opinion.

PR #2013, opened 2018-11-09
ENYO-5718: Add line-height rule for non-latin fonts

viodragon2:
### Checklist

LG Display font, which is used for non-latin characters, has a line-height value of `1` which truncates tall glyphs in unstyled components. (e.g. `Marquee`, `<div>`, etc.)

### Resolution
Add line-height rule for non-latin fonts. The value is set to `1.5 em` to accommodate the tall glyphs.

### Additional Considerations
This problem may have been hidden all along until we made a change in #2007 to use correct rules for the non-latin locale. 

CAUTION: IT may cause a **layout change** and thorough tests are required for the existing apps.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
+### Fixed
+
+- `moonstone` non-latin fonts to display in correct line-height
+
 ### Added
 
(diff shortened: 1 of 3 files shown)


Djspaceg:
I get this feeling that we should also assign the "latin" `line-height` too, so the rulesets stay in sync. Curious to hear what deeper testing reveals though.

viodragon2:
I thought about it adding it, but Museo Sans font-family correctly sets line-height to 1.2 already and I wasn't sure what the "new" value should be. Should it also be `1.5 em`?

PR #2015, opened 2018-11-12
PLAT-64690: Disable scroll animation with 5-way navigation

yeram:
### Checklist

Fix stuttering issue by disabling animation with 5-way navigation

### Resolution
Allow `animate` as a prop and use it to branch whether it animate or not

### Additional Considerations
Not pass to `ui/VirtualListBase` at this time since it doesn't have any logic relate to animation

### Links
PLAT-64690

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,12 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/VideoPlayer` to disable pointer mode when hiding media controls via 5-way
-- `moonstone/VirtualList` and `moonstone/Scrollable` to not to animate with 5-way navigation by default
-- `moonstone/VirtualList.VirtualGridList` and `moonstone/VirtualList.VirtualList` property `childProps` to support additional props included in the object passed to the `itemsRenderer` callback
-
-### Changed
-
-- `moonstone/VirtualList` and `moonstone/Scroller` to scroll without animation when navigating via 5-way
+- `moonstone/VirtualList` and `moonstone/Scroller` to not to animate with 5-way navigation by default
 
 ## [2.2.5] - 2018-11-05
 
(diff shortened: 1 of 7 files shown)


packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -755,7 +755,7 @@ class ScrollableBaseNative extends Component {
 
 		this.lastInputType = 'pageKey';
 
-		this.scrollToAccumulatedTarget(pageDistance, canScrollVertically, this.props.overscrollEffectOn.pageKey);
+		this.scrollToAccumulatedTarget(pageDistance, canScrollVertically, this.props.overscrollEffectOn.pageKey, this.props.animate);
```
yeram: I saw animation while paging feature validation. I pass `animate` here to disable animation explicitly.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -572,7 +581,7 @@ const VirtualListBaseFactory = (type) => {
 							// So we find nextIndex only when currentExtent is different from candidateExtent.
 							if (currentExtent !== candidateExtent) {
 								nextIndex = this.findNearestSpottableItemInExtent(currentIndex, candidateExtent);
-								animate = (wrap === true);
+								animate = this.props.animate && (wrap === true);
```
yeram: Not sure what this line means. But I added checking `animate` to disable it anyway.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -452,7 +461,7 @@ class ScrollableBase extends Component {
 			ev.preventDefault();
 		}
 
-		this.animateOnFocus = true;
+		this.animateOnFocus = this.props.animate;
```
yeram: There is no initial value of `animateOnFocus`. I guess it's safer to assign value, but leave it as-is for now.

ryanjduffy:
In order to isolate the behavior change to moonstone, I've updated the `ui` components to use `noAnimation` instead of `animate` so that the default behavior is still animated scrolling for non-moonstone targets. I've left the prop as `private` because I expect that we'll attempt to improve the performance so that we can re-enable animation by default without altering public APIs.

packages/moonstone/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/VideoPlayer` to disable pointer mode when hiding media controls via 5-way
+- `moonstone/VirtualList` and `moonstone/Scrollable` to not to animate with 5-way navigation by default
```
webOS101: `moonstone/Scrollable` is private, so we don't discuss it in the CHANGELOG.

PR #2016, opened 2018-11-12
PLAT-69511: Add line-height in the base font mixin

Djspaceg:
### Checklist

* [?] This is an API breaking change

Characters are vertically clipped when the font doesn't specify proper metrics.

### Resolution
Unilaterally assign basic, low-level, easy to override, line-height measurements based on our existing definitions.


Djspaceg:
VideoPlayer does actually have a non-Latin line height definition a few lines below, so it should not be a problem.
https://github.com/enactjs/enact/blob/c0f978cc66e442146670e33e0e254d2a5fd6d86b/packages/moonstone/VideoPlayer/VideoPlayer.less#L110

The other two _shouldn't_ be a problem, but still worth verifying that things aren't terribly broken.

ryanjduffy:
That would be great. Should we remove those `line-height` rules in this PR since they have no effect and would only confuse future us?

Djspaceg:
That is a negative, Red Leader. non-Latin header line-height is 1.5em while our standard non-Latin is 1.7em (to account for things like Thai tall characters). Header line-height we can more tightly assign than generalized body text because we're (the Framework is) responsible for the layout and possible region overlap of our Header component.

ryanjduffy:
:+1: I was only thinking of non-latin but the `line-height` rules would still take effect in latin locales.

PR #2017, opened 2018-11-12
ENYO-5710: Fix Scroller to better handle scrolling due to focus

ryanjduffy:
### Checklist

Changing the size of a container within a Scroller causing some unpredictable scrolling behavior.

### Resolution
The logic for scrolling on focus had become overly complex due to changing requirements over time. I've restated the requirements and reimplemented the method in light of those requirements.

### Additional Considerations
My goal for this PR was to standardize behaviors so there may be scenarios that behave differently than before. If we decide on specific deviations to the current behavior, we can add those here as well.

### DIFF
--- a/packages/moonstone/Scroller/Scroller.js
+++ b/packages/moonstone/Scroller/Scroller.js
@@ -16,19 +16,18 @@
  * @exports ScrollerBase
  */
 
+import ri from '@enact/ui/resolution';
 import {ScrollerBase as UiScrollerBase} from '@enact/ui/Scroller';
+import {Spotlight} from '@enact/spotlight';
 import {getTargetByDirectionFromPosition} from '@enact/spotlight/src/target';
 import PropTypes from 'prop-types';
 import React, {Component} from 'react';
-import {Spotlight} from '@enact/spotlight';
 
-import ri from '@enact/ui/resolution';
 import Scrollable from '../Scrollable';
 import ScrollableNative from '../Scrollable/ScrollableNative';
 
 const
 	dataContainerDisabledAttribute = 'data-spotlight-container-disabled',
-	epsilon = 1,
 	reverseDirections = {
 		left: 'right',
 		right: 'left'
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Note: This still needs tested with a few other scenarios in particular with regards to the removal of `scrollInfo` and `scrollPosition` parameters.

packages/moonstone/Scroller/Scroller.js
```diff
@@ -262,16 +232,14 @@ class ScrollerBase extends Component {
...
-		const {
-			top: itemTop,
-			left: itemLeft,
-			height: itemHeight,
-			width: itemWidth
-		} = this.getFocusedItemBounds(item);
-
 		if (this.uiRef.isVertical()) {
-			this.uiRef.scrollPos.top = this.calculateScrollTop(item, itemTop, itemHeight, scrollInfo, scrollPosition);
+			this.uiRef.scrollPos.top = this.calculateScrollTop(item, scrollInfo, scrollPosition);
```
ybsung: Do we need `scrollInfo`, `scrollPosition` parameters because there is a only `item` parameter in `calculateScrollTop`?

ryanjduffy: I'm not sure. I expected to need it but they aren't right now. If the functionality checks out as is, we should remove passing them.

ybsung:
Hi @ryanjduffy 

There is one regression with qa-scroller sample.

Issue
: The Title in Expandable List is gone when the Expandable List opens.

Steps
1. Launch qa-scroller sample with this PR branch
2. 5-way Select the 2nd ExpandableList (not the ExpandableItem)

Expected Result: `ExpandableList` title should be shown and stick to the top.
Actual Result: `ExpandableList` title was gone and the first item `San Francisco` sticks to the top.

When testing http://nebula.lgsvl.com/enyojs/enact-samples/qa-scroller/ sample. I could not find any issue. 

Is it the reason that the following code was gone with your PR?  I could not find any related code in your PR.
```
} else if (newItemTop - nestedItemHeight - currentScrollTop > epsilon) {	
	// set scroll position so that the top of the container is at least on the top as a fallback.	
	newScrollTop = newItemTop - nestedItemHeight;	
}
```

packages/moonstone/Scroller/Scroller.js
```diff
@@ -152,98 +151,69 @@ class ScrollerBase extends Component {
...
-		} else if (newItemTop - currentScrollTop < epsilon && scrollHeightChange <= 0) {
-			// Calculate when 5-way focus up past the top.
 
-			// if the first item is focused and have an invisible area, scroll all the way to the top
-			if (newItemTop < heightThreshold) {
-				newScrollTop = 0;
-			} else {
-				newScrollTop += newItemTop - currentScrollTop;
+			// if the container doesn't fit, adjust the scroll top ...
+			if (containerBounds.top > scrollerBounds.top) {
```
ybsung: I'm not sure but don't we need add the following condition in the as-is code to the above condition because it seems that the top of the container needs to stick to the top of the scroller when only the focused item bottom is less then scrollTop.
```
if (nestedItemTop - currentScrollTop < epsilon)
```

If not, for example,
When opening the 2nd ExpandableList in qa-scroller sample, both of `containerBounds.top > scrollerBounds.top` and `containerBounds.top + containerBounds.height < scrollerBounds.top + scrollerBounds.height` conditions are true, so finally the bottom of the container sticks to the boom of the scroller instead of the top of the scroller even though `containerBounds.top > scrollerBounds.top`  condition is true.

ybsung:
Did you consider the case that we considered in https://github.com/enactjs/enact/pull/1093 to ignore number less than 1 while animating?

ybsung:
Sometimes the ExpandableList in the sample described in JIRA does not close even though 5-way selecting the item in the ExpandableList.

packages/moonstone/Scroller/Scroller.js
ryanjduffy: I'm not following what you expect to see. Right now, it's set up to anchor the container to the top of the viewport if the container is below the top of the viewport and the bottom if the bottom of the container is above the bottom of the viewport.

Does that match your expectation? Can you share more specific steps to repro the problem?

ryanjduffy:
I pushed a commit to work around the first issue you shared. I think scroller is behaving correctly but it doesn't match our expectations for moonstone. Longer term, i think we'll want to add some API to support components opting-out of the "Scroll container into view" logic of scroller rather than relying on a "secret" data attribute.

packages/moonstone/Scroller/Scroller.js
ybsung: When trying to reproduce this issue with the latest code, I could not reproduce it.

PR #2018, opened 2018-11-13
ENYO-5720: Use device locale as default in sampler

ryanjduffy:
### Checklist

Unable to use device language in sampler. Selecting the `'local'` value of `''` resulted in the custom knobs reverting to the default value of `'en-US'`.

### Resolution
* Update sampler to use the local language as the default to avoid the knob overriding with a non-falsey default
* Update I18nDecorator to allow a prop change of `locale` to a falsey value triggering the existing logic in `updateLocale` to use the device language

### DIFF
--- a/packages/i18n/CHANGELOG.md
+++ b/packages/i18n/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact i18n module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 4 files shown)


webOS101:
Will this change all tests executed outside the US to fail due to changes in the default locale?  Does this affect QA Sampler? What will be the impact on testers?

ryanjduffy:
Good questions! It seems likely that we'd introduce some testing churn with that change so I've reverted it and worked around the issue another way.

PR #2019, opened 2018-11-14
ENYO-5724: Fix marquee is not animating

yeram:
### Checklist

Fix marquee is not animating if component has floating number width and scrolling width is same with ceiled width

### Resolution
Use `node.getBoundingClientRect().width` instead of `node.clientWidth` since `node.clientWidth` doesn't support float number

### Additional Considerations
I'm not sure about `node.scrollWidth`

### Links
https://bugs.chromium.org/p/chromium/issues/detail?id=360889
ENYO-5724

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Marquee` to calculate distance precisely
+
 ## [2.2.5] - 2018-11-05
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Noting that this change reverts part of #1300.

ryanjduffy:
Also, the proposed version is what Enyo 2.6.3 uses:

https://github.com/enyojs/moonstone/blob/f494922966f13bccee6324a012261ce56587a067/src/Marquee/Marquee.js#L964-L966

ryanjduffy:
After some discussion, we've decided to revert the "threshold" change because it breaks synchronized marquees. By allowing the decorator to start animation but suppressing the actual `transition` and by extension `onMarqueeComplete`, the marquee instance never informs the controller it has completed so the group will not restart.

The original fix (sorry for mucking with it, @yeram!) is the best compromise we can offer at this point and favors ensuring the content is displayed to the user over the slightly odd edge case UX of removing the ellipsis with nearly no animation.

PR #2020, opened 2018-11-14
ENYO-5725: Fix VideoPlayer to blur slider when hiding media controls

xbinary:
### Checklist

If `mediaControls` disappear when `mediaSlider` has focus and pointer mode is `true`,
focus stays in `mediaSlider`.

### Resolution
Blur focus when hiding `mediaControls`

### Links
Related issues : https://github.com/enactjs/enact/pull/1952

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


ryanjduffy:
Thanks for the report and fix @xbinary! I refined it a bit to cast a broader net so instead of checking if the focused item is within media controls or is the slider individually, it will blur when anything within `css.bottom` which is inclusive of both the slider and media controls.

PR #2022, opened 2018-11-14
ENYO-5724: Fix marquee handling of very small marquees

ryanjduffy:
### Checklist

It's possible to have a Marquee instance where the natural width of the text is larger than the allowed width but both round to the same value. As a result, the browser determines it should display an ellipsis for `text-overflow` but `ui/Marquee` believes the distance to be zero (due to rounding of `clientWidth` and `scrollWidth`).

### Resolution
* Update `ui/Marquee.MarqueeDecorator` to use `getBoundingClientRect()` to retrieve the sub-pixel measurements
* Update `ui/Marquee.MarqueeBase` to only animate when the distance is greater than 1px

The cumulative result is that the text will still appear with an ellipsis but will then be displayed fully when marquee is triggered even though no animation occurs.

### Additional Considerations
I was unable to find a way to definitively determine if the ellipsis should be suppressed altogether so this is a compromise.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Marquee` to handle very small marquees better
+
 ## [2.2.5] - 2018-11-05
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


ryanjduffy:
I realized this had become very similar to @yeram's first PR #2019 so I've updated it and will close this in favor of it.

PR #2026, opened 2018-11-20
ENYO-5731: Fix Marquee to avoid very small animations

ryanjduffy:
### Checklist

The change in #2019 introduced a regression that would allow text that fits in its container to marquee if the subpixel size of the text container were smaller than the rounded `scrollWidth`.

This should resolve both the issues from ENYO-5731 and ENYO-5724:
1. text that is long enough to cause an ellipsis but not enough to cause marquee (the text length using `getBoundingClientRect()` is a subpixel value [e.g. `165.5px`] and would round up) causing it to never fully be visible to the user
2. text that fits in its container but its size is a subpixel value that would round up (e.g. `165.5px`) causing it to marquee just a little

### Resolution
* Add a 1px threshold before triggering animation
* Update state logic to ensure `overflow: clip` is applied after measuring and determining that animation isn't required.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Marquee` to avoid very small animations
+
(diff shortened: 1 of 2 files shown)


yeram:
:+1:

PR #2027, opened 2018-11-29
PLAT-68129: Convert tests to use Jest

JayCanuck:
Requires: https://github.com/enactjs/cli/pull/154, https://github.com/enactjs/cli/pull/155, and https://github.com/enactjs/eslint-config-enact/pull/22

### Checklist

* [X] This is an API breaking change.
Sort of an API breaking change. Does not break Enact APIs, however will require `develop` Enact CLI in order to test, which is the next-gen CLI release and will contain some technically breaking CLI-side changes.

* Desired to transition from the buggy Karma/Webpack text runner towards a Jest-based test runner format. This is both for simplicity and opensource pattern of usage/popularity, as well as external requests to support Jest.

### Resolution
* Converted using `jest-codemods`, and then manually updated individually to correct any failures.
* Removed the need for `dirty-chai` and `mocha` usage in favour of the built-in Jest utilities.
* Replaced any remaining string ref usage into modern ref usage.
* For `ui/Cancelable` spec 'modal instances' tests, use `window.KeyboardEvent`.
* For `moonstone/Input` spec test  'should blur input on enter if dismissOnEnter', use `mouseDown` simulated event on to focus, rather than `node.focus()`.
* For `moonstone/EditableIntegerPiker` spec test 'should take value inputted and navigate to the value on blur', simulate a `focus` event rather than `node.focus()`.
* Skip 39 remaining failing Spotlight tests as they rely on layout measurements/metrics, which Jest's underlying `jsdom` does not support.
* Update library `package.json` test tasks for Enact CLI Jest update.
* Update root monorepo `package.json` for `test` and `clean` shorthand tasks.

### Additional Considerations
* Future task to assess skipped Spotlight tests and mock layout metrics as viable, moving the remaining tests into `ui-tests`.
* Future task to assess removing `sinon` usage in favour or Jest APIs.


ryanjduffy:
Pushed a commit to fix the tests for Job.promise() which likely landed after this work started.

PR #2031, opened 2018-12-04
ENYO-5735: Fix Marquee to show ellipsis when children change

ryanjduffy:
### Checklist

I'm a little fuzzy on the specifics of the bug but it relates to some recent changes to Marquee. There are two related issues it seems.

First, when changing from short text to long text, the long text should display an ellipsis until animation starts. If the long text is set after the `marqueeRenderDelay` timeout fires, it won't show an ellipsis but will start animating after the requisite delay.

Second, when the above occurs, the subsequent state change to update `overflow` to `'ellipsis'` causes marquee to attempt to start the animation which prevents resetting when complete because the `timerState` is pending due to the spurious animation start request.

### Resolution
Resetting the `overflow` state when invalidating metrics resolves the first issue and, by extension, prevents the second. Seems like a reasonable change to make regardless since it would correctly mirror invalidating all of the potential changes from `calculateMetrics'.

### Additional Considerations
We may need to investigate the latter issue further. It's possible that we could expand `shouldStartMarquee` to handle this as well but I think the proposed solution is adequate for now.

### DIFF
--- a/packages/ui/Marquee/MarqueeDecorator.js
+++ b/packages/ui/Marquee/MarqueeDecorator.js
@@ -443,6 +443,10 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			this.distance = null;
 			// Assume the marquee does not fit until calculations show otherwise
 			this.contentFits = false;
+
+			this.setState(state => {
+				return state.overflow === 'ellipsis' ? null : {overflow: 'ellipsis'};
+			});
 		}
 
 		/*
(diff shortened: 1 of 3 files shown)


packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -443,6 +443,10 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			this.distance = null;
 			// Assume the marquee does not fit until calculations show otherwise
 			this.contentFits = false;
+
+			this.setState(state => {
+				return state.overflow === 'ellipsis' ? null : {overflow: 'ellipsis'};
+			});
```
germboy: I think this makes sense and appears safe enough overall. However, we might need to de-couple this state change bit, since `invalidateMetrics` is called in the `constructor`, and we can't call `setState` there.

ryanjduffy: Good call!

ryanjduffy: Moved that initialization to the constructor rather than calling `invalidateMetrics()`. Adding a new method just to reset the overflow seemed like unnecessary decomposition to me and it's appropriate to set up members in the constructor anyway.

PR #2036, opened 2018-12-10
ENYO-5745: Fix unable to navigate into container if first item is disabled

yeram:
### Checklist

Fix unable to navigate into spotlight container that has `overflow` configuration if first item is disabled

### Resolution
Filter element that has `disabled` attribute when finding spottable

### Additional Considerations

### Links
ENYO-5745
https://stackoverflow.com/a/23701672

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `spotlight` to find spottable element correctly if first element in container is disabled
+
+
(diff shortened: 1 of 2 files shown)


ryanjduffy:
@yeram Thanks for the great analysis on this bug. We discussed the behavior further this morning and decided to take a different approach in #2037. Using `disabled` to determine focusability isn't always accurate though it is in this case. Instead, we found a combination of adjusting spotlight behavior and updating `ExpandableItem` to report its spottable state correctly provided a solution more in line with the rest of the spotlight implementation.

I definitely appreciate the proposed fix and attention to minimizing change though. Let me know if you have any questions on the route we took.

yeram:
:+1: Thanks!

PR #2040, opened 2018-12-18
ENYO-5742: Add unit tests of voice control for moonstone components

changgilee:
### Checklist

Components should be checked voice control related props

### Resolution
Add unit test to components

### Additional Considerations

### Links
ENYO-5742

### Comments

### DIFF
--- a/packages/moonstone/Button/tests/Button-specs.js
+++ b/packages/moonstone/Button/tests/Button-specs.js
@@ -59,14 +59,13 @@ describe('Button', () => {
 
 			expect(actual).toBe(expected);
 		});
-	});
 
-	describe('voice control', () => {
 		test('should have "data-webos-voice-intent=Select" in default', () => {
 			const button = mount(<Button>Hello</Button>);
 
 			const expected = 'Select';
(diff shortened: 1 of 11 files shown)


changgilee:
@ryanjduffy It depends on what kind of intent used. For example, **Select** intent should be positioned same node with spottable because Web Engine will be simulate click event to target node. But another case, the node is used for only gateway to receive voice event from Voice Framework.
 
_it be better to generalize the node lookups_ <= Can you provide more info about this? Thanks~

ryanjduffy:
Instead of the following which assumes a particular DOM node or subcomponent
```
button.find('div').at(0).prop('data-webos-voice-intent');
expandableItem.find('LabeledItem').prop('data-webos-voice-disabled');
```

Implement tests that use generalized requirements like "the node with role='button' has the voice intent" or "the spottable node has the voice intent"

changgilee:
@ryanjduffy Thanks for your info. I also think generalized requirement is more clear way and try to apply it to Button component.(I did update TC code) In other case, for example TimePicker, my intention is to check voice disabled prop successfully delegated to each child(hour, meridiem, minute) pickers.(So, I hope it will be checked in current way)

PR #2041, opened 2018-12-21
ENYO-5751: Fixed to focus a list item properly by `scrollTo` API call with the current position

0x64:
### Checklist

A list does not perform scrolling if the new position is the same position as the current target position during the list is scrolling.
After #2030 is applied, a list stops its scrolling after a short period of idle time from the last scroll position change to prevent apps from overwhelming events. So, if another scrolling request (from apps by `scrollTo` API) occurs within the short period of idle time above, a list is treated as `scrolling` status and the new request is ignored when the target position is the same position as the current target position.
In normal cases, this behavior is fine regarding the scrolling position only. But, if a list handles another status (`focus` in this issue) beside a position, this behavior blocks a list from proper status updates since it is possible to check only positions and internal scrolling status inside a list.

### Resolution
Changed internal logic not to ignore the new request when a `focus` option is set with a valid value.

### Additional Considerations

### Links
ENYO-5751

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unrelesed]
+
+### Fixed
+
+- `moonstone/VirtualList.VirtualGridList` and `moonstone/VirtualList.VirtualList` to focus an item properly by `scrollTo` API right after scrolling to the same position
+
 ## [2.2.8] - 2018-12-06
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


seunghoh:
LGTM

ryanjduffy:
I'd like to avoid `force` parameters when possible. They usually indicate a gap in business rules since we're effectively opting out of the built-in rules. In this case, we've tied focusing to the `stop` callback which isn't fired here since `this.scrolling` hasn't been cleared. 

I'm not sure if there's a better approach yet but I'd like to investigate further tomorrow. It's possible that clearing `this.scrolling` before the `onScrollStop` job fires would be appropriate. Or maybe we need to adjust when the focusing is triggered to account for this case.

I'd appreciate ideas and alternatives!

ryanjduffy:
Closing in favor of #2044

PR #2042, opened 2018-12-26
ENYO-5755: Add voice control attribute extractor

changgilee:
### Checklist

In current status, propType should be added in each moonstone components when additional voice control attribute need

### Resolution
Add extractor function for pre-check and delegate voiceProps to sub components.

### Additional Considerations

### Links
ENYO-5755

### Comments

### DIFF
--- a/packages/core/util/util.js
+++ b/packages/core/util/util.js
@@ -111,6 +111,28 @@ const extractAriaProps = function (props) {
 };
 
 /**
+ * Removes voice control related props from `props` and returns them in a new object.
+ *
+ * @function
+ * @param   {Object}    props    Props object
+ *
+ * @returns {Object}             voice control related props
+ * @memberof core/util
+ * @public
+ */
+const extractVoiceProps = function (props) {
+	const obj = {};
+	Object.keys(props).forEach(key => {
+		if (key.indexOf('data-webos-voice-') === 0) {
+			obj[key] = props[key];
+			delete props[key];
+		}
+	});
+
+	return obj;
+};
+
+/**
  * Gets the current timestamp of either `window.performance.now` or `Date.now`
  *
  * @function
@@ -204,6 +226,7 @@ export {
 	coerceArray,
 	coerceFunction,
 	extractAriaProps,
+	extractVoiceProps,
 	isRenderable,
 	Job,
 	memoize,
--- a/packages/moonstone/IncrementSlider/IncrementSlider.js
+++ b/packages/moonstone/IncrementSlider/IncrementSlider.js
@@ -22,7 +22,7 @@
 import {forward} from '@enact/core/handle';
 import {is} from '@enact/core/keymap';
 import kind from '@enact/core/kind';
-import {extractAriaProps} from '@enact/core/util';
+import {extractAriaProps, extractVoiceProps} from '@enact/core/util';
 import Spottable from '@enact/spotlight/Spottable';
 import Changeable from '@enact/ui/Changeable';
 import Slottable from '@enact/ui/Slottable';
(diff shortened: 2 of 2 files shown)


packages/core/util/util.js
```diff
@@ -110,6 +110,28 @@ const extractAriaProps = function (props) {
...
+ * Removes voice control related props from `props` and returns them in a new object.
+ *
+ * @function
+ * @param   {Object}    props    Props object
+ *
+ * @returns {Object}             voice control related props
+ * @memberof core/util
+ * @public
+ */
+const extractVoiceProps = function (props) {
```
ryanjduffy: I'd like to move this method into moonstone and mark it private since voice control is currently platform-specific and `core` should be platform-independent.

packages/moonstone/IncrementSlider/IncrementSlider.js
```diff
@@ -516,12 +497,11 @@ const IncrementSliderBase = kind({
 		return (
 			<div {...rest}>
 				<IncrementSliderButton
+					{...voiceProps}
```
ryanjduffy: Are there any `data-webos-voice-` props that can't be applied to multiple components with the same value?

packages/moonstone/internal/Picker/Picker.js
```diff
@@ -866,11 +851,10 @@ const PickerBase = class extends React.Component {
 				ref={this.initContainerRef}
 			>
 				<PickerButton
+					{...voiceProps}
```
ryanjduffy: Same question here on applying the same voice props to multiple components.

packages/core/util/util.js
changgilee: @ryan I also agree your guide and move the function to moonstone. (I did made moonstone/internal/util.js because I can't find proper file to move it. If you think the function should be placed another js file in moonstone, please mention it and I will change it)

packages/moonstone/IncrementSlider/IncrementSlider.js
changgilee: Thanks for mention about this. As you said, some voice props should not be delegated to IconButtons. So, I did update commit for fix it.

packages/moonstone/internal/Picker/Picker.js
changgilee: This component is also fixed. Thanks~

packages/moonstone/IncrementSlider/IncrementSlider.js
```diff
@@ -507,6 +488,10 @@ const IncrementSliderBase = kind({
 		...rest
 	}) => {
 		const ariaProps = extractAriaProps(rest);
+		const voiceProps = extractVoiceProps(rest);
+		delete voiceProps['data-webos-voice-label'];
+		delete voiceProps['data-webos-voice-labels'];
```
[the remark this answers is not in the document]
changgilee: These are different label and usage

packages/moonstone/internal/Picker/Picker.js
```diff
@@ -807,6 +788,12 @@ const PickerBase = class extends React.Component {
 			...rest
 		} = this.props;
 
+		const voiceProps = extractVoiceProps(rest);
+		const voiceLabelsExt = voiceProps['data-webos-voice-labels-ext'];
+		delete voiceProps['data-webos-voice-label'];
+		delete voiceProps['data-webos-voice-labels'];
+		delete voiceProps['data-webos-voice-labels-ext'];
```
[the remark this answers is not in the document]
changgilee: In case of Picker, specific props(ex. data-webos-voice-disabled, data-webos-voice-group) should be delegated to child buttons. so, unnecessary props is delete like above. (Delete is applied properly)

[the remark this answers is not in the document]
changgilee: In my opinion, Each case seems to have pros and cons. Declaring an argument value to a function can explicitly declare an item to be extracted and has a performance advantage. However, when adding an attribute value, all function call must be modified. Also there are many voice control attributes, the function declaration may become longer. If there is no big problem, I would like to maintain the current structure.

viodragon2:
With these changes, all the information related to `data-voice-*` props is hidden and also removed from the documentation. Can you add props back for the docs?

Also, please resolve merge conflicts as well. Thanks

PR #2044, opened 2018-12-27
ENYO-5751: Fix focusing a VirtualList list item within the viewport via scrollTo

0x64:
### Checklist

A list does not perform scrolling if the new position is the same position as the current target position during the list is scrolling.
After #2030 is applied, a list stops its scrolling after a short period of idle time from the last scroll position change to prevent apps from overwhelming events. So, if another scrolling request (from apps by `scrollTo` API) occurs within the short period of idle time above, a list is treated as `scrolling` status and the new request is ignored when the target position is the same position as the current target position.
In normal cases, this behavior is fine regarding the scrolling position only. But, if a list handles another status (`focus` in this issue) beside a position, this behavior blocks a list from proper status updates since it is possible to check only positions and internal scrolling status inside a list.

### Resolution
Changed internal logic not to ignore the new request when a `focus` option is set with a valid value.


0x64:
Note: the fixed module is `ui` but the behavior of `ui` modules is not changed. The affected scenarios are `moonstone` cases.

ryanjduffy:
Retargeted to `develop`. We will cherry pick into a release branch when required.

PR #2045, opened 2018-12-27
PLAT-68796: Add more aria property values to verify visually (for debug aria feature)

Baekwoo-s:
### Checklist

Currently, only aria-label and aria-valuetext display when 'debug aria' is enabled.
It would be helpful if adding more aria properties to verify aria features visually.
This PR will cause more debug aria properties to be displayed on the screen.

### Resolution
Support the following property categories in `debug aria` feature except for few semantical aria properties (ref: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#States_and_properties)
- Role (role)
- Widget attributes (aria-checked, aria-disabled, aria-hidden, aria-label, aria-pressed, aria-valuetext)
- Live region attributes (aria-live, role)
- Relationship attributes (aria-controls, aria-labelledby, aria-owns)
- All (All of the above-mentioned attributes)
(sampler: 
image
)

If you want to enable debug aria in your app, the aria property of the category is displayed on the screen by putting the names of the above categories after the `debug aria` class.
(e.g.)
- `debug aria role`
- `debug aria widget`
- `debug aria live`
- `debug aria relationship`
- `debug aria all`

### Additional Considerations
- Fixed issue where debug features was not enabled but `debug` class was included in the sampler

issue

AS-IS
```
Line: 151~153
if (Object.keys(classes).length > 0) {	 
	classes.debug = true;
}
```
TO-BE
```
Line: 162~165
for (const key in classes) {
	if (typeof classes[key] !== 'undefined') {
		classes.debug = true;
	}
}
```

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -146,21 +146,25 @@ const StorybookDecorator = (story, config) => {
 	const DevelopmentConfig = {
 		defaultProps: {
 			'debug aria': false,
+			'debug aria type': '',
 			'debug layout': false,
 			'debug spotlight': false
 		},
 		groupId: 'Development'
 	};
 
-	const debugAriaType = select('debug aria type', debugAriaTypes, DevelopmentConfig, getPropFromURL('debugAriaType'));
+	const
+		aria = boolean('debug aria', DevelopmentConfig, (getPropFromURL('debug aria') === 'true')),
+		ariaDebugType = select('debug aria type', debugAriaTypes, DevelopmentConfig, getPropFromURL('debug aria type')),
+		layout = boolean('debug layout', DevelopmentConfig, (getPropFromURL('debug layout') === 'true')),
+		spotlight = boolean('debug spotlight', DevelopmentConfig, (getPropFromURL('debug spotlight') === 'true'));
 
-	let classes = {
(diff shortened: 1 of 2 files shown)


packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -136,20 +145,24 @@ const StorybookDecorator = (story, config) => {
...
-		aria: boolean('debug aria', DevelopmentConfig, (getPropFromURL('debug aria') === 'true')),
+		[`aria ${debugAriaOption}`]: debugAriaOption && true,
 		layout: boolean('debug layout', DevelopmentConfig, (getPropFromURL('debug layout') === 'true')),
 		spotlight: boolean('debug spotlight', DevelopmentConfig, (getPropFromURL('debug spotlight') === 'true'))
 	};
-	if (Object.keys(classes).length > 0) {
-		classes.debug = true;
+
+	for (const key in classes) {
+		if (typeof classes[key] !== 'undefined') {
```
0x64: If we set a fallback value when calling `getPropFromURL`, we don't need to check whether its value is `undefined` or not.

0x64:
LGTM

Baekwoo-s:
I tried to implement display only the aria attributes in the DOM when debug aria is enabled,
but I found that the amount of code is much increased because there are too many combinations of aria attributes to be added. 
I will close this PR because I would like to avoid append a lot of code by adding attributes to debug aria.

PR #2052, opened 2019-01-09
ENYO-5722: Create sample to test tall glyphs as non latin components

haileyR:
ENYO-5722

### Resolution
Added sample to test tall glyphs as non latin components.


ryanjduffy:
@haileyR - Should we add a knob to this sample so we can try with other text values?

@LucieRoy - Where do you think this sort of sample makes the most sense in QA sampler? It's in Divider now but isn't really related to Divider. Should we make a new category that isn't component-centric?

LucieRoy:
@ryanjduffy - a similar issue just came up yesterday actually about a test not being really related to the  component it was under in qa-sampler. You have a good idea moving forward.  A category not component centric will be helpful and not misleading. Thank you!

haileyR:
@ryanjduffy I think that will help. I will add it.

PR #2056, opened 2019-01-17
PLAT-72156: Replace `resize` PubSub with context API

rundmt:
### Checklist

* [] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

We're no longer using our old PubSub, now we're relying more on React's new context API and our own Registry to manage these updates.

### Resolution
Added Registry to replace PubSub.

### Additional Considerations
I checked breaking API change because technically it does break a Private API, but not a public one. So this might not be a huge issue. I just wanted to make sure we document it though.

Also no changelog because these were private APIs

I couldn't go with the context class style for Marquee, because it will probably use multiple contexts so we can't attach it to the class.
I made Registry as an internal module because it is currently used for special cases in the framework. Since PubSub was internal I figure we'll keep it the same with this.

### Links
PLAT-72156


packages/ui/Transition/Transition.js
```diff
@@ -527,6 +526,19 @@ class Transition extends React.Component {
 		}
 	}
 
+	setRegistry = registry => {
```
ryanjduffy: This might not be necessary here. When using `contextType` and the lifecycle methods, we know that `register`/`unregister` are only invoked a mount/unmount. With render props, you need a way to only register the first time which is where this helps a bit.

After typing this comment, it occurs to me that when using render props we might be able to register with: ```
<ResizeContext.Subscriber>
  {value => {
    if (!this.registry) {
      this.registry = value;
      this.registry.register(this.handleResize);
    }
    ...
  }
</ResizeContext.Consumer>
```
and unregister in `componentWillUnmount` with `if (this.registry) this.registry.unregister(this.handleResize)`

File: packages/ui/Transition/Transition.js [hunk 2]
```diff
@@ -432,8 +434,8 @@ class Transition extends React.Component {
 			this.measureInner();
 		}
 
-		if (this.context.Subscriber) {
-			this.context.Subscriber.subscribe('resize', this.handleResize);
+		if (this.context) {
```
ryanjduffy: When I was testing this, this.context was always valued but could be an empty object.

rundmt: Interesting, I just tested it in the unit tests. If you don't have a parent with the context I get `undefined`.

packages/ui/Transition/Transition.js
rundmt: Yeah that makes sense. I've simplified them now.

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -310,12 +309,7 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			this.timerState = TimerState.CLEAR;
 			this.distance = null;
 			this.contentFits = false;
-		}
-
-		componentWillMount () {
-			if (this.context.Subscriber) {
-				this.context.Subscriber.subscribe('resize', this.handleResize);
-			}
+			this.registry = null;
```
ryanjduffy: I think we need to add an `unregister` in `componentWillUnmount`. Otherwise, if an instance is removed from the tree and a resize occurs, we'll get some errors/complaints. Let's add a test for that scenario as well.

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -824,11 +815,23 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		render () {
-			if (this.props.marqueeDisabled) {
-				return this.renderWrapped();
-			} else {
-				return this.renderMarquee();
-			}
+			return (
+
```
ryanjduffy: extra line

packages/moonstone/MoonstoneDecorator/AccessibilityDecorator.js
```diff
@@ -13,8 +14,11 @@ import PropTypes from 'prop-types';
  */
 const AccessibilityDecorator = hoc((config, Wrapped) => {	// eslint-disable-line no-unused-vars
 	return class extends React.Component {
+		static contextType = ResizeContext;
+
 		static displayName = 'AccessibilityDecorator'
 
+
```
ryanjduffy: extra line

packages/core/internal/Registry/Registry.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * Allows components to register parents to cascade context changes and trigger functions
+ *
+ * @class Registry
+ * @memberof ui/Registry
+ * @ui
```
ryanjduffy: `@ui` is indicates components but this is a factory so let's remove.

packages/core/internal/Registry/Registry.js
```diff
@@ -0,0 +1,44 @@
+/**
+ * Allows components to register parents to cascade context changes and trigger functions
+ *
+ * @class Registry
```
ryanjduffy: Let's call it `@type Object` since it isn't instantiable.

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -285,8 +285,7 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		static contextTypes = {
-			...contextTypes,
-			...stateContextTypes
+			...contextTypes
```
ryanjduffy: No need to clone the object here. Pass by ref is okay now that it's only 1 `contextTypes`.

packages/i18n/I18nDecorator/I18nDecorator.js
```diff
@@ -104,8 +103,7 @@ const I18nDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			locale: PropTypes.string
 		}
 
-		static contextTypes = stateContextTypes
-		static childContextTypes = {...contextTypes, ...stateContextTypes}
+		static childContextTypes = {...contextTypes}
```
ryanjduffy: No need to clone the object here. Pass by ref is okay now that it's only 1 `contextTypes`.

packages/ui/Remeasurable/Remeasurable.js
```diff
@@ -1,130 +1,13 @@
 /**
- * Exports the {@link ui/Remeasurable.Remeasurable} and {@link ui/Remeasurable.RemeasurableDecorator}
- * higher-order component (HOC). Adds the ability to broadcast remeasure changes
- * based on a callback. The default export is {@link ui/Remeasurable.Remeasurable}.
+ * Exports the Resize context. This is used internally for things to notify children that they need
+ * to resize because of a parent update.
  *
  * @module ui/Remeasurable
```
ryanjduffy: Maybe move this entire thing into `ui/internal/Resize` to keep our "private" code cleanly separated?

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -348,11 +349,8 @@ class ScrollableBaseNative extends Component {
 		verticalScrollbar: PropTypes.oneOf(['auto', 'visible', 'hidden'])
 	}
 
-	static contextTypes = contextTypesState
-
 	static childContextTypes = {
-		...contextTypesResize,
-		...contextTypesState
+		...contextTypesResize
```
ryanjduffy: No cloning needed.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -362,11 +363,8 @@ class ScrollableBase extends Component {
 		verticalScrollbar: PropTypes.oneOf(['auto', 'visible', 'hidden'])
 	}
 
-	static contextTypes = contextTypesState
-
 	static childContextTypes = {
-		...contextTypesResize,
-		...contextTypesState
+		...contextTypesResize
```
ryanjduffy: No cloning needed.

ryanjduffy:
Also, Travis is still complaining about ESLint failures :(

rundmt:
@ryanjduffy Fixed up a lot of the nits. What should we do for Travis. Should that be it's own ticket or should I just power through those now?

ryanjduffy:
Do we need to bump the version of the plugin in our tooling to resolve the errors?

ryanjduffy:
@rundmt I made a couple API refinements to the Registry factory to support only passing down the "register/unregister" methods via context so that it isn't possible to "notify" from downstream. Let me know if you have any questions/concerns with that change.

packages/sampler/stories/qa/Remeasurable.js
```diff
@@ -30,6 +30,7 @@ storiesOf('Remeasurable', module)
 		() => (
 			<Scroller style={{height: ri.unit(399, 'rem'), width: ri.unit(501, 'rem')}}>
 				<NoUpdate>
+					<Item marqueeOn="render">MARQUEEONRENDER ABCDE</Item>
```
webOS101: @LucieRoy Can you confirm these changes to the QA sample don't cause any of the test cases to change?

ryanjduffy:
Merging on Roy's behalf.

PR #2057, opened 2019-01-18
 PLAT-72725: Update lifecycle methods in Changeable and Toggleable

viodragon2:
### Checklist

Update `UNSAFE_componentWillReceiveProps()` to `static getDerivedStateFromProps()` in `ui/Changeable` and `ui/Toggleable`. Unlike `UNSAFE_componentWillReceiveProps`, `static getDerivedStateFromProps()` runs on mount and thus small change was required in `Changeable` to account for such case.
Also, `Toggleable` tests were not done correctly. `defaultActive` is not the correct default key and was used. By changing the value to `defaultSelected`, I was able to spot a few tests that failed because it was wrongly written.


packages/ui/Changeable/Changeable.js
```diff
@@ -127,35 +128,28 @@ const Changeable = hoc(defaultConfig, (config, Wrapped) => {
...
-					value = props[prop];
-				}
-
-				controlled = true;
-			}
 
 			this.state = {
-				controlled,
-				value
+				value: getValue(props),
```
webOS101: Isn't `getDerivedStateFromProps` called during mounting?  Do we need to set `value` here?

webOS101: Hmm, yes, that makes a lot of sense!

packages/ui/Changeable/Changeable.js
```diff
@@ -67,6 +67,7 @@ const defaultConfig = {
 const Changeable = hoc(defaultConfig, (config, Wrapped) => {
 	const {prop, change} = config;
 	const defaultPropKey = 'default' + cap(prop);
+	const getValue = props => props[prop] != null ? props[prop] : props[defaultPropKey];
```
webOS101: There's a slight change in functionality.  When controlled, it will go back to the default value if the prop is set to `null` or `undefined`.  Before, it would just set the value to `null` or `undefined`.  Is this change OK?

webOS101: I guess this could only happen if someone added the `[defaultPropKey]` value after initialization or specified both at startup.  Alternative is to move the `props[defaultPropKey]` check into `constructor()` and just return `props[prop]`.

PR #2059, opened 2019-01-18
PLAT-72741: ui/RadioDecorator: update life cycle method

sugardave:
### Checklist

This patch moves the controller notification into `componentDidUpdate`.

### Resolution
This change was made to support the overall lifecycle update work for Enact.

### Links
PLAT-72741


viodragon2:
Please change target to `feature/PLAT-53066`

packages/ui/RadioDecorator/RadioDecorator.js
```diff
@@ -85,8 +85,8 @@ const RadioDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			}
 		}
 
-		UNSAFE_componentWillReceiveProps (nextProps) {
-			this.notifyController(nextProps);
+		componentDidUpdate () {
```
webOS101: Is there any way to test whether this change causes something that took 1 refresh cycle to now take 2?  It's possible this always caused two update cycles and changing this to occur before the update will not make any difference.  If it did change this, is it worth coming up with a different solution to prevent an intermediate state where more than one radio item is selected at a time?

sugardave: Just finished discussion with @ryanjduffy about this.  While it is true that now there are two refresh cycles, it is unclear if there is an easy way around it.  Since this PR is addressing just the lifecycle change, we can move the discussion/strategy for optimizing `RadioDecorator` to a new issue.

One interesting thing we found is that removing the `open` prop from `Expandable` and making it uncontrolled results in only one refresh cycle, but we don't want to make that change here because it needs more investigation for potential impacts that we didn't consider.

PR #2061, opened 2019-01-19
PLAT-72376: Add typescript so linked Enact works

webOS101:
### Checklist

Adds `@types/react` so that you can work with linked Enact.

### Resolution

Also, ignore any `.d.ts` files in the enact directory so it doesn't show lots of untracked files

### Additional Considerations

### Links
PLAT-72376

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -35,3 +35,6 @@ jspm_packages
 
 # Optional REPL history
 .node_repl_history
+
+# Typescript definitions (auto generated)
+*.d.ts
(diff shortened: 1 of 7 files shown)


JayCanuck:
I'm not a fan of this.  By virtue of us going the 'generate typedefs at release-time' route, linking is not a valid option to begin with. And adding mandatory react typedef dependencies into the framework itself is extra, unneeded bulk.

ryanjduffy:
I'm not sure this is a necessary change but I can see that value for us. I think the "bulk" concerns would be mitigated if we used `devDependencies` which wouldn't need to be installed by npm consumers.

What alternatives do we have for verifying our type generation?

ryanjduffy:
Found another solution that doesn't require adding to enact dependencies.

PR #2063, opened 2019-01-21
PLAT-72875: Add support for 8k display in Moonstone

Djspaceg:
### Checklist

Loading Moonstone on an 8k display, things look tiny.

### Resolution
Added a new screenType to support the new larger screen resolution


webOS101:
I think it's OK in this case.  There's no proper component to tag this onto, except maybe `moonstone/Decorator`.

PR #2065, opened 2019-01-21
PLAT-72158: Replace contextTypes with contextType in core/kind

haileyR:
Updated contextTypes to contextType. and added warning for deprecated prop.

Deprecate {{contextTypes}} support in {{core/kind}} to be removed in 3.0.

### Resolution
 {{contextType}} is added instead.


packages/core/kind/kind.js
```diff
@@ -135,7 +136,7 @@ const kind = (config) => {
 
 		static propTypes = propTypes
 
-		static contextTypes = contextTypes
```
ryanjduffy: We can't remove `contextTypes` yet (it's only deprecated) but I just verified that if both are present that `contextType` will take precedence so we can leave the both for now and remove `contextTypes` fully in 3.0.

ryanjduffy: Sandbox link to try them out together if you're interested: https://codesandbox.io/s/o53r87r9w9

PR #2066, opened 2019-01-21
PLAT-72736: Update Slider to getDerivedStateFromProps

rundmt:
### Checklist

* [] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [] At least one test case is included for this feature or bug fix

* [] This is an API breaking change

Updated Slider to use getDerivedStateFromProps.

### Links
PLAT-72736

### DIFF
--- a/packages/moonstone/Slider/SliderBehaviorDecorator.js
+++ b/packages/moonstone/Slider/SliderBehaviorDecorator.js
@@ -78,15 +78,15 @@ const SliderBehaviorDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				dragging: false,
 				focused: false,
 				useHintText: false,
-				valueClone: props.value
+				prevValue: props.value
 			};
 		}
 
 		static getDerivedStateFromProps (props, state) {
-			if (props.value !== state.valueClone) {
+			if (props.value !== state.prevValue) {
 				return ({
 					useHintText: false,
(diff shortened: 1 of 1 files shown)


packages/moonstone/Slider/SliderBehaviorDecorator.js
```diff
@@ -77,14 +77,19 @@ const SliderBehaviorDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+				useHintText: false,
+				valueClone: props.value
 			};
 		}
 
-		UNSAFE_componentWillReceiveProps (nextProps) {
-			if (this.props.value !== nextProps.value) {
-				this.setState({useHintText: false});
+		static getDerivedStateFromProps (nextProps, prevState) {
+			if (nextProps.value !== prevState.valueClone) {
```
viodragon2: We don't have a specific coding rule for naming in `getDerivedStateFromProps`, but let's follow `react`'s examples. Use `(props, state)` instead of `(nextProps, prevState)`, and `prev*` instead of `*Clone`
e.g.
```
		static getDerivedStateFromProps (props, state) {
			if (props.value !== state.prevValue) {
				return ({
					useHintText: false,
					prevValue: props.value
				});
			}
```

PR #2067, opened 2019-01-21
PLAT-72737: TimePickerBase - Update lifecycle method

teckliew:
### Checklist

Update lifecycle in `TimePickerBase`

### Links
PLAT-72737

### DIFF
--- a/packages/moonstone/TimePicker/TimePickerBase.js
+++ b/packages/moonstone/TimePicker/TimePickerBase.js
@@ -62,7 +62,9 @@ class HourPicker extends React.Component {
 		const hours = hasMeridiem ? hours12 : hours24;
 
 		return (
-			<DateComponentPicker {...rest} noAnimation={this.state.noAnimation}>{hours}</DateComponentPicker>
+			<DateComponentPicker {...rest} noAnimation={this.state.noAnimation}>
+				{hours}
+			</DateComponentPicker>
 		);
 	}
 }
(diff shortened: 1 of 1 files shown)


teckliew:
Sounds good, I simplified `HourPicker` with `shouldComponentUpdate`. LMK WYT

viodragon2:
Shoot. I now understand why it was setting `noAnimation` momentarily. By blocking the render, it's not going to update the position of the value. You can see the problem by updating the hour after changing meridiem picker. (e.g. 5AM -> 5PM -> press up in hour picker -> 6AM)

packages/moonstone/TimePicker/TimePickerBase.js
```diff
@@ -43,18 +43,21 @@ class HourPicker extends React.Component {
 		};
```
viodragon2: Need to add a default value for `prevValue` state

packages/moonstone/TimePicker/TimePickerBase.js
```diff
@@ -43,18 +43,21 @@ class HourPicker extends React.Component {
...
-	UNSAFE_componentWillReceiveProps (nextProps) {
-		const {children, value} = this.props;
-		const {children: nextChildren, value: nextValue} = nextProps;
+	static getDerivedStateFromProps (props, state) {
+		const hours = props.hasMeridiem ? hours12 : hours24;
 
-		this.setState({
-			noAnimation: children[value] === nextChildren[nextValue]
-		});
+		return {
```
viodragon2: Let's add a condition to return `null` when there's no change in the value.

e.g.
```

		if (state.prevValue !== props.value) {
			return {
				noAnimation: hours[state.prevValue] === hours[props.value],
				prevValue: props.value
			};
		}
		return null;
```

PR #2069, opened 2019-01-22
PLAT-72653 - Converted Cancellable to not use componentWillUpdate

rundmt:
### Checklist

Updating to be inline with new react lifecycle.

### Resolution
I moved this addModal function call into render. I think this is the correct place to do it, because it seems to most simple without breaking every test.

### Links
PLAT-72653


rundmt:
Made some updates to this. We actually do have unit tests to determine the order https://github.com/enactjs/enact/blob/3729b7735e2c75158853d686fb00e7d874dad145/packages/ui/Cancelable/tests/Cancelable-specs.js#L302, but for some reason they pass if we move `addModal` to `componentDidMount` as well. I added a more real world test to `Notification` based on the sample you gave. This uses floating layer. It's more of an integration test, but it can reliably tell ordering. It fails if we put `addModal` in `componentDidMount`. Not sure of a way to simplify the test, so if you have any simple solutions I'd be open to replacing. I'm guess we'd have to use something like floating layer to see it, but I'm not sure.

viodragon2:
You should add a unit test with nested Cancelable components to verify the order of things. The test you mentioned above only tests for the sequential mount. 

Something like this maybe?

```
			const results = [];
			const append = str => () => {
				results.push(str);
				return false;
			};

			const First = Cancelable(
				{modal: true, onCancel: append('first')},
				Component
			);
			const Second = Cancelable(
				{modal: true, onCancel: append('second')},
				Component
			);

			mount(<First><Second /></First>);

			document.dispatchEvent(makeKeyboardEvent(27));

			const expected = ['second', 'first'];
			const actual = results;

			expect(actual).toEqual(expected);
```

rundmt:
Updated!

packages/ui/Cancelable/tests/Cancelable-specs.js
```diff
@@ -351,5 +351,31 @@ describe('Cancelable', () => {
 				expect(actual).toBe(expected);
 			}
 		);
+
+		test('should invoke modal handlers in LIFO order', () => {
```
viodragon2: Let's rename the title so that there's no duplicate.

viodragon2: And also let's move it up and group with the other mounting order test.

rundmt: Whoops. Yeah that makes sense

ryanjduffy:
I think that's a good direction, @viodragon2. One challenge we'll have to overcome is handling key events when nothing is focused to initiate the event bubbling within the react hierarchy. If we can find a suitable way to handle that then removing our own event management infrastructure would be a win.

packages/ui/Cancelable/Cancelable.js
```diff
@@ -216,6 +215,11 @@ const Cancelable = hoc(defaultConfig, (config, Wrapped) => {
 			delete props.onCancel;
 			delete props[onCancel];
 
+			if (this.isFirstRender && modal) {
```
ryanjduffy: This is a clever way to simulate `componentWillMount`! I think we'd like to keep our render methods [idempotent](https://en.wikipedia.org/wiki/Idempotence) but I think this is an acceptable deviation for the time being. We might also think of other ways to trigger subscription to the events rather than tied to rendering (e.g. when an event is fired similar to our discussions on #2070).

viodragon2: Fun fact. `react-transition-group` has similar stuff here https://github.com/reactjs/react-transition-group/blob/master/src/TransitionGroup.js#L73

PR #2070, opened 2019-01-22
PLAT-72733: ExpandablePicker lifecycle update

teckliew:
### Checklist

Update lifecycle in `ExpandablePickerDecorator` 

### Links
PLAT-72733

### Comments

### DIFF
--- a/packages/moonstone/ExpandablePicker/ExpandablePickerDecorator.js
+++ b/packages/moonstone/ExpandablePicker/ExpandablePickerDecorator.js
@@ -1,3 +1,4 @@
+import {forward, handle} from '@enact/core/handle';
 import hoc from '@enact/core/hoc';
 import React from 'react';
 import PropTypes from 'prop-types';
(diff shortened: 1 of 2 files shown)


packages/moonstone/ExpandablePicker/ExpandablePickerDecorator.js
```diff
@@ -16,21 +16,20 @@ const ExpandablePickerDecorator = hoc((config, Wrapped) => {	// eslint-disable-l
...
-		UNSAFE_componentWillReceiveProps (nextProps) {
-			let {value} = this.state;
-
-			// if opening or the props value has changed, use it
-			if (!this.props.open && nextProps.open || this.props.value !== nextProps.value) {
-				value = nextProps.value;
+		static getDerivedStateFromProps (props, state) {
+			if (state.prevOpen !== props.open && state.value !== props.value) {
+				return {
+					value: props.open ? props.value : state.value,
```
ryanjduffy: I don't think this ternary is necessary since the same logic is applied in render. If you can verify, I'd suggest we remove it here so the code is clearer.

viodragon2:
What do you think about handling `onOpen` from `Expandable` and handling `onPick` from `ExpandablePicker` to set `value` state.  This would require reordering the HOC to `ExpandablePickerDecorator` be at the top to handle `onOpen` but I think that makes much more sense. WDYT?

ryanjduffy:
I considered that as well. :) Reordering doesn't work like you'd expect ... but I can't recall why at the moment even though i just tried it. It was possible to handle the `onOpen` event to update the state but I didn't see that as materially "better" than what we had. The advantage to this approach is that it defers setting the value until it's actually open since it's possible for an app to handle `onOpen` but decide to not open the expandable. Unlikely, I know but possible.

viodragon2:
Not sure if the example is right. `onOpen` should never fire unless it's opened. Plus, the fact that we're introducing a new state (`prevOpen`) means that `ExpandablePickerDecorator` HOC is trying to fully control those states when they are controlled in other HOCs. In general, I think it's better to leave it uncontrolled when possible.

ryanjduffy:
`onOpen` is fired when the user takes an action to open the expandable (usually clicking the labeled item). If the app is controlling `open`, it can choose to not set the `open` prop; that is by design.

Also for clarity, the `prevOpen` state member isn't controlling the open state; it only reflects the last value of the `open` prop.

PR #2075, opened 2019-01-23
PLAT-72730: Update ContextualPopupDecorator life cycle methods

viodragon2:
### Checklist

Update `ContextualPopupDecorator` life cycle methods

### Resolution
- Remove `componentWillReceiveProps`
- Instead of positioning the popup container in ref callback function, it positions when `FloatingLayer`'s `onOpen` is invoked. This guarantees that both `this.clientNode` and `this.containerNode` are present
- `onOpen` essentially can replace the `open` condition in `componentWillReceiveProps`. Spotlight caching can be moved here.
- Similarly, `onClose` event can be used for the `close` condition in `componentWillReceiveProps`. Resets Spotlight activator and leavefor
- `shouldSpotActivator` requires to check whether the spotlight was in container node on close condition. This can't be done in either `static getDerivedStateFromProps` or `onClose` handler, because `this.containerNode` is inaccessible or null.
- `getSnapshotBeforeUpdate` allows us to access the `this.containerNode` before `onClose` on closing condition. Therefore` shouldSpotActivator` state is removed and passed down as the snapshot.

### Additional Considerations
- Previously, `ContextualPopopDecorator` only rendered once on closing, but with this change, it will render twice. Once per `open` prop change and then subsequent render caused by `setState` in `onClose` handler.

### DIFF
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -268,7 +268,7 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 						// is on the activator and we want to re-spot it so a11y read out can occur
 						current === prevState.activator ||
 						// is within the popup
-						this.containerNode.contains(Spotlight.getCurrent())
+						this.containerNode.contains(current)
 					)
 				};
 			}
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -482,21 +482,21 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			forward('onClose')
 		)
 
-		handleOpen = () => {
-			forward('onOpen', this.props);
+		handleOpen = (ev) => {
+			forward('onOpen', ev, this.props);
 			this.positionContextualPopup();
 			const current = Spotlight.getCurrent();
 			this.updateLeaveFor(current);
-			this.setState(() => ({
+			this.setState({
 				activator: current
-			}));
+			});
 		}
 
 		handleClose = () => {
 			this.updateLeaveFor(null);
-			this.setState(() => ({
+			this.setState({
 				activator: null
-			}));
+			});
 		}
 
 		handleDirectionalKey (ev) {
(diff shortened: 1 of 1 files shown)


packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -259,47 +258,38 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
...
+				const current = Spotlight.getCurrent();
+				return {
 					shouldSpotActivator: (
 						// isn't set
 						!current ||
 						// is on the activator and we want to re-spot it so a11y read out can occur
-						current === state.activator ||
+						current === prevState.activator ||
 						// is within the popup
-						this.containerNode.contains(current)
```
sugardave: Why does this have to change?

viodragon2: good catch

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -495,6 +482,23 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			forward('onClose')
 		)
 
+		handleOpen = () => {
+			forward('onOpen', this.props);
```
ryanjduffy: You've omitted the event payload here. It should likely be forwarded on from the param passed to `handleOpen`.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -495,6 +482,23 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
...
+			this.positionContextualPopup();
+			const current = Spotlight.getCurrent();
+			this.updateLeaveFor(current);
+			this.setState(() => ({
```
ryanjduffy: This (and on line 497) may be better with passing the object rather than a function since the new state isn't dependent upon the previous. WDYT?

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
viodragon2: :disappear:

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
viodragon2: No objection here. (Though I am curious though how react handles it differently)

ryanjduffy: Probably no effective difference beyond ms or less perf improvement avoiding creating the function and calling it. For me, I think the readability benefit is more compelling.

PR #2077, opened 2019-01-24
PLAT-72734: Update DateTimeDecorator life cycle methods

viodragon2:
### Checklist

Update `moonstone/internal/DateTimeDecorator` life cycle methods

### Resolution
- `date` prop converts from JS `Date()` to unix time and stored as `state.date`. `static getDerivedStateFromProps` handles the job for conversion.
- Since `setState` calls were invoked only when `open` props changed, we can achieve the same effect by handling it in `onOpen` and `onClose` handlers.
- `initI18n` is moved to `render` and added `I18nContextDecorator` to handle `locale` change through props.

### Additional Considerations
- Wonder if `memoize` is necessary for `toTime` function.
- We could use `static getDerivedStateFromProps` to handle `open` state and pass down a new state down the life cycle, but the timing of `emitChange` will then have to happen in `componentDidUpdate`. This isn't wrong by any sense as it happens in the same update life cycle, but `onChange` could fire before the `render` is complete if we handle it in `onOpen`.

### DIFF
--- a/packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
+++ b/packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
@@ -16,6 +16,18 @@ import React from 'react';
 
 import {Expandable} from '../../ExpandableItem';
 
+
+/**
+ * Converts a JavaScript Date to unix time
+ *
+ * @param	{Date}	date	A Date to convert
+ *
+ * @returns	{undefined}
+ */
+const toTime = (date) => {
+	return date && date.getTime();
+};
+
 /**
  * {@link moonstone/internal/DateTimeDecorator.DateTimeDecorator} provides common behavior for
  * {@link moonstone/DatePicker.DatePicker} and {@link moonstone/TimePicker.TimePicker}.
@@ -63,7 +75,7 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
 		constructor (props) {
 			super(props);
 
-			const initialValue = this.toTime(props.value);
+			const initialValue = toTime(props.value);
 			const value = props.open && !initialValue ? Date.now() : initialValue;
 			this.state = {initialValue, value};
 
@@ -73,6 +85,16 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
 					this.handlers[name] = this.handlePickerChange.bind(this, handlers[name]);
 				});
 			}
+		}
+
+		static getDerivedStateFromProps (props, state) {
+			const propValue = toTime(props.value);
+			if (state.value !== propValue) {
+				return {
+					value: propValue
+				};
+			}
+			return null;
 		}
 
 		initI18n () {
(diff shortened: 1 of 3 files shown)


packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
```diff
@@ -211,12 +213,12 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
 			let order = defaultOrder;
 
 			// Guard for isomorphic builds
-			if (this.i18nContext) {
+			if (typeof window !== 'undefined' && i18n) {
 				if (value) {
-					label = this.i18nContext.formatter.format(value);
+					label = i18n().formatter.format(value);
```
ryanjduffy: The `i18n()` callback has some cost to it so we should minimize invoking it. At a minimum, a single call here which stores the return value would be an improvement. Also, I think it'd be a good candidate for memoization based on the locale prop which would give us comparable performance to the current implementation, I expect.

packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
```diff
@@ -169,17 +149,39 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
...
+		}
+
+		handleOpen = (ev) => {
+			forward('onOpen', ev, this.props);
+
+			const newValue = toTime(this.props.value);
+			const value = newValue || Date.now();
+
+			// if we're opening, store the current value as the initial value for cancellation
+			this.setState({
```
ryanjduffy: Unrelated to this, I noticed that cancellation doesn't work any longer for DatePicker and TimePicker. I haven't bisected to find the root cause yet but we should track it down since without it storing the initial value is not productive.

packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
```diff
@@ -76,42 +95,14 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
...
-			const locale = ilib.getLocale();
-
-			if (i18n && this.locale !== locale && typeof window === 'object') {
-				this.locale = locale;
-				this.i18nContext = i18n();
+		static getDerivedStateFromProps (props, state) {
+			const propValue = toTime(props.value);
+			if (state.value !== propValue) {
+				return {
+					value: propValue
```
ryanjduffy: Is the convention `prevValue`?

viodragon2: The intention here is not to cache, but only to convert the current prop value.

packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
viodragon2: Let's create a ticket for the bug. It seems like it's been broken for a very long time.

PR #2078, opened 2019-01-24
PLAT-72739: Update Spottable lifecycle method

germboy:
### Checklist

Updated the `UNSAFE_componentWillReceiveProps` lifecycle method to use `getDerivedStateFromProps` instead.

This new lifecycle doesn't allow access to the `this.isFocused` instance property, so focus-awareness is now moved to `state`. Since we don't want to re-render simply when a component gains focus, I've added a `shouldComponentUpdate` method to verify if any of the other state/prop values has changed and am allowing updates accordingly.


ryanjduffy:
Confirmed by testing visually and monitoring the state values in React dev tools for both `focused` and `focusedWhenDisabled` (for a Picker).

PR #2079, opened 2019-01-24
Spottable: Fix multiple click event when keyboard accessible node is …

germboy:
…selected via 5way enter

### Checklist

`Spottable` was not preventing the native generated click event when using 5way enter to select a keyboard accessible node (ex: `<button>`). This was resulting in 2 click events being emitted per enter press.

### Resolution
We needed to check if the selection key that is pressed is a directional key - if it is not, then we verify its `isKeyboardAccesible` value and can prevent the default behavior.

### DIFF
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -262,6 +262,7 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 
 			const {currentTarget, repeat, type, which} = ev;
 			const {selectionKeys} = this.state;
+			const keyboardAccessible = isKeyboardAccessible(currentTarget);
 
 			const keyCode = selectionKeys.find((value) => (
 				// emulate mouse events for any remote okay button event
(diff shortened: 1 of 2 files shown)


packages/spotlight/Spottable/Spottable.js
```diff
@@ -270,13 +271,15 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
...
 					which === value &&
-					(type !== 'keypress' || !isKeyboardAccessible(currentTarget))
+					(type !== 'keypress' || !keyboardAccessible)
 				)
 			));
 
 			if (getDirection(keyCode)) {
 				preventDefault(ev);
 				stop(ev);
+			} else if (keyboardAccessible) {
```
ryanjduffy: I think we should further limit it to matching selection keys. Right now, it will call `preventDefault` for other key presses which I don't think is desired. Do you agree?

germboy: Ah, good point. I mistakenly thought there was a guard for that in place already. I've updated this line to include it. Good 👀

PR #2080, opened 2019-01-24
PLAT-72715: ViewManager  update lifecyle methods

haileyR:
### Checklist

update ViewManager and View lifecyle methods

### Resolution
Used `getDerivedStateFromProps` and made it use component states.


packages/ui/ViewManager/View.js
```diff
@@ -135,13 +135,19 @@ class View extends React.Component {
...
 		};
 	}
 
-	UNSAFE_componentWillReceiveProps (nextProps) {
-		// changeDirection let's us know we need to switch mid-transition
-		this.changeDirection = this.animation ? this.props.reverseTransition !== nextProps.reverseTransition : false;
+	static getDerivedStateFromProps (props, state) {
+		if (state.reverseTransition !== props.reverseTransition) {
+			return {
+				changeDirection: true
```
viodragon2: I think you meant to use this state here ?https://github.com/enactjs/enact/blob/be9ec0cb07698b0b19fdcb39488a389441dbbc42/packages/ui/ViewManager/View.js#L292

packages/ui/ViewManager/ViewManager.js
```diff
@@ -189,33 +189,33 @@ class ViewManager extends React.Component {
...
-		// null or undefined => determine automatically
-		if (nextProps.reverseTransition != null) {
-			this.reverseTransition = !!nextProps.reverseTransition;
-		} else if (this.props.index !== nextProps.index) {
-			this.reverseTransition = this.props.index > nextProps.index;
+	static getDerivedStateFromProps (props, state) {
+		let reverseTransition;
+		if (props.reverseTransition != null) {
+			reverseTransition = !!props.reverseTransition;
+		} else if (props.index !== state.index) {
```
viodragon2: There's no `index` state. Did you mean `previousIndex`?

packages/ui/ViewManager/ViewManager.js
```diff
@@ -189,33 +189,33 @@ class ViewManager extends React.Component {
 		index: 0
 	}
 
-	UNSAFE_componentWillReceiveProps (nextProps) {
-		this.previousIndex = this.props.index;
-		this.checkReverse(nextProps);
+	constructor (props) {
+		super(props);
+		this.state = {
+			previousIndex: null,
```
viodragon2: minor nit: Let's keep consistency with other changes and name it `prevIndex` to shorten the variable name

packages/ui/ViewManager/ViewManager.js
```diff
@@ -189,33 +189,33 @@ class ViewManager extends React.Component {
...
+		if (props.reverseTransition != null) {
+			reverseTransition = !!props.reverseTransition;
+		} else if (props.index !== state.index) {
+			reverseTransition = state.index > props.index;
 		} else {
-			this.reverseTransition = false;
+			reverseTransition = false;
 		}
+
+		return {
```
viodragon2: We shouldn't always return a new derived state in `getDerivedStateFromProps`. See https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#anti-pattern-unconditionally-copying-props-to-state

packages/ui/ViewManager/View.js
```diff
@@ -135,13 +135,19 @@ class View extends React.Component {
...
+			changeDirection: false,
+			reverseTransition: props.reverseTransition
 		};
 	}
 
-	UNSAFE_componentWillReceiveProps (nextProps) {
-		// changeDirection let's us know we need to switch mid-transition
-		this.changeDirection = this.animation ? this.props.reverseTransition !== nextProps.reverseTransition : false;
+	static getDerivedStateFromProps (props, state) {
+		if (state.reverseTransition !== props.reverseTransition) {
```
viodragon2: I don't see where `reverseTransition` state is being utilized other than the check here. How does this work?

haileyR: to save it to check with previous prop value later?

packages/ui/ViewManager/tests/ViewManager-specs.js
```diff
@@ -203,4 +202,58 @@ describe('ViewManager', () => {
 		expect(actual).toBe(expected);
 		done();
 	});
+
+	test('should update the View reverseTransition prop.', done => {
+		const subject = mount(
+			<ViewManager noAnimation index={0} duration={0}>
```
viodragon2: 1. Testing `reverseTransition` without the `arranger` has no effect. Please update accordingly.
2. Do you think `noAnimation` makes difference as well?

packages/ui/ViewManager/tests/ViewManager-specs.js
```diff
@@ -203,4 +202,58 @@ describe('ViewManager', () => {
...
+				<div className="view">View 1</div>
+				<div className="view">View 2</div>
+				<div className="view">View 3</div>
+				<div className="view">View 4</div>
+				<div className="view">View 5</div>
+			</ViewManager>
+		);
+
+		subject.setProps({index: 1});
+		const actual = subject.find('View').props().reverseTransition;
```
viodragon2: The selector does not refer to the `className`. Should we remove it or should it use `className` as the selector?

packages/ui/ViewManager/View.js
```diff
@@ -135,13 +135,19 @@ class View extends React.Component {
...
 		};
 	}
 
-	UNSAFE_componentWillReceiveProps (nextProps) {
-		// changeDirection let's us know we need to switch mid-transition
-		this.changeDirection = this.animation ? this.props.reverseTransition !== nextProps.reverseTransition : false;
+	static getDerivedStateFromProps (props, state) {
+		if (state.prevReverseTransition !== props.reverseTransition) {
+			return {
+				changeDirection: true
```
viodragon2: Shouldn't we cache `prevReverseTransition` or was this intentional? Plus, it is understandable that there's a rare case when `changeDirection` needs to be reset to `false` once after set to `true`, but shouldn't we have a reset condition somewhere? This leaves a room for an error IMO.

packages/ui/ViewManager/tests/ViewManager-specs.js
haileyR: It has effect. It updates the `reverseTransition` prop of View component and this test is checking View's `reverseTransition`. This test is for testing if ViewManager's `getDerivedStateFromProps` method is setting its own `reverseTransition` state according to its new props and pass it to View. `noAnimation` just followed from copy and paste from the last test, this has nothing to with this test. It can be removed or left as it is. 
Lmk if I'm not understanding how arranger work in ViewManager.

packages/ui/ViewManager/View.js
haileyR: I guess that might happen, but where should I add it?

packages/ui/ViewManager/tests/ViewManager-specs.js
viodragon2: `<div className="view">View 1</div>` ? 
See other tests for comparison. The result would be the same as there's only one `children` rendered unless we try to capture in mid-transition. I was pointing out that you added `className`s for each view but not it was never used as part of the test.

packages/ui/ViewManager/tests/ViewManager-specs.js
viodragon2: I should've put the comment in a different line to make it clear.

From `View`'s perspective, `reverseTransition` prop is only meaningful when `ViewManager` has `arranger` prop declared.  If you look at how `prepareTransition()` is called in `View`, `reverseTransition` is respected only if there's `arranger` prop. And similarly, `noAnimation` will disregard `reverseTransition` in `View` as well. This test verifies whether the prop is passed down correctly to its children, and this is fine. But this would never be the normal usage. I'm on the fence that our test cases should cover the actual usages. WDYT?

packages/ui/ViewManager/View.js
viodragon2: Do you think there's a way to capture mid-transition state?

packages/ui/ViewManager/tests/ViewManager-specs.js
haileyR: I added this test for add unit tests for `reverseTransition` props?" and to test if this lifecycle change is breaking the original logic or not. I think you want to add some other test cases.

viodragon2: Perhaps? This could be in the scope of `View` rather than `ViewManager` depending on how you look at it. But I'm arguing that `reverseTransition` won't have any impact on how the component behaves without these props. Why does it matter if it passes down to children when it doesn't consume?

packages/ui/ViewManager/View.js
haileyR: i don't know

PR #2082, opened 2019-01-25
PLAT-72742: Update ui/Touchable lifecycle method

germboy:
### Checklist

Updating the `ui/Touchable` lifecycle method to not use `UNSAFE_componentWillReceiveProps`

### DIFF
--- a/packages/ui/Touchable/Touchable.js
+++ b/packages/ui/Touchable/Touchable.js
@@ -422,7 +422,7 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
 
 			if (prevDisabled !== disabled) {
 				return {
-					...activeProp && !prevDisabled && disabled && deactivate(state),
+					...(activeProp && !prevDisabled && disabled && deactivate(state)),
 					disabled
 				};
 			}
--- a/packages/ui/Touchable/Touchable.js
+++ b/packages/ui/Touchable/Touchable.js
@@ -379,7 +379,7 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
 
 			this.state = {
 				active: States.Inactive,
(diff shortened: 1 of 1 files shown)


packages/ui/Touchable/Touchable.js
```diff
@@ -415,6 +416,19 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
 			handleGlobalMove.bindAs(this, 'handleGlobalMove');
 		}
 
+		static getDerivedStateFromProps (props, state) {
+			const {disabled} = props;
+			const {disabled: prevDisabled} = state;
+
+			if (prevDisabled !== disabled) {
+				return {
+					...activeProp && !prevDisabled && disabled && deactivate(state),
```
viodragon2: This part is a bit confusing to me.
`activeProp` is a String. Why do we need to spread it?
Also, does returning a function do the right thing?

germboy: This line is a convenience method for conditionally setting an object key on the returned (next) state. If `activeProp && !prevDisabled && disabled` then we spread the value of `deactivate(state)` (which returns an object, ultimately containing the `active` key `{active: false}`) to the next state in the `return`. That's where the `...` spread is used. The `&&` operators take precedence here, so we're not spreading `activeProp`, we're spreading `deactivate(state)` if the line evals. If it doesn't eval, then nothing gets spread. If it helps, it's the same as `...(activeProp && !prevDisabled && disabled && deactivate(state)),` This is being done rather than defining an object, then mutating the object keys/props based on conditions across multiple lines.

germboy: Fair point - I've added the parenthesis for clarity.

PR #2083, opened 2019-01-25
PLAT-72726: Update VideoPlayer life cycle methods

viodragon2:
### Checklist

Update VideoPlayer life cycle methods. 

### Resolution
**General**
_style_ updates to components are moved to `componentDidUpdate` life cycle
- Remove `style` attribute from JSX
- Remove any _state_ keeping track of the DOM values as managing DOM values via _state_ can result in different values depending on which life cycle it's accessed from

**MediaControls**
- Counting the left/right components and resetting width of the controls is moved from `componentWillReceiveProps` to `componentDidUpdate`
- Setting `showMoreComponents` on `visible` prop change is unnecessary as it is set in associated event handlers

**VideoPlayer**
- Add `I18nContextDecorator` to handle `locale` change via prop
- Change `durfmt` to *state* as it depends on locale change
- Set title offset height in `componentDidUpdate`

**MediaTitle**
- Forwards `ref` for VideoPlayer to access the DOM

### Additional Considerations
- It is possible to measure `MediaTitle` height in `getSnapshotBeforeUpdate`, but we're interested in the value of height after it's been rendered in this case. Therefore `componentDidUpdate` is more suitable.

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -669,6 +669,15 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {	// eslint-disable-line
 			}
 		}
 
+		static getDerivedStateFromProps (props) {
+			if (!props.visible) {
+				return {
+					showMoreComponents: false
+				};
+			}
+			return null;
+		}
+
 		componentDidMount () {
 			this.calculateMaxComponentCount(
 				countReactChildren(this.props.leftComponents),
--- a/packages/moonstone/VideoPlayer/MediaControls.js
+++ b/packages/moonstone/VideoPlayer/MediaControls.js
@@ -679,11 +679,11 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {	// eslint-disable-line
 			on('keyup', this.handleKeyUp);
 		}
 
-		UNSAFE_componentWillReceiveProps (nextProps) {
+		componentDidUpdate (prevProps, prevState) {
 			// Detect if the number of components has changed
-			const leftCount = countReactChildren(nextProps.leftComponents),
-				rightCount = countReactChildren(nextProps.rightComponents),
-				childrenCount = countReactChildren(nextProps.children);
+			const leftCount = countReactChildren(prevProps.leftComponents),
+				rightCount = countReactChildren(prevProps.rightComponents),
+				childrenCount = countReactChildren(prevProps.children);
 
 			if (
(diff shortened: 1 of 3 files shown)


packages/moonstone/VideoPlayer/MediaControls.js
```diff
@@ -693,16 +693,6 @@ const MediaControlsDecorator = hoc((config, Wrapped) => {	// eslint-disable-line
 				this.calculateMaxComponentCount(leftCount, rightCount, childrenCount);
 			}
 
-			if (this.props.visible && !nextProps.visible) {
```
ryanjduffy: I'm not sure about removing this. I couldn't see any place that `showMoreComponents` was reset to false beyond the imperative methods exposed by `ApiDecorator`. Can you elaborate on why this is safe?

ryanjduffy: In testing this, it appears that if you click "more", click the video to hide the controls, and click again to show them, it shows the more controls rather than the playback. I'm assuming that's due to this change.

ryanjduffy: Possible fix :)
```js
static getDerivedStateFromProps (props) {
	if (!props.visible) {
		return {
			showMoreComponents: false
		};
	}

	return null;
}
```

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -658,14 +665,26 @@ const VideoPlayerBase = class extends React.Component {
...
 
 		if (props.setApiProvider) {
 			props.setApiProvider(this);
 		}
 	}
 
+	static getDerivedStateFromProps (props, state) {
+		if (typeof window === 'object' && props.locale !== state.prevLocale) {
+			return {
+				durfmt: new DurationFmt({length: 'medium', style: 'clock', useNative: false}),
```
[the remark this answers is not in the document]
viodragon2: :awesome:

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -1818,6 +1813,8 @@ const VideoPlayerBase = class extends React.Component {
 			proportionSelection = selection.map(t => t / this.state.duration);
 		}
 
+		const durFmt = this.getDurFmt(locale);
```
ryanjduffy: Pushed a change to remove setting `durFmt` to an instance member in order to keep our render methods idempotent. Since the method is memoized, the cost of invoking it elsewhere is only the lookup cost.

PR #2084, opened 2019-01-28
PLAT-64279: High Contrast Support

Djspaceg:
### Checklist

Moonstone needs to support alternate color palettes (high contrast mode) without interfering with the existing basic skin assignments.

### Resolution
Added a system to handle "variants" of a skin, with a prop called `skinVariants` to `ui/Skinnable` which allows a theme to add special variations of a skin's rules. 

### Additional Considerations
This collection of changes is built with future implementations in mind. When the browser support for `@media prefers-contrast` is added, that can be a drop-in addition to this implementation to automatically enable high-contrast mode.

### Links
* https://drafts.csswg.org/mediaqueries-5/#prefers-contrast


Djspaceg:
The `Button` and `Item` stories that are modified in this PR will be removed before final merging. They aid in validating that the code is indeed working as expected.

PR #2086, opened 2019-01-30
ENYO-5766: [qa-sampler] Fixed qa-sampler not being refreshed when entering values for knob

Baekwoo-s:
### Checklist

When you change the knob's value in the scroller qa-sampler, the app is refreshed when you press enter key.

### Resolution
Modified input tag to slider tag so that you can test without using Enter key

### Additional Considerations
1. The order of the props to pass to the number function was incorrect.
Incorrect
number('Spacer size', {min: 0, max: 300}, 100);
Correct
number('Spacer size', 100, {min: 0, max: 300});

2. Fixed overlapping of text when `Spacer size` knob is reduced in height

### DIFF
--- a/packages/sampler/stories/qa/Scroller.js
+++ b/packages/sampler/stories/qa/Scroller.js
@@ -324,10 +324,10 @@ storiesOf('Scroller', module)
 		() => {
 			const size = number('Spacer size', 100, {max: 300, min: 0, range: true});
 			return (
-				<Scroller style={{height: 200}}>
+				<Scroller style={{height: ri.unit(200, 'rem')}}>
 					<Item>1</Item>
-					<div style={{height: size, paddingLeft: 40}}>{size}px Spacer</div>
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Scroller.js
```diff
@@ -321,11 +322,11 @@ storiesOf('Scroller', module)
 	.add(
 		'Test scrolling to boundary with small overflow',
 		() => {
-			const size = number('Spacer size', {min: 0, max: 300}, 100);
+			const size = number('Spacer size', 100, {max: 300, min: 0, range: true});
 			return (
 				<Scroller style={{height: 200}}>
 					<Item>1</Item>
-					<div style={{height: size}}>{size}px Spacer</div>
+					<div style={{height: size, paddingLeft: 40}}>{size}px Spacer</div>
```
ybsung: Please apply ri.scale to paddindLeft.

ybsung:
LGTM

PR #2087, opened 2019-01-30
PLAT-72716: Picker and Scrollable - Update lifecycle method

0x64:
### Checklist

Remove _componentWillReceiveProps_ from Picker and _componentWillUpdate_ from Scrollable.

### Resolution
Validation code of Picker is moved to _render_ function to validate values both for mounting and updating cases.
Code for _scrollTo_ method of Scrollable is moved to _componentDidUpdate_.

### Additional Considerations
If we don't need to validate values of Picker for mounting, we can move the code into _getDerivedStateFromProps_ from _componentWillReceiveProps_ instead.

For Scrollable, the behavior of before and after this PR is identical, but I found that the code does not work as intention in some cases due to the changed behavior of React life cycle. We need to take a look at this issue separately.

### Links
PLAT-72716

### DIFF
--- a/packages/moonstone/internal/Picker/Picker.js
+++ b/packages/moonstone/internal/Picker/Picker.js
@@ -419,18 +419,6 @@ const PickerBase = class extends React.Component {
 		}
 	}
 
-	UNSAFE_componentWillReceiveProps (nextProps) {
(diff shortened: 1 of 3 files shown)


ybsung:
LGTM

PR #2093, opened 2019-02-01
PLAT-72718: Scrollable and VirtualList: Update lifecycle method

0x64:
### Checklist

Removed deprecated _componentWillReceiveProps_ methods used to configure Spotlight containers.

### Resolution
Replaced _componentWillReceiveProps_ with _componentDidUpdate_ method.

### Additional Considerations
There was a missing code to reduce the number of calls of _Spotlight.set_ in Scrollable, so I fixed it also.

### Links
PLAT-72718


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -264,6 +258,7 @@ const VirtualListBaseFactory = (type) => {
 		nodeIndexToBeFocused = null
 		preservedIndex = null
 		restoreLastFocused = false
+		lastConfiguredSpotlightId = null
```
ybsung: How about `prevSpotlightId` as you did in `Scrollable` instead of `lastConfiguredSpotlightId`?

ybsung:
LGTM

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -236,10 +236,28 @@ class ScrollableBase extends Component {
...
 
-	UNSAFE_componentWillReceiveProps (nextProps) {
-		configureSpotlightContainer(nextProps);
+	static getDerivedStateFromProps (props, state) {
+		const
+			{'data-spotlight-id': spotlightId, focusableScrollbar} = props,
+			{prevSpotlightId, prevFocusableScrollbar} = state;
+
+		if (prevSpotlightId !== spotlightId || prevFocusableScrollbar !== focusableScrollbar) {
+			configureSpotlightContainer(props);
```
webOS101: We really, really shouldn't be doing side effects within this static method.  Also, there should be a better way to deal with this other than adding state just for the sole purpose of having it available within `getDerivedStateFromProps`.

0x64:
I updated to remove side effects from the static method.

0x64:
The call in constructor looks good for now. To reduce calls to Spotlight, I left code to check the modification of props and the initial Spotlight configuring is needed at least once at the beginning.

0x64:
> Any reason why we can't run `configureSpotlightContainer` in `componentDidUpdate`?

`render` is executed in _render phase_ and `componentDidUpdate` is in _commit phase_, so I just wanted to set up spotlight config in _render phase_ since a focus could be moved out of a scrollable area if a user input occurs during updates. But, I think now that I don't need to consider such an abnormal scenario which cannot happen in apps. So, it looks better to move the code. I will update. Thanks.

0x64:
Updated.

viodragon2:
I was thinking more like this https://github.com/enactjs/enact/compare/develop...feature/PLAT-72718-stephen

0x64:
Oops, what a silly mistake I did! Thanks.

PR #2094, opened 2019-02-01
PLAT-72723: ui/VirtualList update lifecycle method 

ybsung:
### Checklist

`UNSAFE_componentWillMount()`, `UNSAFE_componentWillReceiveProps()`, and `UNSAFE_componentWillUpdate()` will be deprecated.

### Resolution

- Move the code in `UNSAFE_componentWillMount()` to `constructor()`. (https://github.com/enactjs/enact/commit/ecbf47c23fda9cb486288cd2f541151d355a7d56)
- Move the code in `UNSAFE_componentWillReceiveProps()` to render phase. (https://github.com/enactjs/enact/commit/ecbf47c23fda9cb486288cd2f541151d355a7d56)
- Move the code to update state in `UNSAFE_componentWillReceiveProps()` to `getDerivedStateFromProps()` and the code to update DOM elements in `UNSAFE_componentWillReceiveProps()` to render phase.

To move the code to getDerivedStateFromProps(),
- I converted all instance variables to state variables so that those variables could be used in that function. e.g> this.scrollBounds -> this.state.scrollBounds
- I converted all instance functions called in that function to local functions so that those functions could be called in that (static) function, then moved those functions to `metrics.js`.
- I created `hasDataSizeChanged()` and `hasMetricsChanged()` used in `getDerivedStateFromProps()` and `render()`.
- I created `calculateMaxFirstIndex` like other functions with `calculate` prefix.
- I moved `this.syncThreshold()` out of  `calculateScrollBounds` to reduce the responsibility of the function.
- I returned the updated state instead of updating the state directly in the functions with prefix `calculate`.

### Additional Considerations

I added `componentDidUpdate()` function because it would be better to update `this.prevFirstIndex` in it after rendering.

### Links

PLAT-72723


ybsung:
I'm not sure about my last commit - https://github.com/enactjs/enact/pull/2094/commits/a085f09016617142b8154a2928c6cc86e311867f.

`prevProps` state is usually updated in `getDerivedStateFromPops`. But the `prevProps` state before updating is needed in render phase of my PR. So I updated that state in render() function.

0x64:
LGTM

0x64:
Fixed typo.

ybsung:
Closing in favor of #2108

PR #2096, opened 2019-02-01
ENYO-5780: Apply throttle to onmousemove event

changgilee:
### Checklist

CPU usage is increase rapidly during cursor navigation in enact app.

### Resolution
In current status, spotlight use onmousemove event and handler function called many times during cursor move. For resolve this issue, I try to test using throttle.

### Additional Considerations

### Links
ENYO-5780

### Comments

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -456,6 +456,26 @@ const Spotlight = (function () {
 		}
 	}
 
+	let throttleTimer = null;
+	let throttleClientX, throttleClientY;
+
+	function throttleMouseMove (e) {
+		throttleClientX = e.clientX;
+		throttleClientY = e.clientY;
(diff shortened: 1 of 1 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -456,6 +456,26 @@ const Spotlight = (function () {
 		}
 	}
 
+	let throttleTimer = null;
+	let throttleClientX, throttleClientY;
+
+	function throttleMouseMove (e) {
```
sugardave: Can this throttle be implemented using `enact/core/Job` instead of a direct `setTimeout`?  Then we could compare the CPU% numbers from `start` (uses `setTimeout`) and `startRaf` (uses 'requestAnimationFrame`) and see if there is a noticeable difference.

PR #2099, opened 2019-02-07
ENYO-5794: Set default value for Slider

germboy:
### Checklist

A recent change to `Changeable` expects the config `prop` value to be set on creation. `moonstone/Slider` uses this to manage the `value` prop. This value is not set on creation, rather letting a child component `ui/Slider` set the value (default is `min`). `moonstone/Slider`'s use of `Changeable` cannot access that value, so it looks like we also need to set the default value in `moonstone/Slider`.

### DIFF
--- a/packages/moonstone/Slider/Slider.js
+++ b/packages/moonstone/Slider/Slider.js
@@ -254,7 +254,8 @@ const SliderBase = kind({
 			activateOnFocus,
 			active
 		}),
-		tooltip: ({tooltip}) => tooltip === true ? ProgressBarTooltip : tooltip
+		tooltip: ({tooltip}) => tooltip === true ? ProgressBarTooltip : tooltip,
+		value: ({min, value}) => value || min
 	},
 
 	render: ({css, focused, tooltip, ...rest}) => {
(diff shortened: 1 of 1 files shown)


packages/moonstone/Slider/Slider.js
```diff
@@ -254,7 +254,8 @@ const SliderBase = kind({
 			activateOnFocus,
 			active
 		}),
-		tooltip: ({tooltip}) => tooltip === true ? ProgressBarTooltip : tooltip
+		tooltip: ({tooltip}) => tooltip === true ? ProgressBarTooltip : tooltip,
+		value: ({min, value}) => value || min
```
webOS101: This will force a zero value to be `min`.  We have to check against `value == null`.  Can someone whip up a patch quick-like?

germboy: Ah, good call. Looks like setting a default of `min` for `value` (which I did improperly above) would work without us having to type-check. I've updated the logic.

germboy:
Closing in favor of #2103

PR #2100, opened 2019-02-07
ENYO-5796: Fix large text knob value assignment

ryanjduffy:
### Checklist

The knob assignment logic was incorrectly assigning a string to a boolean knob

### Resolution
Move the ternary outside the knob to pass the string to the prop rather than the knob

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -158,7 +158,7 @@ const StorybookDecorator = (story, config) => {
 			title={`${config.kind} ${config.story}`.trim()}
 			description={config.description}
 			locale={selectV2('locale', locales, 'en-US', globalGroup)}
-			textSize={boolean('large text', Config, (getPropFromURL('large text') === 'true')) ? 'large' : 'normal'}
+			textSize={boolean('large text', Config, getPropFromURL('large text') === 'true') ? 'large' : 'normal'}
 			highContrast={boolean('high contrast', Config, (getPropFromURL('high contrast') === 'true'))}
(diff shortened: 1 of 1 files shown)


packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -158,7 +158,7 @@ const StorybookDecorator = (story, config) => {
 			title={`${config.kind} ${config.story}`.trim()}
 			description={config.description}
 			locale={selectV2('locale', locales, 'en-US', globalGroup)}
-			textSize={boolean('large text', Config, (getPropFromURL('large text') === 'true')) ? 'large' : 'normal'}
+			textSize={boolean('large text', Config, getPropFromURL('large text') === 'true') ? 'large' : 'normal'}
```
Djspaceg: I'm not seeing how removing the set of parenthesis changes the result here. The ternary is still in the same place, functionally. Would you mind explaining in more detail?

ryanjduffy: You're right ... I'm not sure why the result was different. I'll debug some more and either revise or explain.

ryanjduffy: Not working now! Glad you noticed that. I wonder what other magic was in play.

ryanjduffy: Back to my original analysis (which I didn't include on the ticket because I thought it was wrong) that `getPropFromUrl` is interfering. Storybook doesn't seem to reset the URL when reset is clicked but instead sometime after render (I didn't debug the specifics). Will close this PR and share my notes on in JIRA.

PR #2101, opened 2019-02-07
ENYO-5793: Restore OPENING state in Popup

viodragon2:
### Checklist

In an attempt to simplify Popup's open logic in #2068, we removed mid-`OPENING` state of the transition container. However, since `ui/Transition` relies on `onTransitionEnd` event to notify whether it is opened or not, changing the `visible` state in the midst of transition makes states to go out of sync.

### Resolution
Restored `OPENING` state in Popup. We can revisit later on cleaning up Popup's code, but I think this is okay FWIW.

### DIFF
--- a/packages/moonstone/Popup/Popup.js
+++ b/packages/moonstone/Popup/Popup.js
@@ -404,7 +404,7 @@ class Popup extends React.Component {
 		if (props.open !== state.prevOpen) {
 			if (props.open) {
 				return {
-					popupOpen: props.noAnimation ? OpenState.OPEN : OpenState.CLOSED,
+					popupOpen: props.noAnimation || state.floatLayerOpen ? OpenState.OPEN : OpenState.CLOSED,
 					floatLayerOpen: true,
 					activator: Spotlight.getCurrent(),
 					prevOpen: props.open
--- a/packages/moonstone/Popup/Popup.js
+++ b/packages/moonstone/Popup/Popup.js
@@ -241,6 +241,12 @@ const checkScrimNone = (props) => {
 		'is not supported. Use a transparent scrim to prevent spotlight focus outside of the popup');
 };
 
+const OpenState = {
+	CLOSED: 0,
+	OPENING: 1,
+	OPEN: 2
+};
+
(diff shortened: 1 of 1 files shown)


packages/moonstone/Popup/Popup.js
```diff
@@ -383,24 +389,34 @@ class Popup extends React.Component {
...
-				floatLayerOpen: true
-			};
-		} else if (!props.open && state.popupOpen) {
-			return {
-				popupOpen: false
-			};
+		if (props.open !== state.prevOpen) {
+			if (props.open) {
+				return {
+					popupOpen: props.noAnimation ? OpenState.OPEN : OpenState.CLOSED,
```
ryanjduffy: I noticed one situation in which this condition doesn't quite work. This isn't actually a regression from your last update but instead my previous change that introduced `OpenState`. In sampler, if you check the `open` knob, let the animation complete, and then quickly toggle `open` off and then on, the dialog will not be displayed.

From my analysis, it appears that since the floating layer isn't closed (because the popup didn't finish closing) we should check its state here as well and, if `true`, set `OpenState.OPEN`.

```
popupOpen: props.noAnimation || state.floatLayerOpen ? OpenState.OPEN : OpenState.CLOSED,
```

@viodragon2 - can you check this out and see if you come to the same conclusion?

viodragon2: Yup. That makes sense. 
On a side note, it would be nice if `Transition` supports mid transition events (e.g. `entering`) so that we can handle these cases.

ryanjduffy: Sure. That's one of the constraints we have with CSS transitions. We only get an "end" event and only if the triggering CSS prop has changed since the last paint/layout (i'm not clear which off hand, tbh).

Djspaceg: This is something I've been wanting to add since Enyo, but we didn't have a use-case, so I couldn't justify the feature when originally writing Enact's Transition. If we're still interested, let's create a ticket and start this process.

PR #2103, opened 2019-02-08
ENYO-5794: Revert logic for controlled state for Changeable/Toggleable

ryanjduffy:
### Checklist

We changed how we determined that an instance of Changeable or Toggleable was controller in #2057 but it has had some unintended side effects in our components and would likely in app usage as well.

### Resolution
Reverting that bit of logic in both coomponents

### Additional Considerations
May consider revising this in a future (e.g. 3.0) release.

### DIFF
--- a/packages/ui/Changeable/tests/Changeable-specs.js
+++ b/packages/ui/Changeable/tests/Changeable-specs.js
@@ -81,13 +81,13 @@ describe('Changeable', () => {
 			expect(actual).toBe(expected);
 		});
 
-		test('should use value prop when value prop is null', () => {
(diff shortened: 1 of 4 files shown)


packages/ui/Changeable/tests/Changeable-specs.js
```diff
@@ -81,13 +81,13 @@ describe('Changeable', () => {
 			expect(actual).toBe(expected);
 		});
 
-		test('should use value prop when value prop is null', () => {
+		test('should use defaultValue prop when value prop is null', () => {
```
viodragon2: Just leaving a note that we need to address this later as `null` can be a valid value.

PR #2105, opened 2019-02-08
PLAT-73583: Consolidated Picker's largeText rules into one section

Djspaceg:
### Checklist

With the changes to the CSS structure of supporting largeText as a skin variant, Picker is generating selectors that don't match the reality of the DOM structure any longer.

### Resolution
Rearrange the `.moon-custom-text` mixin rules in the Picker LESS file to apply the correct classes only to the base element. Secondarily, the rules have been consolidated into a single mixin, replicating the selector structure for better readability.

Includes updates to `ContextualPopup`, `LabeledItem`, and `Slider` as well.


Djspaceg:
Yeah, I encountered that and it seemed like it was ok, but it could easily be changed. Since you mentioned it, I'll switch it over to ensure it will always work as expected.

PR #2106, opened 2019-02-08
ENYO-5795: SliderBehaviorDecorator - prevent setting focus when disabled

germboy:
### Checklist

`SliderBehaviorDecorator` can be focused when it is disabled, which can result in unintended actions - such as a `Slider` displaying a tooltip when a click action occurs on the slider bar.

### Resolution
Preventing forwarding the focus event and setting an internal focused state when the component is `disabled` fixes the issue.


packages/moonstone/Slider/SliderBehaviorDecorator.js
```diff
@@ -49,6 +49,7 @@ const SliderBehaviorDecorator = hoc(defaultConfig, (config, Wrapped) => {
 
 		static propTypes = {
 			'aria-valuetext': PropTypes.oneOfType([PropTypes.string, PropTypes.number]),
+			disabled: PropTypes.boolean,
```
ryanjduffy: `bool` instead of `boolean` :)

PR #2107, opened 2019-02-08
ENYO-5815: Storybook Improvements

Djspaceg:
### Checklist

* Setting background images is not as useful as it should be
* Custom QUERY_STRING handling is unnecessary and can be offloaded into a module
* Several layers of DOM are unnecessary
* getPropFromURL is cumbersome to use for Boolean knobs
* Story description is never set
* Classes aren't being applied to MoonstoneEnvironment correctly

### Resolution
Numerous bits of refactoring. The most noticible bits are the background image is now blured and faded to improve text readability and it responds to the skin background color; as well as the story description now (finally) displays on each story.

### Additional Considerations
We may have to hide the descriptions if it causes problems, however they likely won't be technical, just preferential. The usefulness of the descriptions will improve over time, now that they'd be visible instead of just ignored.

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -108,27 +108,21 @@ const getArgs = (str) => {
 	return qs.parse(str || (typeof window !== 'undefined' ? window.parent.location.search : ''));
 };
 
-// This allows any prop/argument to be taken from the URL.
-const getPropFromArgs = (args, propName, fallbackValue) => {
-	if (args && typeof args[propName] !== 'undefined') {
-		return args[propName];
-	}
-	return fallbackValue;
-};
-
 // This allows any knob to be taken from the URL.
 const getKnobFromArgs = (args, propName, fallbackValue) => {
-	const knobVal = getPropFromArgs(args, 'knob-' + propName, fallbackValue);
-	try {
-		// If it's valid JSON, parse it and return
-		return JSON.parse(knobVal);
-	} catch (e) {
-		// If it's invalid, see what it is, and in the unknown case, just return it (string most likely).
-		if (typeof knobVal === 'undefined' || knobVal === 'undefined' || knobVal === 'void 0') {
-			return void 0;
+	const knob = 'knob-' + propName;
+	let value = fallbackValue;
+
+	if (args && knob in args) {
+		try {
+			// If it's valid JSON, parse it
+			value = JSON.parse(args[knob]);
+		} catch (e) {
+			// no handling required; allow fallbackValue to be used
 		}
-		return knobVal;
 	}
+
+	return value;
 };
 
 const StorybookDecorator = (story, config) => {
@@ -178,9 +172,7 @@ const StorybookDecorator = (story, config) => {
 			highContrast={boolean('high contrast', Config, getKnobFromArgs(args, 'high contrast'))}
 			style={{
 				'--moon-env-background': backgroundLabelMap[select('background', backgroundLabels, Config, getKnobFromArgs(args, 'background'))]
-				// '--moon-env-background-filter': 'blur(' + number('background blur', Config, {range: true, min: 0, max: 10, step: 0.5}, 3) + 'px)'
 			}}
-			// data-background={backgroundLabelMap[select('background', backgroundLabels, Config, getKnobFromArgs(args, 'background'))].background}
 			skin={select('skin', skins, Config, getKnobFromArgs(args, 'skin'))}
 		>
 			{sample}
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -4,10 +4,12 @@ import classnames from 'classnames';
 import kind from '@enact/core/kind';
 import React from 'react';
 import PropTypes from 'prop-types';
+import BodyText from '@enact/moonstone/BodyText';
 import MoonstoneDecorator from '@enact/moonstone/MoonstoneDecorator';
 import {Panels, Panel, Header} from '@enact/moonstone/Panels';
 import {boolean, select} from '../enact-knobs';
 import {selectV2} from '@storybook/addon-knobs';
+import qs from 'query-string';
 
 import css from './MoonstoneEnvironment.module.less';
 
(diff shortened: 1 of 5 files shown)


webOS101:
This does reveal we have two different styles of descriptions for the stories. :(

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -86,44 +89,53 @@ const backgroundLabels = {
...
+// This allows any knob to be taken from the URL.
+const getKnobFromArgs = (args, propName, fallbackValue) => {
+	const knobVal = getPropFromArgs(args, 'knob-' + propName, fallbackValue);
+	try {
+		// If it's valid JSON, parse it and return
+		return JSON.parse(knobVal);
+	} catch (e) {
+		// If it's invalid, see what it is, and in the unknown case, just return it (string most likely).
+		if (typeof knobVal === 'undefined' || knobVal === 'undefined' || knobVal === 'void 0') {
+			return void 0;
```
ryanjduffy: Is it important that it be `undefined` vs `null`? If so, would it be more readable to `return undefined;` with an eslint pragma to suppress the error?

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -157,11 +174,15 @@ const StorybookDecorator = (story, config) => {
...
-			textSize={boolean('large text', Config, (getPropFromURL('large text') === 'true')) ? 'large' : 'normal'}
-			highContrast={boolean('high contrast', Config, (getPropFromURL('high contrast') === 'true'))}
-			style={backgroundLabelMap[select('background', backgroundLabels, Config, getPropFromURL('background'))]}
-			skin={select('skin', skins, Config, getPropFromURL('skin'))}
+			locale={select('locale', locales, Config)}
+			textSize={boolean('large text', Config, getKnobFromArgs(args, 'large text')) ? 'large' : 'normal'}
+			highContrast={boolean('high contrast', Config, getKnobFromArgs(args, 'high contrast'))}
+			style={{
+				'--moon-env-background': backgroundLabelMap[select('background', backgroundLabels, Config, getKnobFromArgs(args, 'background'))]
+				// '--moon-env-background-filter': 'blur(' + number('background blur', Config, {range: true, min: 0, max: 10, step: 0.5}, 3) + 'px)'
```
ryanjduffy: Why have we left this (and L184) in here commented out? Should we remove or add a descriptive comment as to why they exist?

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.module.less
```diff
@@ -16,7 +17,17 @@
...
 
-.moonstone {
+.moonstoneEnvironmentPanels {
+	&::before {
+		content: "";
+		background: var(--moon-env-background);
+		position: absolute;
+		.position(0);
+		filter: blur(3px);
+		// filter: var(--moon-env-background-filter);
```
ryanjduffy: Same here on commented code.

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
Djspaceg: I wanted an opinion on whether a dynamic blur knob would be useful. I left this in place in case we wanted to add it at some point (it would already be there), since the blur feature is new, this would allow configurability to it. Wanted people to use it and think about the effect before answering about the blur amount.

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
Djspaceg: Well, the two are different. To me, undefined means it was never set, and null means it was intentionally set to be blank or no value. I wanted it to be possible to explicitly set `null` and have that persist through this generalized method.
I didn't use `undefined` because I don't like adding linting exceptions. However, in this case, it may make the output easier to read/understand.

Djspaceg:
I too agree that we should read the descriptions of the components' stories. Maybe even (manually) copying the docs one-liner component description into the story, so it's a little more meaningful information.

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
ryanjduffy: I agree with your point on the difference between `undefined` and `null` and that using `return undefined` would be more readable. Let's make that update.

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
ryanjduffy: Good idea to have a conversation about adding the feature. I think it would be helpful in the future to explicitly call that out in the PR description and possibly include it by default. As commented code, it appears to be cruft and doesn't provide the context around which to have a conversation.

What features of the framework will be better demonstrated by adding this knob?

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -157,11 +173,15 @@ const StorybookDecorator = (story, config) => {
...
+			}}
+			// data-background={backgroundLabelMap[select('background', backgroundLabels, Config, getKnobFromArgs(args, 'background'))].background}
```
ryanjduffy: Is there a conversation to be had on this as well?

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
ryanjduffy: You might be able to avoid this entire debate by refactoring the code a bit. I noticed that `getPropFromArgs` can return the fallback value but that value is then passed through `JSON.parse`. If the default value were a parsable string (not sure why it would be but nonetheless), the incorrect fallback would be used (the parsed value rather than the actual string value).

Consider the following which starts with the fallback value and only updates to the parsed value if the knob exists in `args` and can be parsed. The additional advantage being that if `fallbackValue` is `undefined`, we'll correctly return `undefined` without the eslint gymnastics.

```

// This allows any knob to be taken from the URL.
const getKnobFromArgs = (args, propName, fallbackValue) => {
	const knob = 'knob-' + propName;
	let value = fallbackValue;

	if (args && knob in args) {
		try {
			// If it's valid JSON, parse it
			value = JSON.parse(args[knob]);
		} catch (e) {
			// no handling required; allow fallbackValue to be used
		}
	}

	return value;
};
```

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
Djspaceg: Well, I didn't necessarily want to have the conversation right now, since nobody has seen or used storybook with the blurred background yet, so it's really just a passive reminder for a future conversation.

If it _did_ actively demonstrate a real feature of the framework, it would have been included (not commented out), but it doesn't directly demonstrate that. However, the ability to customize the blur amount may be helpful to someone wanting to see just how "crisp" a background could be before a component looks like it's hard to read. It's also sort of _neat_ (read: not mandatory) to see how a background blurring effect can affect foreground readability/usability and be adjusted dynamically.

ryanjduffy: I'm concerned that including commented code as a means to prompt future conversation is not particularly effective. I would recommend either including the feature in the PR for discussion or opening a follow up ticket to discuss and come to a decision.

Djspaceg: Yeah, this wasn't meant to be a conversation topic, just some parked code for a future feature that I didn't have to store, archive, regurgitate or rewrite if the need came up. I'll just delete it. It's already got far more attention than I had intended or even any of the rest of the code or changes.

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
Djspaceg: This is simpler. I'll add it.

Djspaceg:
All requested changes made.

PR #2108, opened 2019-02-11
PLAT-72723: Update ui/VirtualList lifecycle method

ybsung:
### Checklist

`UNSAFE_componentWillMount()`, `UNSAFE_componentWillReceiveProps()`, and `UNSAFE_componentWillUpdate()` will be deprecated.

### Resolution

- Move the code in `UNSAFE_componentWillMount()` to `constructor`.
But `updateStatesAndBounds` was renamed as `getStatesAndUpdateBounds` to get a new state instead of calling `setState` in the `updateStatesAndBounds`.

- Move the code in `UNSAFE_componentWillReceiveProps()` to `componentDidUpdate`.
But we still need the `this.hasDataSizeChanged` variable used in `moonstone/VirtualList`.
In addition, `this.cc = []` for RTL was removed.

- Move the code in `UNSAFE_componentWillUpdate()` and the part of the `positionItems` in render phase to `getDerivedStateFromProps` so that the previous first index and the previous `childProps` are only referenced in the `getDerivedStateFromProps`.

### Links

PLAT-72723

### DIFF
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -281,11 +281,11 @@ const VirtualListBaseFactory = (type) => {
 				!equals(prevProps.itemSize, this.props.itemSize)
 			) {
 				this.calculateMetrics(this.props);
-				// eslint-disable-next-line react/no-did-mount-set-state
+				// eslint-disable-next-line react/no-did-update-set-state
 				this.setState(this.getStatesAndUpdateBounds(this.props));
 				this.setContainerSize();
 			} else if (this.hasDataSizeChanged) {
-				// eslint-disable-next-line react/no-did-mount-set-state
+				// eslint-disable-next-line react/no-did-update-set-state
 				this.setState(this.getStatesAndUpdateBounds(this.props));
 				this.setContainerSize();
 			} else if (prevProps.rtl !== this.props.rtl) {
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -805,12 +805,10 @@ const VirtualListBaseFactory = (type) => {
 				this.positionItems();
 			}
 
-			const {cc} = this;
-
 			return (
 				<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} ref={this.initContainerRef} style={style}>
 					<div {...rest} ref={this.initContentRef}>
-						{itemsRenderer({cc, initItemContainerRef, primary})}
+						{itemsRenderer({cc: this.cc, initItemContainerRef, primary})}
 					</div>
 				</div>
 			);
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -696,8 +696,8 @@ const VirtualListBaseFactory = (type) => {
 			}
 
 			let
-				{primaryPosition, secondaryPosition} = this.getGridPosition(updateFrom),
-				width, height;
+				width, height,
+				{primaryPosition, secondaryPosition} = this.getGridPosition(updateFrom);
 
 			width = (isPrimaryDirectionVertical ? secondary.itemSize : primary.itemSize) + 'px';
 			height = (isPrimaryDirectionVertical ? primary.itemSize : secondary.itemSize) + 'px';
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -229,7 +229,7 @@ const VirtualListBaseFactory = (type) => {
 
 			if (props.clientSize) {
 				this.calculateMetrics(props);
-				nextState = this.getStatesAndUpdateBounds(this.props);
+				nextState = this.getStatesAndUpdateBounds(props);
 			}
 
 			this.state = {
@@ -272,6 +272,7 @@ const VirtualListBaseFactory = (type) => {
 		}
 
 		componentDidUpdate (prevProps) {
+			// TODO: remove `this.hasDataSizeChanged` and fix ui/Scrollable*
 			this.hasDataSizeChanged = (prevProps.dataSize !== this.props.dataSize);
 
 			if (
@@ -285,8 +286,9 @@ const VirtualListBaseFactory = (type) => {
 				this.setState(this.getStatesAndUpdateBounds(this.props));
 				this.setContainerSize();
 			} else if (this.hasDataSizeChanged) {
+				const newState = this.getStatesAndUpdateBounds(this.props, this.state.firstIndex);
 				// eslint-disable-next-line react/no-did-update-set-state
-				this.setState(this.getStatesAndUpdateBounds(this.props));
+				this.setState(newState);
 				this.setContainerSize();
 			} else if (prevProps.rtl !== this.props.rtl) {
 				const {x, y} = this.getXY(this.scrollPosition, 0);
@@ -432,10 +434,9 @@ const VirtualListBaseFactory = (type) => {
 			}
 		}
 
-		getStatesAndUpdateBounds = (props) => {
+		getStatesAndUpdateBounds = (props, firstIndex = 0) => {
 			const
 				{dataSize, overhang, updateStatesAndBounds} = props,
-				firstIndex = this.hasDataSizeChanged ? this.state.firstIndex : 0,
 				{dimensionToExtent, primary, moreInfo, scrollPosition} = this,
 				numOfItems = Math.min(dataSize, dimensionToExtent * (Math.ceil(primary.clientSize / primary.gridSize) + overhang)),
 				wasFirstIndexMax = ((this.maxFirstIndex < moreInfo.firstVisibleIndex - dimensionToExtent) && (firstIndex === this.maxFirstIndex)),
@@ -456,7 +457,7 @@ const VirtualListBaseFactory = (type) => {
 				dataSize,
 				moreInfo
 			}))) {
-				newFirstIndex = this.calculateFirstIndex(props, wasFirstIndexMax, dataSizeDiff);
+				newFirstIndex = this.calculateFirstIndex(props, wasFirstIndexMax, dataSizeDiff, firstIndex);
 			}
 
 			return {
@@ -465,10 +466,9 @@ const VirtualListBaseFactory = (type) => {
 			};
 		}
 
-		calculateFirstIndex (props, wasFirstIndexMax, dataSizeDiff) {
+		calculateFirstIndex (props, wasFirstIndexMax, dataSizeDiff, firstIndex) {
 			const
 				{overhang} = props,
-				{firstIndex} = this.state || {firstIndex: 0},
 				{dimensionToExtent, isPrimaryDirectionVertical, maxFirstIndex, primary, scrollBounds, scrollPosition, threshold} = this,
 				{gridSize} = primary;
 			let newFirstIndex = firstIndex;
@@ -748,7 +748,7 @@ const VirtualListBaseFactory = (type) => {
 
 			if (clientWidth !== scrollBounds.clientWidth || clientHeight !== scrollBounds.clientHeight) {
 				this.calculateMetrics(props);
-				this.setState(this.getStatesAndUpdateBounds(this.props));
+				this.setState(this.getStatesAndUpdateBounds(props));
 				this.setContainerSize();
 				return true;
 			}
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -225,29 +225,34 @@ const VirtualListBaseFactory = (type) => {
 			super(props);
 
 			this.state = {firstIndex: 0, numOfItems: 0};
-		}
-
-		UNSAFE_componentWillMount () {
-			if (this.props.clientSize) {
-				this.calculateMetrics(this.props);
-				this.updateStatesAndBounds(this.props);
-			}
+			this.updatePrevProps(props);
 		}
 
 		// Calculate metrics for VirtualList after the 1st render to know client W/H.
 		componentDidMount () {
 			if (!this.props.clientSize) {
 				this.calculateMetrics(this.props);
-				this.updateStatesAndBounds(this.props);
+				this.updateStatesAndBounds(this.props, false);
 			}
 			this.setContainerSize();
 		}
 
+		componentDidUpdate () {
+			this.updatePrevProps(this.props);
+			this.prevFirstIndex = this.state.firstIndex;
+		}
+
+		updatePrevProps (props) {
+			const {childProps, dataSize, direction, itemSize, overhang, rtl, spacing} = props;
+			this.prevProps = {childProps, dataSize, direction, itemSize, overhang, rtl, spacing};
+		}
+
 		// Call updateStatesAndBounds here when dataSize has been changed to update nomOfItems state.
 		// Calling setState within componentWillReceivePropswill not trigger an additional render.
-		UNSAFE_componentWillReceiveProps (nextProps) {
+		updateStateFromProps () {
 			const
-				{dataSize, direction, itemSize, overhang, rtl, spacing} = this.props,
+				{dataSize, direction, itemSize, overhang, rtl, spacing} = this.prevProps,
+				nextProps = this.props,
 				hasMetricsChanged = (
 					direction !== nextProps.direction ||
 					((itemSize instanceof Object) ? (itemSize.minWidth !== nextProps.itemSize.minWidth || itemSize.minHeight !== nextProps.itemSize.minHeight) : itemSize !== nextProps.itemSize) ||
@@ -259,7 +264,9 @@ const VirtualListBaseFactory = (type) => {
 
 			if (hasMetricsChanged) {
 				this.calculateMetrics(nextProps);
-				this.updateStatesAndBounds(nextProps);
+				if (this.primary !== null ) {
+					this.updateStatesAndBounds(nextProps);
+				}
 				this.setContainerSize();
 			} else if (this.hasDataSizeChanged) {
 				this.updateStatesAndBounds(nextProps);
@@ -273,12 +280,6 @@ const VirtualListBaseFactory = (type) => {
 				} else {
 					this.setScrollPosition(x, y, 0, 0, nextProps.rtl);
 				}
-			}
-		}
-
-		UNSAFE_componentWillUpdate (nextProps, nextState) {
-			if (this.state.firstIndex === nextState.firstIndex || this.props.childProps && this.props.childProps !== nextProps.childProps) {
-				this.prevFirstIndex = -1; // force to re-render items
 			}
 		}
 
@@ -310,6 +311,7 @@ const VirtualListBaseFactory = (type) => {
 		hasDataSizeChanged = false
 		cc = []
 		scrollPosition = 0
+		prevProps = {}
 
 		contentRef = null
 		containerRef = null
@@ -421,7 +423,7 @@ const VirtualListBaseFactory = (type) => {
 			this.state.numOfItems = 0;
 		}
 
-		updateStatesAndBounds = (props) => {
+		updateStatesAndBounds = (props, slient = true) => {
 			const
 				{dataSize, overhang, updateStatesAndBounds} = props,
 				{firstIndex} = this.state,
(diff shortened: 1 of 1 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -666,6 +637,11 @@ const VirtualListBaseFactory = (type) => {
 		}
 
 		positionItems () {
+			this.prevFirstIndex = (
+				this.prevFirstIndex === this.state.firstIndex ||
+				(this.prevProps.childProps || this.props.childProps) && this.prevProps.childProps !== this.props.childProps
+			) ? -1 : this.prevFirstIndex;
```
ybsung: The code above is moved from the following code in `UNSAFE_componentWillUpdate()`.
```
UNSAFE_componentWillUpdate (nextProps, nextState) {
	if (this.state.firstIndex === nextState.firstIndex || this.props.childProps && this.props.childProps !== nextProps.childProps) {
		this.prevFirstIndex = -1; // force to re-render items
	}
}
```

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -684,8 +660,8 @@ const VirtualListBaseFactory = (type) => {
 			}
 
 			let
-				{primaryPosition, secondaryPosition} = this.getGridPosition(updateFrom),
-				width, height;
+				width, height,
+				{primaryPosition, secondaryPosition} = this.getGridPosition(updateFrom);
```
ybsung: I fixed it due to `unexpected reserved word let`.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,10 +771,42 @@ const VirtualListBaseFactory = (type) => {
...
+				} else {
+					this.setScrollPosition(x, y, 0, 0, props.rtl);
+				}
+			}
+
 			if (primary) {
 				this.positionItems();
 			}
 
+			const {cc} = this;
```
ybsung: I could not replace `cc` with`this.cc` in `{itemsRenderer({cc, initItemContainerRef, primary})}` due to `unexpected reserved word this`.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,10 +771,42 @@ const VirtualListBaseFactory = (type) => {
...
+				const {x, y} = this.getXY(this.scrollPosition, 0);
+
+				this.cc = [];
+				if (type === Native) {
+					this.scrollToPosition(x, y, props.rtl);
+				} else {
+					this.setScrollPosition(x, y, 0, 0, props.rtl);
+				}
+			}
+
```
ybsung: I just moved the code in `UNSAFE_componentWillReceiveProps()` to here.

0x64:
LGTM

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,14 +771,44 @@ const VirtualListBaseFactory = (type) => {
...
+				hasMetricsChanged = (
+					direction !== props.direction ||
+					((itemSize instanceof Object) ? (itemSize.minWidth !== props.itemSize.minWidth || itemSize.minHeight !== props.itemSize.minHeight) : itemSize !== props.itemSize) ||
+					overhang !== props.overhang ||
+					spacing !== props.spacing
+				);
+
+			this.hasDataSizeChanged = (dataSize !== props.dataSize);
+
+			if (hasMetricsChanged || this.props.clientSize && this.primary === null && this.secondary === null) {
```
viodragon2: All these conditions seem to happen after the first render. IOW, metrics and data size change would only happen after its first mount and subsequent renders.

I wonder if there are any noticeable perf loss if these changes were pushed back to `componentDidUpdate` phase. By doing so, I think we can remove the hacky `updatePrevProps()` and (yet another) the internal member variable. Plus, we can remove the direct `state` mutation hack that I pointed out above.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,14 +771,44 @@ const VirtualListBaseFactory = (type) => {
...
+				this.setContainerSize();
+			} else if (this.hasDataSizeChanged) {
+				this.updateStatesAndBounds(props, true);
+				this.setContainerSize();
+			} else if (rtl !== props.rtl) {
+				const {x, y} = this.getXY(this.scrollPosition, 0);
+
+				this.cc = [];
+				if (type === Native) {
+					this.scrollToPosition(x, y, props.rtl);
```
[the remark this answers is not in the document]
ybsung: If `rtl` mode changes, we need to update `VirtualList` as well as `VirtualList items` based on the `rtl` mode.

For `VirtualList`, we called the `this.scrollToPosition` or `this.setScrollPosition` here. There might be no problem if calling it in `componentDidUpdate`.

But for `VirtualList items`, `this.cc = [];` was executed at https://github.com/enactjs/enact/blob/31e372d4d38972d7e0448a0f4481a6502dde2efe/packages/ui/VirtualList/VirtualListBase.js#L796, then `this.positionItems();` was called at https://github.com/enactjs/enact/blob/31e372d4d38972d7e0448a0f4481a6502dde2efe/packages/ui/VirtualList/VirtualListBase.js#L805 If moving the `this.cc =[];` to `componentDidUpdate`,
- To render the items, we need to call `this.forceUpdate()` instead of `this.setState()` in `componentDidUpdate` because no state changes. I think calling `forceUpdate` is not a good pattern.
- `render()` function will be called once more.

So I thought it would be better to execute `this.cc = [];` in `render`() with calling `scrollTo` together.

packages/ui/VirtualList/VirtualListBase.js
ybsung: The following condition is for the first render.
- this.props.clientSize && this.primary === null && this.secondary === null

The original code  for the condition above was in `UNSAFE_componentWillMount ()`. The reason why we added the condition was to reduce calling `render()` from twice to once to enhance app launch time. The impact for the patch was so big. Please check the comments in PLAT-37236. If it is ok to call `render()` more, we could just undo the patch by removing the condition. But I'm afraid of increasing the launch time.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -448,7 +414,12 @@ const VirtualListBaseFactory = (type) => {
 				newFirstIndex = this.calculateFirstIndex(props, wasFirstIndexMax, dataSizeDiff);
 			}
 
-			this.setState({firstIndex: newFirstIndex, numOfItems});
+			if (slient) {
+				this.state.firstIndex = newFirstIndex; // eslint-disable-line react/no-direct-mutation-state
```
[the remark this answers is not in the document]
ybsung: I'll try to find a better way.

ybsung:
Hi @viodragon2 
There are lots of bad practices in `VirtualList`. But to enhance the launch time performance and the scroll performance, we manipulate the DOM elements and the states manually in the `VirtualList`. As I mentioned the comments in PLAT-37236, the impact of calling the render function once more in `VirtualList` seems to be pretty large.

IMHO, if possible, we need to reduce calling the render function in `VirtualList`. To do that, we could not avoid updating states directly and referring the previous props from states in render phase.

But it is true that it is the right time to remove all the bad practices. Then we can do improve performance from another perspective for several months.

Please let me know your thought.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,14 +771,44 @@ const VirtualListBaseFactory = (type) => {
...
+					((itemSize instanceof Object) ? (itemSize.minWidth !== props.itemSize.minWidth || itemSize.minHeight !== props.itemSize.minHeight) : itemSize !== props.itemSize) ||
+					overhang !== props.overhang ||
+					spacing !== props.spacing
+				);
+
+			this.hasDataSizeChanged = (dataSize !== props.dataSize);
+
+			if (hasMetricsChanged || this.props.clientSize && this.primary === null && this.secondary === null) {
+				this.calculateMetrics(props);
+				this.updateStatesAndBounds(props, true);
```
viodragon2: As the name of the function suggests, I wonder if we can translate this to `getDerivedStatesFromProps` so that we don't worry about directly mutating states.

This would require for `cbScrollTo` it to be called elsewhere (maybe in `componentDidUpdate`?) and I'm not sure if that would cause a too big of a change.

What do you think?

packages/ui/VirtualList/VirtualListBase.js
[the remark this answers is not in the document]
viodragon2: Is `this.cc` required to be reset when `rtl` prop changes? Doesn't index and key stay the same in RTL layout? Or am I missing something?

packages/ui/VirtualList/VirtualListBase.js
viodragon2: Generic questions:

- Have we promoted `clientSize` props over setting dimensions to developers? I don't see any apps using this, unfortunately :(
- I don't think we can't ignore what happens when VirtualList is used with `Scroller` as `VirtualListBase` has a strong relationship with `Scrollable`. And `syncClientSize` gets called when `Scrollable` updates.
- If the goal is to minimize the initial render when `clientSize` were provided as a prop, can we calculate metrics in `constructor` instead? Essentially we're initializing the metrics via `calculateMetrics` so that we can position items on render.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,14 +771,44 @@ const VirtualListBaseFactory = (type) => {
...
+				this.setContainerSize();
+			} else if (this.hasDataSizeChanged) {
+				this.updateStatesAndBounds(props, true);
+				this.setContainerSize();
```
viodragon2: Why do we need to reset container size when only the `dataSize` has changed?

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,14 +771,44 @@ const VirtualListBaseFactory = (type) => {
 			delete rest.spacing;
 			delete rest.updateStatesAndBounds;
 
+			const
+				{dataSize, direction, itemSize, overhang, rtl, spacing} = this.prevProps,
+				{props} = this,
+				hasMetricsChanged = (
+					direction !== props.direction ||
+					((itemSize instanceof Object) ? (itemSize.minWidth !== props.itemSize.minWidth || itemSize.minHeight !== props.itemSize.minHeight) : itemSize !== props.itemSize) ||
```
viodragon2: nit: we can use ramda's `equals` function to simplify.

`!equals(prevProps.itemSize, this.props.itemSize)`

viodragon2:
I tried a few things here and there with minimal changes in [this branch](https://github.com/enactjs/enact/tree/feature/PLAT-72723-stephen). Here's a diff you can look at https://github.com/enactjs/enact/compare/feature/PLAT-72723-simple...feature/PLAT-72723-stephen

I'm still struggling how we can handle `prevFirstIndex` and `childProps` before `componentDidUpdate` but hopefully you get a generic idea on how we can move a lot of things in `componentDidUpdate`

packages/ui/VirtualList/VirtualListBase.js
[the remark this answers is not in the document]
ybsung: All items usually display properly depending on the RTL mode based on CSS. But app developers could position the content in the item depending on the RTL mode. For this case, we need to re-render the items so that app developers could position them properly in `itemRender`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: > * Have we promoted `clientSize` props over setting dimensions to developers? I don't see any apps using this, unfortunately :(

I found just only one use case in Settings app. Actually the issue related with `clientSize` was raised from Settings Picture panel. But the `clientSize` in the panel was not used.
\com.webos.app.settings\src\views\General\LanguageList.js

```
	setVirtualListProps (listData, comp) {
		const retProps = {
			itemSize: ri.scale(78),
			dataSize: listData.length,
			isItemDisabled: (index) => (listData[index].disabled),
			itemRenderer: comp
		};

		if (typeof window === 'object' && window.testing) {
			retProps.clientSize = {
				clientWidth: 1920,
				clientHeight: 1080
			};
		}

		return retProps;
	}

	const virtualListProps = this.setVirtualListProps(listData, this.getLanguageListItem());

	<VirtualList {...virtualListProps} className={css['general-language-list-virtuallist']} />
```

ybsung: > * If the goal is to minimize the initial render when `clientSize` were provided as a prop, can we calculate metrics in `constructor` instead? Essentially we're initializing the metrics via `calculateMetrics` so that we can position items on render.

Yes. I think it makes sense. As you proposed the code in [your branch](https://github.com/enactjs/enact/tree/feature/PLAT-72723-stephen) I'll try to apply it.

packages/ui/VirtualList/VirtualListBase.js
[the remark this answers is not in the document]
ybsung: But there is no real case until now. Let's move this logic to `componentDidUpdate`.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -797,14 +771,44 @@ const VirtualListBaseFactory = (type) => {
...
+					overhang !== props.overhang ||
+					spacing !== props.spacing
+				);
+
+			this.hasDataSizeChanged = (dataSize !== props.dataSize);
+
+			if (hasMetricsChanged || this.props.clientSize && this.primary === null && this.secondary === null) {
+				this.calculateMetrics(props);
+				this.updateStatesAndBounds(props, true);
+				this.setContainerSize();
```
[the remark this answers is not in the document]
ybsung: When `componentDidMount` is called in `VirtualList`, there is no scrollbar even if the content is long enough to scroll it. So the VirtualList re-renders. Then it seems that we need to update the container size again based on the area except the scrollbar.

In addition, when metrics changes or data size changes, we need to update container size again because the visibility of the scrollbar could changes and the scroll area could changes.

packages/ui/VirtualList/VirtualListBase.js
ybsung: >As the name of the function suggests, I wonder if we can translate this to `getDerivedStatesFromProps` so that we don't worry about directly mutating states.

As I mentioned in the JIRA, my first approach is the one that you mentioned. I moved all the code in `componentWillReceiveProps` to `getDrivedStateFromProps`.

As you can see the following code, I renamed `updateStateAndBounds` as `updateMetrics`, passed `prevProps`, `props`, and `prevState` to the function, then the function returned a new state. To do that, I needed to make all functions called from `updateStateAndBounds` returning a new state.
https://github.com/enactjs/enact/compare/feature/PLAT-72723-enhanced2#diff-decb3acaa813ab261efba3ee57c8b4f9R235

But I thought we needed to update lots of code to change member variables to state variables and member functions to class static functions or global functions because `getDrivedStateFromProps` was a static function and only props and states were passed.

If it is the right way, I'll keep going ahead instead of this PR.

ybsung:
https://github.com/enactjs/enact/pull/2108#discussion_r256657816
> Is `this.cc` required to be reset when `rtl` prop changes? Doesn't index and key stay the same in RTL layout? Or am I missing something?

As I mentioned in https://github.com/enactjs/enact/pull/2108#discussion_r257102041, let's remove this code to reset `this.cc`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: Oh, it looks much better. Thanks.

packages/ui/VirtualList/VirtualListBase.js
ybsung: I just copied my comment in https://github.com/enactjs/enact/pull/2108#discussion_r257104333.
```
In addition, when metrics changes or data size changes, we need to update container size again because the visibility of the scrollbar could changes and the scroll area could changes.
```

packages/ui/VirtualList/VirtualListBase.js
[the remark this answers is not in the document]
ybsung: <On Feb 18th>
The code to change the state directly in `calculateMetrics` was removed.
`getStatesAndUpdateBounds` ( which name was `updateStateAndBounds` before ) returns a new state instead of updating the state directly.

packages/ui/VirtualList/VirtualListBase.js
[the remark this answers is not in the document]
ybsung: <On Feb 18th>
- The code related with RTL including calling `scrollTo` is moved to not render phase but `componentDidUpdate`.
- The code `this.cc = [];` that we discussed is removed. If any issue raised, then we could revisit it.

packages/ui/VirtualList/VirtualListBase.js
ybsung: <On Feb 18th>
- I move the code in `UNSAFE_componentWillMount()` to `constructor` related with `clientSize`.
But `updateStatesAndBounds` was renamed as `getStatesAndUpdateBounds` to get a new state because we could not call `setState` in `constructor`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: <On Feb 18th>
- The code above including `cbScrollTo` is moved to `componentDidUpdate`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: <On Feb 18th>
- I updated the PR.

packages/ui/VirtualList/VirtualListBase.js
ybsung: <On Feb 18th>
- I updated the PR.

ybsung:
Hi @viodragon2 

I updated the code and the description in the PR. Could you review the PR again?

As you recommended, I moved the code in the deprecated lifecycle methods not to `render` but to `constructor`, `getDerivedStateFromProps`, and `componentDidUpdate`.

But I'm not sure whether it would be better to move the code in `componentDidUpdate` to `getDerivedStateFromProps`. Could you check my comment https://github.com/enactjs/enact/pull/2108#discussion_r257106546 above?
Please let me know your thought.

Thank you so much.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -222,66 +223,82 @@ const VirtualListBaseFactory = (type) => {
...
 		}
 
-		UNSAFE_componentWillMount () {
-			if (this.props.clientSize) {
-				this.calculateMetrics(this.props);
-				this.updateStatesAndBounds(this.props);
-			}
+		static getDerivedStateFromProps (props, state) {
+			const
+				isNeedToInvalidate = (
```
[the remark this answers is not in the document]
ybsung: `shouldInvalidate` would be great. I'll update it.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -222,66 +223,82 @@ const VirtualListBaseFactory = (type) => {
...
+					state.prevFirstIndex === state.firstIndex ||
+					(state.prevChildProps || props.childProps) && state.prevChildProps !== props.childProps
```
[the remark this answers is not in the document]
ybsung: I could not remember why I did. The condition that you mentioned should be removed.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -448,13 +459,16 @@ const VirtualListBaseFactory = (type) => {
...
+				firstIndex: newFirstIndex,
+				numOfItems: numOfItems
+			};
 		}
 
 		calculateFirstIndex (props, wasFirstIndexMax, dataSizeDiff) {
 			const
 				{overhang} = props,
-				{firstIndex} = this.state,
+				{firstIndex} = this.state || {firstIndex: 0},
```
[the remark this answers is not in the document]
ybsung: `calculateFirstIndex` is called in `constructor` before defining this.state. So there is an error for {firstIndex} = this.state. That is the reason why adding `|| {firstIndex: 0}.

There is another alternative without adding it like the following way.

AS-IS
```
		constructor (props) {
			let nextState = null;

			super(props);

			if (props.clientSize) {
				this.calculateMetrics(props);
				nextState = this.getStatesAndUpdateBounds(this.props);
			}

			this.state = {
				firstIndex: 0,
				numOfItems: 0,
				prevChildProps: null,
				prevFirstIndex: 0,
				updateFrom: 0,
				updateTo: 0,
				...nextState
			};
		}
```

TO-BE
```
		constructor (props) {
			let nextState = null;

			super(props);

			this.state = {
				firstIndex: 0,
				numOfItems: 0,
				prevChildProps: null,
				prevFirstIndex: 0,
				updateFrom: 0,
				updateTo: 0
			};

			if (props.clientSize) {
				this.calculateMetrics(props);
				nextState = this.getStatesAndUpdateBounds(this.props);
			}

			this.state = {
				...this.state,
				...nextState
			};
		}
```

PR #2110, opened 2019-02-11
ENYO-5798: Update QA Sampler to add marqueeOnRenderDelay knob for Marquee > LTR sample or create new sample

sugardave:
### Checklist

This patch adds a knob for `marqueeOnRenderDelay` for `Marquee` qa stories.

### Additional Considerations
At least one of the stories has multiple `Marquee`s but currently offer no knobs at all, so they have been left out of this patch as well.

### Links
ENYO-5798


webOS101:
Did you verify this knob works?  It never worked before because the initial render doesn't get the update `marqueeOnRenderDelay` setting from Storybook and updating it after the component is rendered will not use the render delay.

sugardave:
I got it to work by toggling {{marqueeOn}} to hover and back to render.

webOS101:
:awesome:

PR #2115, opened 2019-02-14
PLAT-73547: Update ViewManager internal lifecycle methods

ryanjduffy:
### Checklist

Update `TransitionGroup` to use `getDerivedStateFromProps`


ryanjduffy:
Thanks for the catch, @viodragon2! The root cause was a logic failure in how I had determined that views were entering/leaving TransitionGroup. By comparing prevState with current state when multiple children were provided, it wasn't flagging children to leave that should have been when they existed in both states. Switching `componentDidUpdate` to use the current children from props instead of the children from state for the "next children" seems to resolve the issue.

Going to add a test for this and then should be good to go.

PR #2116, opened 2019-02-14
ENYO-5791: Adjust focus timing of ContextualPoup

Baekwoo-s:
### Checklist

ContextualPopup does not focus on popup content when the popup is first opened.

### Resolution
The inner content was not yet rendered at the timing to focus it. so I fixed the timing to focus after it is rendered.

### Additional Considerations

### DIFF
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -233,10 +233,11 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 		constructor (props) {
 			super(props);
 			this.state = {
+				activator: null,
 				arrowPosition: {top: 0, left: 0},
-				containerPosition: {top: 0, left: 0},
 				containerId: Spotlight.add(this.props.popupSpotlightId),
-				activator: null
+				containerPosition: {top: 0, left: 0},
+				prevOpen: false
 			};
 
 			this.overflow = {};
@@ -262,6 +263,7 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			if (prevProps.open && !this.props.open) {
 				const current = Spotlight.getCurrent();
 				return {
+					prevOpen: this.props.open,
 					shouldSpotActivator: (
 						// isn't set
 						!current ||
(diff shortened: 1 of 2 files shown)


packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -282,15 +282,17 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
...
-			if (this.props.open && !prevProps.open) {
-				on('keydown', this.handleKeyDown);
-				on('keyup', this.handleKeyUp);
-				this.spotPopupContent();
-			} else if (!this.props.open && prevProps.open) {
-				off('keydown', this.handleKeyDown);
-				off('keyup', this.handleKeyUp);
-				if (snapshot && snapshot.shouldSpotActivator) {
-					this.spotActivator(prevState.activator);
+			if (prevProps.open && this.state.activator) {
```
0x64: If this component is _updated_ in any reason during a popup is open, this routine runs again.
Also, this code assumed that the second rendering right after opening a popup. Does this component is rendered twice whenever it opens a popup?

0x64:
LGTM

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -282,7 +282,7 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 				this.positionContextualPopup();
 			}
 
-			if (this.props.open && !prevProps.open) {
+			if (this.props.open && this.state.activator && !prevState.activator) {
```
ryanjduffy: This condition will also avoid entering this block when nothing is focused when opening the contextual popup (as is often the case in sampler) since `activator` could be set to `null` even though the floating layer is in fact open.

Baekwoo-s: In Sampler, activator sometimes became 'undefined'. (I can't find 'null') Even if it is 'undefined', the condition of the above if statement does not work as expected.

So I modified the code by adding a new state to your opinion. In this case, ContextualPopup modified the state value to be true after the node was drawn in the FloatingLayer.

ryanjduffy:
Closing in favor of #2124

PR #2118, opened 2019-02-14
PLAT-64279: High Contrast color support

teckliew:
### Checklist

PLAT-64279: High Contrast color support.

### Additional Considerations
The high contrast for dark and light skins. Also alphabethesized the order of variable sections in `colors.less` and `colors-light.less`

### Links
PLAT-64279

### DIFF
--- a/packages/moonstone/Scrollable/Scrollbar.module.less
+++ b/packages/moonstone/Scrollable/Scrollbar.module.less
@@ -65,10 +65,15 @@
 	/* ScrollThumb */
 	.thumbShown {
 		opacity: 1;
-		.applySkins({
+	}
+}
+
+.applySkins({
+	.scrollbar{
+		.thumbShown {
 			&::before {
 				background-color: @moon-scroller-scrollthumb-bg-color;
 			}
-		});
+		}
 	}
-}
+});
--- a/packages/moonstone/Scrollable/Scrollbar.module.less
+++ b/packages/moonstone/Scrollable/Scrollbar.module.less
@@ -1,3 +1,4 @@
+@import "../styles/skin.less";
 @import '../styles/variables.less';
 
 .scrollbar {
@@ -64,5 +65,10 @@
 	/* ScrollThumb */
 	.thumbShown {
 		opacity: 1;
+		.applySkins({
+			&::before {
+				background-color: @moon-scroller-scrollthumb-bg-color;
+			}
+		});
 	}
 }
--- a/packages/moonstone/Scrollable/Scrollbar.module.less
+++ b/packages/moonstone/Scrollable/Scrollbar.module.less
@@ -69,7 +69,7 @@
 }
 
 .applySkins({
-	.scrollbar{
+	.scrollbar {
 		.thumbShown {
 			&::before {
 				background-color: @moon-scroller-scrollthumb-bg-color;
--- a/packages/moonstone/ToggleButton/ToggleButton.module.less
+++ b/packages/moonstone/ToggleButton/ToggleButton.module.less
@@ -29,7 +29,7 @@
 
 	&.selected {
 		&::after {
-			content: @moon-icon-circle;
+			content: @moon-toggle-button-selected-glyph-highcontrast;
 		}
 	}
 
--- a/packages/moonstone/ToggleButton/ToggleButton.module.less
+++ b/packages/moonstone/ToggleButton/ToggleButton.module.less
@@ -10,7 +10,6 @@
 	text-align: center;
 	position: relative;
 
-	&::before,
 	&::after {
 		.position-start-end(auto, @moon-button-h-padding);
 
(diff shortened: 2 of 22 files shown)


ryanjduffy:
Most things looked pretty good to me. I noticed a couple things that I wanted to call out but will do a little more testing tomorrow.
* The header colors appear off to me. In the specs, the border is `#404040` and text color is `#000` but they are `#4d4d4d` for both in the PR. I had to check colors from a PNG export because I don't have photoshop so it's possible my color selection doesn't match the original.
* Input's focused and active states don't match the specs (though I'm suspicious of the thicker stroke for the focused input)
* The dark gradient behind the VideoPlayer's controls in light skin high contrast is making it harder to distinguish the buttons. It should probably be a light gradient (though we lack specs for it).

packages/moonstone/Scrollable/Scrollbar.module.less
```diff
@@ -64,5 +65,10 @@
 	/* ScrollThumb */
 	.thumbShown {
 		opacity: 1;
+		.applySkins({
```
ryanjduffy: Our convention has been to use `.applySkins()` under the root class and duplicate the hierarchy within it. Should we do the same here?

packages/moonstone/ToggleButton/ToggleButton.module.less
```diff
@@ -10,53 +10,94 @@
 	text-align: center;
 	position: relative;
 
-	&::after {
+	&::before, &::after {
```
ryanjduffy: Convention is to have one selector per line so let's move `&::after` to the next line.

packages/moonstone/ToggleButton/ToggleButton.module.less
```diff
@@ -10,53 +10,94 @@
...
+		width: @moon-toggle-button-switch-width - 3px;
+		height: @moon-toggle-button-switch-width - 3px;
+		transform: translate(-36%, -50%);
+	}
 
 	&.small {
 		.padding-start-end(@moon-button-small-h-padding, @moon-toggle-button-small-text-switch-gutter + @moon-button-small-h-padding + @moon-toggle-button-switch-width + @moon-toggle-button-switch-border-width);
 
-		&::after {
+		&::before, &::after {
```
ryanjduffy: + Line

ryanjduffy:
@teckliew How are we handling the questions on GridListImageItem, Input, and VideoPlayer?

packages/moonstone/styles/colors-light.less
```diff
@@ -186,7 +184,7 @@
 // ---------------------------------------
 @moon-spinner-bg-color: @moon-button-bg-color;
 @moon-spinner-text-color: @moon-white;
-@moon-spinner-text-shadow-color: rgba(0, 0, 0, 0.3);
+@moon-spinner-text-shadow-color: fade(@moon-black, 0.30%);
```
Djspaceg: This is a **super** small percentage ;)

packages/moonstone/VideoPlayer/VideoPlayer.module.less
```diff
@@ -422,5 +421,13 @@
 				}
 			}
 		}
+
+		.overlay {
+			&.high-contrast-scrim::after {
+				:global(.enact-a11y-high-contrast) & {
```
Djspaceg: I think this section could be refactored to use the skinning (`.applySkins`) system, rather than its own method.

packages/moonstone/VideoPlayer/VideoPlayer.module.less
```diff
@@ -424,9 +424,7 @@
 
 		.overlay {
 			&.high-contrast-scrim::after {
-				:global(.enact-a11y-high-contrast) & {
-					background: @moon-video-player-high-contrast-scrim;
-				}
+				background: linear-gradient(fade(@moon-video-player-high-contrast-scrim-gradient-color, 0%) 0%, fade(@moon-video-player-high-contrast-scrim-gradient-color, 60%) 100%);
```
Djspaceg: What if this was reconfigured to be like the following:
```
.overlay {
    &::after {
        background: @moon-video-player-overlay-scrim-bg;
    }
}
```
Then in `colors.less`:
```
@moon-video-player-overlay-scrim-bg: initial; // or `transparent none` to affect both bg-color and bg-image.
```
And in `colors-highcontrast.less`:
```
@moon-video-player-overlay-scrim-bg: linear-gradient(fade(@moon-black, 0%) 0%, fade(@moon-black, 60%) 100%);
```
We may not need the `&.high-contrast-scrim` class at all since we can simply _always_ have it and just render it as blank/empty under normal circumstances.

This approach would leverage our new high-contrast code, and remove some old implementation that's now obsolete.

teckliew: 👍 Sounds good. Just updated the scrim.

ryanjduffy:
Getting close! It's looking really good now. Two more revisions:
* Dark skin input's border is thinner than light skin
* The overlay checkbox for VirtualGridList in light skin is hidden when focusing but unselected because it's color is white on a white background. See the middle picture above.

PR #2119, opened 2019-02-15
ENYO-5779: Spottable > prevent updates due to focus changes

germboy:
### Checklist

Due to updating the lifecycle in #2078, we re-introduced an unwanted behavior where simply changing the focus state of a `Spottable` component caused an update. This was due to moving our local focus state from an instance property to state, which is required since our use of `getDerivedStateFromProps` needs to be aware of focus state.

### Resolution
This change re-implements (originally used in #2078 at [this commit](https://github.com/enactjs/enact/pull/2078/commits/3fb6b210261bb36b4b9bd3993e2fb26e58a08348) ) the use of `shouldComponentUpdate` to prevent updates if focus state is the only change to the component.

The use of `shouldComponentUpdate` is being used as a last resort here, as I would like to avoid it since:
- we don't want to inadvertently prevent updates that should be allowed
- it still takes time to run per each potential update vs simply allowing a component to update (whether it needs to or not)

There really doesn't appear to be another option in this case. Since `gDSFP` is a static method, we don't have access to instance variables/props - which is what we would need in order to avoid storing focus state to `this.state`. We also still need `gDSFP` to respond to prop changes.

This use-case doesn't appear to be a valid candidate for memoization, we cannot access or otherwise pass instance variables as values to `gDSFP`, it appears we cannot respond to expected prop changes without `gDSFP`, so I cannot see a better solution than `sCU`.

While `sCU` will prevent updates, it still takes work to perform (it shows up on perf traces) so it doesn't exactly match our performance prior to using `gDSFP`, but it does reduce unnecessary updates and the time spent during focus changes to a simple `Spottable('div')` drops from ~2.4ms (update) to ~0.3ms (`sCU`).

If anyone has any other thoughts or ideas, I'm all 👂


packages/spotlight/Spottable/Spottable.js
```diff
@@ -222,6 +223,18 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 			this.node = ReactDOM.findDOMNode(this);
 		}
 
+		shouldComponentUpdate (nextProps, nextState) {
+			const mutatedState = {...this.state};
```
ryanjduffy: We're incurring some additional cost by cloning both state objects just to delete the `focused` member. Would it be better to compare the state members individually since there is a small known set? Also, is it necessary to keep `selectionKeys` in state? I couldn't see a reason to copy it into state. Removing it would simplify the state comparison further.

germboy: I think that's a good observation - I don't see where it's necessary to keep `selectionKeys` in state, since we won't need to compare them against previous props in `gDSFP`. I'm removing this and updating the `sCU` logic. 👍

PR #2120, opened 2019-02-15
PLAT-73546: ui/Changeable & ui/Toggleable: Add warning if both [defaultProp] and [prop] are declared

sugardave:
### Checklist

Developers should not assign both `[defaultProp]` and `[prop]` at the same time so we add a console warning if that is the case.

### Additional Considerations
- Components and apps that have used the pattern will start seeing warnings in the console
- Existing tests for `ui/Changeable` and `ui/Toggleable` have been updated to expect warnings
- There are some tests in `ui/Toggleable` that still warn even when both props are not present

### Links
PLAT-73546

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
+- `ui/Changeable` and `ui/Toggleable` to warn when both `[defaultProp]` and `[prop]` are provided
+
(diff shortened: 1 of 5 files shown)


ryanjduffy:
I pushed a couple fixes but otherwise LGTM:
* Fixed linting error for unused variable
* Removed default value for Toggleable's `defaultPropKey` which was triggering the warning when not provided by the author.

ryanjduffy:
Also, I wanted to comment that I noticed we were missing `mockRestore` but adding it appears to interfere with the prop type spy. Seems like that is a potential issue for future consideration.

PR #2125, opened 2019-02-22
PLAT-73358: Fix to perform `scrollTo` callback during the list is updated by prop changes

0x64:
### Checklist

After migration to React 16.7, items are not displayed properly if an app calls `scrollTo` during VirtualList is being updated by prop changes.

### Resolution
When props are changed, a list should be reset by data change or bounds change. And if the scroll area is expanded or shrunk by scrollbar visibility, a list needs at least two rendering cycles due to measuring. So we had a logic to defer a `scrollTo` request until internal updating is finished.
Note that the first index to render is managed through `state` and it is very tightly coupled with the scroll position of a list which is _updated by browser outside of React lifecycles_.
Here are changes in this PR;
- After migration to React 16.7, timing to defer is changed, so I moved the code from `componentDidUpdate` to `render`.
- Due to async execution of `setState`, the first index value and the scroll position are not matched if a list is updated by props change. So changed to use intentional last first index.
- Also, the legacy code to pick the first index on scroll is assuming that the previous first index and thresholds are matched to the scroll position. But, it's not true now, so I changed calculation logic to be independent from the assumption.

### Links
PLAT-73358

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -8,6 +8,10 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 - `ui/Changeable` and `ui/Toggleable` to warn when both `[defaultProp]` and `[prop]` are provided
 
+### Fixed
+
+- `ui/VirtualList` to scroll properly by `scrollTo` callback during the list is updated by prop changes
+
 ## [2.3.0] - 2019-02-11
 
 ### Added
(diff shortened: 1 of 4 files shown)


ybsung:
@0x64 
Without https://github.com/enactjs/enact/pull/2125/commits/9a80ee87f2a56c6ddbb2fa3a009ada8f178f15da commit, it seems to work well to me. Please check it.

0x64:
@ybsung Good point. The 3rd commit is not required anymore after #2108 is merged.

ybsung:
LGTM

PR #2126, opened 2019-02-22
ENYO-5839: Fix Storybook story workspace positioning

Djspaceg:
### Checklist

Storybook addons don't adhere to our idealized styling policy, and as such, some addons simply inject unstylable DOM nodes into the hierarchy. This causes layout problems when a child component expects to have a container which doesn't auto-shrink to minimum size.

There is a typo in an enact-knob code.

And some knobs do not properly group themselves when the page is reloaded even though they do properly group when switching stories.

### Resolution
Added better layout support (by using Layout) to the story workspace area, and restored the positioning rules for the unstylable foreign DIV.

Fixed said typo

Added a `groupId` key to the Config object so the component name isn't discovered after several of the knobs are rendered (into the wrong tab-group category). May not be the *best* place to put it, but it seems logical and should cover all of our cases.

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.module.less
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.module.less
@@ -35,6 +35,12 @@
 		}
 	}
 
+	.storyCell > div:not([class]):not([style]):first-child {
+		position: relative;
+		width: 100%;
+		height: 100%;
+	}
+
 	// Hack into the Info - Don't try this at home, kids
 	main > article > section > :nth-child(2) > :nth-child(3) {
 		.enact-selectable();
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.module.less
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.module.less
@@ -35,6 +35,12 @@
 		}
 	}
 
+	// Here, we target the storybook rendering canvas div. It's inserted as a child of our Panel,
+	// but it applies position:relative, which doesn't help us when we have stories that should fill
+	// the entire panel body. This targets that inaccessible node using Panel's structure and
+	// storybook's structure to update that goofy rule to a rational one.
+	// The :not() bit targets only non-enact elements, so if a story doesn't use withInfo, and a
+	// `div` isn't added, the components aren't being told to be a different size/position.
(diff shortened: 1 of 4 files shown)


packages/sampler/src/utils/propTables.js
```diff
@@ -5,6 +5,7 @@ const merge = (components, field) => {
 const mergeComponentMetadata = (displayName, ...components) => {
 	const fn = function () {};
 	fn.displayName = displayName;
+	fn.groupId = displayName;
```
sugardave: Is this supposed to be part of this PR?

Djspaceg: I described it in the description. It resolves a different, but equally important, knob issue with grouping.

sugardave: Ah, my bad.

:LGTM:

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.module.less
```diff
@@ -35,6 +35,12 @@
 		}
 	}
 
+	.storyCell > div:not([class]):not([style]):first-child {
```
ryanjduffy: Can you add an inline comment to describe the purpose of this particular CSS incantation for future us? :)

webOS101: :mindblown:

Djspaceg: Done. Originally added from #1863.

packages/sampler/src/utils/propTables.js
ryanjduffy: Noticing an interesting side effect of this change for samples with multiple components (ProgressBar, Slider, IncrementSlider, and VideoPlayer) that each gets its own knob tab. For example, Slider's sample has a tab for Slider and one for SliderTooltip.

image


I'm not sure if this is a good thing or not but I'm okay with moving ahead with it unless someone objects.

Djspaceg: That's actually intentional. It was built with that in mind and has been present since it was added.

Here's a screenshot from 2.2.8:
screen shot 2019-02-22 at 3 46 19 pm

ryanjduffy: :nice: Even better

PR #2128, opened 2019-02-23
ENYO-5841: Add ref support to Layout and Cell

Djspaceg:
### Checklist

No Layout components support `ref` props. This makes measurement or other necessary things much more difficult.

### Resolution
Added `forwardRef` support to do this safely on `Layout`, `Row`, `Column`, and `Cell`.


ryanjduffy:
I think this is a reasonable change but curious if we should expand the scope a bit to include other components that have a `component` prop in order to be consistent for HOC-like UI components.

Quick search reveals the following candidates:
* `ui/ViewManager` (+ it's internal `TransitionGroup`)
* `ui/ToggleItem`
* `ui/Spinner`
* `ui/SlotItem`
* `ui/Item`
* `moonstone/Panels.Router` - not significant but a possibility

A later consideration might be to either move our HOCs to use this as well or perhaps adding it to the tops of our various HOC stacks to ensure that users get the expected `ref`s. Not suggesting we do this change now as it is potentially more disruptive.

packages/ui/Layout/Cell.js
```diff
@@ -58,6 +59,14 @@ const CellBase = kind({
 		 */
 		component:  PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
 
+		/**
+		 * A ref instance provided by {@link ui/ForwardRef.ForwardRef}.
+		 *
+		 * @type {Function}
+		 * @private
+		 */
+		forwardedRef: PropTypes.func,
```
ryanjduffy: I think that in general we'll want to mark these as `@public` particularly when we export the *Base component. Along that line, I'd like to adopt a more descriptive name to indicate to what you are gaining a reference when using the prop.

What do you think of `componentRef` to align it with the `component` prop?

packages/ui/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ## [unreleased]
 
+### Added
+
+- `ui/Layout` `ref` support so developers can safely get refs to their layout elements
```
ryanjduffy: If we mark the ref props public, we'd need to call them out here.

Djspaceg:
That's a really good point. It would certainly help for consistency if we implemented it the same all over.

HOCs do pollute the conversion a bit, since the refs would get filtered out on any component that is wrapped in a HOC. The user would need to be expecting that both the base component and the HOC stack (in entirety) supports the prop passing.

packages/ui/Layout/Cell.js
Djspaceg: This was marked as private because it's being provided internally by our ForwardRef HOC, which does (or at least should) have a public `ref` method.

I think like `componentRef`, though I'm only hesitant because of my comment above.

ryanjduffy:
The benefit of the `forwardRef` is that is reroute the ref callback into a new prop so if we applied it at the outermost point, it could safely travel through the stack. The issue then is how to put it in the right spot at the end.

For example, if we had the following, `ComponentBase` (which means every Base component in the framework) would need to add a prop for `componentRef`.

```
Decorator = compose(
  ForwardRef({prop: 'componentRef'}),
  Changeable,
  ComponentBase
);
```

We can avoid this with something like this but it adds another component layer just to reroute the prop.


```
Decorator = compose(
  ForwardRef({prop: 'componentRef'}),
  Changeable,
  // new HOC to rewrite componentRef back to ref at the bottom :(
  (Wrapped) => ({componentRef, ...rest}) => <Wrapped {...rest} ref={componentRef} />,
  ComponentBase
);
```

packages/ui/Layout/Cell.js
ryanjduffy: I was concerned that an author using `CellBase` wouldn't be able to capture a ref because `forwardedRef` stomps on it. However, looking at the implementation again, it looks like an author-provided `ref` prop directly to `CellBase` would override the undefined `forwardedRef`.

Given that, keeping it `@private` makes sense as it makes the `ref`s behave seamlessly in both cases. I do still prefer `componentRef` to `forwardedRef` fwiw.

Djspaceg:
And also extra consideration for if a HOC needs to use a ref of a Base component, **then** what?! :mindblown:

packages/ui/Spinner/Spinner.js
```diff
@@ -148,28 +156,31 @@ const Spinner = kind({
...
 				return (
-					<Component {...rest} />
+					<Component ref={componentRef} {...rest} />
 				);
 			}
 		}
 	}
 });
 
+const Spinner = ForwardRef({prop: 'componentRef'}, SpinnerBase);
```
webOS101: For nearly all these components, doc blocks were not added to reflect the new structure of the components (A base + a HOC wrapped version).  Also, the decorators were not updated to add a `@mixes` section for `ForwardRef`.  For consistency with the docs, we should probably update all these instances _or_ do not bother to export the Base version.

PR #2129, opened 2019-02-23
PLAT-72723: Fix first index reset on props update

viodragon2:
### Checklist

When metrics props (e.g. `direction`, `spacing`, etc.) update, children are incorrectly positioned. For example, changing `direction` in `VirtualGridList` after scrolling a bit, it will show a blank page.

Prior to lifecycle method update PR (#2108), it used to reset `firstIndex` state directly prior to making an updated `setState` call, but it's been removed as part of the fix. This left us with incorrect value of the "new first index", thus resulting in wrong positioned items.

### Resolution
Removed references of `this.state` in all related calculation logic and added parameters to each function that accepts "firstIndex" value.

### DIFF
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -229,7 +229,7 @@ const VirtualListBaseFactory = (type) => {
 
 			if (props.clientSize) {
 				this.calculateMetrics(props);
-				nextState = this.getStatesAndUpdateBounds(this.props);
+				nextState = this.getStatesAndUpdateBounds(props);
 			}
 
 			this.state = {
@@ -272,8 +272,6 @@ const VirtualListBaseFactory = (type) => {
 		}
 
 		componentDidUpdate (prevProps) {
-			this.hasDataSizeChanged = (prevProps.dataSize !== this.props.dataSize);
-
 			if (
 				prevProps.direction !== this.props.direction ||
 				prevProps.overhang !== this.props.overhang ||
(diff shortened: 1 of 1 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -272,8 +272,6 @@ const VirtualListBaseFactory = (type) => {
 		}
 
 		componentDidUpdate (prevProps) {
-			this.hasDataSizeChanged = (prevProps.dataSize !== this.props.dataSize);
```
ybsung: `this.hasDataSizeChanged` was used in `componentDidUpdate` function of `ui/Scrollable.js` and `ui/ScrollableNative.js` to know the timing when a VirtualList have finished rendering at last.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -229,7 +229,7 @@ const VirtualListBaseFactory = (type) => {
 
 			if (props.clientSize) {
 				this.calculateMetrics(props);
-				nextState = this.getStatesAndUpdateBounds(this.props);
```
ybsung: I had a mistake. It must be not `this.props` but `props`. Thanks for fixing it.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -284,9 +282,10 @@ const VirtualListBaseFactory = (type) => {
 				// eslint-disable-next-line react/no-did-update-set-state
 				this.setState(this.getStatesAndUpdateBounds(this.props));
 				this.setContainerSize();
-			} else if (this.hasDataSizeChanged) {
+			} else if (prevProps.dataSize !== this.props.dataSize) {
+				const newState = this.getStatesAndUpdateBounds(this.props, this.state.firstIndex);
 				// eslint-disable-next-line react/no-did-update-set-state
-				this.setState(this.getStatesAndUpdateBounds(this.props));
+				this.setState(newState);
```
ybsung: As of now, it seems that we could use the following code here like other code.
```
this.setState(this.getStatesAndUpdateBounds(this.props, this.state.firstIndex));
```

ybsung:
I forgot that I had to remove all references of `this.state` in the function as the first parameter of `setState`. I think that this draft PR looks good to me. @viodragon2 , thanks for your PR.

packages/ui/VirtualList/VirtualListBase.js
viodragon2: Needless to say, this is a terrible way of handling state change. I will revert the change, but we need to clean up this mess.

PR #2131, opened 2019-02-23
ENYO-5843: Added a component prop to Repeater to allow customization

Djspaceg:
### Checklist

`ui/Repeater` always renders a `<span>` tag. It's often useful to be able to change this.

### Resolution
Added a `component` prop to `ui/Repeater` to allow customization of its base level element. This logically makes sense because the child component can already be customized (via `childComponet`) and fits in line with the way other components' base elements are customized (via the `component` prop).

This simple change greatly improves the flexibility and utility of this component.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ## [unreleased]
 
+### Added
+
+- `ui/Repeater` prop `component` to allow customization of its base element
+
 ### Changed
 
 - `ui/Changeable` and `ui/Toggleable` to warn when both `[defaultProp]` and `[prop]` are provided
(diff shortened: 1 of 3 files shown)


ryanjduffy:
Had a wild hair to try to add support for `React.Fragment` as an allowed value for `Component`. It's possible to do without too much trouble but it subverts our build in accessibility role which doesn't seem like a good behavior to encourage from the framework. It's still possible to do with a custom component if an author had a reason to do so.

```js
const Fragment = ({children}) => <React.Fragment>{children}</React.Fragment>;
<Repeater component={Fragment}>{children}</Repeater>
```

Djspaceg:
@ryanjduffy wow, that's a cool idea! Then the Repeater would just melt away, leaving just the generated children. I can definitely imagine several scenarios where that would be awesomely useful and also horribly dangerous. It would be neat to try it out on something practical.

PR #2133, opened 2019-02-27
ENYO-5852: Enact-ui-tests: DatePicker and TimePicker - running same test on different builds gives different result

sugardave:
### Checklist

In certain cases, most notably in `TimePicker`, an `undefined` value is incorrectly overriding a supplied `defaultValue` on render.

### Resolution
I opted to use a state value (`derived`), initially `false`, that is checked in `getDerivedStateFromProps` and prevents updating in the case where a decorated component is rendered `open`.  Once the props have been derived the first time, the state is updated and the component will update properly.

### Additional Considerations
Component and UI tests seem to work fine with this change, but there might be a better way to handle this?

### Links
ENYO-5852

### DIFF
--- a/packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
+++ b/packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
@@ -109,22 +109,16 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
 
 		static getDerivedStateFromProps (props, state) {
 			const propValue = toTime(props.value);
-			if (state.derived) {
-				if (state.value !== propValue) {
-					return {
-						value: propValue
-					};
-				}
+			if (state.derived && state.value !== propValue) {
+				return {
+					value: propValue
+				};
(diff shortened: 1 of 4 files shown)


packages/moonstone/internal/DateTimeDecorator/DateTimeDecorator.js
```diff
@@ -105,10 +109,14 @@ const DateTimeDecorator = hoc((config, Wrapped) => {
 
 		static getDerivedStateFromProps (props, state) {
 			const propValue = toTime(props.value);
-			if (state.value !== propValue) {
+			if (state.derived && state.value !== propValue) {
```
webOS101: Maybe it would be better to use a `typeof props.value !== 'undefined'` to avoid applying bad values?

ryanjduffy:
I pushed an update to move the state determination from the constructor into `getDerivedStateFromProps` which allows us to avoid the extra state prop. I also added into this PR the fix for cancelling which I wanted to test while verifying this change.

PR #2135, opened 2019-02-28
PLAT-69177: Added single-line support to BodyText

Djspaceg:
### Checklist

App developers need a quick and internationalization-safe way to display one line of text that can optionally marquee if the text is too long. It should be easily customizable but come with a built-in set of rules for quickest usage.

### Resolution
A new prop `nowrap` was added to `moonstone/BodyText`, which enables this functionality.

### Additional Considerations
A new prop `component` was added to `ui/BodyText` to allow the changing of the root DOM node to something else. This was necessary due to `DIV` not being allowed to be used inside of a `P` tag, which results in a warning/error. In the case of single-line/marqueeing, the usual `P` tag is swapped out for a `DIV` tag.

### DIFF
--- a/packages/moonstone/BodyText/BodyText.js
+++ b/packages/moonstone/BodyText/BodyText.js
@@ -48,7 +48,6 @@ const BodyTextBase = kind({
 		 * [theming]{@link /docs/developer-guide/theming/}.
 		 *
 		 * @type {Boolean}
-		 * @default false
 		 * @public
 		 */
 		centered: PropTypes.bool,
--- a/packages/moonstone/BodyText/BodyText.js
+++ b/packages/moonstone/BodyText/BodyText.js
@@ -11,15 +11,19 @@
  */
 
 import kind from '@enact/core/kind';
-import UiBodyText from '@enact/ui/BodyText';
 import React from 'react';
 import PropTypes from 'prop-types';
 import compose from 'ramda/src/compose';
+import UiBodyText from '@enact/ui/BodyText';
 import Pure from '@enact/ui/internal/Pure';
 
+import {MarqueeDecorator} from '../Marquee';
 import Skinnable from '../Skinnable';
 
 import componentCss from './BodyText.module.less';
+
+// Create a Marquee using BodyText as the base
+const MarqueeBodyText = MarqueeDecorator(UiBodyText);
 
 /**
  * A simple text block component.
@@ -38,6 +42,18 @@ const BodyTextBase = kind({
 
 	propTypes: /** @lends moonstone/BodyText.BodyTextBase.prototype */ {
 		/**
+		 * Centers the contents.
+		 *
+		 * Applies the `centered` CSS class which can be customized by
+		 * [theming]{@link /docs/developer-guide/theming/}.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		centered: PropTypes.bool,
+
+		/**
 		 * Customizes the component by mapping the supplied collection of CSS class names to the
 		 * corresponding internal Elements and states of this component.
 		 *
@@ -48,7 +64,21 @@ const BodyTextBase = kind({
 		 * @type {Object}
 		 * @public
 		 */
-		css: PropTypes.object
+		css: PropTypes.object,
+
+		/**
+		 * Toggles multi-line (`false`) vs single-line (`true`) behavior. `nowrap` mode
+		 * automatically enables {@link moonstone/Marquee} so long text isn't permanently occluded.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		nowrap: PropTypes.bool
+	},
+
+	defaultProps: {
+		nowrap: false
 	},
 
 	styles: {
@@ -56,10 +86,27 @@ const BodyTextBase = kind({
 		publicClassNames: 'bodyText'
 	},
 
-	render: ({css, ...rest}) => {
+	computed: {
+		className: ({nowrap, styler}) => styler.append({nowrap})
+	},
+
+	render: ({centered, css, nowrap, ...rest}) => {
+		if (nowrap) {
+			return (
+				<MarqueeBodyText
+					component="div" // Assign a new component to BodyText, since DIV is not allowed inside a P tag (the default for BodyText)
+					marqueeOn="render"
+					{...rest}
+					alignment={centered ? 'center' : null} // Centering Marquee
+					centered={centered} // Centering UiBodyText
+					css={css}
+				/>
+			);
+		}
 		return (
 			<UiBodyText
 				{...rest}
+				centered={centered}
 				css={css}
 			/>
 		);
--- a/packages/moonstone/BodyText/BodyText.js
+++ b/packages/moonstone/BodyText/BodyText.js
@@ -67,18 +67,18 @@ const BodyTextBase = kind({
 		css: PropTypes.object,
 
 		/**
-		 * Toggles multi-line (`false`) vs single-line (`true`) behavior. `nowrap` mode
+		 * Toggles multi-line (`false`) vs single-line (`true`) behavior. `noWrap` mode
 		 * automatically enables {@link moonstone/Marquee} so long text isn't permanently occluded.
 		 *
 		 * @type {Boolean}
 		 * @default false
 		 * @public
 		 */
-		nowrap: PropTypes.bool
+		noWrap: PropTypes.bool
 	},
 
 	defaultProps: {
-		nowrap: false
+		noWrap: false
 	},
 
 	styles: {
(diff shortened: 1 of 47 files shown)


packages/moonstone/BodyText/BodyText.js
```diff
@@ -48,18 +64,49 @@ const BodyTextBase = kind({
...
 	},
 
 	styles: {
 		css: componentCss,
 		publicClassNames: 'bodyText'
 	},
 
-	render: ({css, ...rest}) => {
+	computed: {
+		className: ({nowrap, styler}) => styler.append({nowrap})
```
Djspaceg: This is technically not necessary at the moment, but I had intended to put a class stub into the stylesheet for future styling. Maybe this can be deleted. Let's see what direction the reviewers think is best.

packages/ui/BodyText/BodyText.js
```diff
@@ -35,6 +35,16 @@ const BodyText = kind({
 		 */
 		centered: PropTypes.bool,
```
teckliew: If we support centered, should we support right-align as well?

packages/moonstone/BodyText/BodyText.js
teckliew: Probably better to remove this and add them later as needed. Otherwise if we are keeping this, might as well add an empty ruleset for `.nowrap`

packages/ui/BodyText/BodyText.js
Djspaceg: It's possible, but that may be a deeper consideration for BodyText than the intent of this ticket.

To do it the way I'd want, which would be to mirror marquee's `alignment` prop, we'd need to deprecate `centered` and include some handling/conversion for it.

packages/moonstone/BodyText/BodyText.js
```diff
@@ -48,18 +64,49 @@ const BodyTextBase = kind({
...
+
+		/**
+		 * Toggles multi-line (`false`) vs single-line (`true`) behavior. `nowrap` mode
+		 * automatically enables {@link moonstone/Marquee} so long text isn't permanently occluded.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @public
+		 */
+		nowrap: PropTypes.bool
```
ryanjduffy: Why did you choose to use all lower case vs. camel case for this prop?

packages/moonstone/BodyText/BodyText.js
```diff
@@ -11,16 +11,20 @@
...
 
 import kind from '@enact/core/kind';
-import UiBodyText from '@enact/ui/BodyText';
 import React from 'react';
 import PropTypes from 'prop-types';
 import compose from 'ramda/src/compose';
+import UiBodyText from '@enact/ui/BodyText';
 import Pure from '@enact/ui/internal/Pure';
 
+import {MarqueeDecorator} from '../Marquee';
```
ryanjduffy: Wanted to note here that `ui/Marquee` doesn't have any i18n support which is added by the `moonstone` version. We should consider how that support will be added to `moonstone/BodyText`.

ryanjduffy: nvm ... this is the moonstone version ... :nothingtodohere: ...

packages/moonstone/BodyText/BodyText.js
Djspaceg: Habit from CSS. :facepalm: I did think about it, and slipped my mind that we allow camel case in prop names. I'll change this.

packages/ui/BodyText/BodyText.js
teckliew: Sounds good. That could be another issue that's out of this scope.

packages/moonstone/BodyText/BodyText.js
Djspaceg: Addressed by adding the nowrap class to the LESS file.

Djspaceg:
I believe all issues have been addressed, including adding tests to both UI and Moonstone.

packages/moonstone/BodyText/tests/BodyText-specs.js
```diff
@@ -0,0 +1,41 @@
+import React from 'react';
+import {mount, shallow} from 'enzyme';
+import BodyText from '../BodyText';
+import {MarqueeDecorator} from '../../Marquee';
```
teckliew: `shallow` and `MarqueeDecorator` aren't used, could probably clean these up.

Djspaceg: done

packages/moonstone/BodyText/BodyText.js
```diff
@@ -48,18 +64,49 @@ const BodyTextBase = kind({
...
+	computed: {
+		className: ({noWrap, styler}) => styler.append({noWrap})
+	},
+
+	render: ({centered, css, noWrap, ...rest}) => {
+		if (noWrap) {
+			return (
+				<MarqueeBodyText
+					component="div" // Assign a new component to BodyText, since DIV is not allowed inside a P tag (the default for BodyText)
+					marqueeOn="render"
```
viodragon2: Not suggesting to change, but I'm concerned if there's a use case for a `hover`.

packages/moonstone/BodyText/tests/BodyText-specs.js
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+import {mount} from 'enzyme';
+import BodyText from '../BodyText';
+import css from '../BodyText.module.less';
+
+describe('BodyText Specs', () => {
+
+	test('should support multi-line content', () => {
```
viodragon2: This doesn't describe the test case correctly. This is more like verifying the default component is `p` tag. Let's update the description text.

packages/ui/BodyText/tests/BodyText-specs.js
```diff
@@ -50,4 +50,15 @@ describe('BodyText Specs', () => {
...
+	test('should support changing the component element', () => {
+		const componentTag = 'address';
+		const subject = shallow(
+			<BodyText component={componentTag} />
+		);
+
+		const expected = componentTag;
+		const actual = subject.getElement().type;
+		expect(actual).toBe(expected);
+	});
```
viodragon2: Let's add one more test for supporting function components.

packages/moonstone/BodyText/tests/BodyText-specs.js
Djspaceg: The description is actually what I wanted to test. Do you have any advice on how the test should be rewritten to be more accurate to the description?

packages/ui/BodyText/tests/BodyText-specs.js
Djspaceg: Would a functional component being used in the prop be any different than a string?

packages/moonstone/BodyText/BodyText.js
Djspaceg: `rest` is spread after this is applied, so this will be overridden if the developer specifies `marqueeOn="hover" or any other marqueeOn values.

Djspaceg:
Regarding
>One question regarding the styling. Isn't the request Item looking single line which has no smaller top and bottom margin? Do we expect them to apply custom styles for those cases?

`BodyText` is set up in such a way that allows easy overriding of margins due to its low class precedence. This seemed like the cleanest approach: retaining P tag's default margins in single-line mode. It was rather unpleasant switching from multi-line with margins to single-line and no margins. Open to suggestions though.

packages/moonstone/BodyText/BodyText.js
viodragon2: Heh, that's a bit sneaky. The concern is whether developers want to do that and if so should it be documented? Or description in `noWrap` good enough?

We can follow up later if the use case comes up later. I think supporting marquee on hover for non-spottable component is weird IMO.

Djspaceg: I think the existing description is satisfactory. It references and links to marquee, so (to me at least) the marquee functionality, and by extension its props, would be available. Though, you're right, it's not explicit.

packages/ui/BodyText/tests/BodyText-specs.js
Djspaceg: I went ahead and added a test, just for good measure.

viodragon2:
It's a fair argument. I would also prefer this approach rather than adding one more text component. Should we add a description/example on how they can achieve the `Item`-looking single line text in a release note or a somewhere?

Djspaceg:
>Should we add a description/example on how they can achieve the Item-looking single line text in a release note or a somewhere?
That's a really good suggestion. We should include it in our release notification email; in the ticket, as example code; and maybe some other place, like a direct message to someone. I don't think the docs is a great place for it though.

The following should be sufficient:
```
.bodyTextItem {
    margin-top: 0;
    margin-bottom: 0;
    font-weight: normal;
}
```
```
<BodyText noWrap className={css.bodyTextItem}>...</BodyText>
```

packages/moonstone/BodyText/tests/BodyText-specs.js
Djspaceg: I didn't actually know of a "real" way to test this, other than just trusting that if marquee wasn't there (and P was) that the CSS wouldn't clip the additional lines of text.

PR #2137, opened 2019-03-01
Add componentRef to other components with component

ryanjduffy:
Adding `ForwardRef` to other components that have a `component` prop


ryanjduffy:
My informal test: verify each logged an DOM instance of `custom-element`.

```js

			<Panel {...props}>
				<Layout component="custom-element" ref={n => console.log('Layout', n)} />
				<Cell component="custom-element" ref={n => console.log('Cell', n)} />
				<ViewManager component="custom-element" ref={n => console.log('ViewManager', n)} />
				<ToggleItem component="custom-element" ref={n => console.log('ToggleItem', n)} />
				<Spinner component="custom-element" ref={n => console.log('Spinner', n)} />
				<Item component="custom-element" ref={n => console.log('Item', n)} />
				<SlotItem component="custom-element" ref={n => console.log('SlotItem', n)} />
				<Router path="/" component="custom-element" ref={n => console.log('Router', n)}>
					<Route path="/" component="route-element" />
				</Router>
			</Panel>
```

packages/ui/Item/Item.js
```diff
@@ -48,6 +50,14 @@ const ItemBase = kind({
 		 */
 		component: EnactPropTypes.renderable,
 
+		/**
+		 * Called with a reference to `component`
```
Djspaceg: I think these documentation lines are my only issue with this PR. If they could be converted to @link directly to the `component` prop, this would be a little better. `[component]{@link moonstone/Item.ItemBase#component}` (I think that's the syntax). It's a little more involved to write, but I get the impression that it will make the terse docs here clearer.

ryanjduffy: I'm not it adds much value in this case since `componentRef` is private so the docs won't render. Thoughts?

Djspaceg: I like being thorough, just in case. When I read the new docs with the eyes of a new developer, I was a little confused as to what "component" referred to. More words could have helped that, as well as a link to the direct thing it references. Now, also granted, the thing it refers to is directly above it, so it may be clear, as long as our docs are alphabetized and `component` isn't "isRequired" and is bumped to the top of the list. But, that's what I was thinking. We can move forward without it though. I'm probably being too concerned.

PR #2138, opened 2019-03-02
PLAT-74086: Update interop docs for 2.3.0

sugardave:
### Checklist

Documentation is needed for integrating Enact with existing projects and other libraries.

### Resolution
Added documentation for how to integrate Enact with existing projects and other libraries:
- Using third-party components in Enact apps
- Using Enact in React apps
- Special case for using `@enact/i18n` in ejected CRA apps

### Additional Considerations
In order to get the style override example to work, I had to use `!important` which is :sadpanda:

### Links
PLAT-74086

### DIFF
--- a/docs/interoperability.md
+++ b/docs/interoperability.md
@@ -132,6 +132,38 @@ class App extends Component {
 export default MoonstoneDecorator(App);
 ```
 
+### Internationalization Outside of Enact
+
+#### CRA
+In order to use the [`@enact/i18n`](../../developer-guide/i18n) library for internationalization, you can [eject](https://facebook.github.io/create-react-app/docs/available-scripts#npm-run-eject)
+your CRA app, install your required Enact libraries (plus, `@enact/dev-utils`), and update the webpack configuration.
+
+```
+// package.json
+
+...
+"devDependencies" : {
+  "@enact/dev-utils": "2.0.0"
+},
+"dependencies": {
+  ...
+  "@enact/i18n": "2.3.0",
+  "@enact/moonstone": "2.3.0",
+  ...
+}
+...
+
+// webpack.config.js
+...
+const {GracefulFsPlugin, ILibPlugin} = require('@enact/dev-utils');
+...
+plugins: [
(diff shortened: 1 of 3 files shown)


docs/interoperability.md
```diff
@@ -64,6 +68,102 @@ const App = kind({
...
+
+Behind the scenes, Enact supports using CSS modules and most Enact components provide public class
+names that can be overridden to easily change their style.
+
+#### CRA Example
+
+If you are using `react-scripts` version `2.0.0` or later, you can import style overrides by using
+the `[component].module.css` filename convention.  For earlier versions, you will need to eject
+the CRA app and modify the webpack configuration to enable CSS modules.  This [article](https://medium.com/nulogy/how-to-use-css-modules-with-create-react-app-9e44bec2b5c2)
+is a good reference for enabling CSS modules prior to `2.0.0`.
```
webOS101: Should this be `2.3.0`?

docs/theming.md
```diff
@@ -156,7 +156,7 @@ import React from 'react';
 import kind from '@enact/core/kind';
 import UiButton from '@enact/ui/Button';
 
-import componentCss from './Button.less';
+import componentCss from './Button.module.less';
```
webOS101: Nice catch on the `.module.less`.  Do we have any of that in any of the tutorials that we need to update?

docs/interoperability.md
sugardave: It's about the version of `react-scripts`.  Apps made with `create-react-app` that used `react-scripts` prior to 2.0.0 couldn't use this feature.

docs/theming.md
sugardave: Probably!  I will take a quick look.

docs/interoperability.md
webOS101: :disappear:

docs/theming.md
sugardave: I only found one more example in the Enact migration docs, and it's updated now.

PR #2139, opened 2019-03-04
ENYO-5866: Fix isRenderable and hoc to support special React types

ryanjduffy:
### Checklist

Values returned from `React.lazy()`, `React.memo()`, and `React.forwardRef()` did not pass `isRenderable` because they were objects rather than strings or functions.

### Resolution
Update `isRenderable` to check for these special cases allowing `core/hoc` to work correctly with them as well.

### Additional Considerations
With this change, I've also reviewed our existing prop types, standardized them with the following vocabulary, and implemented new internal validators to support this.

Type | Description | Custom Prop Type
-----|--|--
`Component` | A function or class component | `component`
`String\|Component` | A native DOM node or function or class component | `renderable`
`Component\|Element` | A function or class component or a React Element | `componentOverride`
`String\|Component\|Element` | A native DOM node or a function or class component or a React Element | `renderableOverride`

Also, while this doesn't change any APIs, it will update the documentation and I think we should consider it a minor-level change for the purposes of communication.


packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `core/util.isRenderable` to treat values returned by `React.lazy()`, `React.memo()`, and `React.forwardRef` as renderable
```
webOS101: Assuming the parens for `forwardRef()` weren't left off on purpose.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -262,7 +263,7 @@ const VideoPlayerBase = class extends React.Component {
 		 * @default `moonstone/VideoPlayer.MediaControls`
 		 * @public
 		 */
-		mediaControlsComponent: PropTypes.oneOfType([PropTypes.func, PropTypes.object]),
+		mediaControlsComponent: EnactPropTypes.componentOverride,
```
webOS101: Should we update the `@type` here as well?  I think it should be `{Component|Element}`.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -78,11 +79,11 @@ const ToggleItemBase = kind({
 		 * and must therefore respond to it in some way. It is recommended to use
 		 * [ToggleIcon]{@link ui/ToggleIcon} for this.
 		 *
-		 * @type {Component|Element}
+		 * @type {String|Component|Element}
 		 * @required
 		 * @public
 		 */
-		iconComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.element, PropTypes.func]).isRequired,
+		iconComponent: EnactPropTypes.renderableOverride.isRequired,
```
webOS101: This one really shouldn't be a plain string as it receives props.  What was the reasoning for adding `String` here?  Also, other `iconComponents` are generally just `Component`, but this one is also `Element`.

packages/moonstone/ToggleItem/ToggleItem.js
```diff
@@ -61,12 +62,12 @@ const ToggleItemBase = kind({
 		 * This component receives the `selected` prop and value, and must therefore respond to it in some
 		 * way. It is recommended to use [ToggleIcon]{@link moonstone/ToggleIcon} for this.
 		 *
-		 * @type {Component|Element}
+		 * @type {String|Component|Element}
 		 * @default null
 		 * @required
 		 * @public
 		 */
-		iconComponent: PropTypes.oneOfType([PropTypes.string, PropTypes.element, PropTypes.func]).isRequired,
```
webOS101: Same comment as `ui/ToggleItem`

ryanjduffy:
Good catch on ToggleItem. I matched the current prop types and didn't look at the implementation. Agree that they should be `Component|Element` given the non-standard props passed. Updated VideoPlayer docs as well to match the prop type.

PR #2140, opened 2019-03-05
PLAT-73548: TooltipDecorator recalculate layout

teckliew:
### Checklist

`Tooltip` position needed to be updated when its client has changed position. 

### Resolution
Recalculate tooltip layout on key down and when the client changes its attributes

### Additional Considerations
Recalculation occurs when:
- a key is pressed
- an attribute has been changed
- a child of the client has been added or removed
- client resize

### Links
PLAT-73548


packages/moonstone/TooltipDecorator/util.js
```diff
@@ -0,0 +1,155 @@
...
+
+/**
+ * Calculates the top and left position for `Tooltip`.
+ * Takes the output of `adjustAnchor`, `adjustDirection`, and `calcOverflow`.
+ *
+ * @method
+ * @memberof moonstone/TooltipDecorator
+ * @param   {Object} tooltipNode		The `getBoundingClientRect` values for tooltip node
+ * @param   {Object} clientNode			The `getBoundingClientRect` values for client node
+ * @param	{String} arrowAnchor		Anchor position from `adjustAnchor`
```
Djspaceg: :lol: Tab

packages/moonstone/TooltipDecorator/util.js
```diff
@@ -0,0 +1,155 @@
+// TooltipDecorator util.js
+//
+
+/**
```
webOS101: I was going to comment that we shouldn't document these with jsdoc blocks since they aren't reexported... However, VS Code automatically parses these out and makes the docs and types available.  It might be a good idea to revisit some of the other places we disabled the doc blocks and turn them private for this enhanced capability.

packages/moonstone/TooltipDecorator/TooltipDecorator.js
```diff
@@ -443,7 +377,8 @@ const TooltipDecorator = hoc(defaultConfig, (config, Wrapped) => {
 					onBlur: this.handleBlur,
 					onFocus: this.handleFocus,
 					onMouseOut: this.handleMouseOut,
-					onMouseOver: this.handleMouseOver
+					onMouseOver: this.handleMouseOver,
+					onKeyDown: this.handleKeyDown
```
ryanjduffy: Why have you added a key down handler in props with a global handler already attached?

packages/moonstone/TooltipDecorator/TooltipDecorator.js
```diff
@@ -346,13 +277,16 @@ const TooltipDecorator = hoc(defaultConfig, (config, Wrapped) => {
 
 		handle = handle.bind(this)
 
-		handleKeyDown = this.handle(
-			forProp('disabled', true),
-			() => {
+		handleKeyDown = (ev) => {
+			forward('onKeyDown', ev, this.props);
```
ryanjduffy: We shouldn't forward local event callbacks from a global handler since the events might not have originated from this DOM tree.

packages/moonstone/TooltipDecorator/TooltipDecorator.js
```diff
@@ -163,6 +164,9 @@ const TooltipDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			super(props);
 
 			this.TOOLTIP_HEIGHT = ri.scale(18); // distance between client and tooltip's label
+			/* global MutationObserver ResizeObserver */
+			this.mutationObserver = new MutationObserver(this.setTooltipLayoutJob.start);
+			this.resizeObserver = new ResizeObserver(this.setTooltipLayoutJob.start);
```
ryanjduffy: In the interest of cross-platform support, I would like to only conditionally create a `ResizeObserver` since it is relatively new and isn't widely supported. For example, it will throw an error in Firefox and Safari. Realize it means that we won't have the same positioning resilience as in chromium but that's acceptable for now. We can revisit alternatives in the future.

packages/moonstone/TooltipDecorator/TooltipDecorator.js
```diff
@@ -163,6 +164,9 @@ const TooltipDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			super(props);
 
 			this.TOOLTIP_HEIGHT = ri.scale(18); // distance between client and tooltip's label
+			/* global MutationObserver ResizeObserver */
```
ryanjduffy: Let's move this up to the top of the file for consistency.

packages/moonstone/TooltipDecorator/util.js
ryanjduffy: Please fix this :)

ryanjduffy:
Also, please add a few scenarios to a qa-sample to help verify this work now and in the future.

packages/moonstone/TooltipDecorator/util.js
teckliew: Should I take this out?

webOS101: no, no...  It's fine.

packages/moonstone/TooltipDecorator/TooltipDecorator.js
teckliew: I separated the keydown events for two different cases; I left notes/comments on L280 and L289 let me know if this works!

ryanjduffy:
:+1: I tweaked the qa sample a little to add some text and use `ui/Layout` to ensure things fit.

PR #2142, opened 2019-03-06
PLAT-75294: Picker: valueWrapper is flexible depend on overall size

Sangwook-Lee:
I suggests to remove `max-width` of Picker's text value area. 

### Checklist

Actually, I got some questions about that from app developers.
They wanted to expand depend on the full-size of the Picker.
Even if overall size of the Picker gets bigger, text value area's size does not expanded because of `max-width`.
Many applications can set various size to Picker directly, without using `width` prop as fixed size.
And, in some cases, Picker should be same to the size of parent area. (width: 100%)

### Resolution
So, I think text value's area should be flexible depend on overall size.
We can replace of `flex: 1` instead of `max-width`.

### Links
PLAT-75294


packages/moonstone/internal/Picker/Picker.module.less
```diff
@@ -17,7 +17,7 @@
 	.valueWrapper {
 		.moon-text-base();
 		.moon-locale-non-latin({line-height: @moon-button-small-height;});
-		max-width: 300px;
+		max-width: 480px;
```
teckliew: Was it decided that it's supposed to be `400px` or `480px`?

packages/moonstone/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
+### Changed
+
+- `moonstone/Picker` to change `max-width` of valueWrapper
```
ryanjduffy: We try to phrase change log entries in terms of the effect on the developer or user rather than the technical change that was made. We also generally reserve `Changed` for significant behavior or API changes that may require developer action to adopt.

I usually try to answer the question, "What problem might I have seen in my application that would no longer be present?" In this case, maybe something like:

> ### Fixed
>
> - `moonstone/Picker` to display more of the selected value in wide instances

PR #2143, opened 2019-03-06
Removed focused state from spotlight

rundmt:
### Checklist

Spotlight takes a long time to blur and focus, causing things like Scroller to not animate properly.

### Resolution
Optimizing spottable so it doesn't take a long time during focus and blur

### Links
PLAT-64252

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+### Unreleased
+
+### Fixed 
+- `spotlight/Spottable` to prevent unnecessary updates due to blur changes
+
+
 ## [2.4.0] - 2019-03-04
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


ryanjduffy:
I think you've got a good path here but wanted to point out some issues we might see with using a HOC-scoped variable. In most use cases, that will likely work okay but it's important to note that the `isFocused` member would be shared across instances of the same component. This is because `Spottable` is called when defining the component (~ `Button = Spottable(ButtonBase)`) and not when constructing the component (~ `<Button />`). As a result, every Button will share a reference to the same `isFocused`.

There are a couple potential side effects to this. First, if an instance becomes disabled while another of its type is focused, it's internal state will become out of sync as it will determine it was `focusedWhenDisabled` when it wasn't. Second, if an instance becomes unmounted while another of its type is focused, it will incorrectly fire `onSpotlightDisappear`. While we might be able to work around the effects, I think we should reconsider the premise and stick with instance members of some type.

Using this as inspiration, I prototyped an alternate approach that reverts back to an instance member for `isFocused`. It seems to work in my initial testing but could use some more eyes both from a overall suitability and a performance perspective.

https://github.com/enactjs/enact/compare/develop...feature/spottable-focused-state?expand=1

rundmt:
Nice, yeah I'll take a look at this and run it through the perf tests as well as `focusedWhenDisabled` tests.

PR #2144, opened 2019-03-06
ENYO-5764: Fix VideoPlayer to continue to show thumbnail on playback keypress

ryanjduffy:
### Checklist

When hovering the slider to show the thumbnail, pressing playback keys (play/pause) would hide the thumbnail.

### Resolution
Remove the state update in `showFeedback` to `feedbackAction` and instead rely on the other existing triggers (`hideFeedback`, `hideControls`).

### Additional Considerations
This feature was added by #1798. I've checked this against that bug report and didn't notice a regression. cc: @viodragon2 in case you recall specifics on why this code was added originally.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/VideoPlayer` to continue to show thumbnail when playback control keys are pressed
+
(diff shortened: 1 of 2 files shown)


viodragon2:
I don't remember exactly why I added this specific part, but my guess is `feedbackAction` needed to be reset. `feedbackAction` is not in `focus` state anymore whenever `showFeedback` is called in most cases. Feedbacks are shown when actions (e.g. play, pause, rew, ff, etc.) are registered. It's possible that this was an oversight for this specific use case where the action happens while retaining focus. I think this change is okay 👍

PR #2146, opened 2019-03-06
ENYO-5865: Fix remaining in pointer mode when the Escape key is pressed

ryanjduffy:
### Checklist

When pressing the escape key on a PC to "cancel", pointer mode would be turned off which would trigger different focus behaviors between webOS and PC.

### Resolution
Add `'cancel'` key test to spotlight in addition to `'nonModal'`

### Additional Considerations
I considered a couple different locations for this update but landed on updating spotlight.
* Add to `spotlight/src/pointer` - It's the only consumer of `nonModal` but declaring the keymapping there would couple spotlight to `ui/Cancelable` which defines the Escape key as the default `'cancel'` key.
* Add cancel keycode to moonstone's `nonModal` list for webOS - Seems out of place since 27 isn't a webOS key code
* Add `'cancel'` as an additional condition in `spotlight/src/pointer` - Seems to make the most sense as it effectively says: "spotlight will remain in pointer mode for 'nonModal' or 'cancel' key events" rather than distinguishing the particular keycode. It would also be more applicable in other environments/platforms.


packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.js
```diff
@@ -101,7 +101,6 @@ const MoonstoneDecorator = hoc(defaultConfig, (config, Wrapped) => {
 	addAll({
 		cancel: 461,
 		nonModal: [
-			461,
```
webOS101: I haver some concerns here.  Currently `nonModal` is only used in Spotlight, but that doesn't mean that it might not be used by apps or elsewhere in Moonstone in the future.  Is there any harm in leaving it here?

ryanjduffy: The inconsistency between webOS and non-webOS is a "harm" imo. If we were to leave it, I'd want to consider it deprecated and remove it in 3.0.

I think your concern is valid but given the complete absence of any documentation of `nonModal`, it seems like a low risk to me.

webOS101: Which inconsistency are you referring to?  The handling of escape in Spotlight if `moonstone` is not included?

ryanjduffy: That "back" on webOS is nonModal but "back" on PC isn't.

webOS101: Is the code in `spotlight` added below enough to fix that inconsistency or is there a further issue?

webOS101: Or is this just a comment on the weird key handling of the magic remote?

webOS101: My rationale for leaving the key entry in moonstone is that it is a non-modal key on the magic remote.  This might be important at some point.  I think adding the additional fix to spotlight is fine, too.

ryanjduffy: I think our approach for these keys should be revisited but not in the context of this ticket. I'll revert this change and we can revisit later.

In short, our definition of "non-modal" at inception was "key codes that don't cause pointer mode to be disabled." In hindsight, I think we'd have been better served by saying "key names that don't cause pointer mode to be disabled" in which case we would have had a list that included "cancel" and would have worked cross-platform.

PR #2147, opened 2019-03-06
PLAT-68767: Fix Scroll buttons to only update state when needed

haileyR:
### Checklist

Issue: Scrollbar buttons update for every animation frame/ scroll and this takes about 30-50 milliseconds.

### Resolution
This update will remove unnecessary Scrollbar buttons updates.


ryanjduffy:
To clarify, the proposed fix attempts to avoid the call to `setState` when we expect the result wouldn't be a change of state. Is that accurate, @haileyR? Is this a case where we think that calling `setState` at all -- even when the updater callback returns `null`/`undefined` is causing a performance impact?

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -156,19 +155,22 @@ class ScrollButtons extends Component {
...
-				return {nextButtonDisabled: shouldDisableNextButton};
-			}
-		});
+		const
+			updatePrevButton = (this.state.prevButtonDisabled !== shouldDisablePrevButton),
+			updateNextButton = (this.state.nextButtonDisabled !== shouldDisableNextButton);
+
+		if (updatePrevButton || updateNextButton) {
+			this.setState(() => {
+				if (updatePrevButton && updateNextButton) {
```
ryanjduffy: @haileyR Since you've already guarded above to avoid unnecessary updates, this could be simplified to set the object directly without the `if` branches.

```
this.setState({
  prevButtonDisabled: shouldDisablePrevButton,
  nextButtonDisabled: shouldDisableNextButton
});
```

I haven't tested this so it's possible I'm missing an edge case but it should be safe. A consideration for next time, perhaps. :)

PR #2149, opened 2019-03-07
ENYO-5867: Switched Spinner away from pseudo-elements for animation fidelity

Djspaceg:
### Checklist

On some devices, the spinner animation can get out-of-sync after props changes

### Resolution
Due to the way DOM nodes are laid out in react, `::before` and `::after` pseudo-elements get rendered slightly later on some hardware. Being tardy-to-the-party means that their animations aren't perfectly in sync. The `Spinner` is (was) using 3 DOM nodes, the first 2 of which had 2 pseudo-elements each (for a total of 4), which compose the "fan" part of the `Spinner`. The ball was its own DOM node, so it never got out of sync with where it should have been; the DOM node and the pseudo-elements were rendering at different times, causing the effect.

By switching the pseudo-elements to real DOM nodes, they all get rendered at the same time, allowing them to stay in sync when props change on the component.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Spinner` animation synchronization after a rerender
 - `moonstone/VideoPlayer` to continue to show thumbnail when playback control keys are pressed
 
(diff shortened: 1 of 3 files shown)


ryanjduffy:
I clicked the knob checkboxes as furiously as I could but couldn't reproduce the bug any longer. :clap: :clap: @Djspaceg !

PR #2151, opened 2019-03-08
PLAT-54923: Apply createRef and forwardRef to VirtualList and Scroller

ybsung:
### Checklist

To enhance code quality and sync up with the latest React version, it is needed to apply `createRef` and `forwardRef` to VirtualList and Scroller.

### Resolution

Changes
- Replace the function initializing ref variables with `React.createRef()`
- Replace `findDOMNode()` with `React.forwardRef()` if possible
- Move the code in the ref functions to `componentDidMount()`
- Remove ref variable definitions with `null` value because those definitions are initialized in `constructor()` (e.g> prevButtonNodeRef = null)

API changes
- I removed `setRef` public prop in `ui/ScrollThumb`. But it would be fine because the `ui/ScrollThumb` is a private component and it is used in VirtualList and Scroller internal only.

### Additional Considerations

- Remove `this.announce` in `ScrollButtons` and use `this.announceRef.current.announce` directly
- Re-position the following code in `moonstone/ScrollableNative` like the `moonstone/Scrollable`
```
	// voice control
	isVoiceControl = false
	voiceControlDirection = 'vertical'
```

### Links

PLAT-54923


packages/moonstone/Scrollable/ScrollButton.js
```diff
@@ -81,6 +81,8 @@ const ScrollButton = kind({
 	}
 });
 
+const ScrollButton = React.forwardRef((props, ref) => <ScrollButtonBase {...props} ref={ref} />);
```
0x64: Hmm.. A big advantage of using `forwardRef` is to encapsulate a component so that it would be independent from other components' structure. So, I think it may be better to move `forwardRef` code to a consumer component; `ScrollButtons` in our code.

ybsung: That's good idea.

packages/moonstone/Scroller/Scroller.js
```diff
@@ -343,30 +345,30 @@ class ScrollerBase extends Component {
...
+		// lie outside of this.uiRef.current.containerRef but within the spotlight container
 		if (contentsContainer && contentsContainer.contains(target)) {
 			this.scrollToBoundary(direction);
 		}
 	}
 
 	initUiRef = (ref) => {
 		if (ref) {
-			this.uiRef = ref;
+			this.uiRef = {current: ref};
```
0x64: I think we'd better change `uiRef` to another name to prevent any misunderstanding rather than assigning `{current: ref}` to make it look like a result of `forwardRef`.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -56,7 +56,10 @@ const
 		scrollWheelPageMultiplierForMaxPixel
 	} = constants;
 
-const TouchableDiv = Touchable('div');
+const TouchableDiv = React.forwardRef((props, ref) => {
```
0x64: Let's try to use `ui/ForwardRef`.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1188,78 +1194,60 @@ class ScrollableBase extends Component {
...
-	initHorizontalScrollbarRef = (ref) => {
-		if (ref) {
-			this.horizontalScrollbarRef = ref;
-		}
-	}
-
-	initVerticalScrollbarRef = (ref) => {
-		if (ref) {
-			this.verticalScrollbarRef = ref;
+			this.childRef = {current: ref};
```
0x64: Same opinion as `uiRef` above.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1188,78 +1194,60 @@ class ScrollableBase extends Component {
 	// ref
 
 	getScrollBounds () {
-		if (typeof this.childRef.getScrollBounds === 'function') {
-			return this.childRef.getScrollBounds();
+		if (this.childRef && this.childRef.current && typeof this.childRef.current.getScrollBounds === 'function') {
```
0x64: Same opinion as `uiRef` above. We can omit unnecessary guard code here.

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -22,7 +23,7 @@ const
  * @param {String} value - CSS Variable value.
  */
 const setCSSVariable = (element, variable, value) => {
-	element.style.setProperty(variable, value);
+	ReactDOM.findDOMNode(element).style.setProperty(variable, value); // eslint-disable-line react/no-find-dom-node
```
0x64: Do we need `findDOMNode` here?

packages/moonstone/Scroller/Scroller.js
ybsung: Rename `uiRef` as `uiRefCurrent`. Same as `uiChildRef`.

packages/ui/Scrollable/Scrollbar.js
ybsung: Currently, I could not find any solution for it. Let's revisit this with another issue.

0x64:
LGTM

PR #2152, opened 2019-03-08
 PLAT-64255: Optimize ExpandableItem render performance

germboy:
### Checklist

`ExpandableItem` performs an additional update/render after mounting due to its use of `Transition` - which performs an idle job in order to render its children. This behavior is generally ideal in that it optimizes mount time initially. However when used in views (`Panel`), it can block animations and result in longer transition time between panels.

### Resolution
We can utilize `ExpandableSpotlightDecorator` to only mount the transition container when it becomes focused (to prepare for the assumed next open state) or when specifically toggled open. When performing this via a job, we can prevent this transition container rendering when simply moving a pointer over & out of an `ExpandableItem` - where we weren't expecting to open the expandable - saving us the trouble of needlessly rendering.


germboy:
I meant to open this as a draft PR (if it even matters) - but I'd like to discuss whether we think this is the right general approach for the perf mounting issue.

I notice that tests are failing with this change - although it's not immediately clear to me why. But I can clean that up as we progress this ticket.

germboy:
Due to the nature of handling child container rendering in `ExpandableSpotlightDecorator`, it will be necessary to use and test against the full version of our expandable components, rather than the base components when testing contained children. As far as I know, we're expecting developers to use the full component so I don't believe this is a breaking change, but correct me if I'm wrong.
That being the case, I've updated the tests to reflect this behavior. In the case of `DatePicker` and `TimePicker`, the full component also uses i18n, which alters some of the expected results.

ryanjduffy:
It was a bit weird that we were `mount()`-ing base components anyways so not too concerned about the test changes.

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -250,6 +250,15 @@ const ExpandableItemBase = kind({
...
 		open: PropTypes.bool,
 
+		/**
+		 * Prevents rendering the transition container on mount.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @private
+		 */
+		preventTransitionContainerRender: PropTypes.bool,
```
ryanjduffy: This prop name is a bit unconventional in the framework. It's very specific to the implementation details of the effect. What are some other options we should consider that are a bit more generalized? Maybe `hideChildren` as is used in Panel?

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -250,6 +250,15 @@ const ExpandableItemBase = kind({
 		 */
 		open: PropTypes.bool,
 
+		/**
+		 * Prevents rendering the transition container on mount.
```
ryanjduffy: Independent of the prop naming comment, the "on mount" portion isn't really accurate in this context.

ryanjduffy:
Code looks sound to me. I won't have time to test in detail before EOD so I'm going to defer merging for now.

PR #2154, opened 2019-03-08
Spottable: added guards and removed focused state

rundmt:
### Checklist

Spotlight was taking a long time to run focus and blur

### Resolution
Removed focused state and have a guard for focusWhenDisabled

### Additional Considerations
doing some hacky things to avoid lifecycle methods, but this works.

### Links
ENYO-5878


packages/spotlight/Spottable/Spottable.js
```diff
@@ -408,8 +402,10 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		render () {
-			const {disabled, spotlightId, ...rest} = this.props;
-			const spottable = this.state.focusedWhenDisabled || isSpottable(this.props);
+			const {disabled, spotlightId, spotlightDisabled, ...rest} = this.props;
+			this.focusedWhenDisabled = this.isFocused && (disabled || spotlightDisabled);
+			const spottable = this.focusedWhenDisabled || isSpottable(this.props);
```
germboy: I'm concerned that rendering based off an instance property value `this.isFocused`/`this.focusedWhenDisabled` will cause inconsistent renders, since it's not guaranteed to re-render when these values change.

Part of my concern is with the way we're attempting to manage the two different `focusedWhenDisabled` states (`this.focusedWhenDisabled` and `this.state.focusedWhenDisabled`). It looks like it would be easy for these to become out of sync (see my comment above) or not be able to provide the current value needed within the given method or life-cycle. It seems like we could utilize the existing state property and adjust the behavior to allow or prevent updates (resulting in unwanted renders) in `sCU`. Thoughts?

rundmt: Yeah I see the concern with having two separate variables.  What would you think about instead of using state, we could technically do a `this.forceUpdate` and not have any state? The reason for not using state, is because we need to mutate in the render method so `setState` would be bad there, but this way we could avoid `sCU` all together.

rundmt: With `this.forceUpdate` we can address the above comment while still avoiding lifecycle of `setState`

packages/spotlight/Spottable/Spottable.js
```diff
@@ -362,9 +345,13 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 
 		handleBlur = (ev) => {
 			if (this.shouldPreventBlur) return;
-
-			if (ev.currentTarget === ev.target) {
-				this.setState({focused: false, focusedWhenDisabled: false});
+			this.isFocused = false;
```
germboy: We should move this down to be guarded against `ev.currentTarget === ev.target` since if a spottable contains a spottable (it's not a real typical scenario, but there is spotlight support for it) the focused state will be incorrectly set here. That will mean moving the `this.focusedWhenDisabled` to guard the `focusedWhenDisabled` value change and update - like in original `Spottable` implementation: https://github.com/enactjs/enact/blob/237836b72c49ceecd12fd9eda34ac81aae012395/packages/spotlight/Spottable/Spottable.js#L358

germboy:
I've removed the failing 'should not re-render when a non-Spottable prop does not change' test, which was added when we originally added the `shouldComponentUpdate` lifecycle to `Spottable`, so no functionality has been changed from the initial `Spottable` implementation. To run such a test, we would need to consider turning Spottable into a pure component

PR #2156, opened 2019-03-09
ENYO-5874: Bump up clientSize prop to Scrollable

viodragon2:
### Checklist

ScrollBar accesses the container DOM for calculating thumb size.

### Resolution
`VirtualList` has `clientSize` to gain performance by reducing DOM access and is the same size as the size which `ScrollBar` needs for the calculation and it's also used in `Scrollable`. We can bump up the prop to `Scrollable` and reduce layout for `ScrollBar`.

### Additional Considerations
From a local perf test (MBP 15", CPU 6x slowdown) done with https://github.com/enactjs/performance-tests/pull/25, I was able to see ~10ms improvement on average. 


ryanjduffy:
@viodragon2, @seunghoh - How frequently is the `clientSize` prop used by applications? Curious how much impact this change will have in the wild.

viodragon2:
Only in one panel in one app. I believe this prop was not promoted well enough for app developers to explore.

ryanjduffy:
Okay. I can see the perf benefit of avoiding layout calculations but I'm questioning the value as compared to the DX. Seems like specifying a static value (even when using `ri` to scale it) is prone to maintenance problems when layouts change (e.g. when we have UX updates or global changes like line-height that adjusts layout) and is contrast to typical flexible web layouts.

More concretely, it this the sort of API we want to promote or do we need to find a way to minimize the impact of style recalc and layout by the scrollbar?

viodragon2:
It depends. Main use cases for TV apps normally don't require a flexible layout. So if there's a perf benefit and maybe an extra work(?), would developers choose to opt-in for this option?

Also, other virtualized libraries normally require explicit dimensions set. In the case of `react-virtualized`, they provide a HOC for flexible layout (i.e. `AutoSizer`). I don't think one is better than the other API-wise IMO.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -136,6 +136,19 @@ class ScrollableBase extends Component {
...
+		/**
+		 * Client size of the container; valid values are an object that has `clientWidth` and `clientHeight`.
+		 *
+		 * @type {Object}
+		 * @property {Number}    clientHeight    The client height of the list.
+		 * @property {Number}    clientWidth    The client width of the list.
+		 * @public
+		 */
+		clientSize: PropTypes.shape({
+			clientHeight: PropTypes.number.isRequired,
```
webOS101: Is it really necessary to supply both values if your container is only scrolling in one direction?  Should we elaborate that this is an opt-in feature and why you'd want to use it?

ryanjduffy: I think you make a good point but I'm not sure we want to muck with the API right now since we'd want to change VirtualList to match this. We could make it backwards compatible by supporting either a single value or the object but would then add complexity to handle it internally. Guess it depends on how important you feel that DX change is.

Adding docs (here and in VL) makes sense though.

webOS101: I think simply removing the `isRequired` would be sufficient.  Don't need to change it to accept a single value vs. object, imo.

seunghoh: Hm.. probably you are right for scrollbar's perspective but VL needs both for calculating item size. I am not confident it will work only with `clientHeight` or `clientWidth`. Our intention was `clientSize` should have both for saving render cycle, or it will be OK without `clientSize` if the perf. is not too bad

viodragon2: `clientSize` is to specify the dimension of the container's viewport essentially, so it should require both values. I understand it's probably unnecessary for `Scrollbar` to have both values as @seunghoh mentioned, but `Scrollable` is the parent component of both `Scrollbar` and `VirtualList` so it should require both values here.

On a side note, these are all private components, so I suggest we revisit later if you're concerned with APIs. As long as we agree on the fact that it gives a performance boost, I say go with it.

PR #2157, opened 2019-03-11
Prevent focused functions until window ready

webOS101:

### Checklist

If `onWindowFocus` was called without a window or before the window was ready then the callback function could be executed.

### Resolution
Only set `focused` once the window is ready

### Additional Considerations
Current usage guards against a problem but things could easily change.

### Links
ENYO-5881

### DIFF
--- a/packages/i18n/I18nDecorator/windowFocus.js
+++ b/packages/i18n/I18nDecorator/windowFocus.js
@@ -1,7 +1,7 @@
 import {on} from '@enact/core/dispatcher';
 import {onWindowReady} from '@enact/core/snapshot';
 
-let focused = true;
+let focused = false;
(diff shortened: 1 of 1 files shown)


packages/i18n/I18nDecorator/windowFocus.js
```diff
@@ -23,6 +23,7 @@ const onWindowFocus = (handler) => {
 };
 
 onWindowReady(() => {
+	focused = true;	// Treat window as initially focused once ready
```
ryanjduffy: We should call `flush()` as well here.

webOS101: :oops:  Yes, of course.

PR #2160, opened 2019-03-12
ENYO-5728: Fix spotlight to guard against accessing non-existent containers

ryanjduffy:
### Checklist

Attempting to focus a configured container that wasn't mounted threw a runtime error.

### Resolution
Add a couple guards to prevent throwing errors

### Additional Considerations
This was originally manifest in unit testing an application but could be reproduced in standalone scenarios as well.

### DIFF
--- a/packages/spotlight/src/tests/container-specs.js
+++ b/packages/spotlight/src/tests/container-specs.js
@@ -1220,7 +1220,7 @@ describe('container', () => {
 		);
 
 		test(
-			'should return an empty array for a unmounted, configured container',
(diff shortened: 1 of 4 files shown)


packages/spotlight/src/tests/container-specs.js
```diff
@@ -1218,5 +1218,22 @@ describe('container', () => {
 				}
 			)
 		);
+
+		test(
+			'should return an empty array for a unmounted, configured container',
```
Djspaceg: "an". "unmounted" begins with a vowel. I'll just fix this.

PR #2161, opened 2019-03-13
PLAT-65757: Move to a previous or next page via a channel up and down immediately

ybsung:
### Checklist

The UX for page navigation via channel up and down was updated. VirtualList or Scroller jump to a previous or next page via a channel up and down immediately instead of jumping to the first or last visible item in the current page.

### Resolution

There were 3 conditions to jump via channel up and down in VirtualList and Scroller with the previous UX.
Priority 1> If Spotlight was not on the first or last visible item in the current page, they jumped to it.
Priority 2> If Spotlight was on the first or last visible item in the current page and there was the spottable item in the previous or next page, it jumped to the first or last item in the page.
Priority 3> If Spotlight was on the first or last visible item in the current page and there was no the spottable item in the previous or next page, it jumped to the next spottable item.

For the update UX for page navigation, I just moved the Priority 1> to the lowest priority. So VirtualList and Scroll will jump to the previous or next page immediately via channel up and down.

### Additional Considerations

Even though changing the priority, Scroller did not meet the full page navigation UX yet. So I also fixed it together.

### Comments

PLAT-65757

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -9,7 +9,7 @@
 import classNames from 'classnames';
 import {constants, ScrollableBase as UiScrollableBase} from '@enact/ui/Scrollable';
 import {getDirection} from '@enact/spotlight';
-import {getTargetByDirectionFromElement, getTargetByDirectionFromPosition} from '@enact/spotlight/src/target';
+import {getTargetByDirectionFromElement} from '@enact/spotlight/src/target';
 import {Job} from '@enact/core/util';
 import platform from '@enact/core/platform';
 import {forward} from '@enact/core/handle';
@@ -432,6 +432,7 @@ class ScrollableBase extends Component {
 				}
 			}
 
+			// Need to check whether an overscroll effect is needed
 			return true;
 		}
 
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -422,7 +422,12 @@ class ScrollableBase extends Component {
 		// Should skip scroll by page when focusedItem is paging control button of Scrollbar
 		if (focusedItem && childRefCurrent.containerRef.current.contains(focusedItem)) {
 			if (this.childRef.current.scrollByPage) {
-				const next = this.childRef.current.scrollByPage({direction, focusedItem, reverseDirection: reverseDirections[direction], spotlightId: containerRef.current.dataset.spotlightId});
+				const next = this.childRef.current.scrollByPage({
+					direction,
+					focusedItem,
+					reverseDirection: reverseDirections[direction],
+					spotlightId: containerRef.current.dataset.spotlightId
+				});
 
 				if (next !== null) {
 					this.animateOnFocus = false;
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -421,38 +421,17 @@ class ScrollableBase extends Component {
 
 		// Should skip scroll by page when focusedItem is paging control button of Scrollbar
 		if (focusedItem && childRefCurrent.containerRef.current.contains(focusedItem)) {
-			const
-				// VirtualList and Scroller have a spotlightId on containerRef
-				spotlightId = containerRef.current.dataset.spotlightId,
-				rDirection = reverseDirections[direction],
-				viewportBounds = containerRef.current.getBoundingClientRect(),
-				focusedItemBounds = focusedItem.getBoundingClientRect(),
-				endPoint = {
-					x: focusedItemBounds.left + focusedItemBounds.width / 2,
-					y: viewportBounds.top + ((direction === 'up') ? focusedItemBounds.height / 2 - 1 : viewportBounds.height - focusedItemBounds.height / 2 + 1)
-				};
-			let next = null;
+			const spotlightId = containerRef.current.dataset.spotlightId;
 
-			/* 1. Find spottable item in viewport */
-			next = getTargetByDirectionFromPosition(rDirection, endPoint, spotlightId);
-
-			if (next !== focusedItem) {
-				Spotlight.focus(next);
-			/* 2. Find spottable item out of viewport */
-			// For Scroller
-			} else if (this.childRef.current.scrollToNextPage) {
-				next = this.childRef.current.scrollToNextPage({direction, focusedItem, reverseDirection: rDirection, spotlightId});
+			if (this.childRef.current.scrollByPage) {
+				const next = this.childRef.current.scrollByPage({direction, focusedItem, reverseDirection: reverseDirections[direction], spotlightId});
 
 				if (next !== null) {
 					this.animateOnFocus = false;
 					Spotlight.focus(next);
 				}
-			// For VirtualList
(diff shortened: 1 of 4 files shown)


packages/moonstone/Scroller/Scroller.js
```diff
@@ -298,49 +298,55 @@ class ScrollerBase extends Component {
 		}
 	}
 
-	getNextEndPoint = (direction, oSpotBounds) => {
-		const bounds = this.uiRefCurrent.getScrollBounds();
-
+	getNextEndPoint = (direction, scrollerBounds, focusedItemBounds, pageDiff) => {
```
0x64: I prefer a boolean `ignoreCurrentPage` or something than a number `pageDiff` for limitation for clearer understanding.

packages/moonstone/Scroller/Scroller.js
```diff
@@ -298,49 +298,55 @@ class ScrollerBase extends Component {
...
-				oPoint.x = oSpotBounds.left - bounds.clientWidth;
-				oPoint.y = oSpotBounds.top;
+				oPoint.x = focusedItemBounds.left - scrollerBounds.width;
+				oPoint.y = focusedItemBounds.top;
 				break;
 			case 'down':
-				oPoint.x = oSpotBounds.left;
-				oPoint.y = oSpotBounds.top + oSpotBounds.height + bounds.clientHeight;
+				oPoint.x = focusedItemBounds.left;
+				oPoint.y = scrollerBounds.top + scrollerBounds.height * (pageDiff + 1);
```
0x64: What about to use scrollerBounds.bottom?
> oPoint.y = scrollerBounds.bottom + ignoreCurrentPage ? scrollerBounds.height : 0;

packages/moonstone/Scroller/Scroller.js
```diff
@@ -298,49 +298,55 @@ class ScrollerBase extends Component {
...
 				break;
 			case 'right':
-				oPoint.x = oSpotBounds.left + oSpotBounds.width + bounds.clientWidth;
-				oPoint.y = oSpotBounds.top;
+				oPoint.x = focusedItemBounds.left + focusedItemBounds.width + scrollerBounds.width;
```
0x64: I'm really confused about the rule for this case here.

packages/moonstone/Scroller/Scroller.js
```diff
@@ -298,49 +298,55 @@ class ScrollerBase extends Component {
...
+			scrollerBounds = this.uiRefCurrent.containerRef.current.getBoundingClientRect(),
+			focusedItemBounds = focusedItem.getBoundingClientRect();
+		let candidateNode = null, endPoint;
+
+		/* Find a spottable item in the current page */
+		endPoint = this.getNextEndPoint(direction, scrollerBounds, focusedItemBounds, 0);
+		const candidateNodeInCurrentPage = getTargetByDirectionFromPosition(reverseDirection, endPoint, spotlightId);
 
 		/* Find a spottable item in the next page */
+		endPoint = this.getNextEndPoint(direction, scrollerBounds, focusedItemBounds, 1);
```
0x64: If a candidate is found in the current page, do we need to run this line?

packages/moonstone/Scroller/Scroller.js
```diff
@@ -298,49 +298,55 @@ class ScrollerBase extends Component {
...
 		candidateNode = getTargetByDirectionFromPosition(reverseDirection, endPoint, spotlightId);
 
 		/* Find a spottable item in a whole data */
-		if (candidateNode === focusedItem) {
+		if (candidateNode === focusedItem || candidateNode === candidateNodeInCurrentPage) {
```
0x64: When this if clause is true, what is the value of `endPoint`?

packages/moonstone/Scroller/Scroller.js
```diff
@@ -298,49 +298,55 @@ class ScrollerBase extends Component {
...
 			candidateNode = getTargetByDirectionFromPosition(direction, endPoint, spotlightId);
 		}
 
-		/* If there is no spottable item next to the current item */
 		if (candidateNode === focusedItem) {
 			return null;
+		} else if (candidateNode === candidateNodeInCurrentPage) {
+			return candidateNodeInCurrentPage;
```
0x64: The returned value of this line is as same as `candidateNode`, so we will get the same result without this _else if_ clause.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -421,35 +421,15 @@ class ScrollableBase extends Component {
...
-			if (next !== focusedItem) {
-				Spotlight.focus(next);
-			/* 2. Find spottable item out of viewport */
-			// For Scroller
-			} else if (this.childRef.current.scrollToNextPage) {
-				next = this.childRef.current.scrollToNextPage({direction, focusedItem, reverseDirection: rDirection, spotlightId});
+			const spotlightId = containerRef.current.dataset.spotlightId;
+
+			if (this.childRef.current.scrollByPage) {
+				const next = this.childRef.current.scrollByPage({direction, focusedItem, reverseDirection: reverseDirections[direction], spotlightId});
```
0x64: `spotlightId` is used just once. We don't need to define the const variable.

0x64:
I have two general questions. The first one: Is removing a searching logic to find the target in the current page from `scrollByPage` in _Scrollable_ intended? The second one: do this PR affect any behaviors on overscroll effects? (overscroll logic depends on the result of `scrollByPage` method.)

packages/moonstone/Scroller/Scroller.js
ybsung: I replaced `pageDiff` with `isInNextPage`.

packages/moonstone/Scroller/Scroller.js
ybsung: I agree with your opinion.

packages/moonstone/Scroller/Scroller.js
ybsung: We don't need the `right` and `left` cases. So I removed those cases.

packages/moonstone/Scroller/Scroller.js
ybsung: Yes. We need to find the spottable item in the next page or after the next page first.

packages/moonstone/Scroller/Scroller.js
ybsung: The code was not clear. So I cleaned up.

packages/moonstone/Scroller/Scroller.js
ybsung: That's true. I cleaned up the return cases again.

packages/moonstone/Scrollable/Scrollable.js
ybsung: I'll fix it.

ybsung:
@0x64 
>The first one: Is removing a searching logic to find the target in the current page from scrollByPage in Scrollable intended?

We need to search the spottable item in the next page first. So I remove the logic that you mentioned.

> The second one: do this PR affect any behaviors on overscroll effects? (overscroll logic depends on the result of scrollByPage method.)

I didn't change the return value for `scrollByPage` in `Scrollable`. So the behaviors won't be changed.

ryanjduffy:
Closing in favor of #2262

PR #2163, opened 2019-03-13
ENYO-5822: Enact QA-Sampler: create a sample to test long text and verify height of Notification extends

sugardave:
### Checklist

It is difficult to quickly enter enough text with the VKB to make the `Notification` resize.  This QA sample lets you add or remove lines to allow quick verification of height changes.

### Resolution
Added a QA sample for `moonstone/Notification` and dynamic text.

### Links
ENYO-5822

### DIFF
--- a/packages/sampler/stories/qa/Notification.js
+++ b/packages/sampler/stories/qa/Notification.js
@@ -12,13 +12,12 @@ const Config = mergeComponentMetadata('Notification', Notification, Popup);
 
 Notification.displayName = 'Notification';
 
-const filler = '0123456789abcdefghijklmnopqrstuvwxyz';
-
-const messageArray = [filler];
-
 class StatefulNotification extends React.Component {
 	constructor (props) {
 		super(props);
+
+		this.messageFiller = '0123456789abcdefghijklmnopqrstuvwxyz';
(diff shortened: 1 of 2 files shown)


packages/sampler/stories/qa/Notification.js
```diff
@@ -0,0 +1,64 @@
...
+import {boolean} from '../../src/enact-knobs';
+import {mergeComponentMetadata} from '../../src/utils';
+
+const Config = mergeComponentMetadata('Notification', Notification, Popup);
+
+Notification.displayName = 'Notification';
+
+const filler = '0123456789abcdefghijklmnopqrstuvwxyz';
+
+const messageArray = [filler];
```
ryanjduffy: Curious why you chose to use a module-scoped member here vs instance state. Can you elaborate a bit?

sugardave: I didn't really consider it, it was just how I started with a solution and I left it there :D   I'll move them to instance members, they don't really need to be in the state, I think.

webOS101:
Could also just have used [`repeat()`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/repeat) with a number.

PR #2164, opened 2019-03-14
PLAT-73816: Add itemIcon prop to ui/ToggleIcon

Sangwook-Lee:
### Checklist

Actually, there're some use cases (UX scenario) using toggleItem with Images.
So, applications customize component to use `ToggleItem` with `Image/Icon`. 

### Resolution
Added `slotImageAfter` and `slotImageBefore` to `ui/ToggleItem`.

### Links
PLAT-73816


packages/moonstone/ToggleIconItem/ToggleIconItem.js
```diff
@@ -0,0 +1,119 @@
...
+ * @memberof moonstone/ToggleIconItem
+ * @ui
+ * @public
+ */
+
+// eslint-disable-next-line
+const itemIconCreator = (position) => ({css, itemIcon, itemIconPosition}) => {
+	if (position === itemIconPosition) {
+		return (
+			itemIcon !== null ?
```
Djspaceg: We could move this up to the `if` 2 lines above and have a simpler return:
```if (position === itemIconPosition && itemIcon != null) {```
Also, `!= null` will catch `undefined` as well as `null`, rather than `!== null` which only catches `null`.

File: packages/ui/ToggleItem/ToggleItem.js [hunk 2]
```diff
@@ -15,25 +15,30 @@ import kind from '@enact/core/kind';
...
 		return (
-			<ComponentOverride
-				component={iconComponent}
-				disabled={disabled}
-				selected={selected}
-			>
-				{icon}
-			</ComponentOverride>
+			<Fragment>
+				{iconPosition === 'before' && beforeItem ? beforeItem : null}
```
Djspaceg: I think this and the part below can be simplified to just the following:
```
{iconPosition === 'before' && beforeItem}
```
Was there a case you found that I didn't account for?

packages/moonstone/ToggleIconItem/ToggleIconItem.js
```diff
@@ -0,0 +1,119 @@
+import kind from '@enact/core/kind';
```
Djspaceg: I'm not sure creating a new component is necessary for what we're trying to do. I also don't seem to be able to locate where it is used in the framework. I assume it's something that an App would have to import and use, rather than just passing in an existing icon component.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -224,6 +287,7 @@ const ToggleItemBase = kind({
  */
 const ToggleItemDecorator = compose(
 	ForwardRef({prop: 'componentRef'}),
+	Slottable({slots: ['itemIconAfter', 'itemIconAfterText', 'itemIconBefore', 'itemIconBeforeText']}),
```
Sangwook-Lee: I found use cases that using multiple `itemIcon`s in the latest UX scenario. So, I added 4 slots.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -15,26 +15,49 @@ import kind from '@enact/core/kind';
...
 		return (
-			<ComponentOverride
-				component={iconComponent}
-				disabled={disabled}
-				selected={selected}
-			>
-				{icon}
-			</ComponentOverride>
+			<Fragment>
+				{itemIconBefore && <ComponentOverride component={itemIconBefore} className={`${css.itemIcon} ${itemIconBefore.props.className}`} />}
```
Sangwook-Lee: Can we approach to other way to get `className` from applications?

packages/moonstone/ToggleIconItem/ToggleIconItem.js
Sangwook-Lee: I've removed new component `ToggleIconItem`. As you commented, applications do not need to select type of ToggleItem.

packages/ui/ToggleItem/ToggleItem.js
Djspaceg: I have an idea about how to do this better, but am interested in @ryanjduffy's opinion on this matter.

packages/ui/ToggleItem/ToggleItem.js
Djspaceg: The presence of these 4 slots implies that it's completely acceptable to have **all 4 icons** displaying on one ToggleItem at the same time, in addition to the built-in `ToggleIcon`. I would like to _strongly discourage_ this, as it is an unsustainable component design. If there is such a need for an enormous amount of icons on an item, we have either failed in designing the component API or the UX.

packages/ui/ToggleItem/ToggleItem.module.less
```diff
@@ -1,3 +1,7 @@
 .toggleItem {
 	/* Placeholder for CSS overrides */
+	.itemIcon {
+		background-position: 0;
```
Djspaceg: I do not believe it is wise to arbitrarily reassign these rules to different values, and _especially_ not in `ui/ToggleItem`. These are properties that Moonstone has defined and all icons adhere to. If an app developer wants them to be different for whatever reason, it is advised that they do this in their own app on their one component.

packages/moonstone/SlotItem/SlotItem.module.less
```diff
@@ -28,7 +28,7 @@
 			> * {
 				margin-top: 0;
 				margin-bottom: 0;
-				vertical-align: top;
+				vertical-align: center;
```
Djspaceg: What are the implications to other uses of `SlotItem` by making this change?
The comment a few lines above explicitly calls-out vertical alignment, so I have to assume that `vertical-align: top;` was written out _on purpose_.

packages/ui/ToggleItem/ToggleItem.js
Sangwook-Lee: That makes sense.
We have not seen more than 2 displays so far. (Maximum 2) So why do not we just select 2 `itemIconPosition`s and suggest using all use cases equally?

I also think the best way is that there is just 1 slot. And we can choose the position. (before, beforeText, after, afterText)

packages/moonstone/SlotItem/SlotItem.module.less
Sangwook-Lee: If it resized the image from app side, it was not centered.

Sangwook-Lee: [https://jira2.lgsvl.com/secure/attachment/474561/%EC%84%A0%ED%83%9D%20%EC%98%81%EC%97%AD_009.png](url)

packages/ui/ToggleItem/ToggleItem.module.less
Sangwook-Lee: I agree with your opinion. I think I missed that point! So, I've removed this change.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -15,26 +15,49 @@ import kind from '@enact/core/kind';
...
 		return (
-			<ComponentOverride
-				component={iconComponent}
-				disabled={disabled}
-				selected={selected}
-			>
-				{icon}
-			</ComponentOverride>
+			<Fragment>
+				{itemIconPosition === 'before' && itemIcon && <ComponentOverride component={itemIcon} className={`${css.itemIcon} ${itemIcon.props.className}`} />}
```
Sangwook-Lee: @ryanjduffy , Can I get the other approaches for mixing styling of framework with application's `className` ?

packages/moonstone/SlotItem/SlotItem.module.less
Sangwook-Lee: @Djspaceg , @ryanjduffy , Could you please look at the above screenshot?
Is there any method?

packages/ui/ToggleItem/ToggleItem.js
ryanjduffy: It doesn't appear that you need to to because the `itemIcon` isn't defined anywhere. But, assuming that is an oversight, there isn't really another way other than reaching into its props. The challenge working with a React.Element (= the output of `<Component>` or `React.createElement(Component)`) is that it treated as a black box and only passed through other React methods (e.g. `cloneElement`). So while you can do something like the above now, it may not always work.

Other alternatives to solve the problem include:
* Document that `className` is passed to `itemIcon` and will override any specified value on the Element
* Only supporting `string|Object` for `itemIcon` which avoids the issue since the author has no means to specify a className
* Wrap the `itemIcon` with another node on which the class can be set
* Use a alternate attribute (e.g. `data-item-icon`) instead of a CSS class to use in the selector for styling

Each of these has its own trade offs which we can discuss.

Djspaceg: @Sangwook-Lee, how do you feel about consolidating this repeated code into a variable, then just refer to this variable in each of the four locations here and below?

Ex:
```
const itemIconComponent = (itemIcon ? <ComponentOverride component={itemIcon} className={`${css.itemIcon} ${itemIcon.props.className}`} /> : null);
...
{itemIconPosition === 'beforeText' && itemIconComponent}
{itemIconPosition === 'afterText' && itemIconComponent}
etc
```

Also, regarding the discussion from above, if `itemIcon` is a `kind()` component, shouldn't the classNames automatically merge in this case, rather than needing special handling? As in the following code example:
```
<ComponentOverride component={itemIcon} className={css.itemIcon} />
```
I've not tested in this regard, but this was my expectation based on what I recall in the past use cases. `iconComponent` of `SwitchItem.js` does this.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -15,26 +15,69 @@ import kind from '@enact/core/kind';
...
 
 import ComponentOverride from '../ComponentOverride';
 import ForwardRef from '../ForwardRef';
+import Slottable from '../Slottable';
 import Toggleable from '../Toggleable';
 import Touchable from '../Touchable';
 
 import componentCss from './ToggleItem.module.less';
 
+const ItemIcon = kind({
```
Sangwook-Lee: @Djspaceg , Is is correct you expected? As you said, we don't need to handle with specific method.

packages/ui/ToggleItem/ToggleItem.js
Sangwook-Lee: @ryanjduffy , @Djspaceg , Thank you for the review. I've updated PR to use `kind()`. I think this issue was resolved by @Djspaceg 's suggestion.

ryanjduffy: This would be a good candidate for a unit test to validate this assumption.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -15,26 +15,69 @@ import kind from '@enact/core/kind';
...
+	name: 'ItemIcon',
+	propTypes: {
+		/**
```
webOS101: This should not be a jsdoc comment

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -15,26 +15,69 @@ import kind from '@enact/core/kind';
...
+	},
+	render: ({itemIcon}) => {
+		return (
+			<ComponentOverride component={itemIcon} />
+		);
+	}
+});
+
 // eslint-disable-next-line
-const iconCreator = (position) => ({disabled, icon, iconComponent, selected, iconPosition}) => {
```
webOS101: Let's not disable eslint entirely!  Only disable the rule needed here.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -136,6 +179,22 @@ const ToggleItemBase = kind({
...
+		/**
+		 * The `itemIcon` to render in this item.
+		 *
+		 * @type {Node}
+		 * @public
+		 */
+		itemIcon: PropTypes.node,
+
+		/**
+		 * Specifies on which position (`'after'`, `'afterText`, `before` or `'beforeText'`) of the text the `itemIcon` appears.
```
webOS101: This doc needs to be update to be more in line with our other docs.

packages/ui/ToggleItem/ToggleItem.js
Djspaceg: Updated lint rule

Djspaceg:
I think we can't conveniently use Slottables with this, without some HOC composition refactoring. I'm going to remove it for now.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -136,6 +162,31 @@ const ToggleItemBase = kind({
...
+		 * @public
+		 */
+		itemIcon: PropTypes.node,
+
+		/**
+		 * Specifies on which position (`'before'`, `'beforeText'`, `'afterText'`, `'after'`) of the
+		 * text the `itemIcon` appears.
+		 *
+		 * * before - first element in the item
+		 * * beforeText - if a "before" `icon` is present, this will be between the icon and the text
```
webOS101: `ToggleItem` makes no restriction on its children so `'beforeText'` is a tad misleading.  Perhaps `'beforeChildren'`?

Djspaceg: ok.

packages/moonstone/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/SlotItem` to change `vertical-align` from `top` to `middle` to align the center
```
ryanjduffy: This change log entry is pretty implementation-specific. We don't normally call out CSS changes even to root components. Is it necessary to share this level of detail with developers?

Sangwook-Lee: Hmm.. I think so. I'll remove it.

packages/moonstone/SlotItem/SlotItem.module.less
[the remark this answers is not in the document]
Sangwook-Lee: Yes, I'll fix it.

packages/moonstone/Switch/Switch.module.less
```diff
@@ -15,6 +15,7 @@
 	display: inline-block;
 	text-align: left;
 	cursor: default;
+	vertical-align: middle;
```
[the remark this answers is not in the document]
Sangwook-Lee: @Djspaceg , What do you think about this?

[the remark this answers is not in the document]
Djspaceg: It was added to each element because we **don't want** to impose that visual rule at the UI level, affecting any theme using it. The alignment is decided by the theme; it's by no means a universal style. ToggleIcon does not have a stylesheet in Moonstone, so it was added to each consumer. It's mandatory for proper alignment regarding the styling of Moonstone.

packages/moonstone/SlotItem/SlotItem.module.less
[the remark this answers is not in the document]
Djspaceg: Please be exhaustive in testing when changing major fundamentals of the alignment of complex components. In nearly all cases the settings are in place for extremely specific reasons, all of which should be understood before changing them. This is not a change that should be taken lightly.

packages/moonstone/Switch/Switch.module.less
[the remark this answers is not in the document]
Sangwook-Lee: @ryanjduffy , Could you please check my patch and @Djspaceg 's opinion?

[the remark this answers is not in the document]
ryanjduffy: Is a compromise to add a stylesheet to `moonstone/ToggleIcon` to apply the rule at an appropriate layer rather than duplicating it in multiple components? That would better leverage our infrastructure and avoid forgetting it in future changes/additions.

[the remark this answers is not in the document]
ryanjduffy: Nevermind. Seems this was resolved otherwise.

packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -24,17 +24,42 @@ import Touchable from '../Touchable';
 
 import componentCss from './ToggleItem.module.less';
 
-// eslint-disable-next-line
-const iconCreator = (position) => ({disabled, icon, iconComponent, selected, iconPosition}) => {
-	if (iconPosition === position) {
+// eslint-disable-next-line enact/display-name,enact/prop-types
+const iconCreator = (position) => ({disabled, icon, iconComponent, iconPosition, itemIcon, itemIconPosition, selected}) => {
+
+	const itemIconComponent = (itemIcon ? <ComponentOverride component={itemIcon} /> : null);
```
ryanjduffy: I'm not sure this is required but wanted to check on the use cases. The docs show (and the samples above demonstrate) accepting a React Element (`<Icon>star</Icon>`). `ComponentOverride` accepts either a component (e.g. `Icon`) or an element so it "works" here but it's not passing any props down so it's only value is taking a component (which the docs don't say we support) and returning an element. However, without passing any props along, the icon/content would have to be built into the component which is the thing we wanted to avoid requiring by introducing `ComponentOverride` in the first place!

That might not make sense -- it's a somewhat complex arrangement -- but it simplifies down to:
* Do we need to support components as a value for `itemIcon`?

If not, then the above code can be reduce to just using `itemIcon` and we can skip `ComponentOverride` altogether. That would be my inclination given what I see right now.

Sangwook-Lee: @ryanjduffy ,
What did you mean `components as a value for itemIcon`? Did you ask if other components besides the Icon component could be used?
Up to now, all use cases only use icon. (Actually this feature has customized with `Image` by applications. However, once we support this component, it's possible to satisfy that requirements with `Icon`.)

ryanjduffy: I'm trying to distinguish between Components (the constructor of elements) and elements (the instances of a component that are rendered). I want to understand if there's a need to support both or just elements (as the docs currently state).

ryanjduffy: I've removed the `ComponentOverride` usage here based on my above observations. If we later decide it would be valuable to support the use case of a Component rather than an element here, we can add it back and update the docs to reflect that.

PR #2162, opened 2019-03-13
ENYO-5885: Fix that 'removeEventListener' was undefined in VirtualList and Scroller samples

ybsung:
### Checklist

* [X] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [X] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [ ] At least one test case is included for this feature or bug fix
* [ ] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)

When unmounting the page including VirtualList or Scroller, there is the following error.
- Cannot read property 'removeEventListener' of undefined

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)

I passed the wrong object and referenced its wrong property before.
So I fixed it and updated the guard condition. 

### Links
[//]: # (Related issues, references)
ENYO-5885

PR #2165, opened 2019-03-18
PLAT-75865: Fix to resume spotlight pauses when closing with animation

ybsung:
### Checklist

When creating the popup with the {open: true, noAnimation: false} props, Spotlight could not be on any controls when hovering them after closing the popup.

### Resolution

Analysis
- When creating the popup with the {open: true, noAnimation: false} props
\- The `popupOpen` status in the `Popup` was initialized with `OpenState.OPENING`
\- When `FloatingLayer` opened, `this.handleFloatingLayerOpen` in the popup was called, then the `popupOpen` status was reset to `OpenState.OPENING` again.
\- The `popupOpen` status was always updated to `OpenState.OPEN` when the popup animation ended and `this.handlePopupShow` callback in the popup was called.
\- But there was no chance to make the `popupOpen` status the `OpenState.OPEN` when creating with those props because the popup showed without animation and there was no animation end event.

- When closing the popup
\- The `popupOpen` had the wrong value which was `OpenState.OPENING`.
\- So `floatLayerOpen` had `true`, then the popup hides without animtion
\- At that time, `this.paused.resume();` was already called.
\- So spotlight could not resume.
\- It means that `this.paused.pause();` was called, but `this.paused.resume();` was not called. when closing the popup

Solution
- Initialize the `popupOpen` with `OpenState.OPEN` when creating the popup with the {open: true, noAnimation: false} props because the popup shows without animation
- Prevent converting the `popupOpen` variable to `OpenState.OPENING` when FloatingLayer updated and `this.handleFloatingLayerOpen` was called.

### Additional Considerations

Removed the unnecessary changelog commented in https://github.com/enactjs/enact/pull/2162#pullrequestreview-214034340. 

### Links

PLAT-75865

### DIFF
--- a/packages/moonstone/Popup/Popup.js
+++ b/packages/moonstone/Popup/Popup.js
@@ -243,8 +243,9 @@ const checkScrimNone = (props) => {
 
 const OpenState = {
 	CLOSED: 0,
-	OPENING: 1,
-	OPEN: 2
+	OPENINGWIHOUTANIMATION: 1,
+	OPENING: 2,
+	OPEN: 3
 };
 
 /**
@@ -389,7 +390,7 @@ class Popup extends React.Component {
 	constructor (props) {
 		super(props);
 		this.paused = new Pause('Popup');
-		const animateOpen = this.props.noAnimation ? OpenState.OPEN : OpenState.OPENING;
+		const animateOpen = this.props.noAnimation ? OpenState.OPEN : OpenState.OPENINGWIHOUTANIMATION;
 		this.state = {
 			floatLayerOpen: this.props.open,
 			popupOpen: this.props.open ? animateOpen : OpenState.CLOSED,
@@ -454,7 +455,7 @@ class Popup extends React.Component {
 	handleFloatingLayerOpen = () => {
 		if (!this.props.noAnimation) {
 			this.setState({
-				popupOpen: OpenState.OPENING
+				popupOpen: (this.state.popupOpen === OpenState.OPENINGWIHOUTANIMATION) ? OpenState.OPEN : OpenState.OPENING
 			});
 		} else if (this.state.popupOpen === OpenState.OPEN && this.props.open) {
 			this.spotPopupContent();
@@ -575,7 +576,7 @@ class Popup extends React.Component {
 					onCloseButtonClick={onClose}
 					onHide={this.handlePopupHide}
 					onShow={this.handlePopupShow}
-					open={this.state.popupOpen >= OpenState.OPENING}
+					open={this.state.popupOpen >= OpenState.OPENINGWIHOUTANIMATION}
 					spotlightId={this.state.containerId}
 					spotlightRestrict="self-only"
 				/>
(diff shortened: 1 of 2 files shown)


packages/moonstone/Popup/Popup.js
```diff
@@ -575,7 +576,7 @@ class Popup extends React.Component {
 					onCloseButtonClick={onClose}
 					onHide={this.handlePopupHide}
 					onShow={this.handlePopupShow}
-					open={this.state.popupOpen >= OpenState.OPENING}
+					open={this.state.popupOpen > OpenState.CLOSED}
```
ryanjduffy: This condition produces the same result so let's skip this change to keep it clear what was changed in this PR.

packages/moonstone/Popup/Popup.js
```diff
@@ -453,9 +452,11 @@ class Popup extends React.Component {
 
 	handleFloatingLayerOpen = () => {
 		if (!this.props.noAnimation) {
-			this.setState({
-				popupOpen: OpenState.OPENING
-			});
+			if (this.state.popupOpen !== OpenState.OPEN) {
```
ryanjduffy: Can you combine this condition and the one above? Would there be issues with doing so?

```
if (!this.props.noAnimation && this.state.popupOpen !== OpenState.OPEN) {
```

ybsung: No. The condition above looks better.

0x64: I agree with @ybsung since `spotPopupContent` will be called within `componentDidMount` then also will be called in below _else if_ clause when (!this.props.noAnimation && this.state.popupOpen === OpenState.OPEN && this.props.open) is true.

ryanjduffy: Won't that still be the case with the code as is? I'll need to step through it tomorrow to verify.

0x64: I tested code in this PR briefly with qa-sampler, and `spotPopupContent` is called twice due to calling in `handlePopupShow` (not `handleFloatingLayerOpen`) which is called back from `PopupBase`. Hmm...

packages/moonstone/Popup/Popup.js
ybsung: @0x64 could you fix it for me?

ryanjduffy:
@0x64 - I pushed a commit that I think addresses the concerns. I've verified it against QA sampler, my local test app, and our UI tests. Take a look and let me know what you think. If it checks out, we'll get this merged. Thanks for the help!

0x64:
@ryanjduffy Looks good! Let's move on. 👍 (BTW, redundant calling of `spotPopupContent` looks not so harmful, anyway.)

ryanjduffy:
I agree that it likely wouldn't cause any issues but I added a [fix](https://github.com/enactjs/enact/pull/2165/commits/e9d159cec9d56874f2e9ea342e2364b2ca058b12#diff-c794cfe5b95cf4676d04d28483b99d6eR425) that will address the redundant call at this in this reported scenario.

PR #2166, opened 2019-03-18
PLAT-75858: Adjust focus timing when ContextualPopup is closed with a 'back' button

Baekwoo-s:
### Checklist

A11y issue when closing the `ContextualPopup` by pressing the back button on the remote control
- The TV must read 'back' + 'activator button contents' but is reading the 'activator button contents' while reading 'back'. (Sometimes it reads 'B'(part of 'Back') 'activator button contents' button' (where the 'b' is stuttered))

### Resolution
- Modified to read activator button contents after 800 ms after reading 'back'

### Additional Considerations
Code to avoid malfunction
- Added activatorAriaHidden state to prevent the activator from being unnecessarily read.
- I only blur and focus the spotlight when the current focus is the activator.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList` to remove event listeners properly
+- `moonstone/ContextualPopupDecorator` timing to read out `activator` button when the `ContextualPopup` is closed by pressing a back button of remote control
 
 ## [2.4.1] - 2019-03-11
 
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -554,7 +554,6 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			if (activator && activator === Spotlight.getCurrent()) {
 				activator.blur();
 			}
-
 			if (!Spotlight.focus(activator)) {
 				Spotlight.focus();
 			}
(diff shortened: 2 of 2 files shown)


packages/moonstone/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList` to remove event listeners properly
+- `moonstone/ContextualPopupDecorator` timing to read out `activator` button when the `ContextualPopup` is closed by pressing a back button of remote control
```
0x64: No need to wrap 'activator' with backticks. It's just a name for a variable. I prefer 'back key' than 'back button'.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -555,6 +557,17 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
...
+			this.setState({
+				activatorAriaHidden: false
+			});
+
+			setTimeout(() => {
+				if (Spotlight.getCurrent() === activator) {
+					Spotlight.getCurrent().blur();
+					Spotlight.focus(activator);
+				}
+			}, 800);
```
0x64: Could you please leave the reason to choose the time of 800ms in PR description?

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -555,6 +557,17 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			if (!Spotlight.focus(activator)) {
 				Spotlight.focus();
 			}
+
+			this.setState({
+				activatorAriaHidden: false
+			});
+
+			setTimeout(() => {
+				if (Spotlight.getCurrent() === activator) {
```
0x64: We need guard for activator here.
> if (activator && activator === Spotlight.getCurrent()) {

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -555,6 +557,17 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
...
+					Spotlight.getCurrent().blur();
```
0x64: Guard here, too.

Baekwoo-s:
@0x64 
I have all modified.
Thanks.

0x64:
Note: this PR is not a solution for the root cause but a kind of workaround.

PR #2167, opened 2019-03-18
PLAT-74980: Update design of moonstone/Button

Djspaceg:
### Checklist

Buttons are now squared, IconButtons still round.

**Also, there are comments here as a reminder that should be removed before final merging.**
They persist in the commit as migration cues for a continually evolving design.


packages/moonstone/styles/variables.less
```diff
@@ -148,23 +148,36 @@
 @moon-button-height-large: 96px;
 @moon-button-small-height: 60px;
 @moon-button-small-height-large: 72px;
-@moon-button-border-width: 6px;
-@moon-button-border-radius: 9999px;
+@moon-button-border-width: 3px;
+@moon-button-border-radius: 3px;
+// @moon-button-border-radius: 960px;
```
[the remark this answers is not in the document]
Djspaceg: It should be removed. I realized during construction that 999 is a dumb number to use to round corners, while 960 is nice because it translates to an even 40rem. I had written a comment as such, but I removed it at some point. This can just be deleted as cruft.

packages/ui/Button/Button.js
```diff
@@ -154,11 +154,12 @@ const ButtonBase = kind({
 	},
 
 	computed: {
-		className: ({minWidth, pressed, selected, small, styler}) => styler.append({
-			pressed,
-			small,
+		className: ({icon, minWidth, pressed, selected, small, styler}) => styler.append({
+			hasIcon: (!!icon),
```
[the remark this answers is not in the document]
Djspaceg: It was unintentionally omitted. I forgot about the docs. I had added it to the exported classes but forgot above. I'll add it.

packages/moonstone/Button/Button.js
```diff
@@ -137,7 +137,6 @@ const ButtonBase = kind({
  */
 const ButtonDecorator = compose(
 	Pure,
-	Uppercase,
```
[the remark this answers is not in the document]
Djspaceg: We've not been given explicit direction in this regard; simply: "Buttons are no longer all uppercase."

Djspaceg:
To address the bullet points you raised:
* Oversight. I'll fix this.
* There is a 3px translucent border, due to the inner-stroke of the design. 12+3=15.
* I'll update Button for this and un-update IconButton, since now I think they have different rules. We were not given direction on the transparent IconButton disabled background color. (Scroller arrows for example.)

Djspaceg:
I believe I addressed everything with the exception of the broken tests/samples (due to Uppercase removal) since that's still sort of under consideration. However, will address it one way or another before this is merged.

packages/moonstone/Button/Button.js
```diff
@@ -86,7 +86,7 @@ const ButtonBase = kind({
 
 	styles: {
 		css: componentCss,
-		publicClassNames: ['button', 'bg', 'selected', 'small']
+		publicClassNames: ['button', 'bg', 'selected', 'small', 'transparent']
```
ryanjduffy: Should we add all the bg opacity options to this list? Between this and the color change, we're starting to "onesie, twosie" class names onto `publicClassNames` and want to be sure we're making intentional choices. Either way, we'll need to update the `css` docs as well.

Djspaceg: `transparent` was intentionally excluded from the docs. The reasoning being that an app developer shouldn't need to dynamically alter the rules of the IconButton they're using. They can simply assign props and CSS rules as desired. For example, if an app wanted to alter the rules of their transparent button, they'd simply assign `backgroundOpacity="transparent"` and then add rules to their Button's `className`. If they wanted to change to non-transparent, they'd just remove that prop. It's not technically a useful feature for an app.

The framework, on the other hand, needs to be able to conditionally change styling rules without altering props. This is the most direct way to have IconButton respond to the presence of the `transparent` class. At this time we do not offer a `privateClassNames` option.

ryanjduffy: Agree that it doesn't add value but I think it reveals a potential architectural problem when we're using features intended to expose public API but intentionally hiding them. It's not the first time we've done similar things (e.g. private props on base components) but I want to think a bit more on the implications of it.

Djspaceg: Fair point. Thanks for bringing it up. Much better to discuss now than when it's too late.

ryanjduffy: Leaving this as is for now.

packages/moonstone/Button/tests/Button-specs.js
```diff
@@ -4,19 +4,6 @@ import Button from '../Button';
 
 describe('Button', () => {
 
-	test('should render with button text upper-cased', () => {
```
webOS101: We'll need to update all the UI and screenshot test cases that rely on the presence of the `Uppercase` HOC.

packages/sampler/stories/default/Button.js
```diff
@@ -29,7 +29,6 @@ storiesOf('Moonstone', module)
 			<Button
 				onClick={action('onClick')}
 				backgroundOpacity={select('backgroundOpacity', prop.backgroundOpacity, Config)}
-				casing={select('casing', prop.casing, Config, 'upper')}
```
webOS101: Casing still in prop above, will remove.

packages/sampler/stories/default/ToggleButton.js
```diff
@@ -25,7 +25,6 @@ storiesOf('Moonstone', module)
 			<ToggleButton
 				aria-label="toggle button"
 				backgroundOpacity={select('backgroundOpacity', prop.backgroundOpacity, Config)}
-				casing={select('casing', prop.casing, Config, 'upper')}
```
webOS101: ditto.

PR #2168, opened 2019-03-19
ENYO-5880: Better layout of overflowing Tooltip

teckliew:
### Checklist

Better support for tooltips that are too big or that still overflows after adjusting for overflow.
Add qa-sample for `Tooltip` overflows

### Resolution
Starting with `calcOverflow`, we calculate the to see if the tooltip width is:
1. bigger than the viewport or
2. bigger than the space on the left side of the `clientNode` and the space on the right side of the `clientNode`

Once we determine that the tooltip will overflow no matter how we adjust the position, then we position the tooltip so that the tooltip can be read from the beginning and it'll take up the whole width of the viewport, and adjust the position of arrow anchor.

### Additional Considerations
Open to suggestions on naming `isOverWide`; something that suggests, the tooltip will overflow no matter where we put it.

Also would like to hear suggestions on how to customize the tooltip arrow anchor without adding a new prop. 

### Links
ENYO-5880

### DIFF
--- a/packages/moonstone/TooltipDecorator/Tooltip.js
+++ b/packages/moonstone/TooltipDecorator/Tooltip.js
@@ -44,6 +44,17 @@ const TooltipBase = kind({
 		arrowAnchor: PropTypes.oneOf(['left', 'center', 'right', 'top', 'middle', 'bottom']),
 
 		/**
+		 * Style object for tooltip arrow position.
+		 *
+		 * @type {Object}
+		 * @public
+		 */
+		arrowPosition: PropTypes.shape({
+			left: PropTypes.number,
+			right: PropTypes.number
+		}),
+
+		/**
 		 * Direction of label in relation to the activator.
 		 *
 		 * * Values: `'above'`, `'below'`, `'left'`, and `'right'`
@@ -109,14 +120,14 @@ const TooltipBase = kind({
 		}
 	},
 
-	render: ({children, tooltipRef, arrowType, width, ...rest}) => {
+	render: ({children, tooltipRef, arrowType, width, arrowPosition, ...rest}) => {
 		delete rest.arrowAnchor;
 		delete rest.direction;
 		delete rest.position;
 
 		return (
 			<div {...rest}>
-				<svg className={css.tooltipArrow} viewBox="0 0 3 5">
+				<svg style={arrowPosition} className={css.tooltipArrow} viewBox="0 0 3 5">
 					<path d={arrowType} />
 				</svg>
 				<TooltipLabel tooltipRef={tooltipRef} width={width}>
--- a/packages/moonstone/TooltipDecorator/util.js
+++ b/packages/moonstone/TooltipDecorator/util.js
@@ -26,6 +26,10 @@ const adjustAnchor = function (arrowAnchor, tooltipDirection, overflow, rtl) {
 		} else if (overflow.isOverLeft) {
 			arrowAnchor = 'right';
 		}
+
+		if (overflow.isOverWide) {
+			arrowAnchor = 'center';
+		}
 	}
 
 	return arrowAnchor;
@@ -53,9 +57,9 @@ const adjustDirection = function (tooltipDirection, overflow, rtl) {
 		tooltipDirection = 'below';
 	} else if (overflow.isOverBottom && tooltipDirection === 'below') {
 		tooltipDirection = 'above';
-	} else if (overflow.isOverLeft && tooltipDirection === 'left') {
+	} else if (overflow.isOverLeft && tooltipDirection === 'left' && !overflow.isOverWide) {
 		tooltipDirection = 'right';
-	} else if (overflow.isOverRight && tooltipDirection === 'right') {
+	} else if (overflow.isOverRight && tooltipDirection === 'right' && !overflow.isOverWide) {
 		tooltipDirection = 'left';
 	}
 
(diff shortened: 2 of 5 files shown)


packages/moonstone/TooltipDecorator/util.js
```diff
@@ -75,19 +79,26 @@ const adjustDirection = function (tooltipDirection, overflow, rtl) {
...
+	const isTooltipWide = (tooltipNode.width > window.innerWidth) ||
+		(leftDelta && rightDelta);
+
 	if (tooltipDirection === 'above' || tooltipDirection === 'below') {
 		return {
 			isOverTop: clientNode.top - tooltipNode.height - tooltipHeight < 0,
 			isOverBottom: clientNode.bottom + tooltipNode.height + tooltipHeight > window.innerHeight,
 			isOverLeft: clientNode.left - tooltipNode.width + clientNode.width / 2 < 0,
-			isOverRight: clientNode.right + tooltipNode.width - clientNode.width / 2 > window.innerWidth
+			isOverRight: clientNode.right + tooltipNode.width - clientNode.width / 2 > window.innerWidth,
```
Djspaceg: I've never really understood why these lines use this math on several parts when a check to see if something overflows the right would be `tooltipNode.getBoundingClientRect().right > window.innerWidth` or more directly in this context, since we already have the `getBoundingClientRect` object: `tooltipNode.right > window.innerWidth`. Perhaps it's because the tooltip isn't in that position yet, so `right` isn't accurate at the time of this execution.

packages/moonstone/TooltipDecorator/util.js
```diff
@@ -144,12 +157,37 @@ const getPosition = function (tooltipNode, clientNode, arrowAnchor, tooltipDirec
 		}
 	}
 
+	// When tooltip is too wide, shift the tooltip so that the first part of the tooltip is always visible. Does not affect tooltips with `tooltipDirection` of `left` and `right`
+	if (overflow.isOverWide && !(tooltipDirection === 'left' || tooltipDirection === 'right') && arrowAnchor !== 'right') {
+		position.left = rtl ? window.innerWidth - tooltipNode.width : 12;
```
Djspaceg: This `12` should probably be REM converted and documented as to what the 12 represents. Spotlight-outset? TV Keepout?

packages/moonstone/TooltipDecorator/util.js
```diff
@@ -144,12 +157,37 @@ const getPosition = function (tooltipNode, clientNode, arrowAnchor, tooltipDirec
...
 	return position;
 };
 
+/**
+ * Adjusts the `Tooltip` arrow anchor when the tooltip is too wide.
+ * Takes the output of `calcOverflow`.
+ *
+ * @method
+ * @memberof moonstone/TooltipDecorator
+ * @param   {Object}  clientNode		The `getBoundingClientRect` values for client node
```
Djspaceg: It seems nitpicky, but we should probably use spaces to align these since the variable width of tabs can mess up the alignment if it's not consistently set everywhere. There was another section above that did the same thing.

packages/moonstone/TooltipDecorator/Tooltip.js
```diff
@@ -109,14 +120,14 @@ const TooltipBase = kind({
...
-	render: ({children, tooltipRef, arrowType, width, ...rest}) => {
+	render: ({children, tooltipRef, arrowType, width, arrowPosition, ...rest}) => {
 		delete rest.arrowAnchor;
 		delete rest.direction;
 		delete rest.position;
 
 		return (
 			<div {...rest}>
-				<svg className={css.tooltipArrow} viewBox="0 0 3 5">
+				<svg style={arrowPosition} className={css.tooltipArrow} viewBox="0 0 3 5">
```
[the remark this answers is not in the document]
teckliew: I updated this where there `Tooltip` doesn't just pass straight through to `style`. Now it takes 0 and 1 for `arrowPosition` and it is converted to `left` percentage.

packages/moonstone/TooltipDecorator/util.js
teckliew: It is because we want to start the tooltip positioning in the middle of the client node. Which is why we need to add the extra calculation with `clientNode.width / 2` and account for that positioning difference.

packages/moonstone/TooltipDecorator/util.js
```diff
@@ -180,7 +181,17 @@ const getPosition = function (tooltipNode, clientNode, arrowAnchor, tooltipDirec
  */
 const getArrowPosition = function (tooltipNode, clientNode, overflow, rtl) {
 	if (overflow.isOverWide) {
-		return rtl ? 1 - ((window.innerWidth - clientNode.right + (clientNode.width / 2)) / tooltipNode.width) : (clientNode.left + (clientNode.width / 2)) / tooltipNode.width;
+		const tooltipWidth = tooltipNode.width;
+		// finding out where the middle of the clientNode is and figuring out where that is in relation to the tooltip node in percentage between 0 and 1
+		const arrowPosition = rtl ? 1 - ((window.innerWidth - clientNode.right + (clientNode.width / 2)) / tooltipWidth) : (clientNode.left + (clientNode.width / 2)) / tooltipWidth;
+
+		const borderRadius = 33; // based on @moon-tooltip-border-radius
```
Djspaceg: This value needs to be RI converted so it represents the correct amount of pixels on 720p and 4k: `ri.scale(33)`

Djspaceg:
Regarding the sample, that's a good point. Perhaps we leave the new sample alone and add a new one which doesn't have (much?) interactivity but displays all tooltip orientations and positions at the same time, for quick spot-checking of the tooltip's construction. The tooltip placement and positioning could be essentially static since that would be covered by the sample you've created already. The tooltip text and `IconButtons` could be rather small too, like just one or two words ("top left", "center bottom", etc) so there would be little chance for tooltip overlap.

packages/moonstone/TooltipDecorator/util.js
webOS101: FWIW, we use tabs a good bit in the docs.

PR #2169, opened 2019-03-19
ENYO-5890: Fix to move Spotlight between scroll buttons properly via 5-way keys

ybsung:
### Checklist

When pressing the 5-way Up on the Down scroll button, Spotlight jumps not to the Up scroll button but to the item in the VirtuaList.

### Resolution

We have to compare the current spotted DOM element and the Up / Down scroll button DOM elements to navigate between the Up / Down scroll buttons. But we compared the current spotted DOM element and the scroll button ref (Not DOM element). So I used `React.findDONNode() to find the scroll button DOM element.

### Links
ENYO-5890

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,6 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList` to move Spotlight between paging controls properly via 5-way keys
 - `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList` to remove event listeners properly
 
 ## [2.4.1] - 2019-03-11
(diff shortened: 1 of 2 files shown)


ryanjduffy:
Thanks for the fix, @0x64! The proposed solution seem to work as expected and I think you're right to rely on `findDOMNode` in this case. I wanted to share an alternative that might be more in line with the original implementation from #2151.

The issue as you discovered is comparing a DOM node with a React component instance. The reason `forwardRef` wasn't mitigating that was two-fold. First, the `ref` was mapped back to the `ref` prop on `ScrollButton` effectively nullifying `forwardRef`. Instead, we need to reassign the provided `ref` arg to a different prop that can be mapped to a native DOM node downstream. Unfortunately, the second gap is that we lack the ability to do that in `IconButton`.

My alternative attempts to insulate `ScrollButtons` from that gap by using `findDOMNode` in `ScrollButton` to return the native DOM node allowing the existing logic to work as designed.

https://github.com/enactjs/enact/compare/develop...feature/ENYO-5890-ryan?expand=1

If you agree with this approach, you can cherry pick it into your branch here and we'll move this ahead. If you have questions/concerns, please share them!

0x64:
@ryanjduffy Thank you for suggesting better way. 👍 `fowardRef` is always an object not a function in our scrollbar but I like that pattern. Hope prop types for refs will be added.

BTW, I think we'd better to make another PR from your code since the new approach is not based on this PR and no commit is shared between two branches.

ybsung:
Closing in favor of https://github.com/enactjs/enact/pull/2174

PR #2172, opened 2019-03-21
PLAT-75887: VirtualList Perf - Manage focus within VirtualList

germboy:
### Checklist

When focus changes between items within a `VirtualList`, the spotlight and/or keydown-driven events take too long - which add to poor performance when scrolling. This is especially the case when holding 5-way to navigate - where focus changes accelerate and occur rapidly.

### Resolution
When using spotlight to direct 5-way navigation, there is some inherent overhead that comes along with it. Spotlight must curate the available spottable controls and calculate where to set focus next. In the case of `VirtualList`, we have a contained/known amount of components, so we can specify where to set focus - saving some time. Ultimately, we can shave 1/3 of the time off a single event.

There are some other micro-optimizations that have been made in the event-handling logic to help speed up those rapidly-occurring events during a hold.

Chrome 6x slow
`keydown` (single event, not scrolling, not animating): 25ms > 17ms
`keydown` (holding, scrolling, not animating): 64ms > 44ms

I added a convenience variable in the `onAcceleratedKeyDown` where we can turn on animations while scrolling (they had been turned off previously due to performance concerns). This should result in better animation/scrolling compared to the original scrolling implementation.

Handling focus within `VirtualList` also gives us the benefit of being able to sync our `keydown` events with focus changes, rather than letting spotlight direct focus via its accelerator then responding to that change in `VirtualList` via the `keydown` event.

### Additional Considerations
I'm not as familiar with `VirtualList` as others may be, so there is a chance I have not accounted for all scenarios in my focus-changing logic. I'd appreciate any feedback or thoughts on this implementation.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -5,16 +5,19 @@ import React, {Component} from 'react';
 import Spotlight, {getDirection} from '@enact/spotlight';
 import Pause from '@enact/spotlight/Pause';
 import Spottable from '@enact/spotlight/Spottable';
+import Accelerator from '@enact/spotlight/Accelerator';
 import {VirtualListBase as UiVirtualListBase, VirtualListBaseNative as UiVirtualListBaseNative} from '@enact/ui/VirtualList';
 
 import {Scrollable, dataIndexAttribute} from '../Scrollable';
 import ScrollableNative from '../Scrollable/ScrollableNative';
 
+const SpotlightAccelerator = new Accelerator();
 const SpotlightPlaceholder = Spottable('div');
 
 const
 	dataContainerDisabledAttribute = 'data-spotlight-container-disabled',
 	isDown = is('down'),
+	isEnter = is('enter'),
 	isLeft = is('left'),
 	isRight = is('right'),
 	isUp = is('up'),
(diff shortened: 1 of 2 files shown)


germboy:
I've refactored the `nextIndex` logic out of `onAcceleratedKeyDown` and was able to clean up some of the unclear bits, so inline docs may not be needed as much now. I also went ahead and removed the old methods that this PR is moving away from (`jumpToSpottableItem`, `setSpotlightContainerRestrict `, `setRestrict`). I don't believe these are being used in any way now, so it should be safe to remove them.

seunghoh:
I've found one issue. Spotlight doesn't work on pointer hover after pressing page up/down. I couldn't found the cause yet

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -695,7 +651,12 @@ const VirtualListBaseFactory = (type) => {
 		focusOnItem = (index) => {
 			const item = this.uiRefCurrent.containerRef.current.querySelector(`[data-index='${index}'].spottable`);
 
-			this.pause.resume();
+			if (this.isWrappedBy5way) {
+				SpotlightAccelerator.reset();
+				this.isWrappedBy5way = false;
+				this.pause.resume();
```
seunghoh: `this.pause.resume()`  should go out of `if` statement. This cause spot lost issue on hover

germboy: Good catch - I've updated this so that spotlight resumes properly.

PR #2173, opened 2019-03-21
ENYO-5893: Fix ContextualPopupDecorator to allow positioning from app code

xbinary:

### Checklist

If method that can be accessed using `ApiDecorator` is not arrow function, this is bound to the object where the method is called.

For example, `positionContextualPopup` of `ContextualPopupDecorator` is not arrow function and will cause problem

### Resolution
`positionContextualPopup` is changed to arrow function.

### Links
ENYO-5893

### DIFF
--- a/packages/core/internal/ApiDecorator/ApiDecorator.js
+++ b/packages/core/internal/ApiDecorator/ApiDecorator.js
@@ -12,7 +12,9 @@ import hoc from '../../hoc';
 
 // Gets a property from `provider`
 const get = (provider, name) => () => {
-	return provider[name];
+	if (provider) {
+		return provider[name];
+	}
 };
 
 // Sets a property on `provider`
--- a/packages/core/internal/ApiDecorator/ApiDecorator.js
+++ b/packages/core/internal/ApiDecorator/ApiDecorator.js
@@ -12,7 +12,9 @@ import hoc from '../../hoc';
 
 // Gets a property from `provider`
 const get = (provider, name) => () => {
-	if (provider) {
+	if (typeof provider[name] === 'function') {
+		return provider[name].bind(provider);
+	} else {
 		return provider[name];
 	}
 };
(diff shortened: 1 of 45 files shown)


packages/core/internal/ApiDecorator/ApiDecorator.js
```diff
@@ -12,7 +12,9 @@ import hoc from '../../hoc';
 
 // Gets a property from `provider`
 const get = (provider, name) => () => {
-	if (provider) {
+	if (typeof provider[name] === 'function') {
```
webOS101: If we do want to go this route, I would like to see a test that validates this is working.  Perhaps a better check would be to see if the function is bound or not (can check to see if it has an own property of `prototype`, I think).

webOS101:
@ryanjduffy and I were discussing whether or not this is the right solution to the problem.  I think the subtlety of the class of bugs that can be introduced this way does warrant a solution like this.  It improves the developer experience with little to no downside.

The various scenarios are:
1) The passed function is already bound.   Result: no change
2) The passed function is not bound but does not rely on `this`.  Result: no change
3) The passed function is not bound but does rely on `this`.  Result: works after fix, broken without

ryanjduffy:
I'll call out the rare case that the provider intends to return an unbound function (I'm not sure why ...) which then can't be bound because it was already bound by `ApiDecorator`.

webOS101:
An intentionally unbound method on the prototype?

ryanjduffy:
ApiDecorator doesn't prescribe what is being passed back. Currently, it's a reference to the underlying component instance but could be anything. I'm not implying this is a common case only that ApiDecorator was designed to be intentionally unaware of underlying implementation details.

webOS101:
If we did decide not to take this, it would be worth pointing out the subtle error that developers could fall into if they do not bind their functions and intend to be able to use `this`.

webOS101:
I suppose my other argument is the 99% use case is to be able to access a property or function on an inner instance.  If you really wanted to be able to do something different, as a developer, you could just choose not to use `ApiDecorator` to expose it.

ryanjduffy:
Also point out that `ApiDecorator` is not public API so we're only really protecting ourselves :)

webOS101:
Who better? :)

PR #2174, opened 2019-03-26
ENYO-5890: Fix to move Spotlight between scroll buttons properly via 5-way keys

ybsung:
### Checklist

When pressing the 5-way Up on the Down scroll button, Spotlight jumps not to the Up scroll button but to the item in the VirtuaList.

### Resolution

We have to compare the current spotted DOM element and the Up / Down scroll button DOM elements to navigate between the Up / Down scroll buttons. But we compared the current spotted DOM element and the scroll button ref (Not DOM element). So Ryan fixed to pass the ref to the DOM element properly.

### Links

ENYO-5890


ybsung:
LGTM. It works with VirtualList and Scroller samples.

PR #2176, opened 2019-03-27
ENYO-5897: Viewport: it tries to make panel when child is null

Sangwook-Lee:
### Checklist

This issue is related to ENYO-5630. So, I can reproduce it since `2.4.0`.
In viewport, it tries to make Panel when child is null.

If application defines child(panel) depend on any condition or state, Panel child array have a null value.
Viewport generate `spotlightId` using with information from props. But, it shouldn't get props when child is null.

### Resolution
I added condition whether child is valid or not before generating spotlightId of Panel.

### Links
ENYO-5897

### DIFF
--- a/packages/moonstone/Panels/Viewport.js
+++ b/packages/moonstone/Panels/Viewport.js
@@ -128,17 +128,21 @@ const ViewportBase = class extends React.Component {
 	)
 
 	mapChildren = (children, generateId) => React.Children.map(children, (child, index) => {
-		const {spotlightId = generateId(index, 'panel-container', Spotlight.remove)} = child.props;
-		const props = {
-			spotlightId,
-			'data-index': index
-		};
+		if (child) {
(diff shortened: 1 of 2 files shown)


packages/moonstone/Panels/Viewport.js
```diff
@@ -128,17 +128,21 @@ const ViewportBase = class extends React.Component {
...
+				'data-index': index
+			};
+
+			if (child.props.autoFocus == null && this.state.direction === 'forward') {
+				props.autoFocus = 'default-element';
+			}
+
+			return React.cloneElement(child, props);
+		} else {
+			return null;
```
webOS101: I wonder if other `falsy` values other than `null` would give different results.  Empty string?  The number 0?  This seems like a potential (albeit very, very small) oversight.  Maybe we should return `child`?  Leaving this here as food for thought.

PR #2178, opened 2019-03-28
PLAT-73234: Update design of moonstone/Input

teckliew:
### Checklist

Input update

### Additional Considerations

### Links
PLAT-73234

### DIFF
--- a/packages/moonstone/ExpandableInput/ExpandableInput.module.less
+++ b/packages/moonstone/ExpandableInput/ExpandableInput.module.less
@@ -5,6 +5,6 @@
 .expandableInput {
 	.decorator {
 		display: flex;
-		margin: @moon-spotlight-outset;
+		margin: @moon-expandableinput-input-margin;
 	}
 }
--- a/packages/moonstone/ExpandableInput/ExpandableInput.module.less
+++ b/packages/moonstone/ExpandableInput/ExpandableInput.module.less
@@ -1,9 +1,10 @@
 // ExpandableInput.module.less
 //
+@import "../styles/variables.less";
 
 .expandableInput {
 	.decorator {
 		display: flex;
-		margin: 6px;
+		margin: @moon-spotlight-outset;
 	}
 }
--- a/packages/moonstone/Input/Input.module.less
+++ b/packages/moonstone/Input/Input.module.less
@@ -118,24 +118,13 @@
 	// Skin colors
 	.applySkins({
 		background-color: @moon-input-decorator-bg-color;
+		caret-color: @moon-input-caret-color;
 
 		.input {
 			background: transparent;
-			color: @moon-input-text-color;
-			caret-color: @moon-input-caret-color;
-
-			&::selection {
-				color: @moon-spotlight-text-color;
-				background-color: @moon-spotlight-bg-color;
-			}
 
 			.input-placeholder({
 				color: @moon-input-placeholder-color;
-			});
-
-			.focus({
-				color: @moon-input-text-color;
-				background-color: inherit;
 			});
 
 			.focused& {
@@ -143,10 +132,6 @@
 					color: @moon-input-placeholder-active-color;
 				});
 			}
-
-			.disabled({
-				color: @moon-input-text-disabled-color;
-			});
 		}
 
 		&.invalid input {
@@ -171,12 +156,12 @@
 
 		.disabled({
 			.vendor-opacity(@moon-input-disabled-opacity);
-		});
 
-		.decoratorIcon,
-		&:global(.spottable):focus .decoratorIcon {
-			color: @moon-input-text-color;
-		}
+			.input,
+			.decoratorIcon {
+				color: @moon-input-text-disabled-color;
+			}
+		});
 	});
 }
 
--- a/packages/moonstone/Input/Input.module.less
+++ b/packages/moonstone/Input/Input.module.less
@@ -56,7 +56,7 @@
 	position: relative;
 	padding: @moon-input-padding;
 	border-radius: @moon-input-border-radius;
-	margin: 0 (@moon-spotlight-outset / 2);
+	margin: @moon-input-margin;
 	box-sizing: border-box;
 	vertical-align: middle;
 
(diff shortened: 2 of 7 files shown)


packages/moonstone/Input/Input.module.less
```diff
@@ -136,23 +137,41 @@
...
+			});
 		}
 
 		&.invalid input {
 			color: @moon-red;
 		}
 
-		&:global(.spottable):focus {
-			color: @moon-input-text-color;
-			border-color: @moon-input-border-focus-border-color;
```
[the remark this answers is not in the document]
teckliew: We've replaced border with box-shadow. Unless we want to keep border and have it transparent?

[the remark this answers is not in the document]
Djspaceg: That's right! I'd forgotten about that change. How about the `color` rule?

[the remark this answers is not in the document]
teckliew: According to the new specs, color doesn't change with the exception of `disabled`; also `color` is handled in the `.input` block.
I was wondering if we should take it to the upper scope and have the rule in `.decorator`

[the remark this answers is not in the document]
Djspaceg: re: color: gotcha! Thanks for explaining.
My personal perspective is that everything that _could_ be moved to a higher/upper scope _should_ be moved there. Reasons being that it's easier to override; propagates rules down to children automatically, which reduces direct assignments; minimizes selector length (and overridability), and consolidates rules into more easily readable chunks in the files. This is the reason that some components use techniques like setting a color on the base DOM node and then children elements use `color: inherit` when that color must be assigned on that node directly. Button is one of those.

packages/moonstone/Input/Input.module.less
```diff
@@ -69,8 +73,6 @@
 	}
 
 	&.small {
-		padding: 0 (@moon-input-small-height / 2);
```
Djspaceg: Small should probably have its own padding rather than inheriting using large's padding. What do you think?

packages/moonstone/Input/Input.module.less
```diff
@@ -53,13 +54,16 @@
 .decorator {
 	display: inline-flex;
 	position: relative;
-	padding: @moon-input-padding (@moon-input-height / 2);
-	border-radius: 1008px;
+	padding: @moon-input-padding;
+	border-radius: @moon-input-border-radius;
 	margin: 0 (@moon-spotlight-outset / 2);
```
Djspaceg: I think after almost 6 years, it's time to update this margin to match our standard margin. Let's finally convert this to simply use `@moon-spotlight-outset` rather than half of it. Better still would be to make it into a variable: `@moon-input-h-margin: @moon-spotlight-outset`.
With the margin the way it is (and has been), it means when the Input is against the edge of the panel, or vertically next to buttons or other components, it doesn't horizontally line-up.

This 6px margin (originally 5px, and before that 0px) was added here: https://github.com/enyojs/moonstone/commit/3513a05601c400f1802851a0aab6f373b3534e86#diff-2010cd3d397ac502f284f6d890966042R4

packages/moonstone/Input/Input.module.less
teckliew: I don't think there's specs for padding differences between normal and small. I think the normal padding still works for small, unless you have another padding in mind.

Djspaceg: Nah, I'm ok with that. I love consistency. :)

ryanjduffy:
I noticed a couple "major" items that I wanted to discuss first. Happy to discuss these live and summarize back here since the comment back-and-forth won't be very efficient.

The "normal" size input seems rather large and looks out of place with nearly everything else since its font size is scaled to its height and therefore larger than body/item text. I'm not using the new fonts nor have I viewed this against the other changes but if the sizing remains comparable, I wonder if we should remove the `small` property altogether and only have a small input.

EditableIntegerPicker mimics the old Input style since it's a joined picker. Wondering if we should suggest updating joined pickers to use the squared shape of input and buttons which would apply to ExpandableIntegerPicker as well.

Also noticed a few other things that may or may not be issues.

* The Icons are not evenly spaced between the text and the edge
image
* The left margin compared to ExpandableInput's input do not match so they don't line up vertically
image
* ExpandableInput using the normal input and should probably use the small version
* Placeholder color in high-contrast seems much harder to distinguish with the darker background color
* Selection color in dark high-contrast is barely visible
image
* Focus ring in dark high-contrast may need emphasized more
image

ryanjduffy:
> The "normal" size input seems rather large and looks out of place with nearly everything else since its font size is scaled to its height and therefore larger than body/item text. I'm not using the new fonts nor have I viewed this against the other changes but if the sizing remains comparable, I wonder if we should remove the small property altogether and only have a small input.

Decisions:
* Adjust the font size down of the larger input to be more in line with other content (specific size TBD)
* Replace `small` with a `size` prop, defaulting to `'normal'` with a `'large'` option. Suggest we move this into a separate PR since we'll likely want to adopt this change more broadly.

teckliew:
Regarding the uneven space, it's the result of of text indent from https://github.com/enactjs/enact/pull/1475 to support glyphs that are wider than their bounding box. Also deferring the issue with high contrast colors and `ExpandableIntegerPicker`.

packages/moonstone/ExpandableInput/ExpandableInput.module.less
```diff
@@ -1,9 +1,10 @@
 // ExpandableInput.module.less
 //
+@import "../styles/variables.less";
 
 .expandableInput {
 	.decorator {
 		display: flex;
-		margin: 6px;
+		margin: @moon-spotlight-outset;
```
Djspaceg: Let's set this to be an input variable: `@moon-expandableinput-input-margin`.
And define that as the following in variables.less:
```
@moon-expandableinput-input-margin: 6px @moon-spotlight-outset;
```
This way, we maintain the existing vertical sizing, and we update to the standard horizontal spacing, while using only component-variables in each of the component modules files.

PR #2179, opened 2019-03-28
PLAT-75397 & PLAT-75394 && PLAT-75395: Fixed to blur on handlePointerUp for ExpandableInput.

haileyR:
### Checklist

ExpandableInput does not close when touching the second ExpandableInput.
This solution also resolves the PLAT-75394 issue. `ExpandableInput`'s `Input` field did not always collapse, but with this fix, it correctly collapses.

### Resolution
 It is fixed by adding blur for handlePointerUp(). Adding blur() on handlePointerUp() ended up opening all the inputs.


ryanjduffy:
@haileyR Can you update the title to the usual format (TICKET: Fix XXX to be less broken) and add to the description to share what the problem was and how the proposed solution resolves it? Thanks!

PR #2182, opened 2019-03-29
PLAT-75350: Colored Button UI update

teckliew:
### Checklist

Remote button ui update

### Links
PLAT-75350


ryanjduffy:
Couple more notes while you're updating, @teckliew:

* The dot is squished when rendered when pushing up against the `max-width` either due to long text or "medium" text + and icon.
* The dot is also squished and left-aligned in RTL locales.
* Should the measurements for the dot be externalized into LESS variables?
* Curious if you explored any alternatives that didn't require us to expose the `client` CSS class in Button. That begins to expose some of the internal structure of the component and could limit us in the future. If we do keep it, we should document in on the `css` prop for Button along with the others.

packages/moonstone/Button/Button.module.less
```diff
@@ -67,26 +67,41 @@
...
-	.client {
-		padding: @moon-button-border-width;	// We match the amount removed by the line-height above, so the text doesn't overlap the border
+	.clientContainer {
+		padding: 0 @moon-button-h-padding;
 
-		// Colored under-bar
+		// Colored dot for remote buttons
 		&::before {
-			content: "";
+			content: "O";
```
sugardave: I'm not sure if it's because I don't have fonts installed correctly, but I see differences in the width of the dot depending on what locale I'm setting.  With en-US, it isn't round, but with something like ar-SA it is.

Also not sure if it is a problem, but I wanted to get comments going.

Djspaceg: I'm skeptical that setting this rule is the direction we should be taking with this.

packages/ui/Button/Button.js
```diff
@@ -177,7 +177,9 @@ const ButtonBase = kind({
 		return (
 			<div role="button" {...rest} aria-disabled={disabled} disabled={disabled}>
 				<div className={css.bg} />
-				<div className={css.client}>{icon}{children}</div>
+				<div className={css.clientContainer}>
```
Djspaceg: It was my expectation that this change would not necessitate adding any new DOM.

Djspaceg:
I found a way to implement this without exposing any additional classes, without altering IconButton, and without interfering with the small tap-area, while also providing automatic spacing support for the text and customizable spacing between the color-dot, the icon, and the text. I also found a bug (from my previous Button PR) where the transparent button was no longer transparent. That has been fixed as well.

packages/moonstone/styles/colors-light.less
```diff
@@ -15,10 +15,10 @@
 @moon-white:       #fff;
 @moon-red:         #f00;
 
-@moon-remote-button-red-color:     #ff6d78; // #da3c3c; // #e71b1c;
-@moon-remote-button-green-color:   #6ff467; // #21c171; // #339966;
-@moon-remote-button-yellow-color:  #ffc600; // #dac139; // #fddb24;
-@moon-remote-button-blue-color:    #0084ff; // #5677d0; // #4764b4;
+@moon-remote-button-red-color:     #e1777e; // #ff6d78; // #da3c3c; // #e71b1c;
```
ryanjduffy: Do we need a complete history of "moonstone red" in source? :)

Djspaceg: I thought it was nice, it's small, and shows some evolution over time. It didn't hurt anybody and is pruned out during compile. Just a nice little easter egg, especially if you have an editor set up to display hex colors as colors.

ryanjduffy:
Everything seems to check out. Good work @teckliew + @Djspaceg!

I noticed that when using `color` with an image icon source, the icon doesn't shift up as it does when using the icon font. However, this occurs on `develop` as well so I'll write up a bug for future consideration.

Djspaceg:
@ryanjduffy, That's certainly a good consideration. The current shift up occurs due to changing the line-height property, which block elements don't respond to the same way as inline elements do. Admittedly we need to make our image support a bit more robust for the wide variety of image use-cases out there.

PR #2183, opened 2019-03-29
PLAT-75393 - Added stopPropagation to ToggleIcon

rundmt:
### Checklist

RadioItem was firing twice when being tapped on.

### Resolution
Add stopPropagation to `ToggleIcon` to stop it from firing twice.

### Additional Considerations
Seems to work so far. I'm making this a draft for now while I write tests. I'm having some trouble reproducing the bug in unit tests.

### Links
PLAT-75393

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -8,6 +8,10 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 - `ui/Item`, `ui/Layout`, `ui/Repeater`, `ui/SlotItem`, `ui/Spinner`, `ui/ToggleItem`, and `ui/ViewManager` support for `ref` to gain access to the wrapped `component`
 
+### Fixed
+
+- `ui/ToggleIcon` to not bubble `touchend` event to not double fire
+
 ## [2.4.1] - 2019-03-11
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


rundmt:
@ryanjduffy Yeah that is strange I'll take it look at it.

ryanjduffy:
@rundmt - I don't think the updated approach is going to work. I wasn't able to test it on my Surface table (it's reporting `false` for touch for some reason ...) but when i force `touch` to true to simulate a dual touch/pointer scenario, Slider doesn't drag.

It does seem to work on the reported problem so that might give us a hint that the issue lies in the handling of both mouse and touch events.

webOS101:
@ryanjduffy Can you share what your user agent is for the surface?  Curious why we're not detecting it properly.  I suspect it might be because Edge now supports this:  https://developer.mozilla.org/en-US/docs/Web/API/Navigator/maxTouchPoints instead of the prefixed version.  Looks like we need to update our logic in `hasTouch()`

ryanjduffy:
I tried edge but it was so wonky on first render i didn't even try it really. I was testing in Chrome which wasn't reporting `ontouchstart` in window.

webOS101:
Does it have `maxTouchPoints`?

ryanjduffy:
@rundmt I think I've tracked down the issue. Here's my alternative:

https://github.com/enactjs/enact/pull/2192

rundmt:
@ryanjduffy Ahh I see, this does look better. I totally forgot we had the timestamp problems with mouse. It does seem to work, I'll approve the other PR and use that one to replace this one.

rundmt:
Closing in favor of #2192

PR #2184, opened 2019-04-01
ENYO-5905: Fix QA Tooltip sample typo

ryanjduffy:
### Checklist

We were referencing an undefined key in state thereby breaking the "toggle" logic.

### Resolution
Reference the correct state member

### DIFF
--- a/packages/sampler/stories/qa/Tooltip.js
+++ b/packages/sampler/stories/qa/Tooltip.js
@@ -59,7 +59,7 @@ class ChangeableTooltip extends React.Component {
 	}
 
 	changeTooltipText = () => {
-		const {stringText} = this.state;
+		const {text: stringText} = this.state;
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Tooltip.js
```diff
@@ -59,7 +59,7 @@ class ChangeableTooltip extends React.Component {
 	}
 
 	changeTooltipText = () => {
-		const {stringText} = this.state;
+		const {text: stringText} = this.state;
```
Djspaceg: Odd that we'd rename the property `stringText` rather than just change the references to the actual name of the state (`text`)

ryanjduffy: `text` is defined already in the module scope for the knob so this avoids shadowing

PR #2186, opened 2019-04-02
PLAT-74078: Enable drag support for regular javascript scroller

sugardave:
### Checklist

This patch changes `moonstone/Scrollable` to enable touch/drag scrolling by default.

### Resolution
All `ui/Scrollable` instances have touch/drag scrolling enabled while `moonstone/Scrollable` used prop `noScrollByDrag`.

### Additional Considerations
This change should ideally be tested on touch-capable hardware.

With emulated touch testing and without this patch, Chrome will scroll a `moonstone/Scroller` instance, but not send `onScrollStart` and `onScrollEnd` events.  With this patch, those events are now sent.

### Links
PLAT-74078


sugardave:
Makes sense, updated to only enable on touch platforms.

PR #2188, opened 2019-04-03
ENYO-5906: Add check for demote to avoid setState call

viodragon2:
### Checklist

`MarqueeDecorator` queues up `demoteJob` and fires `setState` calls when it's not promoted.

idlecallbacks

### Resolution
Add check before queuing up

### DIFF
--- a/packages/ui/Marquee/MarqueeDecorator.js
+++ b/packages/ui/Marquee/MarqueeDecorator.js
@@ -393,7 +393,9 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 
 		demote () {
 			this.promoteJob.stop();
-			this.demoteJob.idle();
+			if (this.state.promoted) {
+				this.demoteJob.idle();
+			}
 		}
 
 		promote (delay = this.props.marqueeDelay) {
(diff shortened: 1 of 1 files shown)


packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -567,7 +567,9 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				});
 			}
 
-			this.demote();
+			if (this.state.promoted) {
```
ryanjduffy: Is it safe to skip all of `demote()` since it also stops the `promoteJob` or should this check just be applied around the `idle()` call?

ryanjduffy: Doesn't seem like any change is required. I tried to repro and wasn't able to and couldn't find a way through reading the code that seemed to indicate a problem.

ryanjduffy:
:nice: catch, @viodragon2!

packages/ui/Marquee/MarqueeDecorator.js
webOS101: I'm not so sure. I think I would prefer this to check the state in the `demote` method so we can safely abort the promotion job even if it hasn't finished yet.

viodragon2:
Changed condition to be placed in `demote` method as suggested.

PR #2189, opened 2019-04-03
ENYO-5904: Fix TooltipDecorator to position tooltips within non-root floating layers

ryanjduffy:
### Checklist

When tooltips are rendered within non-root `FloatingLayer`s as is the case in `VideoPlayer`, they are positioned incorrectly based on the viewport rather than the layer into which they are rendered.

### Resolution
* Add `node` to the `FloatingLayer.onOpen` payload to use as a reference point for positioning elements within the layer
* Update the positioning logic of Tooltip to "relativize" the bounds based on the floating layer

### Additional Considerations
Instead of updating `onOpen`, we could "find" the node using `closest()` DOM API but would require breaking encapsulation since there's no way from TooltipDecorator to "tag" the floating layer. Any props added to `FloatingLayerBase` are attached to a child of the node that is establishing the positioning context. Given that, exposing the node seemed like the right API but open to alternatives.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/TooltipDecorator` to correctly position tooltips within the bounds of non-root floating layers
+
 ## [2.5.0] - 2019-04-01
 
 ### Fixed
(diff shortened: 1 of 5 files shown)


Djspaceg:
This may be simplified by leveraging the new `relative` positioning prop in the updated Tooltip branch.
See: https://github.com/enactjs/enact/compare/feature/ui-update...feature/PLAT-76464-blakestephens#diff-b6e1e31265f4f342f40103f18a87f737R88

PR #2190, opened 2019-04-04
PLAT-73229: Update Header to match latest designs

Djspaceg:
### Checklist

Header's got some new visuals and features.

### Resolution
* Updated Header's sizing and positioning
* Accounted for light/dark themes, LTR/RTL, `noCloseButton`/no controls, compact/standard types.
* `Panels` is now smarter about & when to apply the (newly renamed) `.hasControls` class.
* Dropped the `casing` prop, but *did not* update the samples or tests, and don't plan to until we get confirmation that it should be gone for good.
* Added Measurable HOC and Hook
* Panels now supports a `<controls>` slot, which is intended to have app-level `IconButton` controls in it which remain stationary and are accessible from any panel.

### Additional Considerations
Also updated the font names to match the current version.

**Uppercase has been removed from Header, but the tests have not been updated. They should be updated before merging. _Left in place until we get verification that the feature is going away._** If the feature is retained, this PR will no longer be considered a "breaking" change.


ryanjduffy:
Proposed change looks good! :clap: :clap: @Djspaceg 

In addition to what we discussed earlier, I noticed a couple differences from the specs.

* The `padding-right` on the header components `<nav>` element is leaving some unusable space for long titles
image
* Appears to be missing the 60px margin between the `children` controls and the text
image

Djspaceg:
I think I addressed each of the issues we discussed "in person" and above.

I noticed a rendering issue that occurs during first load of Storybook samples. The header title is "blank" because the padding is set too large, and it pushes the text off screen. It must be measuring something too soon in the render cycle and not updating after render. Every other after initial Storybook load seems fine, just that first one seems messed up. This will require further investigation, however, I did not notice this in sample apps.

packages/moonstone/Panels/Header.module.less
```diff
@@ -90,6 +88,7 @@
 	// Compact Header
 	&.compact {
 		height: @moon-header-height-small;
+		height: 87px;
```
ryanjduffy: Guessing this wasn't supposed to be here.

packages/moonstone/styles/colors.less
```diff
@@ -127,6 +127,7 @@
 @moon-header-full-bleed-border-color: fade(@moon-white, 25%);
 @moon-header-input-bg-color:          @moon-spotlight-color;
 @moon-header-input-text-color:        @moon-spotlight-text-color;
+@moon-header-controls-border-color:   #383838;
```
ryanjduffy: Noticed that we used an specific value here and the `@moon-border-color` variable above and wanted to verify it was by design.

packages/ui/Measurable/Measurable.js
```diff
@@ -0,0 +1,88 @@
+/* eslint-disable react/sort-prop-types */
```
ryanjduffy: Is this linting directive still required? I wouldn't expect so given the code.

packages/ui/Measurable/Measurable.js
```diff
@@ -0,0 +1,88 @@
...
+	 * @memberof ui/Measurable.Measurable.defaultConfig
+	 */
+	refProp: 'forwardRef'
+};
+
+function useMeasurable () {
+	const [measurement, setMeasurement] = useState();
+	const ref = useCallback(node => {
+		if (node != null) {
+			setMeasurement(node ? node.getBoundingClientRect() : null);
```
ryanjduffy: Is this ternary or the `if` above redundant? Is using just this line better since it would invalidate the `measurement` state when the node were unmounted?

ryanjduffy:
Made a small doc commit to fix up Measurable. The storybook rendering issue seems to be caused by measurement occurring before the CSS has been applied. Also seems unique to `serve` so I think we can defer the fix from this ticket.

I would like to add some unit tests for Measurable and fix up the failed tests before moving forward.

Good work, @Djspaceg!

packages/moonstone/Panels/Header.module.less
Djspaceg: Does seem out of place, doesn't it.

packages/moonstone/Panels/ApplicationCloseButton.module.less
```diff
@@ -1,12 +1,5 @@
 // ApplicationCloseButton.module.less
 //
-@import '../styles/mixins.less';
-@import '../styles/variables.less';
 
 .applicationCloseButton {
```
ryanjduffy: I can see some value in keeping this for future use but there would be a minor perf win for removing it by avoiding `styles` logic in `kind`.

packages/moonstone/Panels/Header.js
```diff
@@ -45,18 +43,6 @@ const HeaderBase = kind({
...
-		 * Configures the mode of uppercasing for the `title`.
-		 *
-		 * * Values: `'upper'`, `'preserve'`, `'word'`, `'sentence'`
-		 *
-		 * @see i18n/Uppercase#Uppercase.casing
-		 * @type {String}
-		 * @default 'upper'
-		 * @public
-		 */
-		casing: PropTypes.oneOf(['upper', 'preserve', 'word', 'sentence']),
```
ryanjduffy: I'll add a change log entry for this.

packages/ui/Measurable/Measurable.js
ryanjduffy: @Djspaceg can you respond to this please?

packages/ui/Measurable/Measurable.js
ryanjduffy: @Djspaceg Also this

packages/moonstone/styles/colors.less
Djspaceg: Yes, it's intentional. The Header border is not using our standard colors, I believe for perceptual reasons.

packages/moonstone/Panels/Header.js
Djspaceg: Thank you, yes I forgot that.

packages/moonstone/Panels/ApplicationCloseButton.module.less
Djspaceg: Yeah, this was one of the ways I thought of to solve the emptiness problem. I could also simply remove the import statement, comment out the import, or remove it _and_ delete the file. Do you have a preference?

ryanjduffy: I don't think it'll move the needle much in either direction so I'll leave it to you to decide how to handle.

packages/ui/Measurable/Measurable.js
Djspaceg: So if `node` is null or undefined, `setMeasurement` won't be executed at all. If it's some other value, like false or something, then the measurement will be set to null. What may make this more meaningful would be changing `node` to the more detailed `node.getBoundingClientRect`. WDYT?

packages/ui/Measurable/Measurable.js
Djspaceg: This looks like cruft from wherever I copied this starting-point code. I'll fix it up now.

packages/ui/Measurable/Measurable.js
ryanjduffy: `node` should either be a Node or `null`. if it's set to `null` because the `ref` is set to a Node that is unmounted, we should call `setMeasurement` with `null` to clear the measurements since they aren't accurate any longer. As I read the code now, `measurement` will be out of sync.

PR #2191, opened 2019-04-04
PLAT-76463: Use lightweight components to improve scroll performance

germboy:
### Checklist

When holding a 5way direction in a `VirtualList`, too much time is spent updating components (data or position of the component in the list)

### Resolution
1) This fix relies on https://github.com/enactjs/enact/pull/2172
2) If we use lightweight components (alternate, light versions of Item or GridListImageItem), we can reduce the amount of time spent updating them. When these components are focused (using pointer) or receive a `keyup` directional event (indicating focus and 5way navigation has stopped), they will be updated to the full component - allowing actions like marquee to occur.
3) Focus changes have been moved to the `keydown` event, reducing long frames when updating the list/components during a 5way hold.

### DIFF
--- a/packages/moonstone/Item/Item.js
+++ b/packages/moonstone/Item/Item.js
@@ -10,19 +10,16 @@
  * @exports ItemDecorator
  */
 
-import {forward} from '@enact/core/handle';
 import kind from '@enact/core/kind';
-import {Job} from '@enact/core/util';
-import Spotlight from '@enact/spotlight';
 import Spottable from '@enact/spotlight/Spottable';
 import {ItemBase as UiItemBase, ItemDecorator as UiItemDecorator} from '@enact/ui/Item';
 // import Pure from '@enact/ui/internal/Pure';
 import PropTypes from 'prop-types';
 import compose from 'ramda/src/compose';
 import React from 'react';
-import ReactDOM from 'react-dom';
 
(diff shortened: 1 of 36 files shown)


seunghoh:
Overall, looks good idea to improve render performance via using light item. I left some small comment. 
One issue this PR is page up/down doesn't work. Spotlight focus lost.
One question is, what if we add event handler to the Item or GridListImageItem? Don't we lose handler if light change to full?

viodragon2:
I'm still investigating on why scroll by page doesn't work correctly, but in the meantime, I have noticed the issue of `white-space: pre` applied by `Marquee` and wanted to make a note here.

See https://codesandbox.io/s/o9891n87o9 for this possible bug.

packages/moonstone/Item/Item.js
```diff
@@ -79,13 +78,19 @@ const ItemBase = kind({
...
-	Pure,
 	UiItemDecorator,
 	Spottable,
 	MarqueeDecorator({invalidateProps: ['inline', 'autoHide']}),
 	Skinnable
 );
 
+const ItemFull = ItemDecorator(ItemBase);
+const ItemLightDecorator = compose(
+	Spottable,
```
viodragon2: Since `UiItemDecorator` is not wrapped, click/touch will not work until updated.

PR #2194, opened 2019-04-05
PLAT-75396: Signage Enact -  Sampler allows zoom

sugardave:
### Checklist

Added platform name classes for the moonstone environment to enable platform-specific styling.

### Resolution
Multi-touch actions (i.e. pinch-zoom) are disabled on webOS platforms in moonstone samples.

### Additional Considerations
Since this is only applied for the moonstone panels, it is still possible to pinch zoom Sampler app itself from the story list section or the knobs section.

### Links
PLAT-75396


LucieRoy:
@webOS101 - tested the opened branch. Can zoom the components/stories menu list but not the viewport, as you described. Is it possible to stop the zoom also on the menu list? I think it would be best to turn off the zoom there too if possible?
What would be the use case of keeping the zoom only on the menu list?

webOS101:
This was fixed at the platform level.

PR #2198, opened 2019-04-10
PLAT-76786: Change Touchable to not gain focus when activated by touch

haileyR:
### Checklist

Spotlight is on item when 'data-spotlight-container-disabled' is touched.

### Resolution
Removing focus and blur from Touchable resolved this issue. Removing these don't affect other issues. Spotlight is taken care of from Spottable.

### DIFF
--- a/packages/ui/Touchable/Touchable.js
+++ b/packages/ui/Touchable/Touchable.js
@@ -8,7 +8,6 @@
 
 import {adaptEvent, call, forward, forwardWithPrevent, forProp, handle, oneOf, preventDefault, returnsTrue} from '@enact/core/handle';
 import hoc from '@enact/core/hoc';
-import {Job} from '@enact/core/util';
 import {on, off} from '@enact/core/dispatcher';
 import complement from 'ramda/src/complement';
 import platform from '@enact/core/platform';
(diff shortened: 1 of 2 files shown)


packages/ui/Touchable/Touchable.js
```diff
@@ -501,18 +501,10 @@ const Touchable = hoc(defaultConfig, (config, Wrapped) => {
...
-			if (ev && 'changedTouches' in ev) {
-				this.target.focus();
-			}
-
 			return true;
 		}
 
-		deactivate (ev) {
-			if (ev && 'changedTouches' in ev) {
-				this.blurJob.start(this.target);
```
ryanjduffy: Is `this.blurJob` used elsewhere or should we clean it up as well?

ryanjduffy: Looks like this it's only job (:rimshot:) so let's clean up it and the `Job` import too.

haileyR: removed!

ryanjduffy:
This does match Enyo's behavior though I did prefer the momentary touch feedback we had. I suggested we investigate adding this back in spotlight rather than Touchable so the focus rules could be better respected.

PR #2199, opened 2019-04-11
PLAT-76981: Fix Touchable to prevent handling events on multiple nodes

ryanjduffy:
### Checklist

When touching on a component that handles `onTap` *and* reveals another Touchable instance in the same position as the tap, the mouse events could occur on a different target than the touch events.

The event order for touch events and mouse events in Touchable:

* `onTouchStart`
* `onDown`
* `onTouchEnd`
* `onUp`
* `onTap`
* `onMouseDown`
* `onMouseUp`
* `onClick`

If `onTap` changes the DOM, the subsequent `onMouseDown` might happen over a different DOM node causing this bug.

### Resolution;
Add (back? possibly a regression from #2195) a global Touchable variable to verify that the `mousedown` came from the same `target` as `touchend` before handling it.

### Additional Considerations
Doesn't appear to be any good way to test this in unit tests. Perhaps we could simulate it in UI tests but it could be difficult.

### DIFF
--- a/packages/ui/Touchable/ClickAllow.js
+++ b/packages/ui/Touchable/ClickAllow.js
@@ -1,13 +1,18 @@
 // It's possible that emitting `onTap` will cause a DOM change before the mousedown fires resulting
 // in multiple tap/click events for the same user action. To avoid this, we store the last touchend
-// target to compare against the next mouse down. If the target is the same (or no previous target
-// was set if no touch events are emitted), we allow the mousedown *across Touchable instances*.
-let _lastTouchEndTarget = null;
+// target and timestamp to compare against the next mouse down. If the timestamp is different (e.g
+// we're on a hybrid device that emitted a touch event but the next was a mouse event) or the target
+// is the same (or no previous target was set if no touch events have been emitted), we allow the
+// mousedown *across Touchable instances*.
+let _lastTouchEnd = {
+	target: null,
+	timeStamp: 0
+};
 
 const shouldAllowMouseDown = (ev) => {
-	return (
-		ev.target === _lastTouchEndTarget ||
-		_lastTouchEndTarget === null
+	return ev.timeStamp !== _lastTouchEnd.timeStamp || (
+		ev.target === _lastTouchEnd.target ||
+		_lastTouchEnd.target === null
 	);
 };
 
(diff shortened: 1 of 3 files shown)


packages/ui/Touchable/ClickAllow.js
```diff
@@ -1,3 +1,16 @@
+// It's possible that emitting `onTap` will cause a DOM change before the mousedown fires resulting
+// in multiple tap/click events for the same user action. To avoid this, we store the last touchend
+// target to compare against the next mouse down. If the target is the same (or no previous target
+// was set if no touch events are emitted), we allow the mousedown *across Touchable instances*.
+let _lastTouchEndTarget = null;
+
+const shouldAllowMouseDown = (ev) => {
+	return (
+		ev.target === _lastTouchEndTarget ||
```
webOS101: Should something clear `_lastTouchEndTarget`?  e.g. platform with both touch and mouse.

ryanjduffy: I considered this but I'm not sure how best to do it. We could add a global event listener but that would add overhead. I'm fine to defer this for now since that class of device isn't on our immediate support radar.

webOS101: You mentioned (I think??) that we could use timestamps.  Would that fix the issue or is it not guaranteed?

ryanjduffy: Maybe. I had them, then removed them when I combined with `ClickAllow` because they were duplicate, but adding it back alongside target might do the trick. I'm not sure how to verify it works though so I'm hesitant to make the change.

webOS101: Maybe I can test to see if this causes any issues with Chrome switching from touch to not-touch.  If it doesn't, then I'm happy.  If it does, I don't want to leave a known bug in, unless we document it for our future selves.

webOS101: I tested on Chrome and it's possible to 'disable' components for clicking once they've been touched.  For example, `CheckboxItem`, if you tap on the check in Touch mode then switch to pointer mode and try to click, it no longer will respond to the clicks on the check mark (the body still works until after you touch on it).  Can you find a solution that does not have this issue?

ryanjduffy: Thanks for the catch! Adding back the `timeStamp` to the check seems to due the trick since it allows a subsequent mouse event that wasn't preceded by a touch event as you described. Toggling in either touch or mouse events on the marquee or the icon works as well.

PR #2200, opened 2019-04-12
PLAT-75346: Update design of moonstone/Header with Input

ryanjduffy:
### Checklist

`moonstone/Header` needs to be updated to reflect the new designs when using a `headerInput`

### Resolution
* Expose the `focused` class on `moonstone/Input` to adjust the necessary colors from `Header`
* Update the CSS in `Header` for the new designs

### Additional Considerations
* Changed the `.title` margin to 2px which, though non-standard and not divisible by 3, allows the input and title text to match in all resolutions except HD (where 3px does the trick). Would like to find a better alternative to this approach.

### DIFF
--- a/packages/moonstone/Input/Input.js
+++ b/packages/moonstone/Input/Input.js
@@ -231,7 +231,7 @@ const InputBase = kind({
 	styles: {
 		css: componentCss,
 		className: 'decorator',
-		publicClassNames: ['decorator', 'input']
+		publicClassNames: ['decorator', 'focused', 'input']
 	},
 
 	handlers: {
--- a/packages/moonstone/Input/Input.js
+++ b/packages/moonstone/Input/Input.js
@@ -49,7 +49,6 @@ const InputBase = kind({
 		 * The following classes are supported:
 		 *
 		 * * `decorator` - The root class name
-		 * * `focused` - Applied to the root when the <input> is focused
 		 * * `input` - The <input> class name
 		 *
 		 * @type {Object}
@@ -79,15 +78,6 @@ const InputBase = kind({
 		 * @public
 		 */
 		dismissOnEnter: PropTypes.bool,
-
-		/**
-		 * Adds a `focused` class to the input decorator.
-		 *
-		 * @type {Boolean}
-		 * @default false
-		 * @public
-		 */
-		focused: PropTypes.bool,
 
 		/**
 		 * The icon to be placed at the end of the input.
@@ -232,7 +222,7 @@ const InputBase = kind({
 	styles: {
 		css: componentCss,
 		className: 'decorator',
-		publicClassNames: ['decorator', 'focused', 'input']
+		publicClassNames: ['decorator', 'input']
 	},
 
 	handlers: {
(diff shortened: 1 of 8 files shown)


packages/moonstone/Panels/Header.module.less
```diff
@@ -113,24 +113,30 @@
 
 		&.standard {
 			.decorator {
-				background-color: transparent;
+				&:not(.focused) {
```
Djspaceg: Is it possible to do the opposite of this? We're setting background-color, input color, and placeholder color below already in the `focus` state section.

packages/moonstone/Panels/Header.module.less
```diff
@@ -62,12 +62,12 @@
...
 
 		.title {
-			// This `margin-top: -3px;` is intended to account for the height disparity and visual
+			// This `margin-top: 2px;` is intended to account for the height disparity and visual
 			// offset when switching between Header+Input and a Standard Header's title height.
 			// The height is determined by the line-height below, but we want to maintain that as a
 			// relative unit (`em`) rather than dictate specific pixels. The -3px offset may need to
 			// be updated or removed to accomodate taller glyphs.
-			margin-top: -3px;
+			margin-top: 2px;
```
Djspaceg: This is indeed not ideal. Things that affect layout really shouldn't have a non-multiple of 3. :\

packages/moonstone/Input/Input.js
```diff
@@ -231,7 +232,7 @@ const InputBase = kind({
 	styles: {
 		css: componentCss,
 		className: 'decorator',
-		publicClassNames: ['decorator', 'input']
+		publicClassNames: ['decorator', 'focused', 'input']
```
Djspaceg: Is this absolutely necessary to achieve the desired style?

ryanjduffy:
Switching to `:focus-within` simplified things quite a bit. Thanks for the suggestion, @Djspaceg!

ryanjduffy:
I've removed the two color variables and updated the change log in which i included a `Removed` entry for the `focused` prop and a generic `Changed` entry for the Header visual updates.

I'll make another PR targeting `develop` for the deprecation notice for `focused`. I doubt I'll be able to safely warn on usage since it's a prop that's actively used by `InputSpotlightDecorator` in that version so it would warn all the time.

packages/moonstone/Input/Input.module.less
```diff
@@ -132,7 +132,7 @@
 				color: @moon-input-placeholder-color;
 			});
 
-			.focused& {
+			&:focus-within {
```
webOS101: Is this OK to use?  It's not supported in Blink 53. Do we need a fallback or alternative solution for older browsers?  Or, should we document that 3.x can't be used on older Blink?

ryanjduffy: Hmm. Good question. I think we're safe but we should confirm that apps built with 3.0 will not be deployed to older platforms.

Djspaceg: I'm curious what other JavaScript and CSS features we already use that are not compatible with such an old browser (released in August 2016).

ryanjduffy: We've been pretty careful out of necessity but there is at least one guarded case (e.g. `ResizeObserver`). Not sure how many others though.

PR #2203, opened 2019-04-15
PLAT-75399: Updated font names

Djspaceg:
### Checklist

### Resolution
We've received new font files with their (perhaps) final names.

### Additional Considerations
Updated the main fonts file and the references to the "Global" font name in the fontGenerator code.

### DIFF
--- a/packages/moonstone/MoonstoneDecorator/fontGenerator.js
+++ b/packages/moonstone/MoonstoneDecorator/fontGenerator.js
@@ -11,8 +11,8 @@ const fontName = 'Moonstone LG Display';
 // Locale Configuration Block
 const fonts = {
 	'NonLatin': {
-		regular: 'LG Display-Light',
-		bold:    'LG Display-Regular'
+		regular: 'LGSmartUIGlobal-Light',
+		bold:    'LGSmartUIGlobal-Regular'
 	},
 	'am': {
 		regular: 'LG Display_Amharic'
--- a/packages/moonstone/MoonstoneDecorator/fontGenerator.js
+++ b/packages/moonstone/MoonstoneDecorator/fontGenerator.js
@@ -11,8 +11,8 @@ const fontName = 'Moonstone LG Display';
 // Locale Configuration Block
 const fonts = {
 	'NonLatin': {
-		regular: 'LGSmartUIGlobal-Light',
-		bold:    'LGSmartUIGlobal-Regular'
+		regular: 'LG Display-Light',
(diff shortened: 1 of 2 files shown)


packages/moonstone/MoonstoneDecorator/fontGenerator.js
```diff
@@ -11,8 +11,8 @@ const fontName = 'Moonstone LG Display';
 // Locale Configuration Block
 const fonts = {
 	'NonLatin': {
-		regular: 'LG Display-Light',
-		bold:    'LG Display-Regular'
+		regular: 'LGSmartUIGlobal-Light',
+		bold:    'LGSmartUIGlobal-Regular'
```
seonmiJin: One question, Isn't it should be `LG Smart UI Global`?

Djspaceg: If it refers to just the family name, we can't target specific font-variants.

packages/moonstone/styles/internal/fonts.less
```diff
@@ -1,8 +1,8 @@
 // ----- SYSTEM FONT NAMES ------
-@font-system-src-moonstone-condensed-300:         local("LGSmartUICondTest-Light"), local("Miso"), url("../../fonts/Miso/Miso-Light.ttf") format("truetype");
-@font-system-src-moonstone-condensed-400:         local("LGSmartUICondTest-Regular"), local("Miso"), url("../../fonts/Miso/Miso-Regular.ttf") format("truetype");
-@font-system-src-moonstone-condensed-600:         local("LGSmartUICondTest-SemiBold"), local("Miso"), url("../../fonts/Miso/Miso-Bold.ttf") format("truetype");
-@font-system-src-moonstone-condensed-700:         local("LGSmartUICondTest-Bold"), local("Miso"), url("../../fonts/Miso/Miso-Bold.ttf") format("truetype");
+@font-system-src-moonstone-condensed-300:  local("LGSmartUICond-Light"), local("Miso"), url("../../fonts/Miso/Miso-Light.ttf") format("truetype");
```
Seonghyup: I think We should use font family instead of PS name in this PR.

Djspaceg: You know... I think you're right. The TV appears to not leverage the PostScript names, but does seem to be able to access the full-name, rather than just the family-name. I'll make a PR to update our fonts file to refer to the full-names instead. Thanks.

PR #2208, opened 2019-04-18
ENYO-5920: Fix Pickers to not error when min > max

ryanjduffy:
### Checklist

When `min` exceeded `max`, `EditableIntegerPicker`, `Picker`, and `RangePicker` would throw a runtime error from ramda's `clamp` method.

### Resolution
* Add a `clamp` method to `core/util` that safely clamps values
* Use `core/util.clamp` in the Pickers and Slider (which doesn't currently error but should use the same logic)

### Additional Considerations
The logic used in clamp mirrors the logic used by `<input type="range">` in that it uses `min` as the value when `min` exceeds `max`.

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `core/util.clamp` to safely clamp a value between min and max bounds
+
 ## [2.5.1] - 2019-04-09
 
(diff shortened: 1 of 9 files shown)


ryanjduffy:
As much as is possible, I think it's best to warn but to fail gracefully. This impl follows how HTML behaves and, as such, seems appropriate. Happy to discuss more!

ryanjduffy:
@teckliew thanks for the note! `validateRange` warns for that scenario and is already in use by the various pickers so I think we're okay. Thoughts?

PR #2209, opened 2019-04-18
PLAT-77661: QA Sampler-Button with tap area displayed: added taparea to IconButton and a divider to viewport

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/qa/Button.js
+++ b/packages/sampler/stories/qa/Button.js
@@ -1,5 +1,6 @@
 import Button, {ButtonBase} from '@enact/moonstone/Button';
 import IconButton from '@enact/moonstone/IconButton';
+import Divider from '@enact/moonstone/Divider';
 import UIButton, {ButtonBase as UIButtonBase} from '@enact/ui/Button';
 import {icons} from '@enact/moonstone/Icon';
 import React from 'react';
@@ -95,6 +96,7 @@ storiesOf('Button', module)
 		'with tap area displayed',
 		() => (
 			<div>
+				<Divider>Button</Divider>
 				<Button
 					className={css.tapArea}
 					onClick={action('onClick')}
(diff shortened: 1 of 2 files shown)


Djspaceg:
I think this may need to be merged into `feature/ui-update` rather than `develop`. I'll wait for confirmation before switching it.

LucieRoy:
Correct! good catch! was my first fix on this branch! Thank you!

Djspaceg:
Retargeted. Much smaller change-set now. Also, this means that this doesn't need to go into the 2.x release. If you *do* want it in the 2.x release branch, it will need to be rebased on `develop` and re-targeted back to `develop` as well.

Djspaceg:
Why were tap areas added to large buttons? Large Buttons don't have or need tap areas.

webOS101:
For consistency, it makes sense to me.

Djspaceg:
But it's a sample/story explicitly for showing tap areas. If a component doesn't have a tap area, it shouldn't _show_ a non-existent tap area. That's misleading as to what's going on and what is being tested. The point is to verify and validate that large buttons don't have a mistakenly or unintentionally applied tap area and to show that small buttons _have_ a tap area and what size it is.

webOS101:
That was my point.  If some CSS applied a tap area around large buttons but we didn't show it, that's be an issue, wouldn't it?  Or, am I misunderstanding?

Djspaceg:
If something had unintentionally applied a tap area, it would appear and be visually apparent in this test. If we artificially insert a tap area in the sample, we can't tell if a tap area was added to the framework component.

Large buttons don't have tap areas because they don't need them. They're already above the minimum tap-target dimensions. A small button is only 60px tall, which is 20px smaller than the minimum tap-target dimensions, so we add an invisible tap area to make it easier to tap with the Magic Remote. This sample makes the invisible tap area visible.

webOS101:
I understand what you are saying.  What about the case of the large `IconButton` which has a square tap target?  It was not rendered appropriately with this test without modification.  I actually suspect the change in the border radius could possibly mislead, given that the small button used to have an expanded, but round, tap area.

LucieRoy:
The contrast of the tap area color blue is not as obvious on black as a brighter color. Red? Bright green? Orange?Why was blue chosen?

PR #2210, opened 2019-04-18
PLAT-76789: Add size prop and deprecate small prop

teckliew:
### Checklist

### Resolution
Deprecate `small` prop and use `size` prop instead. The components that use `small` prop: `ui/Button`, `ui/Icon`,  `ui/IconButton`, `ui/LabeledIcon`, `moonstone/Input` and `moonstone/ToggleButton`.

### Additional Considerations
Minor refactor for `LabeledIcon` so that `Cell`'s `size` prop does not conflict with `Icon`'s `size` prop.
In order to support `small` for 2.x, default props was not set for `size`, but instead the default is set in `getSize()`.

### Links
PLAT-76789


ryanjduffy:
To clarify, if we retargeted to `develop`, we'd have to ensure that both `small` and `size` were supported. If both were provided, `size` would override `small`.

Something like:
```js
let effectiveSize = size || 'medium';
if (!size && small) {
  effectiveSize = 'small';
}
```

teckliew:
Oh yea the deprecation is supposed to be for 2.6.0
I'll retarget this to `develop` and add in the override logic.

packages/moonstone/Input/Input.js
```diff
@@ -247,7 +263,7 @@ const InputBase = kind({
 			const title = (value == null || value === '') ? placeholder : '';
 			return calcAriaLabel(title, type, value);
 		},
-		className: ({focused, invalid, small, styler}) => styler.append({focused, invalid, small}),
+		className: ({focused, invalid, size, small, styler}) => styler.append({focused, invalid}, !size && small ? 'small' : size || 'medium'),
```
webOS101: If we include `size` in the `append`, will it be easier to update later?

```
styler.append({focused, invalid}, size, !size && (small ? 'small' : 'medium'))
```

Although, should we specify a default for a prop and not enforce it through `defaultProps`?  Could we implement this so it took that into account?

packages/moonstone/LabeledIconButton/LabeledIconButton.js
```diff
@@ -58,6 +58,7 @@ const LabeledIconButtonBase = kind({
 		 * * `labeledIconButton` - The root component class
 		 * * `icon` - The icon component class
 		 * * `label` - The label component class
+		 * * `medium` - Applied to a `medium` button
```
webOS101: As above.

teckliew:
> I believe there are additional required followup tickets with this one being structured the way it is. We'll need one to remove the deprecation warnings added here. We'll need another to change Button's default from "medium/large" to "small", and we'll need to remove any reference to size="small" since that will now the new default.

Agree need another ticket to remove the deprecations for 3.0 in addition to making the default size of "small". For now we'll keep the default "large" for 2.x

packages/moonstone/ExpandableInput/ExpandableInput.js
```diff
@@ -315,6 +315,7 @@ class ExpandableInputBase extends React.Component {
 					onDeactivate={this.handleDeactivate}
 					onKeyDown={this.handleInputKeyDown}
 					placeholder={placeholder}
+					size="small"
```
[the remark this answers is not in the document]
teckliew: I think we should start implementing this change in 2.x, but we could hold off until 3.0 if that makes more sense.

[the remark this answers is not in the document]
ryanjduffy: Let's avoid implementing unnecessary visual changes at this point. I'll remove it

packages/moonstone/Input/Input.js
```diff
@@ -27,6 +27,19 @@ import InputDecoratorIcon from './InputDecoratorIcon';
...
+const deprecateSmall = deprecate(() => 'small',  {
+	name: 'moonstone/Input.InputBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
+	since: '2.6.0',
+	until: '3.0.0'
+});
+
+function getSize (size, small) {
+	small = small ? deprecateSmall() : 'large';
```
ryanjduffy: This implementation will fail to warn with small is defined by falsey. I'll update to test for any defined value.

packages/moonstone/Input/Input.js
```diff
@@ -192,11 +205,21 @@ const InputBase = kind({
...
+		 * @public
+		 */
+		size: PropTypes.string,
+
 		/**
 		 * Applies the `small` CSS class.
 		 *
 		 * @type {Boolean}
 		 * @default false
+		 * @deprecated replaced by prop `size='small'`
```
webOS101: Should we remove the default here?

packages/moonstone/ToggleButton/ToggleButton.js
```diff
@@ -20,6 +21,19 @@ import Skinnable from '../Skinnable';
 
 import css from './ToggleButton.module.less';
 
+const deprecateSmall = deprecate((small) => small ? 'small' : 'large',  {
+	name: 'moonstone/ToggleButton.ToggleButtonBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
```
webOS101: back tick misplaced

packages/moonstone/Input/Input.js
```diff
@@ -27,6 +27,19 @@ import InputDecoratorIcon from './InputDecoratorIcon';
 import InputSpotlightDecorator from './InputSpotlightDecorator';
 import {calcAriaLabel, extractInputProps} from './util';
 
+const deprecateSmall = deprecate((small) => small ? 'small' : 'large',  {
+	name: 'moonstone/Input.InputBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
```
webOS101: back tick misplaced

packages/moonstone/ToggleButton/ToggleButton.js
```diff
@@ -94,13 +108,23 @@ const ToggleButtonBase = kind({
 		selected: PropTypes.bool,
 
 		/**
-		 * Reduces the size of the button.
+		 * The size of the button.
```
webOS101: The below text needs some reworking.  Probably above, too.

packages/moonstone/ToggleButton/ToggleButton.js
```diff
@@ -94,13 +108,23 @@ const ToggleButtonBase = kind({
...
+		 * @public
+		 */
+		size: PropTypes.string,
+
+		/**
+		 * Reduces the size of the button.
+		 *
 		 * @type {Boolean}
 		 * @default false
+		 * @deprecated replaced by prop `size='small'`
```
webOS101: remove default here, too

packages/ui/Button/Button.js
```diff
@@ -16,6 +17,19 @@ import Touchable from '../Touchable';
 
 import componentCss from './Button.module.less';
 
+const deprecateSmall = deprecate((small) => small ? 'small' : 'large',  {
+	name: 'ui/Button.ButtonBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
```
webOS101: here, too

packages/ui/Button/Button.js
```diff
@@ -127,13 +142,22 @@ const ButtonBase = kind({
...
+		 * @public
+		 */
+		size: PropTypes.string,
+
+		/**
+		 * Reduces the size of the component.
+		 *
 		 * @type {Boolean}
 		 * @default false
+		 * @deprecated replaced by prop `size='small'`
```
webOS101: remove default here, too

packages/ui/Icon/Icon.js
```diff
@@ -116,20 +131,33 @@ const Icon = kind({
...
+		 * @public
+		 */
+		size: PropTypes.string,
+
 		/**
 		 * Applies the `small` CSS class.
 		 *
 		 * @type {Boolean}
 		 * @default false
+		 * @deprecated replaced by prop `size='small'`
```
webOS101: here too

packages/ui/Icon/Icon.js
```diff
@@ -6,13 +6,27 @@
...
 import React from 'react';
 
 import ri from '../resolution';
 
 import componentCss from './Icon.module.less';
 
+const deprecateSmall = deprecate((small) => small ? 'small' : 'large',  {
+	name: 'ui/Icon.IconBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
```
webOS101: here too

packages/ui/IconButton/IconButton.js
```diff
@@ -21,6 +22,19 @@ import Touchable from '../Touchable';
 
 import componentCss from './IconButton.module.less';
 
+const deprecateSmall = deprecate((small) => small ? 'small' : 'large',  {
+	name: 'ui/IconButton.IconButtonBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
```
webOS101: ditto

packages/ui/IconButton/IconButton.js
```diff
@@ -122,11 +137,24 @@ const IconButtonBase = kind({
...
+		 * @public
+		 */
+		size: PropTypes.string,
+
 		/**
 		 * Applies the `small` CSS class.
 		 *
 		 * @type {Boolean}
 		 * @default false
+		 * @deprecated replaced by prop `size='small'`
```
webOS101: ditto

packages/ui/LabeledIcon/LabeledIcon.js
```diff
@@ -19,6 +20,19 @@ import Slottable from '../Slottable';
 
 import componentCss from './LabeledIcon.module.less';
 
+const deprecateSmall = deprecate((small) => small ? 'small' : 'large',  {
+	name: 'ui/LabeledIcon.LabeledIconBase#small',
+	replacedBy: 'the `size` prop',
+	message: 'Use `size="small" instead`.',
```
webOS101: ditto

packages/ui/LabeledIcon/LabeledIcon.js
```diff
@@ -129,13 +143,26 @@ const LabeledIconBase = kind({
...
+		size: PropTypes.string,
+
 		/**
 		 * Reduces the size of the icon component.
 		 *
 		 * The value of `small` is forwarded on to `iconComponent`.
 		 *
 		 * @type {Boolean}
 		 * @default false
+		 * @deprecated replaced by prop `size='small'`
```
webOS101: ditto

PR #2213, opened 2019-04-18
PLAT-77150: Fix moonstone to suppress tap highlighting on Android

sugardave:
### Checklist

Android browser adds a tap highlight color to tapped elements and it clashes with Moonstone.

### Resolution
Make the tap highlight color be `transparent`.

### Links
PLAT-77150

### DIFF
--- a/packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.module.less
+++ b/packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.module.less
@@ -3,22 +3,15 @@
 
 // Core Rules
 @import "~@enact/ui/styles/core.less";
-@import '~@enact/ui/styles/mixins.less';
+@import "~@enact/ui/styles/mixins.less";
 
 // Moonstone Internal Rules
-@import '../styles/internal/fonts.less';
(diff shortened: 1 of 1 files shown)


sugardave:
Unfortunately, only the prefixed tap highlight style keeps it from happening.

packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.module.less
```diff
@@ -13,9 +13,11 @@
 @import "../styles/skin.less";
 @import "../styles/variables.less";
 
-/* Prevent browser's default focus treatment (at least in Chrome) */
 :focus {
+	// Prevent Chrome's default focus treatment
 	outline: none;
+	// Prevent Android Chrome from highlighting tapped elements
+	-webkit-tap-highlight-color: transparent;
```
webOS101: Should we apply this into the root to prevent this action on all elements?

Djspaceg: We should apply this rule and the `:focus` rule to `.root` to keep the rules together and to ensure they only affect our components.

PR #2215, opened 2019-04-19
PLAT-75756: Add paging support to Scrollable when focus is not set

0x64:
### Checklist

The UX for the page navigation for `VirtualList` and `Scroller` component is updated, so those components should scroll 2/3 of the component's height by page up/down keys when the pointer is over the component and there is no item or scroll button is focused. 

### Resolution
Since the component should scroll by `keydown` of page keys without no focused node inside, it cannot catch an event if the `keydown` event handler is added to the node belongs to it. And, to check if the key event occurs when the pointer is over the component, we need to track the position of the pointer since we have no way to get the current position of the pointer in `keydown` handler.
To resolve these conditions, module-level event handlers for `keydown` and `mousemove` are added to `document`. `mousemove` handler keeps tracking the position of the pointer, and `keydown` handler informs the right component (a list or a scroller) that it should handle the key input.

### Additional Considerations
To minimize any impact by event handlers and to be safe in prerendering, module-scope handlers are added.

### Links
PLAT-75756


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -481,15 +536,17 @@ class ScrollableBase extends Component {
...
+		if (!repeat && (globalHandler || this.hasFocus())) {
 			const {overscrollEffectOn} = this.props;
 			let
 				overscrollEffectRequired = false,
 				direction = null;
 
 			if (isPageUp(keyCode) || isPageDown(keyCode)) {
 				if (this.props.direction === 'vertical' || this.props.direction === 'both') {
-					Spotlight.setPointerMode(false);
+					if (!globalHandler) {
```
ryanjduffy: If you're able to refactor as suggested above, this guard will likely go away since this method can always disable pointer mode and the global handler will skip this step.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -73,6 +75,45 @@ const configureSpotlightContainer = ({'data-spotlight-id': spotlightId, focusabl
...
+	scrollables = new Set(),
+	pageKeyHandler = (ev) => {
+		if (Spotlight.getPointerMode() && !Spotlight.getCurrent()) {
+			const
+				{x, y} = lastPointer,
+				elem = document.elementFromPoint(x, y);
+
+			if (elem) {
+				for (let s of scrollables) {
+					if (ReactDOM.findDOMNode(s).contains(elem)) { // eslint-disable-line react/no-find-dom-node
```
ryanjduffy: It'd be nice to avoid this `findDOMNode` if possible. Since we already have various node references in Scrollable, can we provide it to this construct instead. For example, instead of passing the React internal instance, we could use a WeakMap to map the node reference to the handler.

Something (loosely) like:

```
scrollables = new WeakMap();
pageKeyHandler = (ev) => {
  // existing code ...
  for (let node of scrollables.keys()) {
    if (node.contains(elem)) {
      scrollables.get(node)(ev);
      break;
    }
 }

// in the component
scrollables.set(this.someRef.current, this.onGlobalKeyDown);
```

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -73,6 +75,45 @@ const configureSpotlightContainer = ({'data-spotlight-id': spotlightId, focusabl
...
+		lastPointer.x = ev.clientX;
+		lastPointer.y = ev.clientY;
+	};
+
+const
+	// An app could have lists and/or scrollers more than one,
+	// so we should test all of them when page up/down key is pressed.
+	scrollables = new Set(),
+	pageKeyHandler = (ev) => {
+		if (Spotlight.getPointerMode() && !Spotlight.getCurrent()) {
```
ryanjduffy: It might be best to include the keycode checks here to avoid the potentially expensive `getElementFromPoint` call for keys we don't care about.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -470,7 +525,7 @@ class ScrollableBase extends Component {
 		return current && this.uiRef.current.containerRef.current.contains(current);
 	}
 
-	onKeyDown = (ev) => {
+	onKeyDown = (ev, globalHandler = false) => {
```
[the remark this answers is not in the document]
0x64: Updated!

packages/moonstone/Scrollable/Scrollable.js
0x64: Updated!

packages/moonstone/Scrollable/Scrollable.js
0x64: Updated!

packages/moonstone/Scrollable/Scrollable.js
0x64: Unfortunately `WeakMap` keys are not enumerable, so I can use `Set` or `Map`. And it looks not good to use _node_ as a key if we don't use `WeakMap`. So I kept _this_ as a key on `Map`.

0x64:
Hmm... Is it better to update a native version code also to match with JS version?

ryanjduffy:
@0x64 - I've also rebased this on `feature/ui-update` and retargeted the PR in order to land this with the rest of the UX/GUI visual and behavior changes.

PR #2218, opened 2019-04-19
PLAT-76982: Cross-Platform Enact Ipad -Tablet: VKB does not dismiss on tap in blank area

sugardave:
### Checklist

On ios, keyboard activity dismissal activity is handled by the developer (see "Dismissing the Keyboard" in https://developer.apple.com/library/archive/documentation/StringsTextFonts/Conceptual/TextAndWebiPhoneOS/KeyboardManagement/KeyboardManagement.html ).  This means that we need to either listen for touchstart events on non-text input elements after being focused on text input elements, or otherwise blur the focused input.

Having a style `cursor: pointer` causes the focused element to blur when tapping elsewhere, so it will allow the keyboard to dismiss.

### Additional Considerations
It does make the pointer different for browser apps, so this might need to be limited to ios platform somehow.  We could add `[platform]` to the root of `MoonstoneDecorator` instead and put the cursor style in `.root&.ios`.

### Links
PLAT-76982

### DIFF
--- a/packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.module.less
+++ b/packages/moonstone/MoonstoneDecorator/MoonstoneDecorator.module.less
@@ -72,6 +72,7 @@
 
 .root {
 	.moon-text-base(@moon-sub-header-font-size, self);
+	cursor: pointer;
 	font-weight: normal;
 	font-style: normal;
 	letter-spacing: normal;


webOS101:
Can you add some info to the PR text that describes why this fixes the problem?

ryanjduffy:
Closing in favor of #2232

PR #2221, opened 2019-04-22
PLAT-77661: Add IconButton to Taparea QA-sample

Djspaceg:
### Checklist

The clickable area in IconButton is too large (exceeds the round bounds of the visible part of IconButton). Update QA sample to include IconButton, now that IconButton and Button are different shapes.

### Resolution
Fixed IconButton showing a false positive when focusing a corner.
Update QA sample to include headings and IconButtons of both sizes.

### DIFF
--- a/packages/sampler/stories/qa/Button.js
+++ b/packages/sampler/stories/qa/Button.js
@@ -1,5 +1,6 @@
 import Button, {ButtonBase} from '@enact/moonstone/Button';
 import IconButton from '@enact/moonstone/IconButton';
+import Divider from '@enact/moonstone/Divider';
 import UIButton, {ButtonBase as UIButtonBase} from '@enact/ui/Button';
 import {icons} from '@enact/moonstone/Icon';
 import React from 'react';
@@ -95,6 +96,7 @@ storiesOf('Button', module)
 		'with tap area displayed',
 		() => (
 			<div>
+				<Divider>Button</Divider>
 				<Button
 					className={css.tapArea}
 					onClick={action('onClick')}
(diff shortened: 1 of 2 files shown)


ryanjduffy:
@Djspaceg Can you confirm that the icon button tap area should be squared even though the button is still round? I'm not clear on the expected behavior from the PR description.

image

Djspaceg:
Yes, apparently still squared is fine. It does increase the total amount of surface area for the button to be tapped, which is great because that's the purpose of the .taparea in the first place.

webOS101:
Does that imply we should _not_ fix the tap area of the large icon button to be round?

ryanjduffy:
@webOS101 I'm not clear on your question since large IconButtons do not have a tap area.

It is odd to me a bit that small IconButton's now have a larger tappable target than large IconButtons due to the square edges but /shrug.

webOS101:
It did have a square tappable area at the beginning of this PR.  I think you addressed that by making it round.  I, too, think it's weird that small now would have a larger tap area so perhaps we should keep it square for consistency?  Or, just go ahead and add a tap area to make it obvious and consistent?

ryanjduffy:
PS - I pushed a commit to add the knobs on Button to IconButton in that sample for consistency. They're not terribly important to the sample but it's odd to set `backgroundOpacity` (for example) and it only to have an effect on the buttons.

Djspaceg:
Yes, the knobs should have been replicated across each of the components so they all show the same state/props at the same time. Thanks @ryanjduffy.

packages/sampler/stories/qa/Button.js
```diff
@@ -120,9 +122,30 @@ storiesOf('Button', module)
...
+				</Button>
+				<Divider>IconButton</Divider>
+				<IconButton
+					backgroundOpacity={select('backgroundOpacity', prop.backgroundOpacity, Config)}
+					className={css.tapArea}
+					color={select('color', prop.color, Config)}
+					disabled={boolean('disabled', Config)}
+					onClick={action('onClick')}
+					selected={boolean('selected', Config)}
+
```
Djspaceg: Extra blank line here

Djspaceg:
> It did have a square tappable area at the beginning of this PR. I think you addressed that by making it round. I, too, think it's weird that small now would have a larger tap area so perhaps we should keep it square for consistency? Or, just go ahead and add a tap area to make it obvious and consistent?

It still has a squared tap area. Actually, it has a border-radius that matches the Base button's border-radius, so it's slightly rounded.

The small IconButtons have a square tap area because the UX team wanted the user to be able to pan from side to side across the buttons and have the focus retained on one of them (never drop focus). They did not feel that was important or necessary for the large version of the IconButton.

PR #2223, opened 2019-04-23
PLAT-77050: Update to check marqueeOn prop behavior with long title header

Baekwoo-s:
### Checklist

Fixed to check whether marquee is flowing without touch or hover on cross-platform(iOS, Android) when the title of header component is long.

### Resolution
[qa-sampler] Added selector of marqueeOn prop to header sample to make it possible to test marquee with long title.

### DIFF
--- a/packages/sampler/stories/qa/Header.js
+++ b/packages/sampler/stories/qa/Header.js
@@ -5,7 +5,7 @@ import React from 'react';
 import {storiesOf} from '@storybook/react';
 
 import {mergeComponentMetadata} from '../../src/utils';
-import {boolean, text} from '../../src/enact-knobs';
+import {boolean, select, text} from '../../src/enact-knobs';
 
 Header.displayName = 'Header';
 const Config = mergeComponentMetadata('Header', HeaderBase, Header);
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Header.js
```diff
@@ -102,9 +102,10 @@ storiesOf('Header', module)
 		'with long text and headerComponent',
 		() => (
 			<Header
-				title={text('title', Config, 'Title')}
+				title={text('title', Config, 'Title is very very very very very very very long.')}
 				titleBelow={text('titleBelow', Config, 'This is a header sample with long titleBelow text and header components to test positioning of header components.')}
 				subTitleBelow={text('subTitleBelow', Config, 'This is a header sample with long titleBelow, subTitleBelow text and header components to test positioning of header components.')}
+				marqueeOn={select('marqueeOn', ['', 'hover', 'render'], Header, '')}
```
seunghoh: Please remove empty and use with `default` value

Baekwoo-s: @seunghoh 
I removed the empty default value.

PR #2226, opened 2019-04-24
PLAT-77851: Change Notification appears to match new designs

viodragon2:
### Checklist

Update `moonstone/Notification` styles to match the latest GUI design doc

### Resolution

Changes:
- border changed
- add box shadow
- button gaps updated
- padding updated

### Additional Considerations
- GUI doc specifies #EFEFEF to be the background color, but I left it as-is to match all light-skin color background

### DIFF
--- a/packages/moonstone/styles/colors.less
+++ b/packages/moonstone/styles/colors.less
@@ -173,7 +173,7 @@
 
 // Notification
 // ---------------------------------------
-@moon-notification-border-color: #bbb;
+@moon-notification-border-color: fade(@moon-black, 30%);
 @moon-notification-box-shadow-color: fade(@moon-black, 30%);
 
 // Panels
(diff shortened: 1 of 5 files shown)


packages/moonstone/styles/colors.less
```diff
@@ -171,7 +171,8 @@
 
 // Notification
 // ---------------------------------------
-@moon-notification-border-color: fade(@moon-black, 30%);
+@moon-notification-border-color: #bbb;
```
ryanjduffy: @viodragon2 Can you verify the same color should be used for the border in both light and dark skins? This is different than ContextualPopup.

viodragon2: I was waiting for light skin doc for the final update, but I would assume the border color would be the same. I'll make an update.

ryanjduffy: Let's revert this change if it wasn't specced for consistency.

PR #2229, opened 2019-04-25
PLAT-73522: Change ContextualPopup appearance to match new designs

viodragon2:
### Checklist

Update contextual popup GUI

### Resolution
- change border radius
- change border color
- change border width
- adjust arrow SVG accordingly to match the new border width
- add drop shadow

### DIFF
--- a/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
+++ b/packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
@@ -20,6 +20,7 @@ import ri from '@enact/ui/resolution';
 import compose from 'ramda/src/compose';
 import PropTypes from 'prop-types';
 import React from 'react';
+import ReactDOM from 'react-dom';
 
 import {ContextualPopup} from './ContextualPopup';
 
@@ -243,9 +244,10 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 			this.overflow = {};
 			this.adjustedDirection = this.props.direction;
 
-			this.ARROW_WIDTH = ri.scale(30);
-			this.ARROW_OFFSET = ri.scale(18);
-			this.MARGIN = ri.scale(12);
+			this.ARROW_WIDTH = ri.scale(30); // svg arrow width. used for arrow positioning
+			this.ARROW_OFFSET = ri.scale(18); // actual distance of the svg arrow displayed to offset overlaps with the container
+			this.MARGIN = ri.scale(9); // margin from an activator to the contextual popup
+			this.KEEPOUT = ri.scale(12); // keep out distance on the edge of the screen
 
 			if (props.setApiProvider) {
 				props.setApiProvider(this);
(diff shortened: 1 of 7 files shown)


packages/moonstone/styles/variables.less
```diff
@@ -351,8 +351,8 @@
 
 // ContextualPopup
 // ---------------------------------------
-@moon-contextual-popup-border-width: @moon-button-border-width;
-@moon-contextual-popup-border-radius: 15px;
+@moon-contextual-popup-border-width: 1px;
```
ryanjduffy: `1px` is negligible at higher resolutions (4k and 8k) and isn't currently adjusted by our RI plugin. I think we'll need to resolve this eventually but not necessarily in this PR as it's a broader issue than is being solved here.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -615,9 +617,7 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
 							<PopupComponent {...popupPropsRef} />
 						</ContextualPopupContainer>
 					</FloatingLayer>
-					<div ref={this.getClientNode}>
-						<Wrapped {...rest} />
-					</div>
+					<Wrapped ref={this.getClientNode} {...rest} />
 				</div>
```
ryanjduffy: Also unrelated to this PR but I wonder if we really need this wrapping DOM. The only CSS that is applied here is `display: inline-block` and I'm not sure why that is even required. Appears this was ported from Enyo (https://github.com/enyojs/moonstone/commit/9b5ac029913e5804f146dd9c80b7cd2af0e5a938) but the rationale might not be relevant any longer.

packages/moonstone/styles/colors-light.less
```diff
@@ -78,7 +78,8 @@
 // ContextualPopup
 // ---------------------------------------
 @moon-contextual-popup-bg-color: @moon-bg-color;
-@moon-contextual-popup-border-color: fade(@moon-black, 50%);
+@moon-contextual-popup-border-color: #bbb;
```
ryanjduffy: Any insight on why we need a new color here (`#bbbbbb`) vs using `@moon-gray` (`#a6a6a6`)?

viodragon2: #bbbbb is a new color for border colors. #a6a6a6 is close, but we decided to keep it separate.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
viodragon2: The CSS part might be irrelevant now, but we still need `ref` in order to calculate the position of the popup.

packages/moonstone/styles/variables.less
viodragon2: This is intentional. The border isn't considered as a GUI element, but rather a functional "line" that differentiates between the popup and the outside.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
ryanjduffy: I don't think I was clear in my comment. The "wrapping div" to which I was referring wasn't the removed one but the one around Wrapped and FloatingLayer. No concerns with the proposed change.

PR #2230, opened 2019-04-25
PLAT-77739: [sampler, qa-sampler] Add css to prevent features that interfere with iOS, Android testing

Baekwoo-s:
### Checklist

During cross-platform testing, the basic functionality of iOS safari and android chrome apps hinders testing.
iOS safari: When I try to scroll something in the sample, the entire viewport is moved, which interferes with the test.
Android chrome: When I try to scroll something downward, the page reloads with the reload icon, which interferes with the test.
- This issue is caused by the height of the samples being rendered larger than the viewport.

### Resolution
Add CSS to prevent sampler and qa-sampler apps from scrolling by placing CSS in the app body
- I changed samples to render to the viewport size, added CSS to disable the ios viewport movement and the android reload feature.

Code of css to prevent iOS viewport from being scrolled
`
body {
position: fixed;
width: 100%;
height: 100%;
}
`
Code of css that disable Android reload feature
`
body {
overflow: hidden;
}
`

### Additional Considerations

### DIFF
--- a/packages/sampler/.qa-storybook/addons.js
+++ b/packages/sampler/.qa-storybook/addons.js
@@ -1,1 +1,2 @@
 import '../src/addons';
+import '../src/global.module.less';
--- a/packages/sampler/.storybook/addons.js
+++ b/packages/sampler/.storybook/addons.js
@@ -1,1 +1,2 @@
 import '../src/addons';
+import '../src/global.module.less';
--- a/packages/sampler/src/global.module.less
+++ b/packages/sampler/src/global.module.less
@@ -1,0 +1,6 @@
+body {
+	position: fixed;
+	width: 100%;
+	height: 100%;
+	overflow: hidden;
+}


0x64:
Not an actual issue but it would be better to add some comment that which CSS attribute is for which platform.

0x64:
For Chrome on Android, did you try `overscroll-behavior`?
https://developers.google.com/web/updates/2017/11/overscroll-behavior
https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior
There is an npm module to use this CSS. https://www.npmjs.com/package/prevent-pull-refresh

Baekwoo-s:
> Not an actual issue but it would be better to add some comment that which CSS attribute is for which platform.

I updated the PR description.

Baekwoo-s:
> For Chrome on Android, did you try `overscroll-behavior`?
> https://developers.google.com/web/updates/2017/11/overscroll-behavior
> https://developer.mozilla.org/en-US/docs/Web/CSS/overscroll-behavior
> There is an npm module to use this CSS. https://www.npmjs.com/package/prevent-pull-refresh

The `overscroll-behavior` that you pointed out prevents Android's ability to reload as mentioned.
However, I recommend using overflow: hidden.
sampler and qa-sampler have a slightly larger height than the viewport, so if you try to scroll the virtuallist, the viewport will also move.
I think you need to decide which property to use after you confirm this.

0x64:
@Baekwoo-s  I got your point. So, the root problem is not a pull-down-to-refresh behavior but an overflowed body.

0x64:
LGTM

PR #2232, opened 2019-04-25
PLAT-76982: Fix spotlight to unspot when tapping on a non-spottable target

ryanjduffy:
### Checklist

On iOS, tapping non-spottable targets when another element is focused will not remove focus from it. Tapping on another focusable element will move focus to the new target.

Appears that when tapping a focusable element, iOS emits `mouseover` events triggering spotlight logic but it does not do so on non-focusable elements.

### Resolution
* Add a `touchend` handler in spotlight for touch platforms to blur the currently focused element if it doesn't contain the target of the touch event.

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -504,6 +504,13 @@ const Spotlight = (function () {
 		}
 	}
 
+	function onTouchEnd (evt) {
+		const current = getCurrent();
+		if (current && !current.contains(evt.target)) {
+			current.blur();
+		}
+	}
+
 	/*
 	 * public methods
 	 */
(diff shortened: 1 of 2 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -524,10 +531,16 @@ const Spotlight = (function () {
 				window.addEventListener('keyup', onKeyUp);
 				window.addEventListener('mouseover', onMouseOver);
 				window.addEventListener('mousemove', onMouseMove);
+
+				if (platform.touch) {
+					window.addEventListener('touchend', onTouchEnd);
```
webOS101: This is never cleaned up.  I'll take a stab at fixing it.

packages/spotlight/src/spotlight.js
```diff
@@ -504,6 +504,13 @@ const Spotlight = (function () {
 		}
 	}
 
+	function onTouchEnd (evt) {
+		const current = getCurrent();
+		if (current && !current.contains(evt.target)) {
+			current.blur();
```
webOS101: What are the chances this will affect code that manually handles spotlight when dismissing?  If we blur here, will we prevent code like that in `ExpandableInput` or `ExpandableItem` that tries to figure out how to set the focus?  I'll try to do some more testing in that area but if you have some thoughts it'd be good to hear.

ryanjduffy: Most of that code already deals with blur events as triggers so we should be okay. This was tested against Input and ExpandableInput for that reason.

PR #2234, opened 2019-04-26
PLAT-76988: Add noScrollByWheel to prevent scrolling on wheel events

0x64:
### Checklist

If a scroller/list is inside another scroller/list, wheel input invokes scrolling on both scrollers.

### Resolution
- Fixed a scroller/list to consume the wheel input when it scrolls on the input. If it does not scroll since it already reached to the edge on the direction, then the outer scroller/list can scroll by the wheel input.
- Added a prop `noScrollHorizontalByWheel` to disable horizontal scrolling by wheel which may be useful to disable the inner scroller for UX reasons

### Links
PLAT-76988

### DIFF
--- a/packages/moonstone/Scrollable/ScrollableNative.js
+++ b/packages/moonstone/Scrollable/ScrollableNative.js
@@ -142,7 +142,7 @@ class ScrollableBaseNative extends Component {
 		 * @default false
 		 * @private
 		 */
-		noScrollHorizontalByWheel: PropTypes.bool,
+		noScrollByWheel: PropTypes.bool,
 
 		/**
 		 * Specifies overscroll effects shows on which type of inputs.
@@ -206,7 +206,7 @@ class ScrollableBaseNative extends Component {
 		'data-spotlight-container-disabled': false,
 		animate: false,
 		focusableScrollbar: false,
-		noScrollHorizontalByWheel: false,
+		noScrollByWheel: false,
 		overscrollEffectOn: {
 			arrowKey: false,
 			drag: false,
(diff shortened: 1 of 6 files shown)


packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -135,6 +135,15 @@ class ScrollableBaseNative extends Component {
...
 		focusableScrollbar: PropTypes.bool,
 
+		/**
+		 * Prevents scroll by wheeling on the list or the scroller.
+		 *
+		 * @type {Boolean}
+		 * @default false
+		 * @private
+		 */
+		noScrollHorizontalByWheel: PropTypes.bool,
```
seunghoh: What about `noHorizontalScrollByWheel` or `noScrollHorizontallyByWheel`? :)

0x64: It looks better. :) But I want a much better name!

seunghoh: We need @webOS101 's opinion!

ryanjduffy:
Boolean props are terse but can be inflexible if we want to expand. For example, might we/authors want to disable scroll by wheel altogether? By axis?

0x64:
@ryanjduffy 
I wanted to make a prop to disable scroll by wheel for both axis but I found that there is a problem for a native vertical list/scroller. To disable a native vertical scroller, the most proper solution is to call `preventDefault` on `wheel` event. And it will also block scrolling of the outer scroller. This is not a solution for the requirement: blocking only the inner scroller. That is why I added the inflexible prop and why I made it _private_.

seunghoh:
@ryanjduffy is right. We also discussed what you mentioned but wasn't able to find a general solution. I proposed Seungcheon to make narrow on the requirement though. :(

ryanjduffy:
I see your problem with native scrollers. I haven't had a chance to run this yet but I'd like to spend a little time on it tomorrow. Sorry for not getting you feedback today as I had hoped.

ryanjduffy:
Couple considerations for you. First, I don't think we can call the prop private if it's expected to be used by an application. If we need the capability to suppress scroll by wheel, let's own it and support it.

Second, I'm curious if if we really need the "horizontal" part of the prop. Seems that it might make more sense to disable scroll by wheel completely via prop and rely on the direction of the scroller to dictate what that means. In that light, `noScrollByWheel` might be an option.

There might be a way to combine some of the "scroll by" options (drag, wheel, scrollbars) into a single API but we'd need to think more on that so no need to solve that in this PR.

0x64:
I updated the prop to `noScrollByWheel` and it blocks scrolling if the prop value is true.

Notes & Questions
1. For nested vertical scrollers, there is no way to block only the inner vertical list.
Should I add this limitation to changelog?
1. Is it okay to make this prop public _now_?
Or, should I wait to update _scroll-by-options_ (including scrollByDrag) more elegant?

PR #2235, opened 2019-04-26
PLAT-78104: Fix Input refocusing on touch on iOS

ryanjduffy:
### Checklist

Regression from #2187 in which the pointer lock thought it was capturing focus when tapping off of an input but wasn't properly cleaned up because iOS doesn't issue a click event if the target of the tap isn't focusable.

### Resolution
Add a `touchend` handler to ensure we clean up properly.


ryanjduffy:
I'm seeing the same. Thanks for the report!

ryanjduffy:
I think it's fixed up. I added `preventDefault` to the `handleTap` handler. For `click` events, it shouldn't have any effect (afaik) but for `touchend` it will prevent dispatching a click event in Chrome which "reopens" the ExpandableInput in the same update cycle.

PR #2237, opened 2019-04-29
PLAT-73523: Change Tooltip appearance to match new designs

Djspaceg:
### Checklist

New designs for 3.0.0

### Resolution
Tooltips now have angled tails, a border, and stand-off a bit from their assigned component.

### References
This PR depends on the merging of #2214.


Djspaceg:
Thanks for noticing those issues @viodragon2! I got the first two fixed, and while fixing the second, I noticed 2 other little glitches that I was also able to resolve.
* z-index layering was incorrect for the arrow if the label was offset
* relative positioned middle aligned left/right arrows didn't have their appearance mirrored correctly and looked ugly. The positioning/layout for these cases actually got even simpler too, which is great.

As for the color issue, I'm not sure about that. The design doc that I have does indeed show the #ccc color for both the tooltip text and the button text, which this PR replicates. This relationship is further codified by the following line: `@moon-tooltip-text-color: @moon-button-text-color;` so if the button rule changes in a future PR, the tooltip will also update. I believe we can move forward on this as it stands, and just make a correction later if necessary. For reference, the design doc I'm referencing was last updated on `May 2, 2019 2:45 AM` (post-workshop date, and does include the tooltip background color being darker than the button background color)

haileyR:
Should we make tooltip to wrap text when the width is not enough?

image

Djspaceg:
@haileyR, this is a nice suggestion. While I'd like to support this, and the Tooltip does _technically_ support this, line-wrapping not part of the Moonstone design at this time.

PR #2238, opened 2019-04-29
PLAT-78054: Updated the appearance of FormCheckboxItem (and friends)

Djspaceg:
### Checklist

New design for FormCheckbox and FormCheckboxItem

### Resolution
Who wants to look at crusty old designs? Not me, that's for sure. These components were updated to the latest look. :sogood:
We finally have square checkboxes. WOO!

### DIFF
--- a/packages/moonstone/styles/colors.less
+++ b/packages/moonstone/styles/colors.less
@@ -105,8 +105,8 @@
 
 // FormCheckbox
 // ---------------------------------------
-@moon-formcheckbox-bg-color: @moon-button-bg-color;
-@moon-formcheckbox-text-color: @moon-button-text-color;
+@moon-formcheckbox-bg-color: fade(@moon-white, 22%);
+@moon-formcheckbox-text-color: @moon-white;
 @moon-formcheckbox-border-color: fade(@moon-white, 15%);
 
 // GridListImageItem
--- a/packages/moonstone/styles/variables.less
+++ b/packages/moonstone/styles/variables.less
@@ -224,11 +224,12 @@
 
 // FormCheckbox
 // ---------------------------------------
-@moon-formcheckbox-size: (@moon-item-height * 0.8);	// Relative to `Item` height
-@moon-formcheckbox-size-large: (@moon-item-height-large * 0.8);	// Relative to `Item` height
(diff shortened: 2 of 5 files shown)


packages/moonstone/styles/colors.less
```diff
@@ -105,8 +105,8 @@
 
 // FormCheckbox
 // ---------------------------------------
-@moon-formcheckbox-bg-color: @moon-button-bg-color;
-@moon-formcheckbox-text-color: @moon-button-text-color;
+@moon-formcheckbox-bg-color: fade(@moon-white, 22%);
```
viodragon2: needs documentation to verify

packages/moonstone/styles/variables.less
```diff
@@ -224,11 +224,12 @@
...
 // FormCheckbox
 // ---------------------------------------
-@moon-formcheckbox-size: (@moon-item-height * 0.8);	// Relative to `Item` height
-@moon-formcheckbox-size-large: (@moon-item-height-large * 0.8);	// Relative to `Item` height
+@moon-formcheckbox-size: (@moon-item-height * 0.55);	// Relative to `Item` height
+@moon-formcheckbox-size-large: (@moon-item-height-large * 0.55);	// Relative to `Item` height
 @moon-formcheckbox-check-icon-size: 72px;
 @moon-formcheckbox-check-icon-size-large: 84px;
-@moon-formcheckbox-border-size: 3px;
+@moon-formcheckbox-border-size: 2px;
```
viodragon2: why 2px? Is this to match radio item's border?

Djspaceg: That's a GUI decision. It was preferable to GUI to have this be specifically 2px rather than 3 or to match radio (Radio is a circle, while form FormCheckbox is (now) a square, so the perceived thickness is more relevant than the actual.)

packages/moonstone/styles/colors.less
Djspaceg: This was _"funny"_ to me too, –that it's not just a round number–, but IIRC, 22% opacity was the level that visually matched the same color tone of GUI's existing graphic design.

PR #2239, opened 2019-04-29
PLAT-77054: Change components to allow disabled components to be focusable

webOS101:

### Checklist

Disabled spottable elements should be focusable.

### Resolution
Update `Spottable` to allow focus onto disabled items.  Disabled items themselves are still non-interactive but can have spotlight.  Updated focus CSS for many items as disabled + focused was not used before.  Also, I removed the `isItemDisabled` property from VL and VGL.

### Additional Considerations
Design of some components may still be in flux so it is possible these designs will change further.  We don't have any specific guidelines for how disabled + focused should look, but in most cases it will be an opacity applied to the focused element.

I did not add a PR for deprecating the `isItemDisabled` property.

### Links
PLAT-77054

### DIFF
--- a/packages/sampler/stories/default/MediaOverlay.js
+++ b/packages/sampler/stories/default/MediaOverlay.js
@@ -3,7 +3,7 @@ import React from 'react';
 import {storiesOf} from '@storybook/react';
 import {withInfo} from '@storybook/addon-info';
 
-import {select} from '../../src/enact-knobs';
+import {boolean, select} from '../../src/enact-knobs';
 import {mergeComponentMetadata} from '../../src/utils';
 
 const prop = {
@@ -77,6 +77,7 @@ storiesOf('Moonstone', module)
 			const placeholder = prop.placeholder[placeholderName];
 			return (
 				<MediaOverlay
+					disabled={boolean('disabled', Config)}
 					imageOverlay={imageSource}
 					placeholder={placeholder}
 					text={select('text', prop.text, Config, prop.text[0])}
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -46,7 +46,7 @@ const isKeyboardAccessible = (node) => {
 	);
 };
 
-const isSpottable = (props) => !props.disabled && !props.spotlightDisabled;
+const isSpottable = (props) => !props.spotlightDisabled;
 
 // Last instance of spottable to be focused
 let lastSelectTarget = null;
@@ -359,7 +359,7 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 
 		handleFocus = (ev) => {
-			if (this.props.disabled || this.props.spotlightDisabled) {
+			if (this.props.spotlightDisabled) {
 				this.shouldPreventBlur = true;
 				ev.target.blur();
 				this.shouldPreventBlur = false;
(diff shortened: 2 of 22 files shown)


packages/spotlight/Spottable/Spottable.js
```diff
@@ -393,7 +393,7 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 
 		render () {
 			const {disabled, spotlightId, spotlightDisabled, ...rest} = this.props;
-			this.focusedWhenDisabled = this.isFocused && (disabled || spotlightDisabled);
+			this.focusedWhenDisabled = this.isFocused && spotlightDisabled;
```
viodragon2: Maybe we could unwind this and remove `forceUpdate` that happens in blur event.

webOS101: After re-reviewing, I realized this also handles the case where a component becomes `spotlightDisabled` when focused, which is not changed with the PR.  I think it should remain.

ryanjduffy:
Few visual notes from sampler:
* Input - Field is white with full opacity
* ExpandableList/DatePicker/TimePicker - LabeledItem doesn't change
* MediaOverlay - Focus ring appears to be full opacity

Djspaceg:
@ryanjduffy would you mind re-verifying that "ExpandableList/DatePicker/TimePicker - LabeledItem doesn't change"? I just checked again on my machine and it appears to be behaving correctly with yesterday's set of commits.

ryanjduffy:
This is what I see with the latest. It shows the focused+disabled state okay but disabled+unfocused is the same as enabled+unfocused.

image

Djspaceg:
Ah! Focused+notDisabled. Thanks for clarifying! I'll get that sorted out too.

ryanjduffy:
One more visual note: FormCheckbox's text becomes white when focused but it doesn't change when disabled and focused which is inconsistent with other components that still change color but are "muted" by the opacity change.

Djspaceg:
The above (previous 3) concerns were addressed, and a new knob was added to the MediaOverlay story.

I don't have a good answer for the FormCheckboxItem scenario. The white focus text is a recent change, and we've not received complete designs on exactly how those colors should be handled in every situation (like when it's disabled).

Several visual changes to FormCheckbox and Checkbox may be changing back to the 2.x style, which may affect this issue.

What do you want to do?

ryanjduffy:
My opinion is to be consistent in the absence of specific designs but I will support whatever decision you make.

ryanjduffy:
Also, please update the description of the PR to describe the changes.

packages/sampler/stories/default/MediaOverlay.js
```diff
@@ -77,6 +77,7 @@ storiesOf('Moonstone', module)
 			const placeholder = prop.placeholder[placeholderName];
 			return (
 				<MediaOverlay
+					disabled={boolean('disabled', Config)}
```
webOS101: I'm not sure `MediaOverlay` is _supposed_ to be disable-able.  So, I'm not sure this makes sense.

Djspaceg: 🤷‍♂ I just did what I was told. I'm not familiar with how MediaOverlay is used.

webOS101:
I was not able to observe the `VirtualList` disabled 5-way navigation issue.  It seems like it's up to the app to pass in a `isItemDisabled()` function so it's possible the app you were testing supplied this.  I updated the joined pickers and the PR description.

ryanjduffy:
I should have been more clear. You're right that the app must provide that function but I don't think the function should be supported since its purpose was to assist in focus management.

webOS101:
Hmm...  I guess we could still supply it for a `spotlightDisabled` sort of thing...  but...  I can remove it.

webOS101:
Removed.

PR #2243, opened 2019-04-29
PLAT-77089: Add touch move handlers for VideoPlayer

viodragon2:
### Checklist

`moonstone/VideoPlayer` doesn't support touch moves properly.
1. The media controls hiding timeout does not get reset when a user drags the knob and would eventually be closed. 
2. When `preview` is available, the knob would not follow the dragging position

Both of these are due to lack of touch move event handling.

### Resolution
Add `touchmove` event handlers that match the `mousemove` events.

### DIFF
--- a/packages/moonstone/VideoPlayer/MediaSliderDecorator.js
+++ b/packages/moonstone/VideoPlayer/MediaSliderDecorator.js
@@ -1,10 +1,10 @@
 import {adaptEvent, call, handle, forKey, forward, oneOf, preventDefault, returnsTrue, stopImmediate} from '@enact/core/handle';
 import hoc from '@enact/core/hoc';
+import platform from '@enact/core/platform';
 import {calcProportion} from '@enact/ui/Slider/utils';
 import clamp from 'ramda/src/clamp';
 import PropTypes from 'prop-types';
 import React from 'react';
-import platform from '@enact/core/platform';
 
 // decrements the MediaKnob position if we're tracking
 const decrement = (state) => {
@@ -196,6 +196,10 @@ const MediaSliderDecorator = hoc((config, Wrapped) => {	// eslint-disable-line n
 
 			delete rest.onKnobMove;
 
+			if (platform.touch) {
+				rest.onTouchMove = this.handleTouchMove;
+			}
+
 			return (
 				<Wrapped
 					{...rest}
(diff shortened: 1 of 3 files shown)


packages/moonstone/VideoPlayer/MediaSliderDecorator.js
```diff
@@ -198,6 +207,7 @@ const MediaSliderDecorator = hoc((config, Wrapped) => {	// eslint-disable-line n
 					onMouseOver={this.handleMouseOver}
 					onMouseOut={this.handleMouseOut}
 					onMouseMove={this.handleMouseMove}
+					onTouchMove={platform.touch ? this.handleTouchMove : void 0}
```
ryanjduffy: I'm not sure this ternary is really buying us much particularly for how limited VP is used. I think I'd slightly prefer either always attaching the handler (which would require always binding it at construction) or adding an `if` block and appending it `rest`. That said, not a blocker so just for consideration.

packages/moonstone/VideoPlayer/VideoPlayer.js
```diff
@@ -688,6 +688,9 @@ const VideoPlayerBase = class extends React.Component {
 
 	componentDidMount () {
 		on('mousemove', this.activityDetected);
+		if (platform.touch) {
+			on('touchmove', this.activityDetected);
```
webOS101: I wonder if touching and releasing in the same spot over and over would be activity?  :D

viodragon2: I believe the `RootContainer`'s `onClick` event handler will capture all tap events so it will be redundant to handle `touchstart`

webOS101: Yes, I realized that.  :P  It just seemed funny to me.

packages/moonstone/VideoPlayer/MediaSliderDecorator.js
```diff
@@ -174,6 +178,11 @@ const MediaSliderDecorator = hoc((config, Wrapped) => {	// eslint-disable-line n
 			this.move(ev.clientX);
 		}
 
+		handleTouchMove (ev) {
+			// ignores multi touch
+			this.move(ev.touches[0].clientX);
```
webOS101: Should we round these numbers?  As opposed to mouse movements these appear to contain floats.

viodragon2: I'm not sure if it would be useful to round numbers. The proportion calculation will end up with floating points most of the time even with the rounded numbers. (https://github.com/enactjs/enact/blob/46677e3b64a25bac680e62227d9d7065d85db523/packages/moonstone/VideoPlayer/MediaSliderDecorator.js#L135)

webOS101: I figured you might say that.  I suspect we won't see a lot of switching between touch and mouse in the same session so it's not really a big issue for optimization.

PR #2245, opened 2019-04-30
PLAT-78043: Updated Checkbox style a little bit

Djspaceg:
### Checklist

Checkbox and CheckboxItem are round and those shenanigans just don't fly anymore! #ComeOnNowHey

### Resolution
Updated the styles to match the latest designs, squaring the corners and tuning the colors for better contrast. :sogood:

### DIFF
--- a/packages/moonstone/styles/variables.less
+++ b/packages/moonstone/styles/variables.less
@@ -235,13 +235,13 @@
 @moon-formcheckbox-check-icon-size: 72px;
 @moon-formcheckbox-check-icon-size-large: 84px;
 @moon-formcheckbox-border-size: 2px;
-@moon-formcheckbox-border-radius: 3px;
+@moon-formcheckbox-border-radius: 960px; // Generic, sufficiently large value that converts to a nice amount of rem
 
 // Checkbox
 // ---------------------------------------
 @moon-checkbox-size: 27px;
 @moon-checkbox-size-large: 33px;
-@moon-checkbox-border-radius: @moon-button-border-radius;
+@moon-checkbox-border-radius: 960px; // Generic, sufficiently large value that converts to a nice amount of rem
 
 // DaySelector
 // ---------------------------------------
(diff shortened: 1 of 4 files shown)


packages/moonstone/styles/variables.less
```diff
@@ -235,12 +235,13 @@
...
 @moon-formcheckbox-check-icon-size-large: 84px;
 @moon-formcheckbox-border-size: 2px;
-@moon-formcheckbox-border-radius: 3px;
+@moon-formcheckbox-border-radius: 960px; // Generic, sufficiently large value that converts to a nice amount of rem
 
 // Checkbox
 // ---------------------------------------
-@moon-checkbox-size: 30px;
-@moon-checkbox-size-large: 36px;
+@moon-checkbox-size: 27px;
```
viodragon2: Should these size values update if we're keeping the round shape?

Djspaceg: I had considered that myself. It seemed (visually) fine when I looked at it after the switch back to round.

My interpretation of the GUI<->UX conversation response was that squares had visually implied something that they didn't want to possibly visually confuse, rather than the design not being able to be updated at all. I'm content with leaving the size as-is until we hear otherwise. What do you think?

viodragon2: The only concern would be the parity with other apps (e.g. QML, web apps, etc.). Given that 3px width is a subtle difference, I'm okay to move forward with the change. I'll suggest to follow up with GUI to make sure we match the guideline, though.

Djspaceg: Good point. Will do followup.

ryanjduffy:
The checkmark seems tight, primarily for FormCheckbox, compared to the spec. Are you seeing the same?

image

ryanjduffy:
Also, should `DaySelector` return to circular checkboxes as well?

Djspaceg:
Yes, this is expected, as is DaySelector. The tight tolerance is why @viodragon2, above, suggested verifying with GUI, which I signed up to do. The direction we were given during the Tuesday CC was to change _exclucively_ the shape, not any other sizes/colors.

PR #2250, opened 2019-04-30
PLAT-77085: Remove data-spotlight-* props from Scrollable

viodragon2:
### Checklist

Turning on `data-spotlight-container-disabled` in qa-sampler still allows touch devices to focus elements. The problem is two-fold. 1) Touch hold will focus the element on touch devices and 2) `SpotlightContainerDecorator` handles the focus event and prevents by checking `spotlightDisabled` prop, not `data-spotlight-container-disabled` attribute.

### Resolution
- Switch `data-spotlight-container-disabled` to `spotlightDisabled` in the sampler

Technically this would be enough to resolve the immediate issue, but also cleaned up the structure so that it respects `SpotlightContainerDecorator` props properly.

- Removed `data-spotlight-*` props from `Scrollable`

### Additional Considerations

- When digging into as to why it was structured this way, I could only find this [comment](https://github.com/enactjs/enact/pull/1985#issuecomment-430105171). 
- This change requires related tests to be updated.
- As a fallback option, the simpler way to fix this issue may be to prevent focus event to propagate by checking the attribute instead of the prop. See  https://github.com/enactjs/enact/compare/develop...feature/PLAT-77085-stephen-1

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -76,7 +76,7 @@ const navigableFilter = (elem) => {
 	}
 };
 
-const configureSpotlightContainer = ({'data-spotlight-id': spotlightId, focusableScrollbar}) => {
+const configureSpotlightContainer = ({spotlightId, focusableScrollbar}) => {
 	Spotlight.set(spotlightId, {
 		navigableFilter: focusableScrollbar ? null : navigableFilter
 	});
@@ -200,7 +200,7 @@ class ScrollableBase extends Component {
 		 *
 		 * @type {Boolean}
 		 * @default false
-		 * @private
+		 * @public
 		 */
 		spotlightDisabled: PropTypes.bool,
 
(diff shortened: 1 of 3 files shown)


viodragon2:
Good catch! Both lines were supposed to be updated.

Regarding making public, I don't have any problem since `moonstone/Scrollable` is a private module it really doesn't matter to the public anyways. Hence no CHANGELOG.

However, `moonstone/VirtualList` which is public and composes `moonstone/Scrollable` has `spotlightId` as a private prop and no mention of `spotlightDisabled`. (https://github.com/enactjs/enact/blob/346d3344375938d6c090675ce29a67b4f8cf8673/packages/moonstone/VirtualList/VirtualListBase.js#L176)

It probably makes sense to have both props documented in VirtualList, but I'm not 100% sure. It seems a bit weird to update data-* attribute directly like we do in https://github.com/enactjs/enact/blob/346d3344375938d6c090675ce29a67b4f8cf8673/packages/moonstone/VirtualList/VirtualListBase.js#L283 and should use props instead. But it could have been done this way to bypass `setState` which is costly. But for the scope of this PR, I'd suggest creating a follow-up ticket.

PR #2253, opened 2019-05-01
PLAT-76359: Add Dropdown

teckliew:
Using mostly ContextualPopup for Dropdown.

### Additional Considerations
Added `iconPosition` prop to Button for the positioning of the icon.
Added `noArrow` config for ContextualPopup

### Links
PLAT-76359


packages/moonstone/Button/Button.js
```diff
@@ -50,7 +51,7 @@ const ButtonBase = kind({
 		 * * `'lightTranslucent'`, and
 		 * * `'transparent'`.
 		 *
-		 * @type {String}
+		 * @type {translucent'|'lightTranslucent'|'transparent'}
```
sugardave: Minor typo here, should be `@type {'translucent'|'lightTranslucent'|'transparent'}`

packages/moonstone/DropDown/DropDown.js
```diff
@@ -0,0 +1,301 @@
+/**
+ * Moonstone styled DropDown components
+ *
+ * @example
```
sugardave: The example is throwing `ReferenceError: DropDown is not defined` in the docs

packages/moonstone/DropDown/tests/DropDown-specs.js
```diff
@@ -0,0 +1,71 @@
+import React from 'react';
+import {mount, shallow} from 'enzyme';
+import {DropDown, DropDownBase} from '../DropDown';
+
+const title = 'Dropdown select';
+const children = ['option1', 'option2', 'option3'];
+
+describe('DropDownBase', () => {
```
sugardave: Just curious why you're testing the base component separately instead of running them all on the stateful one?

packages/moonstone/DropDown/DropDown.js
```diff
@@ -0,0 +1,301 @@
...
+		css
+	},
+
+	render: ({children, onSelect, selected, ...rest}) => {
+		return (
+			<div {...rest}>
+				<Group
+					childComponent={Item}
+					className={css.group}
+					component={Scroller}
```
sugardave: Is anything gained using `Scroller`?  When I run the sample, there isn't really any scrolling to do.  If I resize the sample window small enough and re-run it, the items will render past the sample viewport and you can't scroll them.

Is the expanded dropdown area supposed to be a fixed size or is it supposed to grow depending on its children?

teckliew: The dropdown height depends on its items up to 5; thus the 5 item limit max. So any more than 5 items, there'll be a Scroller.

packages/moonstone/Button/Button.js
Djspaceg: Other examples in the framework show these being in `()`. Should this one also be? (I'm not really familiar with this new format yet.)
```
@type {('translucent'|'lightTranslucent'|'transparent')}
```

webOS101: I've never tried without and the examples I've seen (and used) all have the parens.

packages/moonstone/ContextualPopupDecorator/ContextualPopup.js
```diff
@@ -114,7 +114,7 @@ const ContextualPopupBase = kind({
 		 *
 		 * Can be one of: `'up'`, `'down'`, `'left'`, or `'right'`.
 		 *
-		 * @type {String}
+		 * @type {'up'|'down'|'left'|'right'}
```
Djspaceg: Parenthesis around these.

packages/moonstone/DropDown/DropDown.js
Djspaceg: Should we consider having this be conditional to not add a `Scroller` (and its cost) if the list length is 5 or less?

packages/moonstone/DropDown/DropDown.js
teckliew: Made a branch in docs `feature/ui-update` and added the import for `DropDown`. Just need to update the `sample-runner` and run `npm run make-runner`.

packages/moonstone/Button/Button.js
teckliew: It still works, but fixed the typo and added the parens.

packages/moonstone/styles/variables.less
```diff
@@ -176,6 +184,13 @@
 @moon-divider-border-width: 3px;
 @moon-divider-height: 39px;
 
+// DropDown
+// ---------------------------------------
+@moon-dropdown-border-radius: 3px;
```
Djspaceg: Let's base this value on the Button border-radius of `@moon-button-border-radius`, so their relationship is established from the start.
And let's also do that with the next line: `@moon-dropdown-width: @moon-button-max-width`

packages/sampler/stories/default/DropDown.js
```diff
@@ -0,0 +1,34 @@
...
+import {withInfo} from '@storybook/addon-info';
+
+import {boolean, select, text} from '../../src/enact-knobs';
+import {mergeComponentMetadata} from '../../src/utils';
+
+const Config = mergeComponentMetadata('DropDown', DropDown, DropDownBase);
+DropDown.displayName = 'DropDown';
+
+
+
```
Djspaceg: Nit: Ooo :sogood: so much whitespace. (Let's drop this down to 1 line)

packages/sampler/stories/default/DropDown.js
```diff
@@ -0,0 +1,34 @@
...
+import React from 'react';
+import {storiesOf} from '@storybook/react';
+import {action} from '@storybook/addon-actions';
+import {withInfo} from '@storybook/addon-info';
+
+import {boolean, select, text} from '../../src/enact-knobs';
+import {mergeComponentMetadata} from '../../src/utils';
+
+const Config = mergeComponentMetadata('DropDown', DropDown, DropDownBase);
+DropDown.displayName = 'DropDown';
```
Djspaceg: I believe this works better if it is _before_ the `Config` line. We should just update every story all at once and be done with it. #soemothertime

Djspaceg:
Looks great @teckliew! I added a new knob to control the number of items in the dropdown. It revealed that we throw a big warning when the option list is empty. Let's add in some failure handling for having no options (like just swallow the click and don't open the contextualpopup). In practice, maybe the data that populates a dropdown isn't ready at page-load, and it comes in a moment later, so no options does sorta make sense as something to allow for.

I'll sign off after that. Cheers!

teckliew:
I went with preventing from opening and disabling `Dropdown` when there are no children.

Djspaceg:
:nice: I even like the visual feedback that the dropdown is disabled when the children/list is empty.

PR #2258, opened 2019-05-06
PLAT-76082: Add SharedStateDecorator

ryanjduffy:
### Checklist

We have requirements to "remember" the scroll position when navigating forward (positive `index` change) in a Panels stack so it can be restored when navigating back (negative `index` change) to it.

### Resolution
* Add a new internal HOC, `SharedStateDecorator`, which maintains a private key/value data store and makes it available via a `SharedState` context object
* Add the `SharedStateDecorator` to `Panels` and `Panel` thereby establishing localized data stores for their respective subtree
* Add a handler to `Scroller` to notify its upstream `SharedState` of scroll position changes

The process is as follows:
1. `Panels` creates a new data store (via `SharedStateDecorator`) to track the state of each of its panel instances
2. `Panel` creates its own data store keyed by the `data-index` value passed by `Panels` (via `Viewport`)
3. `Scroller` uses the data store created and passed by `Panel` to store the current scroll position in an `onScroll` handler
4. When navigating forward, new keys are appended to the `Panels` data store for the new Panels. When navigating backward, all keys greater than the current index are deleted thereby erasing their shared state.
5. When returning to a prior `Panel`, `SharedStateDecorator` retrieves the shared state from its parent (`Viewport`). If that `Panel` contains a `Scroller`, the `Scroller` retrieves the scroll position in `componentDidMount` and imperatively scrolls to the cached position.


seunghoh:
@ryanjduffy Restore scroll logic is in the `scroller` This means VirutalList in the Panel won't support. Is this intentional?

ryanjduffy:
I started with that since VirtualList has its placeholder to help restore scroll position. If we move ahead with this, we'll need to decide how to reconcile those two approaches.

seunghoh:
Overall, I like your approach 😀
One minor concern is `sharedStateId = 'scroller'`. Channel Edit has 3 List components, but we can guide app developer to add prop explicitly. 
It looks like logic in `scroller` can move to `scrollerable` + scroll to index and remove placeholder ?

ryanjduffy:
@germboy, @viodragon2, @Djspaceg, @webOS101 - I've pushed a couple more commits to this and think it's getting close. I've moved the position caching logic into Scrollable, converted to use `id` instead of a new prop, and added a `noSharedState` prop to `SharedStateDecorator` to allow some control.

Outstanding items:
* Determine if we should document `noSharedState` as a public prop on Panels/Panel
* Document `id` prop on Scroller/Panels/Panel
* Document `SharedStateDecorator`
* Test with Virtual(Grid)List(Native) incl async data loading scenarios
* Create a sample for testing
* Unit tests

packages/moonstone/Panels/Panels.js
```diff
@@ -184,22 +208,24 @@ const PanelsBase = kind({
...
 
 		return (
-			<div {...rest}>
+			<div {...rest} id={id}>
 				{applicationCloseButton}
 				<Viewport
 					arranger={arranger}
 					childProps={childProps}
 					generateId={generateId}
+					id={`viewport-${id}`}
```
ryanjduffy: Fix this

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -73,6 +74,8 @@ const configureSpotlightContainer = ({'data-spotlight-id': spotlightId, focusabl
 	});
 };
 
+const isValidId = (id) => id !== '';
```
ryanjduffy: Probably don't need this

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -276,10 +287,14 @@ const VirtualListBaseFactory = (type) => {
 			this.setContainerSize();
 		}
 
-		componentDidUpdate (prevProps) {
+		componentDidUpdate (prevProps, prevState) {
 			// TODO: remove `this.hasDataSizeChanged` and fix ui/Scrollable*
 			this.hasDataSizeChanged = (prevProps.dataSize !== this.props.dataSize);
 
+			if (prevState.firstIndex !== this.state.firstIndex && this.props.onUpdateItems) {
```
ryanjduffy: remove onUpdateItems condition

packages/moonstone/internal/SharedStateDecorator/tests/SharedStateDecorator-specs.js
```diff
@@ -0,0 +1,307 @@
...
+			</Component>
+		);
+
+		const expected = {
+			// "data" is grouped by id so the top level key is the current id
+			outer: {
+				// this is set by value.set('key', 'value');
+				key: 'value',
+				// when a descendant is added, it pushes its data upstream stored by its id
+				inner: {
```
ryanjduffy: Might be a problem if `inner` were renamed to an existing `key` value. Need to consider more.

packages/moonstone/CHANGELOG.md
```diff
@@ -16,6 +17,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 - Fonts to correctly use the new font files and updated the international font name from "Moonstone LG Display" to "Moonstone Global"
 - `moonstone/Dropdown` `children` propType so it supports the same format as `ui/Group` (an array of strings or an array of objects with props)
+- `moonstone/FormCheckbox`, `moonstone/Input`, `moonstone/ProgressBar`, `moonstone/RadioItem`, `moonstone/SwitchItem`, and `moonstone/Tooltip` light skin colors.
```
webOS101: This looks like a stray

ryanjduffy: it was in the wrong spot and i moved it when resolving the merge conflict

PR #2260, opened 2019-05-07
PLAT-78026: Adjust Scrollable container height

germboy:
### Checklist

The `Scrollable` component's container was using an incorrect value which could result in the horizontal scroll bar being rendered out of the viewport of the scroller.

### Resolution
We need to remove the specified `100%` height value and let the flex layout handle the height instead.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -14,7 +14,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/Scrollable` to adjust its container height
+- `ui/Scroller`, `ui/VirtualList`, and `ui/VirtualGridList` to size properly
 
 ## [2.5.2] - 2019-04-23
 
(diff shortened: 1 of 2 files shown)


packages/ui/Scrollable/Scrollable.module.less
```diff
@@ -16,7 +16,7 @@
 	.container {
 		display: flex;
 		width: 100%;
-		height: 100%;
+		height: 0;
```
webOS101: Is there any reason to believe we won't have the same issue in the `width`?

webOS101: Guess it doesn't work without `width: 100%`

packages/ui/CHANGELOG.md
```diff
@@ -12,6 +12,10 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 - `ui/Button`, `ui/Icon`, `ui/IconButton`, and `ui/LabeledIcon` prop `size`
 
+### Fixed
+
+- `ui/Scrollable` to adjust its container height
```
webOS101: Will update this to be for non-private component.

PR #2261, opened 2019-05-07
 ENYO-5949: Header hides components block if there are none

Djspaceg:
### Checklist

Extra padding in the header when there are no header components

### Resolution
Don't render the headerComponents node if there are no children in it


Djspaceg:
Based on #2252 branch.

PR #2262, opened 2019-05-08
PLAT-65757: Changed lists and scroller to scroll by page on page up/down keys

0x64:
### Checklist

This draft PR is a proof of concept for a new UX of scrolling by page up/down keys.
When a page up/down key pressed,
1. Blur the current spot if exist and block focusing on items
1. Scroll a list or scroller with animation for a certain amount of distance
1. Stop blocking focusing after stop scrolling
1. Restore the focus on the same position as before pressing page up/down _only when_ the current mode is 5-way mode

### Resolution
Re-implemented related logic in `Scrollable`s due to a big change.

### Additional Considerations
In this draft PR, native lists/scroller will be scrolled by web runtime on page up/down key press.

### Links
PLAT-78242
PLAT-65757


seunghoh:
LGTM

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -324,6 +324,7 @@ class ScrollableBase extends Component {
 	lastScrollPositionOnFocus = null
 	indexToFocus = null
 	nodeToFocus = null
+	pointToFocus = null
```
ryanjduffy: I haven't had time to review this ticket in detail yet but wanted to ask about this specifically. Can you share why you chose to use a non-state member here rather than using `setState`?

0x64: @ryanjduffy 
Related methods to store and restore that info are invoked by key input and already out of life cycle. If I make it a _state_, I have to check it is safe to _scroll_ when `setState` is invoked by key press. So, simply I made it a variable which is managed by imperative calls.

packages/spotlight/src/spotlight.js
```diff
@@ -747,6 +747,24 @@ const Spotlight = (function () {
...
 
+		/**
+		 * Focuses the specified position.
+		 *
+		 * @param {Object} point Position info consists of `x` and `y` values
+		 * @param {String} direction Direction to find a spottable target when no spottable target
+		 *	is found at the specified position, one of `'left'`, `'right'`, `'up'` or `'down'`
+		 * @private
+		 */
+		focusFromPoint: function (point, direction) {
```
ryanjduffy: @germboy - What do you think of this as pubic API?

germboy: @ryanjduffy This appears to work similarly to `Spotlight.move(direction, selector)` - just providing a pointer-based type of alternative. While I can consider this as a contrasting alternative, I'm a little hesitant.

When using 5way, we're sometimes expected to direct focus somewhere other than the "natural" element in the given direction - however, in pointer mode, I don't feel we'd ever expect to fake or force focus away in any direction other than a true pointer position. It feels like this may be solving a component problem at the spotlight-level. I don't see this being beneficial for many other use-cases.

It seems to me that the component could be aware of its contained `Spottable` components, then it could determine which component existed at a certain location, then use that component's selector in a call to `Spotlight.move(direction, selector)`.

0x64: @germboy 
The current UX requirement is to restore focus on the stored position _without moving at all_ for both 5way mode and pointer mode. The second parameter `direction` is a kind of fallback to cover cases that there is no focusable element in the stored position. (So, `Spotlight.move` is useful only for fallback cases.) And this rule is not only for lists but also for scrollers, so it looks better to get some help of Spotlight.

0x64: @ryanjduffy @germboy 
I'm just wondering `clear` and `unmount` are included as private intentionally.

germboy: @0x64 @ryanjduffy  The way this functions seems ok, I'm just not sure this is the right place for it. All methods here are accessible via `Spotlight` and should probably add value to more use-cases. This seems to solve a very specific platform/component-level use-case, and would likely be a better fit if relocated to `Scrollable` (if possible) or perhaps a spotlight submodule (`spotlight/src/pointer`?), since it appears we may need to access other internal spotlight methods.

As for the 2 "private" methods pointed out:
- `clear` should not be used by developers, and I believe is only used when calling `Spotlight.terminate()`. It appears to be a convenience method that is left over and could be removed.
- `unmount` simply appears to not have inline docs associated. It should be listed as public, since developers are expected to use it when working with containers that they expect to be re-mounted. This restores the previous container data in those cases.

ryanjduffy: I think we left `unmount` private since it was facaded by `SpotlightContainerDecorator` to handle that.

ryanjduffy: Looked into this a bit more and I don't think we can safely use `move` since the element could be the scroller container (if the position doesn't fall on a spottable component) and positioning based on it's bounds will not give the expected result.

ryanjduffy:
I think some refactoring of the spotlight code is warranted but it's not egregious enough in my view to block on it. We can't only move that method out elsewhere as it relies on `spotNextFromPoint` which is used elsewhere in this module. Rather than bog ourselves down in refactoring, let's land this and circle back later.

PR #2267, opened 2019-05-09
PLAT-77740: Fix ViewManager children reconciliation

ryanjduffy:
### Checklist

Quickly changing a ViewManager with only two children could cause the last one to enter to be rendered out of view.

meridiem

### Resolution
I'm a bit fuzzy on the root cause but it appears there are a couple things off about the ViewManager implementation.

* In `TransitionGroup.componentDidUpdate` we were passing `prevState.children` and `this.state.activeChildren` to `reconcileChildren` which means we were comparing the previous- "previous children" (`prevState.children`) with the current children rather the previous-"active children" (`prevState.activeChildren`) with the current children. The side effect appears to be that because `this.state.activeChildren` was in `prevState.children` (since it left and re-entered quickly), that it was treated as "leaving" instead of "entering" causing the animation to be wrong.
* In `ViewManager`, we were attempting to track the previous index in state but the logic was wrong. By setting `prevIndex` in `getDerivedStateFromProps` to `props.index`, the value of `this.state.prevIndex` in `render` doesn't equal the previous index but the current index (since gDSFP is called before every render). This didn't seem to bother `Picker` but broke `BreadcrumbDecorator` with the `TransitionGroup` change. The resolution is to track the current index in `state.index` and when it doesn't match the `index` from props, move `state.index` to `prevIndex` and refresh `state.index` with the new value.

### Additional Considerations
There's a bit of refactoring included in `reconcileChildren` that isn't actually required but I had done in the course of attempting to resolve the bug. The first batch (in 9d44e1b) fixed the picker bug but helped surface the breadcrumb issue. As a result, I think they good changes to make since there was some wonkiness in there as well.


ryanjduffy:
@germboy Pushed a fix for the issue you found. I discovered that changing the duration of `ViewManager` for panels to a larger number (e.g. 1000ms) made it very easy to repro. I included some inline comments on the root cause but, to summarize, it was possible for views to be unmounted before they fired their callback causing the `currentlyTransitioningKeys` map to get out of sync effectively suppressing the `onTransition` event which Panels used to release spotlight pause.

Easy ... right? ... 😓

PR #2269, opened 2019-05-09
PLAT-74475: Restore animation in Scrollable family

viodragon2:
### Checklist

Restore animation in `Scrollable` family

### Resolution
Revert changes from #2015. This entails removing animate/noAnimation props from the components. If we want to leave this option, then the fix should only fix the defaultProp value.


ryanjduffy:
Should this target `feature/ui-update`?

PR #2271, opened 2019-05-09
PLAT-78459: Spotlight focus not applied to Slider knob after dragging

sugardave:
### Checklist

On touch platforms, dragging an element before a press applies focus will result in focus not being applied with the drag/touch ends.

### Resolution
Rather than have `Spotlight` blur an incorrectly focused element, make it always focus the element that ends the touch.

### Links
PLAT-78459


webOS101:
I suspect this will be very tricky to resolve, particularly with the code in scroller that blurs on `touchstart`.

sugardave:
From the conversation I had with @germboy, I was under the impression that we did not want to focus on touchstart and let the platform determine what to do.

PR #2278, opened 2019-05-13
PLAT-78617: Update default size to "small"

Djspaceg:
### Checklist

The latest design calls for all smaller components.

### Resolution
Default `size="small"` set for all components with a size prop.

### Additional Considerations
Several tests/samples needed updates to support properly detecting of default values.

### DIFF
--- a/packages/moonstone/ToggleButton/ToggleButton.js
+++ b/packages/moonstone/ToggleButton/ToggleButton.js
@@ -155,7 +155,6 @@ const ToggleButtonBase = kind({
 		disabled: false,
 		minWidth: true,
 		selected: false,
-		// size: 'large', // we won't set default props for `size` yet to support `small` prop
 		toggleOffLabel: '',
 		toggleOnLabel: ''
 	},
@@ -184,7 +183,7 @@ const ToggleButtonBase = kind({
 		delete rest.small;
 
 		return (
-			<Button data-webos-voice-intent="SetToggleItem" {...rest} aria-pressed={selected} selected={selected} />
+			<Button css={css} data-webos-voice-intent="SetToggleItem" {...rest} aria-pressed={selected} selected={selected} />
 		);
 	}
 });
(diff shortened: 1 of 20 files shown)


packages/moonstone/ToggleButton/ToggleButton.js
```diff
@@ -155,7 +155,6 @@ const ToggleButtonBase = kind({
 		disabled: false,
 		minWidth: true,
 		selected: false,
```
webOS101: Need to add a `defaultProp` here or else `getSize` will change to 'large'.  We could also update all the `setSize()` calls to be 'small' by default now.  However, just assigning a `defaultProps` should be fine.

Djspaceg: Interesting. That wasn't the case that I had seen on `ToggleButton`. It renders small, inheriting the default size from `Button`.

I was thinking that at a later time, we'd simply remove the `getSize` method entirely, but maybe it should just be deleted now. What do you think? (I'd prefer deleting it entirely, rather than change the code to better support it, as it's a temporary workaround to support the old `small` prop.)

Djspaceg:
Hmm, that's a good point about the docs, @webOS101. I think I agree that the `size` docs need to be cloned into each component (with the moonstone default size value).

packages/moonstone/ToggleButton/ToggleButton.js
webOS101: I didn't look deeply, it could be the default comes from how it's constructed/used.  I don't mind if we remove `getSize()` and just leave the deprecation code in.  For `ui`, we should remove the `getSize()` and use `defaultProps`, though it can wait until we remove the deprecations.

Djspaceg:
Updated the samples to import their ancestors (and their default values) in a more consistent way to hopefully provide better initial values for the stories.

PR #2279, opened 2019-05-13
ENYO-5950: Fix Input text to be visible when disabled and focused

haileyR:
…ocused.

### Checklist

The value was not visible when it's is disabled and focused. The color of the input is white and we lower the opacity of white color which doesn't affect anything.

### Resolution
Changed the color to black for disabled input so that it's value is visible for both not-focused and focused states.

### Additional Considerations
Maybe we have applied low opacity on color white somewhere else.

### DIFF
--- a/packages/moonstone/Input/Input.module.less
+++ b/packages/moonstone/Input/Input.module.less
@@ -168,7 +168,7 @@
 
 			.input,
 			.decoratorIcon {
-				color: @moon-input-text-disabled-color;
+				color: @moon-input-text-disabled-focus-color;
 			}
 		});
 	});
--- a/packages/moonstone/Input/Input.module.less
+++ b/packages/moonstone/Input/Input.module.less
@@ -168,7 +168,7 @@
 
 			.input,
 			.decoratorIcon {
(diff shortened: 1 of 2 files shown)


packages/moonstone/Input/Input.module.less
```diff
@@ -168,7 +168,7 @@
 
 			.input,
 			.decoratorIcon {
-				color: @moon-input-text-disabled-color;
+				color: @moon-input-text-disabled-focus-color;
```
webOS101: @Djspaceg Should we move this into a `.focus` section?  Also, do we need to update the `.focus()` to also or instead us `focus-within`?

Djspaceg: This new variable does appear to not be in the ideal place (its name suggests it should be inside a disabled block and a focus block, while it's currently just inside a disabled block). It also seems to delete any reference to `@moon-input-text-disabled-color` which I don't think is the intent.

ryanjduffy:
Color change :lgtm:. Thanks @haileyR + @Djspaceg!

PR #2282, opened 2019-05-14
ENYO-5964: Restore full height Scrollable

Djspaceg:
### Checklist

This reverts code added by a previous commit to restore dropdown's sizing

### Additional Considerations
I'm unclear as to what this change was made to support, or what the consequences outside of dropdown are for reverting this.

### DIFF
--- a/packages/ui/Scrollable/Scrollable.module.less
+++ b/packages/ui/Scrollable/Scrollable.module.less
@@ -17,6 +17,7 @@
 		display: flex;
 		width: 100%;
 		height: 100%;
+		min-height: 0;
 		flex: auto;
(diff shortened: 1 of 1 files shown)


webOS101:
This fix was added so that the scroller would show the horizontal scrollbars.  The sizing was being handled by the `flex`.  Here's the original PR: https://github.com/enactjs/enact/pull/2260

0x64:
Tested. LGTM!

Djspaceg:
As a general note for the future, I'd like to switch our custom implementations of flex in various places in the framework to Layout/Cell, since flex has so many invisible gotchas that are hard to check in every instance.

PR #2283, opened 2019-05-14
ENYO-5975: Fix Disabled+Focused IconButton and Switch

Djspaceg:
as well as large-text Switch

### Checklist

IconButton and Switch are too translucent when disabled and focused.
Switch is misshapen when in large-text mode.

### Resolution
Updated/simplified the disabled focused rules for IconButton and Switch, added proper sizes for Switch in large-text mode

### DIFF
--- a/packages/moonstone/Switch/Switch.module.less
+++ b/packages/moonstone/Switch/Switch.module.less
@@ -123,10 +123,6 @@
 		.toggleIcon {
 			.applySkins({
 				opacity: @moon-disabled-opacity;
-
-				// .disabled({
-				// 	opacity: 1;
-				// });
 			});
 		}
 	});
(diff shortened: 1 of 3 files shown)


packages/moonstone/Switch/Switch.module.less
```diff
@@ -101,13 +118,16 @@
...
-			opacity: @moon-disabled-opacity;
-		});
-	}
-});
+	.disabled({
+		.toggleIcon {
+			.applySkins({
+				opacity: @moon-disabled-opacity;
+
+				// .disabled({
```
ryanjduffy: Is this needed?

Djspaceg: "needed" may be an overstatement. It can be safely removed.

ryanjduffy:
Looks like the switch doesn't _quite_ make it when selected :soclose:

image

PR #2286, opened 2019-05-14
PLAT-78725:  Fix VirtualGridList to work properly with `wrap` prop

0x64:
### Checklist

`wrap` feature is broken after #2239. (A kind of regression)

### Resolution
Added missing condition.

### Additional Considerations
Fixed the regression before any release, so there is no change log.

### Links
PLAT-78725


seunghoh:
LGTM

ryanjduffy:
I'm expecting this is broken in `develop`. If so, this should either be retargeted to `develop` or cherry picked back there after merging.

webOS101:
I believe this regression was introduced when we removed the disabled skipping code.  But, I can't test VGL on `develop` right now.

ryanjduffy:
Based on @webOS101 and @0x64 comments that this was regression from #2239, this is the right branch for the fix. :nothingtodohere:

webOS101:
Confirmed issue does not reproduce on `develop`.

PR #2287, opened 2019-05-15
ENYO-5961: Fix slider to not scroll viewport on touch platforms

ryanjduffy:

### Checklist

Dragging a Slider on touch platforms would scroll the viewport (if it could scroll).

### Resolution
Add `touch-action: none` to slider to prevent scroller when interacting.

### Additional Considerations
Does allow scrolling when the slider is disabled. It _shouldn't_ scroll the sampler in this case either but that's a side effect of the sampler being a tad larger than the viewport. I have not fixed that issue here since it is sampler-specific and not a priority.


webOS101:
This seems OK... but would it have been better to just prevent the default action inside the scroller so that it has no effect or to prevent it from bubbling out?

ryanjduffy:
I'd expect that it would be broken in 2.5.2 as well but I didn't test.

webOS101:
you are correct, sadly.

ryanjduffy:
It should when we cherry pick. I put it in that section

PR #2296, opened 2019-05-17
ENYO-5986: Fix the color of the translucent family of buttons

Djspaceg:
### Checklist

The background colors for translucent and lightTranslucent Buttons are incorrect.

### Resolution
Updated translucent and lightTranslucent background colors in light and dark skins to match the design-intent.

In dark skin, the lightTranslucent button should appear identical to the normal button when on the default black background. Conversely, the sight skin translucent button should appear identical to the normal button when on the standard off-white background. All versions should have a translucent background, through which, you can see a background-image, if one is present behind the button.

### DIFF
--- a/packages/moonstone/Button/Button.module.less
+++ b/packages/moonstone/Button/Button.module.less
@@ -257,11 +257,11 @@
 		}
 
 		&.translucent .bg {
-			background-color: fade(@moon-button-bg-color, @moon-button-translucent-opacity);
+			background-color: @moon-button-translucent-bg-color;
 		}
 
 		&.lightTranslucent .bg {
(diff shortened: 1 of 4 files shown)


ryanjduffy:
I do wonder how clear those differences are to authors as it seems unpredictable that translucent == normal on a light background but lightTranslucent == normal on the black background.

Djspaceg:
Honestly, it's not predictable if you haven't seen/tried the visuals, since it's more "color theory" than something that's obviously intuitive. A gray thing on a black background could be opaque gray or translucent white and visually appear the same. Conversely, a gray thing on a white background could be opaque gray or translucent black, and visually appear the same. This is the basis for these properties/values. It would be nice to have agnostic terms, or auto-flipping terms, but I fear that would just make it more confusing.

PR #2297, opened 2019-05-17
PLAT-78894: Updated caret color for latest designs

Djspaceg:
### Checklist

Caret color is too pink.

### Resolution
Caret color has been reverted to natural (default) state.

This is based on #2293 to allow validation of the caret color on a non-white background color. It's not 💯 necessary, but this hierarchy is helpful for testing.


ryanjduffy:
I'm confused how I didn't see the regression in the last ticket as I distinctly remember focusing the decorator and seeing the outline with a black bg. Nevertheless, I could see it broken now but fixed with this PR. 🤷‍♂

PR #2301, opened 2019-05-20
PLAT-78786: Fix 5way navigation in VirtualList

germboy:
### Checklist

There are some cases where 5way navigation in a VirtualList does not behave as expected

### Resolution
When a VirtualList calculates the next target to set focus, there were cases where it was not taking into account whether there were available controls to focus before attempting to set focus to them (whether the controls were disabled or no further controls were available in the specified direction).

I've attempted to modify how VirtualList decides where focus should change to by determining whether or not spottable controls are first available in the 5way direction pressed. It's tough determining where to set focus in some cases where there are disabled components - strictly from a VirtualList POV - since we want to set focus on the nearest eligible component using spotlight proximity rules. In cases where the next target is disabled, we fallback to allowing spotlight to move focus where appropriate. 

There are other things to consider, whether or not any further controls are disabled and if/when wrapping should take place. There may be other edge-cases that need to be considered that I am not aware of or have not come across. Thorough testing of this proposed change would be appreciated.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -503,33 +503,42 @@ const VirtualListBaseFactory = (type) => {
 		}
 
 		getNextIndex = ({index, keyCode, repeat}) => {
-			const {dataSize, rtl, wrap} = this.props;
+			const {dataSize, isItemDisabled, rtl, wrap} = this.props;
 			const {isPrimaryDirectionVertical, dimensionToExtent} = this.uiRefCurrent;
+			const column = index % dimensionToExtent;
+			const row = (index - column) % dataSize / dimensionToExtent;
 			const isDownKey = isDown(keyCode);
 			const isLeftMovement = (!rtl && isLeft(keyCode)) || (rtl && isRight(keyCode));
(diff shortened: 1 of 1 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -579,11 +588,17 @@ const VirtualListBaseFactory = (type) => {
 				} else {
 					this.isScrolledBy5way = true;
 
-					if (isWrapped && wrap === true) {
+					if (isWrapped) {
```
ryanjduffy: Refactoring suggestion. Thoughts?

```js
					this.isWrappedBy5way = isWrapped;
					if (isWrapped && wrap === true && (
						this.uiRefCurrent.containerRef.current.querySelector(`[data-index='${nextIndex}'].spottable`) == null
					)) {
						this.pause.pause();
						target.blur();
					} else {
						this.focusOnItem(nextIndex);
					}
```

germboy: I like it - still appears to cover all the bases. Allowing `isWrapped` to set the value for `this.isWrappedBy5way` really allows that bit to be cleaned up nicely :nice:

PR #2305, opened 2019-05-20
ENYO-5994: Allow keydown events in VirtualList to propagate

germboy:
### Checklist

`keydown` events in a `VirtualList` do not propagate, causing events handlers that are passed via props to not fire.

### Resolution
We initially stopped propagation of `keydown` events in an attempt to bypass spotlight's logic, so that `VirtualList` can specifically set focus to its next intended target - this was done for better focus-change performance. This had the unintentional side-effect of preventing event handlers (passed as props) to be acted on.

As an alternative method, we can pause spotlight on our keydown events - allowing events to bubble, then re-enabling spotlight when we're finished navigating in `VirtualList`.


ryanjduffy:
I'm concerned adding spotlight pausing will introduce bugs. In this case, we "capture" spotlight on keydown and release on keyup but we can't guarantee that we'll see the keyup since the pointer could be activated or the focused node could be removed or blurred. It might be possible to manage these cases but it's risky and I'm not sure it's the right purpose for pausing.

Are there other alternatives?

germboy:
I completely understand the concern. I wasn't entirely confident that this change wouldn't cause any side effects. That being said, I'm not certain there are any obvious alternatives since we want to prevent default spotlight handling while allowing the events to bubble. I'm open to thoughts/feedback.

PR #2306, opened 2019-05-21
ENYO-5990: Update prop-types to variable range dependency.

JayCanuck:
### Checklist

* We previously restricted `prop-types` to a locked version in https://github.com/enactjs/enact/pull/2111/ as a way to avoid an issue with null values in `oneOf`.
* However the latest versions of React have greater dependencies of `prop-types`. As a result, there are multiple copies of `prop-types` within `./node_modules`

### Resolution
* Updates `prop-type` dependencies to `^15.7.2`.
* The `prop-types@15.7.2` release fixes the null `oneOf` issue, so we are safe to use that as a base for the variable-range.

### DIFF
--- a/packages/core/package.json
+++ b/packages/core/package.json
@@ -30,7 +30,7 @@
   "dependencies": {
     "classnames": "^2.2.5",
     "invariant": "^2.2.2",
-    "prop-types": "15.6.2",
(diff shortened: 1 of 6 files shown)


JayCanuck:
Not sure how I missed that one. Updated :+1:

PR #2310, opened 2019-05-22
ENYO-5997- Increased size of button label to test marquee

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/qa/Spotlight.js
+++ b/packages/sampler/stories/qa/Spotlight.js
@@ -145,7 +145,7 @@ class DisableOnClick extends React.Component {
 			<div>
 				<p>Pressing the marqueeable button will disable it. The marquee should continue and restart while the button is focused and disabled.</p>
 				<Button disabled={this.state.disabled} onClick={this.handleButtonDisable}>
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Spotlight.js
```diff
@@ -145,7 +145,7 @@ class DisableOnClick extends React.Component {
 			<div>
 				<p>Pressing the marqueeable button will disable it. The marquee should continue and restart while the button is focused and disabled.</p>
 				<Button disabled={this.state.disabled} onClick={this.handleButtonDisable}>
-					Marqueeable Button
+					Marqueeeeeeeeeeeeeeeeeeeeeeable Button
```
Djspaceg: We do this in several of our samples, but we really shouldn't. By "this" I mean repeating a long string of characters. While this _does_ invoke a marquee, it's not helpful for monitoring the scroll rate or tracking movement, since you can't tell one from the next when it's moving. Let's change this into multiple distinctly different words so it meets all of the marquee needs.

PR #2312, opened 2019-05-23
ENYO-5999: Extended Dropdown to support objects, like Group

Djspaceg:
Enact-DCO-1.0-Signed-off-by: Blake Stephens <blake.stephens@lge.com>

### Checklist

`Dropdown` uses `Group` internally, but doesn't accept the same set of features as `Group`

### Resolution
Updated the propType definition for `children` of `Dropdown` to include the same options as `Group`.

### DIFF
--- a/packages/moonstone/Dropdown/Dropdown.js
+++ b/packages/moonstone/Dropdown/Dropdown.js
@@ -81,7 +81,6 @@ const DropdownList = Skinnable(
 			 * The selections for Dropdown
 			 *
 			 * @type {String[]|Object[]}
-			 * @required
 			 * @private
 			 */
 			children: PropTypes.oneOfType([
@@ -89,7 +88,7 @@ const DropdownList = Skinnable(
 				PropTypes.arrayOf(PropTypes.shape({
 					key: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired
 				}))
-			]).isRequired,
+			]),
 
 			/**
 			 * Called when an item is selected.
@@ -151,7 +150,6 @@ const DropdownBase = kind({
 		 * the generated components as the readable text.
 		 *
 		 * @type {String[]|Object[]}
-		 * @required
 		 * @public
 		 */
 		children: PropTypes.oneOfType([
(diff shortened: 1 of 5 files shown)


packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -80,10 +80,15 @@ const DropdownList = Skinnable(
 			/**
 			 * The selections for Dropdown
 			 *
-			 * @type {String[]}
+			 * @type {String[]|Object[]}
```
webOS101: @ryanjduffy What should we use to indicate the object must have a `key` prop?   `Array.<{key}>`?  `Array.<{key: Number|String}>`?

ryanjduffy: the first looks okay to me if it renders int the docs well. I don't think we need to maintain the specific type of key in our docs unless necessary

webOS101: I was thinking for TypeScript generation.

ryanjduffy: Ah. The latter would be more correct in that regard. Not sure it's critical but maybe not too painful since we only reference it a couple spots.

webOS101: I'll test it to see what drops out.

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -140,15 +145,21 @@ const DropdownBase = kind({
...
+		 * be passed onto an `Item` component and a unique `key` property is required.
+		 *
+		 * @type {String[]|Array.<{key: (Number|String)}>}
 		 * @public
 		 */
-		children: PropTypes.node,
+		children: PropTypes.oneOfType([
+			PropTypes.arrayOf(PropTypes.string),
+			PropTypes.arrayOf(PropTypes.shape({
+				key: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired
```
webOS101: Should we also require a `children` prop of type `string` or `Component`?

ryanjduffy: Yes. Probably for both this and ExpandableList

PR #2313, opened 2019-05-23
PLAT-79212: Updated the Touchable sample to better demonstrate its intent

Djspaceg:
### Checklist

The Touchable sample is confusing regarding what it's attempting to demonstrate.

### Resolution
Added a visual element to show where the tap-start radius is checking for when pulses should be fired.


Djspaceg:
Updated to hopefully avoid the linting issues and provide slightly better separation. It's a little wonky but works as expected.

PR #2322, opened 2019-05-28
ENYO-6007: Corrected spacing between Header and content

Djspaceg:
### Checklist

There's no gap between the Sampler Header and content.

### Resolution
`Slottable` does not recognize/handle `React.Fragment`. The Sampler Environment component was updated to use a more traditional component layout to allow `Slottable` to continue to function normally in this scenario.


Djspaceg:
It may be possible, however, I'm not really convinced that it's worth it the effort and/or perf cost to update `Slottable` to recursively reach into one or more nested `React.Fragments` to find slot(s).

PR #2323, opened 2019-05-28
fix emitting extra hold pulse event on re-enter

ryanjduffy:

### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/ui/Touchable/Hold.js
+++ b/packages/ui/Touchable/Hold.js
@@ -10,6 +10,14 @@ class Hold {
 
 	isHolding = () => this.holdConfig != null
 
+	isWithinTolerance = ({x, y}) => {
+		const {moveTolerance} = this.holdConfig;
+		const dx = this.startX - x;
+		const dy = this.startY - y;
+
+		return Math.sqrt(dx * dx + dy * dy) < moveTolerance;
+	}
+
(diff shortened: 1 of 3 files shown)


packages/ui/Touchable/Hold.js
```diff
@@ -88,12 +94,12 @@ class Hold {
...
 	}
 
-	enter = () => {
+	enter = (coords) => {
 		if (!this.isHolding()) return;
 
 		const {resume} = this.holdConfig;
 
-		if (resume) {
+		if (resume && this.isWithinTolerance(coords)) {
```
webOS101: Do you need to check if `cancelOnMove` is set before checking coordinates??

ryanjduffy: It doesn't seem to matter in this case but I think there is still some unexpected behaviors with the config options. I thought this was necessary only because it was noticeable in the current test case whereas `cancelOnMove` isn't used in the TC.

webOS101: Hmm, something seems off to me from that description.  Maybe I'll have to dig into it some.  I thought that the move tolerance only applied when `cancelOnMove` was set.

ryanjduffy:
:+1:

PR #2324, opened 2019-05-28
PLAT-78406: Fix EditableIntegerPicker docs for min and max props

haileyR:
More details on how to set the `min` and `max` prop values.


packages/moonstone/EditableIntegerPicker/EditableIntegerPicker.js
```diff
@@ -52,6 +53,9 @@ const EditableIntegerPickerBase = kind({
 		/**
 		 * The maximum value selectable by the picker (inclusive).
 		 *
+		 * The value should be evenly divisible by
```
webOS101: Going to tweak this to say 'the range between `min` and `max` should be divisible by `step`'

PR #2327, opened 2019-05-30
ENYO-5994: Forward key events in VirtualList

germboy:
### Checklist

Applications cannot respond to `onKeyDown` events when passed to a `VirtualList`, since we blindly stop event propagation when capturing it in the list.

### Resolution
We can use the react dispatcher to turn our key events into synthetic events. This will allow us to properly forward events and allow them to be acted on by the application. Also, I've added a change to only stop event propagation when focus actually moves within the `VirtualList`.

### Additional Considerations
I've created this as a draft PR, since I have real performance concerns about this change. Converting the key events to synthetic events means that we're introducing the overhead of the react dispatcher for every key event.

It's fairly critical that key events in `VirtualList` be as lightweight as possible, this helps scrolling performance when navigating via 5way - especially during a hold, where key events occur rapidly. Attaching an event listener directly to the dom allows us to immediately respond to the event - whereas using synthetic events adds time to the amount of work that needs to be done per event.

When using synthetic events, the time it takes to perform a single keydown event increases by 20ms. When performing a 5way hold where the speed of focus change is near its maximum, focus change takes place in the scroll event, so less time is required in the key events themselves. That being said, synthetic events increase the time of each keydown event by 3x the amount (from ~3ms to ~9ms).
Measurements are using chrome 6x slow.

This does not seem ideal to me, but would like thoughts or alternative solutions.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -1,5 +1,6 @@
 import clamp from 'ramda/src/clamp';
 import {is} from '@enact/core/keymap';
+import {forward, forwardWithPrevent, handle} from '@enact/core/handle';
 import PropTypes from 'prop-types';
 import React, {Component} from 'react';
 import Spotlight, {getDirection} from '@enact/spotlight';
@@ -25,7 +26,8 @@ const
 	Native = 'Native',
 	isItemDisabledDefault = () => (false),
 	// using 'bitwise or' for string > number conversion based on performance: https://jsperf.com/convert-string-to-number-techniques/7
-	getNumberValue = (index) => index | 0;
+	getNumberValue = (index) => index | 0,
+	nop = () => {};
 
 /**
  * The base version of [VirtualListBase]{@link moonstone/VirtualList.VirtualListBase} and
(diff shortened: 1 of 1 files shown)


ryanjduffy:
@germboy 

* What is the usability impact of the change on device?
* Do we have a "safe" threshold under which we should stay for handling key events to prevent jank?
* Would it be possible to only support the synthetic event callback on the first `keydown` and then swap to native listeners for subsequent events? This might be super weird but perhaps worth considering.

We may have to consider an alternative impl for the use case than `forwardWithPrevent` if the perf cost is too great. Maybe handling `keydown` in the capture phase for this scenario would be sufficient since the app could stop it then and avoid key handling by VL.

germboy:
To address your first 2 points:
- The impact appears minor - however, it would easily negate the minor gains we made in our last round of optimizing key handlers for VL. The minor gains we previously made were what allowed us to restore animations to VL - and even then, @viodragon2 shared concerns that it may not be improved enough.
- I don't think we have identified a "safe" threshold. I think we're still bordering on not optimized enough, so adding any additional time to key events concerns me.
- Supporting an initial synthetic event then passing off to native listeners is indeed weird ... it has potential to introduce errors, but theoretically could be an option. I'll investigate it.
Thanks for the input - I'll investigate the options suggested 👍

PR #2329, opened 2019-05-30
ENYO-5979: Fix VL unexpected focus change to to last focused items

germboy:
### Checklist

There are cases where `VirtualList` focus could unexpectedly change to a last-focused item.

### Resolution
Each time when setting focus to "next" items, `VirtualList` was needlessly setting the current index value to `this.nodeIndexToBeFocused`. This value only needs to be set prior to explicitly setting scroll position/focus elsewhere (e.g. a wrap or page event). As a result, `this.nodeIndexToBeFocused` was not being nullified, causing unexpected focus changes when items were mounted.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -517,11 +517,12 @@ const VirtualListBaseFactory = (type) => {
 						this.isWrappedBy5way = true;
 						this.pause.pause();
 						target.blur();
+						this.nodeIndexToBeFocused = nextIndex;
 					} else if (!isWrapped || wrap !== 'noAnimation') {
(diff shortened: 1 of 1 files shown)


ryanjduffy:
@germboy Is this a regression from the changes in `develop`? If so, should it be fixed there?

germboy:
@ryanjduffy Perhaps you're right. I think the side-effect of the issue is only seen in `feature/ui-update`, but yeah - the source of the problem still exists in `develop`.

germboy:
@ryanjduffy I opened #2336 against `develop`. That PR also correctly sets the `this.lastFocusedIndex` value, since it was not updated in `develop`

ryanjduffy:
Closing this in favor of #2336 which has been merged into `feature/ui-update` as well.

PR #2330, opened 2019-05-30
ENYO-6019 Fixed ExpandableList with selected array

webOS101:
### Checklist

It was possible for an `ExpandableList` to error if an array of selected items was passed to
a non-multi-select.

### Resolution
Refactor the logic to deal better with `selected` as array.

### Additional Considerations
Added tests

### Links
ENYO-6019


ryanjduffy:
I think we should fix this in 2.x. WDYT?

webOS101:
I think it would be fine to fix in either, so I wouldn't mind rebasing to `develop`.

webOS101:
I rebased to `develop`

PR #2338, opened 2019-05-31
PLAT-77708: Reassigned several newly supported languages

Djspaceg:
### Checklist

The system font now supports more language scripts

### Resolution
Updated several languages to get "non-latin" overrides so they get the correct font assignments.
Also added a basic font/language checking sample in QA Sampler.


Djspaceg:
@webOS101, you'll need to clarify what your difficulties are. The fallbacks are indeed applied in the fonts definition file, and locale-specific fonts are applied in fontGenerator. What are you seeing? Do you see a better way to "couple" things?

webOS101:
Aren't certain font or CSS properties applied depending on whether it's a latin or non-latin locale?

Djspaceg:
That is correct.

webOS101:
So, when the new fonts are not used, the following locales:  'el', 'bg', 'mk', 'mn', 'ru', 'uk', 'kk' will now render differently, will they not?

Djspaceg:
They will indeed be rendered differently.

Followup question: What scenario are you referring to where the new fonts will not be used?

webOS101:
I was thinking of cases of using Moonstone in the browser.

Djspaceg:
You're describing a scenario which: runs a moonstone app in a generic Chrome browser, without LG TV fonts, but with a locale specified as one of the above, and are concerned that Miso or Museo Sans (the old fonts) don't have the correct glyphs available. The OS will provide the specific missing glyph in this case.

webOS101:
:nothingtodohere:  :shipit:

PR #2339, opened 2019-06-03
ENYO-6022: Fix setting initial focus of short VirtualList

changgilee:
### Checklist

When items of list is small, initial focus is not exist in VirtualList

### Resolution
When items is small, componentDidUpdate is not called.
In current status, restoreFocus() is only declared in componentDidUpdate.

### Additional Considerations

### Links
https://jira2.lgsvl.com/browse/ENYO-6022

### Comments

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -194,9 +194,7 @@ const VirtualListBaseFactory = (type) => {
 				containerNode.addEventListener('keyup', this.onKeyUp);
 			}
 
-			setTimeout(() => {
-				this.restoreFocus();
-			}, 0);
+			this.forceUpdate();
 		}
 
 		componentDidUpdate (prevProps) {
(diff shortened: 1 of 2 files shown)


ryanjduffy:
@changgilee - Thanks for the report! The `setTimeout` is problematic as it can't guarantee that focus should be set when it fires and was interfering with focus timing previously. We should look for alternatives that are tied to events like the `onUpdateItems` event recently added to `ui/VirttualList`.

changgilee:
@ryanjduffy Thanks for your feedback~

changgilee:
@ryanjduffy I have question. When I did check onUpdateItems of VirtualListBase, it seems that the handler is not called when VirtualList is initialization.(In usual case, the handler is called when new item is created during item navigation) Should we call it enforce way for resolve this issue?

ryanjduffy:
Possibly. I don't know off hand but I suspect we need to call it on mount as well.

ryanjduffy:
`forceUpdate` is too heavy handed in this case. I've updated this to correctly emit the `onUpdateItems` event on mount (when `clientSize` is set) and on update (when either `firstIndex` or `numOfItems` changes). The latter is the fix for this bug and the former addresses a similar but as yet unreported bug.

PR #2346, opened 2019-06-05
ENYO-6009: ui/ToggleItem value prop is not sent to onToggle

sugardave:
### Checklist

Components that use `ui/Toggleable` HOC do not receive the `value` prop in their toggle handlers as detailed in the docs for `ui/ToggleItem`.  This patch addresses that.

### Resolution
`ui/Toggleable` HOC receives a new `eventProps` config option and `ui/ToggleItemDecorator` now configures `{toggleProp: 'onTap', eventProps: ['value']}` when it is composing.

### Additional Considerations
There is something odd about the way `ui/Toggleable` overwrites `ui/ToggleItem`'s `onToggle` handler but doesn't forward the event properly that I haven't finished puzzling out.

### Links
ENYO-6009


packages/ui/ToggleItem/ToggleItem.js
```diff
@@ -11,6 +11,7 @@
  * @exports ToggleItemDecorator
  */
 
+import {adaptEvent, handle, forward} from '@enact/core/handle';
```
Djspaceg: This is no longer needed, right?

packages/ui/Toggleable/Toggleable.js
```diff
@@ -204,7 +212,7 @@ const ToggleableHOC = hoc(defaultConfig, (config, Wrapped) => {
 
 		handle = handle.bind(this)
 
-		forwardWithState = (evName) => (ev, props) => forward(evName, {[prop]: !this.state.active}, props)
+		forwardWithState = (evName) => (ev, props) => forward(evName, {...pick(eventProps, props), [prop]: !this.state.active}, props)
```
Djspaceg: If `eventProps` is somehow undefined/null, does `pick` fall over? Probably an over-engineered edge case, but I'm curious.

packages/ui/Toggleable/Toggleable.js
```diff
@@ -63,6 +64,13 @@ const defaultConfig = {
 	 */
 	deactivate: null,
 
+	/**
+	 * Configures additional props to attach to the event that is sent when toggled.
+	 *
+	 * @type {String[]}
```
webOS101: Should we add a `@default` here?  I think, probably, yes.  I can do it.

webOS101: Also, need a `@memberof` here to prevent parse errors.

PR #2351, opened 2019-06-10
PLAT-76792: Remove Uppercase from components

viodragon2:
### Checklist

Remove Uppercase HOC and default casing in `moonstone` components. 

### Resolution
- Remove `i18n/Uppercase` hoc
- Remove Uppercase HOC from `Divider`, `Dialog` and `Heading`
- Updated sampler accordingly
- Updated docs accordingly


packages/i18n/Uppercase/Uppercase.js
```diff
@@ -1,86 +0,0 @@
...
-		casing: PropTypes.oneOf(['upper', 'preserve', 'word', 'sentence'])
-	},
-
-	defaultProps: {
-		casing: 'upper'
-	},
-
-	computed: {
-		children: ({casing, children}) => {
-			if (casing !== 'preserve' && React.Children.count(children)) {
```
ryanjduffy: Do you think it would be useful to expose this computed function as a utility function to simplify converting React children to the prescribed case?

viodragon2: No.

The idea was to let the developer use `i18n/util` explicitly instead of `i18n/Uppercase`, because there are no default cases where a component needs to change letter cases. When do you think this function would be useful?

Plus, this function felt like a workaround to me when I first looked at it as it doesn't do recursively.

ryanjduffy: Okay. I can live with that. If we (or an author) wanted to implement the `casing` prop as was here, it would likely require this sort of logic but the specific requirements might vary. Maybe in the future we could consider a version of `formatContent` which accepts a casing arg and the string which might be generalized enough to be useful.

PR #2353, opened 2019-06-11
Deprecate Uppercase HOC

viodragon2:
### Checklist

Deprecate `i18n/Uppercase` HOC for `i18n/util`

### Additional Considerations
Should `casing` prop added by the Uppercase HOC also be marked deprecated and warn?

### DIFF
--- a/packages/i18n/Uppercase/Uppercase.js
+++ b/packages/i18n/Uppercase/Uppercase.js
@@ -26,8 +26,6 @@ const formatContent = (casing, content) => {
 	}
 };
 
-deprecate({name: 'i18n/Uppercase', until: '3.0.0', replacedBy: 'i18n/util', since: '2.6.0'});
-
 /**
  * A higher-order component that is used to wrap an element to provide locale-aware uppercasing of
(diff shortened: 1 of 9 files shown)


packages/i18n/Uppercase/Uppercase.js
```diff
@@ -25,6 +26,8 @@ const formatContent = (casing, content) => {
 	}
 };
 
+deprecate({name: 'i18n/Uppercase', until: '3.0.0', replacedBy: 'i18n/util', since: '2.6.0'});
```
ryanjduffy: This is questionable value but it'd be "nice" to only warn when authors use `Uppercase` but not the framework. Previously, we've accomplished this by exporting a `private` version unwrapped by `deprecate` use it within the framework.

```
const WrappedUppercase = deprecate( ... );
export default WrappedUppercase;
export {
  WrappedUppercase as Uppercase,
  Uppercase as privateUppercase
};
```

I'm having trouble justifying the extra work but something to consider.

viodragon2: Done!

PR #2354, opened 2019-06-11
ENYO-6034: Fix Scroller to scroll once by page up/down

0x64:
### Checklist

After #2262, a list scrolls twice by a single page up/down key press.

### Resolution
We already have a scrolling logic by page keys on scrollbars to cover its own behavior, and another logic is added from #2262 for a general scrolling feature originally designed for contents not for scrollbars.
So, I changed a newly added general logic to be applied to the contents area only.

### Links
ENYO-6034
PLAT-65757

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -403,6 +403,8 @@ class ScrollableBase extends Component {
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


seunghoh:
You missed removing the login when an item in the contents area has a spot.

0x64:
> You missed removing the logic when an item in the contents area has a spot.

The issue that you pointed out is a bug not related to scrollbars, and I was tracking on it to make a new ticket separately. Thanks!

PR #2355, opened 2019-06-11
PLAT-79783: Apply accessibility on Dropdown

seunghoh:
### Checklist

Apply accessibility on new Dropdown component

### Resolution
Remove `role=alert` from the ContextualButton for avoiding read all internal items when drop opens
Add `role=checkbox` and `aria-checked` for reading selected status and X of Y

### DIFF
--- a/packages/moonstone/Dropdown/Dropdown.js
+++ b/packages/moonstone/Dropdown/Dropdown.js
@@ -35,8 +35,6 @@ import Skinnable from '../Skinnable';
 
 import css from './Dropdown.module.less';
 
-const filterChildren = children => children ? children.filter(Boolean) : [];
-
 const compareChildren = (a, b) => {
 	if (!a || !b || a.length !== b.length) return false;
 
@@ -263,7 +261,9 @@ const DropdownBase = kind({
 
 	computed: {
 		children: ({children, selected}) => {
-			return filterChildren(children).map((child, i) => {
+			if (!Array.isArray(children)) return [];
+
+			return children.map((child, i) => {
 				const aria = {
 					role: 'checkbox',
 					'aria-checked': selected === i
@@ -285,9 +285,7 @@ const DropdownBase = kind({
 		},
 		className: ({width, styler}) => styler.append(width),
 		title: ({children, selected, title}) => {
-			const isSelectedValid = !(typeof selected === 'undefined' || selected === null || selected >= children.length || selected < 0);
-
-			if (children && children.length && isSelectedValid) {
+			if (Array.isArray(children) && children[selected] != null) {
 				const child = children[selected];
 				return typeof child === 'object' ? child.children : child;
 			}
@@ -299,8 +297,9 @@ const DropdownBase = kind({
 	render: ({children, disabled, onOpen, onSelect, open, selected, width, title, ...rest}) => {
 		const popupProps = {children, onSelect, selected, width, role: ''};
 
-		// `ui/Group`/`ui/Repeater` will throw an error if empty so we disable the Dropdown and prevent Dropdown to open if there are no children.
-		const hasChildren = filterChildren(children).length;
+		// `ui/Group`/`ui/Repeater` will throw an error if empty so we disable the Dropdown and
+		// prevent Dropdown to open if there are no children.
+		const hasChildren = children.length > 0;
 		const openDropdown = hasChildren ? open : false;
 		delete rest.width;
 
@@ -322,7 +321,8 @@ const DropdownBase = kind({
 });
 
 /**
- * Applies Moonstone specific behaviors and functionality to [DropdownBase]{@link moonstone/Dropdown.DropdownBase}.
+ * Applies Moonstone specific behaviors and functionality to
+ * [DropdownBase]{@link moonstone/Dropdown.DropdownBase}.
  *
  * @hoc
  * @memberof moonstone/Dropdown
@@ -349,10 +349,10 @@ const DropdownDecorator = compose(
 /**
  * A Moonstone Dropdown component.
  *
- * By default, `Dropdown` maintains the state of its `selected` property.
- * Supply the `defaultSelected` property to control its initial value. If you
- * wish to directly control updates to the component, supply a value to `selected` at creation time
- * and update it in response to `onSelected` events.
+ * By default, `Dropdown` maintains the state of its `selected` property. Supply the
+ * `defaultSelected` property to control its initial value. If you wish to directly control updates
+ * to the component, supply a value to `selected` at creation time and update it in response to
+ * `onSelected` events.
  *
  * @class Dropdown
  * @memberof moonstone/Dropdown
--- a/packages/moonstone/Dropdown/Dropdown.js
+++ b/packages/moonstone/Dropdown/Dropdown.js
@@ -26,6 +26,7 @@ import Pure from '@enact/ui/internal/Pure';
 import PropTypes from 'prop-types';
 import {compose, equals} from 'ramda';
 import React from 'react';
+import warning from 'warning';
 
 import Button from '../Button';
 import ContextualPopupDecorator from '../ContextualPopupDecorator/ContextualPopupDecorator';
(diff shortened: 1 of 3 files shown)


packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -275,7 +279,8 @@ const DropdownBase = kind({
 	},
 
 	render: ({children, disabled, onOpen, onSelect, open, selected, width, title, ...rest}) => {
-		const popupProps = {children, onSelect, selected, width};
+		const role = '';
```
Djspaceg: Is the intent here to pre-define a default?
```
		const popupProps = {children, onSelect, selected, width, role: ''};
```

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -133,7 +133,11 @@ const DropdownList = Skinnable(
 					select="radio"
 					selected={selected}
 				>
-					{children}
+					{children.map((o, i) => {
```
Djspaceg: This may be better to place in the `computed` block.

ryanjduffy: Agree. Additionally, the computed block would make it a bit easier to manage `null` which will throw an error now.

ryanjduffy: Also, let's add a `filter` before the map to guard against `null` array members.

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -133,7 +133,11 @@ const DropdownList = Skinnable(
...
+						// eslint-disable-next-line no-shadow
+						const {children = o, key = i} = o;
```
ryanjduffy: This is clever but as a result it might not be obvious what the purpose is. Some inline docs are warranted at least i think.

Djspaceg: I think I'd prefer this to use much less clever code, for easier readability:
```
{children.map((o, i) => 
	({children: o, key: o.key || `item${i}`, role: 'checkbox', 'aria-checked': selected === i})
```

webOS101: `0` is a valid key value for `o.key`, which would get replaced with `item${i}`

seunghoh: @Djspaceg I've started exactly the same as you mentioned :) and then changed since I didn't want to make one line long but yeah.. readability is more important. Thanks~

Djspaceg: Feel free to place each prop on its own line. That's much easier to read too.
```
```javascript
{children.map((o, i) => ({
	children: (o.children || o),
	key: (o.key != null ? o.key : `item${i}`),
	role: 'checkbox',
	'aria-checked': (selected === i)
})
```

seunghoh:
Thank you guys for the detailed feedback. So, I understand as we agreed to go ahead with my proposed a11y UX. Will update codes with regard to the comments

packages/moonstone/Dropdown/Dropdown.js
seunghoh: Nope :) Changed

packages/moonstone/Dropdown/Dropdown.js
seunghoh: Yeap! moved to `computed` block and also added a `filter` to guard `null`

packages/moonstone/Dropdown/tests/Dropdown-specs.js
```diff
@@ -75,4 +101,69 @@ describe('Dropdown', () => {
...
+			</DropdownBase>
+		);
+
+		const expected = true;
+		const actual = dropDown.prop('popupProps').children[0].disabled;
+
+		expect(actual).toBe(expected);
+	});
+
+	test('should allow members in child object ts to override injected aria values', () => {
```
teckliew: typo?

ryanjduffy: yup

ryanjduffy: Fixed!

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -261,6 +262,27 @@ const DropdownBase = kind({
...
+						key: `item${i}`
+					};
+				}
+
+				return {
+					...aria,
+					...child
+				};
+			});
+		},
```
webOS101: Should we apply the filter method on the `title` computed function below?

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -275,10 +297,10 @@ const DropdownBase = kind({
 	},
 
 	render: ({children, disabled, onOpen, onSelect, open, selected, width, title, ...rest}) => {
-		const popupProps = {children, onSelect, selected, width};
+		const popupProps = {children, onSelect, selected, width, role: ''};
 
 		// `ui/Group`/`ui/Repeater` will throw an error if empty so we disable the Dropdown and prevent Dropdown to open if there are no children.
-		const hasChildren = children && children.length;
+		const hasChildren = filterChildren(children).length;
```
webOS101: Why don't we use computed here?  No real reason to change it, I guess.  Just curious.

ryanjduffy: Actually don't need to do it here. It can't be computed because `children` is computed and the computations are independent. However, _because_ `children` is computed and filters the incoming value, we need not do so here.

packages/moonstone/Dropdown/Dropdown.js
ryanjduffy: We should. Good catch!

ryanjduffy:
That's a good point on `selected`. It's definitely an edge case but if we are going to filter out children, we'll have to handle it.

packages/moonstone/Dropdown/Dropdown.js
webOS101: Both good points.

webOS101:
We should probably document that invalid children will be discarded _or_ do our best to render _something_.  Is there anything we can borrow from `Group` that would be useful?  Why is this component different from `Group`?

ryanjduffy:
I'm not sure how to handle this. Tempted to just throw/warn since it likely points to a design-time problem if you pass a selected index for an invalid option.

The notable difference between this and Group is that Dropdown prescribes the component that is rendered and the behavior for selection whereas Group doesn't. You could pass a component to Group that expected `null` values and they carried special meaning but that isn't the case here.

webOS101:
Then `ExpandableList`?

ryanjduffy:
Has the same concern. We currently do nothing to handle it.

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -261,6 +262,27 @@ const DropdownBase = kind({
 	},
 
 	computed: {
+		children: ({children, selected}) => {
```
webOS101: Do we need to account for passing `data-webos-voice-disabled` on to the children?

ryanjduffy: Good question. Agreed on following up separately.

webOS101:
Also curious why we have special aria handling for children in `DropDown` but not in `ExpandableList`.  Given the author can easily add it, do we need to do it manually?  Also, why don't we use `role: 'radiogroup'` for this since it uses radio select?

ryanjduffy:
The items in ExpandableList provide the ARIA props since it uses either `CheckboxItem` or `RadioItem` but Dropdown use `Item` which isn't "selectable" normally.

Not sure about using `radiogroup`. If we were to support that, it'd be best to due so automatically within group based on the selection mode I think.

webOS101:
How do `aria-selected` and `aria-selctable` play into this, if at all?

ryanjduffy:
Not sure off hand

ryanjduffy:
@webOS101 I've updated this to no longer filter children and instead warn in dev mode when an invalid child is passed. The prospect of adjusting `selected` for invalid children is too undefined to handle successfully at run time, imo.

PR #2357, opened 2019-06-11
ENYO-6045: Deprecate Divider

viodragon2:
### Checklist

Deprecate `moonstone/Divider` which will be moved to `moonstone/Heading` in 3.0.0

### DIFF
--- a/packages/moonstone/Divider/Divider.js
+++ b/packages/moonstone/Divider/Divider.js
@@ -15,6 +15,7 @@
  * @exports DividerDecorator
  */
 
+import deprecate from '@enact/core/internal/deprecate';
 import kind from '@enact/core/kind';
 import Uppercase from '@enact/i18n/Uppercase';
 import Pure from '@enact/ui/internal/Pure';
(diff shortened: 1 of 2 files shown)


packages/moonstone/Divider/Divider.js
```diff
@@ -92,6 +94,13 @@ const DividerBase = kind({
 	render: (props) => {
 		delete props.spacing;
 
+		deprecate({
```
ryanjduffy: This will cause the deprecation to warn on every render which is rather noisy. Would like to limit it to a single invocation is possible.

Maybe something like: https://codesandbox.io/s/weathered-fire-9xo6z

viodragon2: updated

packages/moonstone/Divider/Divider.js
```diff
@@ -39,6 +40,7 @@ import css from './Divider.module.less';
  * @memberof moonstone/Divider
  * @ui
  * @public
+ * @deprecated since 2.6.0. Will be moved to {@link moonstone/Heading} in 3.0.0
```
ryanjduffy: Since `Heading` doesn't exist here, we'll have to just reference it statically without the link. I'll update it here.

PR #2359, opened 2019-06-11
PLAT-78609: Remove deprecated props

viodragon2:
### Checklist

Remove deprecated props
- `small` prop in `moonstone/Input`, `moonstone/ToggleButton`, `moonstone/Button`, `moonstone/Icon`, `moonstone/IconButton`, and `moonstone/LabeledIcon`
-  `core/kind` config property `contextTypes`


Djspaceg:
It may be helpful for us and app developers to discover places where `small` wasn't removed (since it won't get a React warning) for us to add an interesting global CSS rule, something like:
```
[small] {
   outline: 3px red solid;
   &::before {
      position: absolute;
      outline: 3px red solid;
      content: "A stray `small` prop exists on this element";
   }
}
```
This will highlight any element which still has a `small` prop that blead onto the DOM node ("illegal").

webOS101:
That's a neat idea... but it'd have to be in develop mode only.  I can't suggest we break the appearance of deployed apps.  I think we could use that in the future during the deprecation phase, maybe?

Djspaceg:
Could be done behind the `.debug` global class.

webOS101:
I'm merging this as-is.  I'd like to discuss that css thing in another ticket or separately so we can get this removal in for today's release.

PR #2361, opened 2019-06-12
ENYO-6047: Header: title is not center-aligned

Sangwook-Lee:
### Checklist

The title is not center-aligned when centeted is set.

### Resolution
Fixed center alignment issue of Header by adding CSS `padding-start` when centered.

### Links
ENYO-6047

### DIFF
--- a/packages/moonstone/Panels/Header.module.less
+++ b/packages/moonstone/Panels/Header.module.less
@@ -155,6 +155,12 @@
 		.padding-start-end(0, var(--moon-panels-controls-width));
 	}
 
+	.header.centered.compact,
+	.header.centered.standard .title,
+	.header.centered.standard .headerInput {
+		.padding-start-end(var(--moon-panels-controls-width), var(--moon-panels-controls-width));
+	}
+
 	.header.compact {
 		.headerComponents {
 			border-right-width: @moon-header-controls-border-width;
--- a/packages/moonstone/Panels/Header.module.less
+++ b/packages/moonstone/Panels/Header.module.less
@@ -155,7 +155,6 @@
 		.padding-start-end(0, var(--moon-panels-controls-width));
 	}
 
-	.header.centered.compact,
 	.header.centered.standard .title,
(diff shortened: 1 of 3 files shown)


packages/moonstone/Panels/Header.module.less
```diff
@@ -155,6 +155,12 @@
 		.padding-start-end(0, var(--moon-panels-controls-width));
 	}
 
+	.header.centered.compact,
```
Djspaceg: I don't think compact should be included here. Compact Header isn't intended to be centered and would look weird if it were, so I don't think this is necessary.

Sangwook-Lee: I've attached screenshot on compact header in the ticket.

Djspaceg:
The Travis failure can be resolved by merging-in `develop` to this branch.

packages/moonstone/Panels/Header.module.less
```diff
@@ -155,6 +155,11 @@
 		.padding-start-end(0, var(--moon-panels-controls-width));
 	}
 
+	.header.centered.standard .title,
+	.header.centered.standard .headerInput {
+		.padding-start-end(var(--moon-panels-controls-width), var(--moon-panels-controls-width));
```
ryanjduffy: We're incurring a bit of overhead by using the mixin when both values are the same since it will produce extra selectors for rtl with the same rules.

@teckliew + @Djspaceg - worth revising?

Djspaceg: The use of the mixin, while seemingly verbose, allows this rule to overcome the rule on line 155. If we simply applied `padding` rules here, we would not be targeting the higher precedence RTL rules established on line 155 above.

If we instead used a newer (less supported) rule like `padding-block-start` and `padding-block-end`, we wouldn't need to use the mixin, and could use a simpler override on line 160.

ryanjduffy: :+1: Thanks for the explanation.

PR #2362, opened 2019-06-12
PLAT-78555: Handle resize event in Scrollable

ybsung:
### Checklist

VirtualList is not responsive when resizing a window.

### Resolution

VirtualList is recalculated and rendered again when receiving `resize` event.

### Additional Considerations

This PR also covered the case when rotating a window.

### Links

PLAT-78555


packages/ui/Scrollable/Scrollable.js
```diff
@@ -1243,6 +1250,12 @@ class ScrollableBase extends Component {
 		if (this.props.addEventListeners) {
 			this.props.addEventListeners(childRefCurrent.containerRef);
 		}
+
+		if (this.childRefCurrent.cc) { // For VirtualList
```
0x64: Is there any better way to distinguish the type of child component rather than checking `cc`?

ybsung: While we discussed this issue, we found that Scroller also need to be resizable to show or to hide scrollbars. I'll remove the condition.

0x64:
LGTM

seunghoh:
GridListNative position is wrong when you focus on scroller button and rotate the view.

0x64:
@seunghoh Good point! A browser keeps `scrollTop` value for natively scrollable elements. But our components reset on resize (int this PR), so temporarily the scroll position may not match with components' internal status.

seunghoh:
@ybsung on Mobile (Android and iPhone both) selected item is changed.
LIke, spot on item 63 on portrait, rotate device to landscape then selected item changed to item 3

ybsung:
Hi @seunghoh 
When rotating a device, Spotlight is hidden now with my patch.

PR #2364, opened 2019-06-13
PLAT-80079: Cleanup code in VirtualList and Scroller

ybsung:
### Checklist

We need to cleanup VirtualList and Scroller because we missed some patches and applied uncompleted patches in them.

### Resolution

* Remove unused `findSpottableItemWithPositionInExtent()`
* Apply the patch for `Scrollable` in ENYO-5957 to `ScrollableNative`
* Cleanup `onAcceleratedKeyDown` parameters
* Reduce calling `getNextIndex() in VirtualList once
* Focus an item immediately when an app mounted and called the `cbScrollTo` in `VirtualList`

### Additional Considerations

* Cleanup Scrollable and ScrollableNative code.
* Cleanup VirtualList code.

### Links

PLAT-80079

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -594,7 +594,7 @@ class ScrollableBase extends Component {
 						this.checkAndApplyOverscrollEffectByDirection(direction);
 					}
 				}
-			} else if (getDirection(keyCode)) {
+			} else if (!Spotlight.getPointerMode() && getDirection(keyCode)) {
 				const element = Spotlight.getCurrent();
 
 				this.uiRef.current.lastInputType = 'arrowKey';
@@ -877,8 +877,8 @@ class ScrollableBase extends Component {
 				onFlick={this.onFlick}
 				onKeyDown={this.onKeyDown}
 				onMouseDown={this.onMouseDown}
-				onWheel={this.onWheel}
 				onScroll={this.handleScroll}
+				onWheel={this.onWheel}
 				ref={this.uiRef}
 				removeEventListeners={this.removeEventListeners}
 				scrollTo={this.scrollTo}
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -7,18 +7,17 @@
  */
 
 import classNames from 'classnames';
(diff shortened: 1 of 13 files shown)


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -594,7 +594,7 @@ class ScrollableBase extends Component {
 						this.checkAndApplyOverscrollEffectByDirection(direction);
 					}
 				}
-			} else if (getDirection(keyCode)) {
+			} else if (!Spotlight.getPointerMode() && getDirection(keyCode)) {
```
0x64: I'm confusing that it is possible to catch 5-way key input in pointer mode here. Is `onKeyDown` handler called before the mode is changed to 5-way mode?

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -364,7 +364,7 @@ class ScrollableBaseNative extends Component {
 			isHorizontalScrollButtonFocused = horizontalScrollbarRef.current && horizontalScrollbarRef.current.isOneOfScrollButtonsFocused(),
 			isVerticalScrollButtonFocused = verticalScrollbarRef.current && verticalScrollbarRef.current.isOneOfScrollButtonsFocused();
 
-		if (focusedItem && !isHorizontalScrollButtonFocused && !isVerticalScrollButtonFocused) {
+		if (!Spotlight.isPaused() && focusedItem && !isHorizontalScrollButtonFocused && !isVerticalScrollButtonFocused) {
```
0x64: @ybsung What about to remove checking code from `Scrollable` instead of adding this to native one? It looks safe to blur for touch inputs event if Spotlight is paused.

ybsung: We added this condition in `Scrollable` because we found some issue.
https://github.com/enactjs/enact/pull/2314

0x64: 👍

packages/moonstone/Scrollable/Scrollable.js
ybsung: Let's investigate this one more and apply it again if needed.

0x64:
LGTM

PR #2367, opened 2019-06-17
ENYO-6051: Go next line via 5-way Down even though there is no item below the current item

ybsung:
### Checklist

If pressing a 5-way Down key when there is no item below the current item but there is the next line in VirtualGridList, Spotlight was gone.

### Resolution

If there is no item below the current item and there is the next line in VirtualGridList when pressing 5-way Down, Spotlight moves to the last item in it.

### Links

ENYO-6051

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -366,8 +366,14 @@ const VirtualListBaseFactory = (type) => {
 			if (isPrimaryDirectionVertical) {
 				if (isUpKey && row) {
 					targetIndex = index - dimensionToExtent;
-				} else if (isDownKey && isNextRow) {
-					targetIndex = index + dimensionToExtent;
+				} else if (isDownKey) {
+					if (isNextRow) {
+						targetIndex = index + dimensionToExtent;
+					} else if ( // if the next line is the last line and there is no item below the current item
+						Math.floor((index + dimensionToExtent ) / dimensionToExtent) === Math.floor((dataSize - 1) / dimensionToExtent)
+					) {
+						targetIndex = dataSize - 1;
+					}
 				} else if (isLeftMovement && column) {
 					targetIndex = index - 1;
 				} else if (isRightMovement && isNextAdjacent) {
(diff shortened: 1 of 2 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -366,17 +366,29 @@ const VirtualListBaseFactory = (type) => {
 			if (isPrimaryDirectionVertical) {
 				if (isUpKey && row) {
 					targetIndex = index - dimensionToExtent;
-				} else if (isDownKey && isNextRow) {
-					targetIndex = index + dimensionToExtent;
+				} else if (isDownKey) {
+					if (isNextRow) {
+						targetIndex = index + dimensionToExtent;
+					} else if ( // if the next line is the last line and there is no item below the current item
+						Math.floor((index + dimensionToExtent ) / dimensionToExtent) === Math.floor((dataSize - 1) / dimensionToExtent)
```
0x64: @ybsung  What about this?
`targetIndex = Math.min(index + dimensionToExtent, dataSize - 1)`
This has a little different behavior from the PRed code and it may be sufficient to work if the behavior is acceptable.

germboy:
It looks like the issue here is a result of a fallback in `VirtualListBase` where spotlight is used (via `Spotlight.move()`) in order to set focus to the nearest "next" eligible target, but the scroller is not triggered, so the focused item is not scrolled into view. This occurs when the next target in the 5way direction pressed is not immediately next to the currently focused item (whether it is disabled or does not exist - as in our use-case).

In this instance, we don't want to use our current index to calculate which index is next because based on the size of the list - and the size of the items used - the next calculated index may not be the index of the dom that spotlight would spot next. This was more of an issue when disabled components could not be focused, but we should still respect those rules, since the rest of the logic used here still supports those rules - allowing `VirtualList` to be more flexible. Perhaps we could entirely remove any need for that support, but we'd need to refactor a bit more.

For now, the issue can be resolved by getting the index of the newly-focused item (caused by the call to `Spotlight.move()` and passing that to the accelerator method so it can be processed and the dom can be scrolled into view. Could you look at this branch I created and tell me your thoughts?
https://github.com/enactjs/enact/compare/develop...feature/ENYO-6051-jeremythomas?expand=1

ybsung:
Hi @germboy 
I agree with your opinion. It seems to me that your PR is much flexible and maintainable. Let's close this PR and use yours. I verified yours. Thanks for your detail comment.

Closing in favor of https://github.com/enactjs/enact/pull/2377

PR #2369, opened 2019-06-17
PLAT-80381: Apply basic spatial-navigation process model and api for spotlight

jeonghee27:
### Checklist

The standardization of spatial navigation is progressing for the lightweight spotlight.
(https://drafts.csswg.org/css-nav-1/)
We need to check up the standard is appropriate to implement lightweight spotlight.
In the POC, We aim to apply the latest standard to spotlight and replace the spotlight code to spatial-navigation-polyfill.
As the first step of this POC, In this ticket, the basic arrow key function of spotlight will be replaced by precess model of standard spatial navigation-polyfill.

### Resolution
Below features was implemented. 
1) Include spatial-navigation-polyfill to spotlight package.
2) Remove replaceable APIs in spotlight/target.js
3) Remove unused APIs in spotlight/utils.js
4) Replace Spotlight.move APIs in moonston componenets
(Default spatial-navigation processing is recommended than explicit invoke of Spotlight.move api.)
5) Support container configutation. 
(leave-for, reverse, enter-to :[default, last-focused], restrict-self)
6) Support pause and accelerator.

### Additional Considerations

### Links
PLAT-80381


packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
```diff
@@ -546,16 +546,9 @@ const Decorator = hoc(defaultConfig, (config, Wrapped) => {
...
-			const direction = getDirection(ev.keyCode);
-
-			if (!direction) return;
-
-			this.handleDirectionalKey(ev);
-
+		handleContainerBlur = (ev) => {
 			// if focus moves outside the popup's container, issue the `onClose` event
-			if (Spotlight.move(direction) && !this.containerNode.contains(Spotlight.getCurrent())) {
+			if (!this.containerNode.contains(Spotlight.getCurrent())) {
```
0x64: I'm not sure that `Spotlight.getCurrent()` is valid here since this handler is called on `blur`.

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -309,16 +289,16 @@ const ExpandableItemBase = kind({
 				// case here in which the children of the container are spottable and the
 				// ExpandableList use case which has an intermediate child (Group) between the
 				// spottable components and the container.
-				if (autoClose && onClose && isUp(keyCode) && wouldDirectionLeaveContainer('up', target)) {
-					onClose();
-					ev.nativeEvent.stopImmediatePropagation();
-					ev.preventDefault();
-				} else if (isDown(keyCode) && wouldDirectionLeaveContainer('down', target)) {
+				if (autoClose && onClose && isUp(keyCode)) {
+					target.getSpatialNavigationContainer().addEventListener('navnotarget', onClose, {once: true});
```
0x64: The handler will be removed once it is called. Is it intentional?

packages/moonstone/Panels/Panel.js
```diff
@@ -173,7 +173,7 @@ const PanelBase = kind({
 		delete rest.hideChildren;
 
 		return (
-			<article role="region" {...rest} aria-labelledby={headerId} ref={spotOnRender}>
+			<article role="region" {...rest} aria-labelledby={headerId} ref={spotOnRender} style={{'--spatial-navigation-contain': 'contain', '--spatial-navigation-action': 'focus'}} >
```
0x64: I prefer to use a constant in this case, but it looks okay to see what is needed for now.

packages/moonstone/Popup/Popup.js
```diff
@@ -489,6 +482,18 @@ class Popup extends React.Component {
 		}
 	}
 
+	handleOnNavNoTarget = (ev) => {
+		const {onClose} = this.props;
+		if (onClose && this.pressUpKey) {
+			ev.preventDefault();
+			// stop propagation to prevent default spotlight behavior
```
0x64: What about adding some comments on what would happen when `preventDefault` and `stopProgatation` here?

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -208,55 +208,6 @@ class ScrollButtons extends Component {
 		Spotlight.focus(isPrev ? this.prevButtonRef.current : this.nextButtonRef.current);
 	}
 
-	focusOnOppositeScrollButton = (ev, direction) => {
```
0x64: Some UX requirements will be missing by removing this code.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -602,18 +610,35 @@ class ScrollableBase extends Component {
...
+				if (overscrollEffectOn.arrowKey) {
+					if (!element) {
+						const {horizontalScrollbarRef, verticalScrollbarRef} = this.uiRef.current;
 
-					if (!(horizontalScrollbarRef.current && horizontalScrollbarRef.current.getContainerRef().current.contains(element)) &&
-						!(verticalScrollbarRef.current && verticalScrollbarRef.current.getContainerRef().current.contains(element))) {
-						this.checkAndApplyOverscrollEffectByDirection(direction);
+						if (!(horizontalScrollbarRef.current && horizontalScrollbarRef.current.getContainerRef().current.contains(element)) &&
+							!(verticalScrollbarRef.current && verticalScrollbarRef.current.getContainerRef().current.contains(element))) {
+							this.checkAndApplyOverscrollEffectByDirection(direction);
```
0x64: This code will be called in different conditions from the original intention.

packages/spotlight/SpotlightContainerDecorator/SpotlightContainerDecorator.js
```diff
@@ -168,6 +168,15 @@ const SpotlightContainerDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			 */
 			spotlightMuted: PropTypes.bool,
 
+			/**
+			 * When `grid`, the grid navigation rule apply to the container.
+			 *
+			 * @type {String}
+			 * @public
+			 */
+			spotlightRule: PropTypes.string,
```
0x64: I wonder what are valid values for this prop.

packages/moonstone/Panels/Panel.js
jeonghee27: OK, I'll fix it in the next PR

packages/moonstone/Popup/Popup.js
jeonghee27: I'll add more annotation now.

packages/moonstone/Scrollable/ScrollButtons.js
jeonghee27: I intended to use `spotlightRule="grid"` option to move focus oppositeButton.
But I missed the code for normal scroller. I'll fix it now.

packages/moonstone/Scrollable/Scrollable.js
jeonghee27: Before is 
`if (overscrollEffectOn.arrowKey && !(element ? getTargetByDirectionFromElement(direction, element) : null)) `
After is
`if (overscrollEffectOn.arrowKey) && (!element) `
And, 
`if (overscrollEffectOn.arrowKey && element && getTargetByDirectionFromElement(direction, element) `
is covered on onNavNoTarget handler. 
So, I think it's same.

Did I miss something?

packages/moonstone/ExpandableItem/ExpandableItem.js
0x64: The expandable container is closed but not unmounted, then the container could be open again. Then, the handler would be required again.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
0x64: As I understand, when `Spotlight.getCurrent` is called, the returned value will be always `null`. So I think that we actually don't need the latter condition here. That was my question.

packages/moonstone/ExpandableItem/ExpandableItem.js
jeonghee27: Yes, It need unmount. I missed it , I'll fix it now.

packages/moonstone/ContextualPopupDecorator/ContextualPopupDecorator.js
jeonghee27: `Spotlight.getCurrent()` will return next target element instead `null`. because _current element will set on navbeforefocus.

packages/moonstone/Scrollable/Scrollable.js
0x64: Looks right.

packages/moonstone/Panels/Panel.js
0x64: 👍

packages/moonstone/ExpandableItem/ExpandableItem.js
jeonghee27: I tried to call removeEventListener when the onHide() is called.

To call removeEventListener, I need store the handler function in some variable.
But It's impossible with current component structure Because ExpandableItem is stateless (kind based) component. :(
Using state component is solution, I will takes more time. 
Thus, I'll fix it at the end of POC.

0x64:
Looks good.

PR #2373, opened 2019-06-18
PLAT-79871: Fix spotlight support for Dropdown

germboy:
### Checklist

Added/fixed focus management for `Dropdown`

### Resolution
- When dropdown opens, the previously-selected item (if available) will be scrolled to (vertically-centered) and focused
- Pressing 5way horizontally will close the popup and move focus to the next eligible horizontal target
- Pressing 5way vertically (when at the top/bottom bounds of the `Dropdown` list) will not allow focus to leave the list
- Pointer can be used to begin an interaction with other components while a `Dropdown` is opened, effectively closing the `Dropdown` (removing the scrim)

### DIFF
--- a/packages/moonstone/Dropdown/Dropdown.js
+++ b/packages/moonstone/Dropdown/Dropdown.js
@@ -19,6 +19,7 @@
 import kind from '@enact/core/kind';
 import EnactPropTypes from '@enact/core/internal/prop-types';
 import {handle, forKey, forward} from '@enact/core/handle';
+import {I18nContextDecorator} from '@enact/i18n/I18nDecorator';
 import Spotlight from '@enact/spotlight';
 import Changeable from '@enact/ui/Changeable';
 import Toggleable from '@enact/ui/Toggleable';
@@ -56,6 +57,7 @@ const compareChildren = (a, b) => {
 	return true;
 };
 const isSelectedValid = ({children, selected}) => Array.isArray(children) && children[selected] != null;
+const scrollOffset = 2;
 
 const DropdownButton = kind({
 	name: 'DropdownButton',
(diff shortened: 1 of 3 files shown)


packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -140,6 +151,54 @@ const DropdownList = Skinnable(
...
+	}
+
+	componentDidMount () {
+		// eslint-disable-next-line react/no-find-dom-node
+		this.node = ReactDOM.findDOMNode(this);
+		Spotlight.set(this.node.dataset.spotlightId, {leaveFor: {up: '', down: ''}});
+	}
+
+	componentDidUpdate () {
+		if (isSelectedValid(this.props)) {
```
ryanjduffy: I'm assuming this should only on open so we should add a guard to that end. I would suggest refactoring into a new method too. We seem to muddy up lifecycle methods over time by mixing concerns and starting with separate methods will help future us keep things clear.

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -249,6 +308,11 @@ const DropdownBase = kind({
 	},
 
 	handlers: {
+		onKeyDown: handle(
+			forward('onKeyDown'),
+			(ev, props) => forKey('left', ev, props) || forKey('right', ev, props),
```
ryanjduffy: I think we'll need a bit more logic to handle this safely. For example, if you hover a paging control and press 5-way left (or right in RTL), the dropdown will close but should focus an item.

packages/moonstone/Dropdown/Dropdown.js
germboy: Point taken about refactoring into a new method.

I didn't add the guard for `open`, since the list component is only mounted/updated in the popup when the list is open, so it seemed redundant. Are there cases where it is still needed?

ryanjduffy: It's unlikely but possible to break it. To test it, I added a `setInterval` in a view holding a drop down and used it to toggle the `width` every half second. When the popup was open, focus would get reset and rescrolled each time.

packages/moonstone/Dropdown/Dropdown.js
```diff
@@ -167,20 +169,15 @@ class DropdownList extends React.Component {
 		// eslint-disable-next-line react/no-find-dom-node
 		this.node = ReactDOM.findDOMNode(this);
 		Spotlight.set(this.node.dataset.spotlightId, {leaveFor: {up: '', down: ''}});
+		this.isScrolledIntoView = false;
```
ryanjduffy: It's possible that `getDerivedStateFromProps` could assist here rather than using a class member. Something to consider in a future refactor. :)

PR #2376, opened 2019-06-18
ENYO-6064: Derive groupId from displayName, when one is not set

JayCanuck:
### Checklist

* The Sampler `select` knob doesn't assign a groupId when none if found. As a result, `select` knobs with no explicit group are appearing in an "Other" category. This can be observed in the Moonstone->Icon story.

### Resolution
* Add a fallback to use `displayName` as the default `groupId` value when not found. This is the technique already used on the other knob proxies.

### DIFF
--- a/packages/sampler/stories/qa/Spotlight.js
+++ b/packages/sampler/stories/qa/Spotlight.js
@@ -203,7 +203,6 @@ class DisableTest extends React.Component {
 }
 
 class PopupFocusTest extends React.Component {
-	static groupId = 'Popup'
 	static propTypes = {
 		noAnimation: PropTypes.bool,
 		noAutoDismiss: PropTypes.bool,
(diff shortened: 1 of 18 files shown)


packages/sampler/stories/qa/Spotlight.js
```diff
@@ -203,6 +203,7 @@ class DisableTest extends React.Component {
 }
 
 class PopupFocusTest extends React.Component {
+	static groupId = 'Popup'
```
ryanjduffy: Why use `groupId` here vs `displayName`?

Djspaceg: I'd expect that this should indeed be `displayName`, so I'm curious about this too.

JayCanuck: The PopupFocusTest is passed in to the knobs later on as the config object. Given the knob syntax uses `groupId` as the default value and `displayName` as the fallback, I figured it made sense to go this route. Additionally, given `displayName` has both implicit-based and API-based implications, I thought it made sense to designate the `PopupFocusTest` as part of the "Popup" group (rather than assigning it a `displayName` of "Popup"). If we really want to go the `displayName` route, can always use `static displayName = 'PopupFocusTest'`; just thought it looked and made more sense to have the knob category named simplistically as "Popup"

Djspaceg: I don't actually see a problem with using `displayName` here. In terms of consistency, I think the benefits of displayName are better, as well as linting. The `groupId` is really intended to be more of an internal key, abstracted away by our enact knobs. "Popup" makes sense too since all of the knobs flow into the Popup component, instead of into the wrapping div or the buttons. If they were assigned into those elements, I'd probably have a different opinion on the name.

JayCanuck: Looking at the PopupFocusTest, that makes sense. Simplifying even further, I wonder if could we use Popup itself for the config object rather than the PopupFocusTest?

Djspaceg:
There's a lot of clean-up here. Thanks for taking this on @JayCanuck

PR #2379, opened 2019-06-19
ENYO-6067: Fix ExpandableItem to guard against undefined event handlers

Sangwook-Lee:
### Checklist

No action is taken even if clicking or pressing any keys when there is no `onClose` if component in the app.

### Resolution
added exception about onClose handler

### Additional Considerations

### Links
ENYO-6067


packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -332,7 +332,7 @@ const ExpandableItemBase = kind({
 			// When disabled, don't attach an event
 			if (!disabled) {
 				if (open) {
-					onClose(ev);
+					onClose && onClose(ev);
```
webOS101: Should this guard `onOpen`, too?

Sangwook-Lee: @webOS101 , Currently, there's no problem.

webOS101: There would be, though.  Since it's not required, please add the same guard below.

Sangwook-Lee: Done

packages/moonstone/ExpandableItem/tests/ExpandableItem-specs.js
```diff
@@ -36,6 +36,42 @@ describe('ExpandableItem', () => {
...
+					</ExpandableItemBase>
+				);
+				const item = expandableItem.find('LabeledItem');
+				item.simulate('click');
+				const expected = 1;
+				const actual = handleOpen.mock.calls.length;
+
+				expect(actual).toBe(expected);
+			});
+		});
```
Sangwook-Lee: It just tests calling event handlers. I think we can test user scenarios in `qa-sampler` below.

packages/sampler/stories/qa/ExpandableList.js
```diff
@@ -162,4 +162,16 @@ storiesOf('ExpandableList', module)
 		() => (
 			<ExpandableListWithAddedChildren />
 		)
+	)
+	.add(
```
ryanjduffy: `ExpandableList` doesn't produce this bug since it is wrapped by `Toggleable` which provides the handlers. As a result, I'm going to remove the sample and rely on the unit tests to address the concern.

packages/moonstone/ExpandableItem/ExpandableItem.js
```diff
@@ -332,9 +332,9 @@ const ExpandableItemBase = kind({
...
```
ryanjduffy: Although this is valid code, it does violate our linting rules and needs to be updated.

ryanjduffy:
There isn't a change log but I think that is okay given the narrowness of the fix.

PR #2381, opened 2019-06-20
PLAT-80081: Scroll and Move Spotlight to the paging control properly via 5-way Down if the current item sticking to the top is only spottable

ybsung:
### Checklist

When
* only the first item is spottable in Scroller,
* the item gets spot,
* the item locates at (0,0), 
* the item height is larger than the list height,
* paging controls can be spottable, and
* pressing a Down key,

Scroller does not scroll and Spotlight stays on the first item.

### Resolution

This was the edge case for the requirement. I fixed the condition, then it worked.

### Additional Considerations

I added the qa Scroller sampler for the TC.

### Links

PLAT-80081

### DIFF
--- a/packages/moonstone/Scroller/Scroller.js
+++ b/packages/moonstone/Scroller/Scroller.js
@@ -312,7 +312,7 @@ class ScrollerBase extends Component {
 				pos = isVerticalDirection ? top : left,
 				max = isVerticalDirection ? maxTop : maxLeft;
 
-			if (pos > 0 && pos < max) {
+			if (pos >= 0 && pos < max) {
 				this.props.scrollAndFocusScrollbarButton(direction);
 			}
 		}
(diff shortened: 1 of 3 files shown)


packages/moonstone/Scroller/Scroller.js
```diff
@@ -312,7 +312,7 @@ class ScrollerBase extends Component {
 				pos = isVerticalDirection ? top : left,
 				max = isVerticalDirection ? maxTop : maxLeft;
 
-			if (pos > 0 && pos < max) {
+			if (pos >= 0 && pos < max) {
```
0x64: `pos` can be `max` also. `maxTop` is `scrollHeight` - `clientHeight` and it is the maximum position. So I wonder it is possible that the issue occurs on the last item, too.

ybsung:
@0x64 
>  So I wonder it is possible that the issue occurs on the last item, too.
As you mentioned, I found there was also a problem with the last item. So I updated the condition. Thanks.

0x64:
LGTM

PR #2382, opened 2019-06-20
PLAT-79069: Update Header with Input focus

teckliew:
### Checklist

Update Header with Input to match new designs. 

### Resolution
Added a span, `.inputHighlight`, with the input value and background color to mimic highlighted text when the `InputDecorator` receives focus.
`.inputHighlight`'s opacity and background color are left up to `Header` to handle otherwise they are transparent in normal input because in normal input, they are not needed.

### Links
PLAT-79069

### DIFF
--- a/packages/moonstone/styles/variables.less
+++ b/packages/moonstone/styles/variables.less
@@ -268,7 +268,7 @@
 @moon-input-decorator-small-height: @moon-input-small-height;
 @moon-expandableinput-input-margin: 6px @moon-spotlight-outset;
 @moon-input-highlight-height: (@moon-input-height - 3px);
-@moon-input-highlight-max-width: 379px;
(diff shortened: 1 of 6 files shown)


packages/moonstone/styles/variables.less
```diff
@@ -266,6 +267,9 @@
 @moon-input-small-height: @moon-button-small-height;
 @moon-input-decorator-small-height: @moon-input-small-height;
 @moon-expandableinput-input-margin: 6px @moon-spotlight-outset;
+@moon-input-highlight-height: (@moon-input-height - 3px);
+@moon-input-highlight-max-width: 379px;
```
ryanjduffy: @teckliew + @Djspaceg - any concerns with these not being divisible by 3?

PR #2384, opened 2019-06-24
PLAT-79985: Fix VirtualList event bubbling

germboy:
### Checklist

Keydown events in `VirtualList` do not bubble expectedly.

### Resolution
Fix `keydown` event bubbling when pressing 5way directional keys from either a list item or scroll buttons.

The changes in this PR have the potential to affect a number of 5way behaviors (all for the better hopefully 🤞 ). Since we need to be aware of `keydown` events on the scroll buttons, we need to change the dom where we are listening for events - from the wrapper to the spotlight container. This change means we need to act on the capture phase of the event in order to prevent an application from being able to act on the event when attached to the `VirtualList`, as was the intent behind the existing behavior. We should spend time thoroughly testing this change.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -11,8 +11,6 @@ import React, {Component} from 'react';
 import {Scrollable, dataIndexAttribute} from '../Scrollable';
 import ScrollableNative from '../Scrollable/ScrollableNative';
 
-import scrollbarCss from '../Scrollable/Scrollbar.module.less';
-
 const SpotlightAccelerator = new Accelerator();
 const SpotlightPlaceholder = Spottable('div');
 
@@ -459,8 +457,13 @@ const VirtualListBaseFactory = (type) => {
 				} else {
 					const {repeat} = ev;
 					const {dimensionToExtent, isPrimaryDirectionVertical} = this.uiRefCurrent;
-					const isScrollButton = target.classList.contains(scrollbarCss.scrollButton);
 					const targetIndex = target.dataset.index;
+					const isScrollButton = (
+						// if target has an index, it must be an item so can't be a scroll button
+						!targetIndex &&
+						// if it lacks an index and is inside the scroller, it must be a button
+						target.matches(`[data-spotlight-id="${this.props.spotlightId}"] *`)
+					);
 					const index = !isScrollButton ? getNumberValue(targetIndex) : -1;
 					const {isDownKey, isUpKey, isLeftMovement, isRightMovement, isWrapped, nextIndex} = this.getNextIndex({index, keyCode, repeat});
 					const directions = {};
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -29,7 +29,13 @@ const
 	Native = 'Native',
 	// using 'bitwise or' for string > number conversion based on performance: https://jsperf.com/convert-string-to-number-techniques/7
 	getNumberValue = (index) => index | 0,
-	nop = () => {};
+	nop = () => {},
+	moveFocusStraight = ({direction, id}) => {
+		Spotlight.set(id, {straightOnly: true});
+		const moved = Spotlight.move(direction);
+		Spotlight.set(id, {straightOnly: false});
+		return moved;
+	};
 
 /**
  * The base version of [VirtualListBase]{@link moonstone/VirtualList.VirtualListBase} and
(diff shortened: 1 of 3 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -434,32 +451,62 @@ const VirtualListBaseFactory = (type) => {
...
+								ev.stopPropagation();
+
+								if (typeof nextTargetIndex === 'string') {
+									this.onAcceleratedKeyDown({keyCode, nextIndex: getNumberValue(nextTargetIndex), repeat, target});
+								}
+							}
+
 						}
+					} else if (
+						directions.right && repeat ||
```
0x64: Don't we need to test `Spotlight.move` for `directions.right`? For example, for scenes like two lists in a single screen in channel edit?

0x64:
Is there any reason to pass the prop to VirtualList directly rather than via ui/Scrollable? The very first commit is comfortable to me. :)

I found one changed behavior; when I `up` or `down` on scrollbars, spotlight no longer moves to the opposite button but moves into items. Looks side effect of moving handlers to a container node.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -434,32 +451,62 @@ const VirtualListBaseFactory = (type) => {
...
-					const index = getNumberValue(target.dataset.index);
-					const {isBackward, isForward, isLeftMovement, isRightMovement, isWrapped, nextIndex} = this.getNextIndex({index, keyCode, repeat});
-
-					if (nextIndex >= 0) {
-						ev.preventDefault();
-						ev.stopPropagation();
-						this.onAcceleratedKeyDown({isWrapped, keyCode, nextIndex, repeat, target});
+					const {repeat} = ev;
+					const {dimensionToExtent, isPrimaryDirectionVertical} = this.uiRefCurrent;
+					const isScrollButton = target.classList.contains(scrollbarCss.scrollButton);
```
ybsung: I concerned that we started adding some code related with scrollbar in `VirtualListBase` including `scrollAndFocusScrollbarButton`. That might increase the complexity of the `VirtualListBase`. If possible, it would be better to add the code related with scrollbar in not `VirtualListBase` but `ScrollableBase`.

ybsung:
Do we have the same behavior in Scroller like VirtualList with this PR?

If there are VirtualList and Scroller and they have the same items, then it looks same. So the VirtualList and the Scroller should generate same events with same behavior.

germboy:
@ybsung Currently, the behavior is different between `VirtualList` and `Scroller`. Ideally, you would receive the same events for both, but as part of our performance work, we were able to gain perf benefits by strictly handling focus (via key events) in `VirtualList`, rather than allowing the key events to be handled by spotlight. This ultimately means preventing key event bubbling - which is what prevents spotlight from also handling focus. We aren't able to get around this behavior difference by instead using react synthetic events, since that adds too much overhead when processing key events in rapid succession (like when holding a 5way key & scrolling a list) - causing poorly performing scroll performance.

That being the case, we only want to allow key event bubbling when focus attempts to leave the list - which includes scrollbars (when using `focusableScrollbar`) - this allows app developers to respond to key events only when focus leaves the list. This means we need to be aware of key events on the scroll buttons and conditionally prevent event bubbling on the `VirtualListBase`-level, by listening for events on the container. I agree, it's less than ideal - but it appears to be necessary to enable better performance. I open to any alternative thoughts or other ideas and concerns.

germboy:
@0x64 I needed to pass the prop to `VirtualList`, since `ui/Scrollable` (and all components in `ui`) should not be aware of spotlight and the `focusableScrollbar` prop. This additionally led to prop bleed in `ui` where the prop was unintended. While it worked fine in `moonstone/VirtualList` (and we could prevent prop bleed there), simply using a `moonstone/Scroller` would unintentionally pass the prop down to `ui/Scroller`.

packages/moonstone/VirtualList/VirtualListBase.js
germboy: I don't believe so. When we reach this point in the code (the `else if` in the line above), focus is on a scroll button. The conditionals here determine only when we should prevent event bubbling. In this line `directions.right && repeat`, we want to prevent bubbling on repeated events, since focus should stop at the border during a key-hold. If we release and press 5way right (remembering that "right" means "down" in a horizontal list) again then none of these conditions would be truthy, so we wouldn't prevent event bubbling. At that point, spotlight would process the key event and move focus in the direction pressed.

Let me know if this makes sense or if I'm overlooking an edge-case.

0x64: 👍 It makes sense.

0x64:
> since ui/Scrollable (and all components in ui) should not be aware of spotlight and the focusableScrollbar prop.
Good point. I missed it.

> Regarding the changed behavior, I had not noticed that - good catch. Strangely, it does not reproduce for me in a vertical list - only a horizontal list. I will try to fix this.
Really strange! I saw in a vertical list!

germboy:
@0x64 I was able to remedy the issue you found by configuring the container to only target "next" items in a straight direction, if the key event was triggered via a scroll button. This configuration immediately gets reverted after the call to `Spotlight.move()`, so it will not affect other 5way behavior within the list.

0x64:
@germboy 👍

ybsung:
LGTM

PR #2385, opened 2019-06-24
ENYO-6042: Add qa sample for VirtualList in panels

ryanjduffy:
### Checklist

We lack a sample to test VirtualList in Panels

### Resolution
Adds a sample in QA sampler to test variable length virtual lists in an instance of panels

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -48,7 +48,7 @@ const PanelsBase = kind({
 					</Column>] : children
 				}
 			</Panel> : children}
-		</Panels> : children
+		</Panels> : <div {...rest}>{children}</div>
 	)
 });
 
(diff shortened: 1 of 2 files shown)


Djspaceg:
@viodragon2, by "decorators" do you mean the Storybook Panel Header and Component description? If so, we now have the ability to remove those elements to get cleaner/clearer samples for Header, Panel, and Panels.

viodragon2:
More or less. The main idea here is that we want to "test" and not "show" how apps would use the component, so the environment we have set up should reflect that. Not sure if the current storybook decorators with options are sufficient enough to do such a task, but it's definitely a good start. And if I were to extend the argument, `qa-sampler` really doesn't require any metadata (i.e. title, description) presented in the rendered space, so I would consider removing an extra layer entirely if we can. Food for thought.

Djspaceg:
Our Storybook setup now easily has the capability to remove any of those layers. Just set the render function to use the `context` object argument and set one of the new options: 
```javascript
context: {
    noHeader: Boolean,
    noPanel: Boolean,
    noPanels: Boolean,
    panelsProps: {
        // Object of props to pass to Panels if you keep Panels.
        // Useful for setting `noCloseButton:true`
    }
}
```
https://github.com/enactjs/enact/blob/develop/packages/sampler/stories/default/Header.js#L31

PR #2388, opened 2019-06-25
ENYO-6061: Warn only once for validation

viodragon2:
### Checklist

`Slider` warns on every render if values/max are not divisible by steps.

### Resolution
Add functions in `validators` to display only once per validation.


teckliew:
Looks good otherwise I think there's missing validation for min and max as well.

PR #2392, opened 2019-06-26
ENYO-6073: Fix Scrollable to focus outside the container

viodragon2:
### Checklist

`scrollAndFocusScrollbarButton` function only accounts for vertical direction spotlight leave to focus to the closest scroller button. This results in an error where focus moving out horizontally from the container fails to find scroller to focus.

### Resolution
Checks both the vertical and horizontal direction and check if the scroller exists in the direction. A simple condition check whether scroller exists or not would probably be enough, but it's probably better to get the logic down correctly.

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -636,9 +636,15 @@ class ScrollableBase extends Component {
 			isHorizontalDirection = direction === 'left' || direction === 'right',
 			isVerticalDirection = direction === 'up' || direction === 'down';
 
-		this.onScrollbarButtonClick({isPreviousScrollButton, isVerticalDirection});
-
 		const {focusableScrollbar, direction: directionProp} = this.props;
+
+		this.onScrollbarButtonClick({
+			isPreviousScrollButton,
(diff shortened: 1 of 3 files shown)


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -633,15 +633,24 @@ class ScrollableBase extends Component {
 		const
 			isRtl = this.uiRef.current.state.rtl,
 			isPreviousScrollButton = direction === 'up' || (isRtl ? direction === 'right' : direction === 'left'),
-			isVerticalScrollBar = direction === 'up' || direction === 'down';
+			isHorizontalDirection = direction === 'left' || direction === 'right',
+			isVerticalDirection = direction === 'up' || direction === 'down';
 
-		this.onScrollbarButtonClick({isPreviousScrollButton, isVerticalScrollBar});
+		this.onScrollbarButtonClick({isPreviousScrollButton, isVerticalDirection});
```
ryanjduffy: We're changing the args to `onScrollbarButtonClick` but that function hasn't been updated to use the new member (which is relevant because we're passing as an obj rather than fn args).

PR #2393, opened 2019-06-28
PLAT-71282: Add NewVirtualList to support different height items

ybsung:
### Checklist

Needed a new feature to support different height items in VirtualList.

### Resolution

We created `moonstone/VirtualList.NewVirtualList` and `moonstone/VirtualList.NewVirtualListNative`(tentative name) based on `moonstone/VirtualList.VirtualList` to support different height items.

**Usage**
```
	<NewVirtualListNative
		dataSize={items.length}
		focusableScrollbar
		itemRenderer={this.renderItem}
		itemSize={{minSize: defaultItemSize, size: this.state.itemSize}}
		spacing={10}
	/>
```

The `itemSize` object with `minSize` and `size` properties should be passed to `NewVirtualList`. It means that an app should give the list all item height information with `size` property. If not, the list will render the items with `minSize` height.

`minSize`: The default item height(number) if `size` with an index is not defined
`size`: The real item height(array)

**Restriction**
* Not support grid list
* Not support horizontal list
* Not support `cbScrollTo` in `NewVirtualList`

### Additional Considerations

### Links

PLAT-71282

### Comments

To check a sample for `NewVirtualListNative`, please check https://github.com/enyojs/enact-internal-samples/pull/84 PR.


ybsung:

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -800,7 +800,7 @@ const listItemsRenderer = (props) => {
 	const {
 		cc,
 		handlePlaceholderFocus,
-		initItemContainerRef: initUiItemContainerRef,
+		itemContainerRef: initUiItemContainerRef,
```
0x64: It seems to be a bug fix not an implementation for a new feature. Am I correct?

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -363,12 +395,35 @@ const VirtualListBaseFactory = (type) => {
...
+				const
+					{dimensionToExtent, primary, secondary, variableGridPositions} = this,
+					extent = Math.floor(index / dimensionToExtent),
+					firstIndexInExtent = extent * dimensionToExtent,
+					primaryPosition = variableGridPositions[firstIndexInExtent] ? variableGridPositions[firstIndexInExtent].position : extent * primary.gridSize,
+					secondaryPosition = (index % dimensionToExtent) * secondary.gridSize;
+
+				return {primaryPosition, secondaryPosition};
+			} else {
+				const {dimensionToExtent, primary, secondary} = this,
```
0x64: You missed a new line. 🤓

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -363,12 +395,35 @@ const VirtualListBaseFactory = (type) => {
...
+					primaryPosition = Math.floor(index / dimensionToExtent) * primary.gridSize,
+					secondaryPosition = (index % dimensionToExtent) * secondary.gridSize;
+
+				return {primaryPosition, secondaryPosition};
+			}
+		}
+
+		// For NewVirtualList
+		getVariableGridPosition = (index) => {
```
0x64: Note: this function assumes that a list always scrolls from the starting point 0.

0x64: The name of this method indicates that it will return _**the starting position**_ corresponding to the input parameter `index`, but this returns _**the end position**_ in fact. It looks better to rename this method or to change the behavior.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -451,7 +506,11 @@ const VirtualListBaseFactory = (type) => {
 
 			primary.gridSize = primary.itemSize + spacing;
 			secondary.gridSize = secondary.itemSize + spacing;
-			thresholdBase = primary.gridSize * 2;
+			if (this.props.type === 'NewVirtualList') {
+				thresholdBase = primary.gridSize * Math.ceil(overhang / 2);
```
0x64: I believe this line is the right code as we intended originally. I want this for other types of lists, too.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -633,20 +693,65 @@ const VirtualListBaseFactory = (type) => {
...
+							newThresholdMin = variableGridPosition.position;
+							newThresholdMax = newThresholdMin + size + spacing;
+						} else {
+							const diffToFirstIndex = Math.ceil((variableGridPosition.position - pos) / gridSize);
+							newFirstIndex = firstRenderedIndex - (diffToFirstIndex + overhangBefore) * dimensionToExtent;
+							newThresholdMin = variableGridPosition.position - diffToFirstIndex * gridSize;
+							newThresholdMax = newThresholdMin + gridSize;
+						}
+					} else { // calculate the first index based on assuming that all items have minimal size
+						const firstExtent = Math.max(
```
0x64: Let's refine this redundant code later.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -660,6 +765,60 @@ const VirtualListBaseFactory = (type) => {
...
+						childNode.style.transform = `translate3d(0, ${variableGridPositions[index].position}px, 0)`;
+					}
+				}
+
+				// Update threshold
+				this.threshold.min = firstIndex === 0 ? -Infinity : this.getVariableGridPosition(firstIndex + numOfUpperLine * 1);
+				this.threshold.max = lastIndex === maxFirstIndex ? Infinity : this.getVariableGridPosition(firstIndex + (numOfUpperLine + 1));
+
+				// Update scroll bounds
+				if (variableGridPositions.filter(Boolean).length === dataSize) { // all item sizes are known
```
0x64: We already assume that there is no jumping, so we can replace this comparison with `variableGridPositions.length === dataSize` or some other simpler code. But I think it looks better to leave this code as is. 👍

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -660,6 +765,60 @@ const VirtualListBaseFactory = (type) => {
...
+					this.scrollBounds.scrollHeight = variableGridSizes.reduce((acc, cur) => acc + cur, 0) + (dataSize - 1) * spacing;
```
0x64: If a list scrolls from the starting point 0 as we assume, we're able to get the last position from the last entry of size's array.

packages/moonstone/VirtualList/VirtualListBase.js
ybsung: Yes. Curious why we could not find this bug until now. Anyway NewVirtualList will work after fixing this bug. So I added this patch to my PR.

packages/ui/VirtualList/VirtualListBase.js
ybsung: Yeah, I missed it.

packages/ui/VirtualList/VirtualListBase.js
ybsung: Good point! I will rename `getVariableGridPosition` as `getVariableGridBottomPosition`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: `Math.ceil(overhang / 2)` is equal to `2` if this.props.type !== 'NewVirtualList'.
So let's remove `if`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: Let's use `getVariableGridBottomPosition` function to get the last position.

packages/ui/VirtualList/VirtualListBase.js
ybsung: With our more discussion, it would be better to fix this code as `variableGridPositions.length === dataSize`.

packages/ui/VirtualList/VirtualListBase.js
ybsung: We found there was a problem using`getVariableGridBottomPosition`. Items were not shown properly. Let's try to fix this issue later when enhancing performance.

PR #2395, opened 2019-07-01
Enyo 6073: Added QA Sample to Scroller to test focus in a container

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/qa/Scroller.js
+++ b/packages/sampler/stories/qa/Scroller.js
@@ -127,34 +127,6 @@ class ScrollerWithLargeContainer extends React.Component {
 	}
 }
 
-class ScrollerWithFocusInContainer extends React.Component {
-	constructor (props) {
-		super(props);
-	}
-
-	render = () => {
-		const {...rest} = this.props;
-
-		return (
-			<Panel {...rest} >
-				<Button style={{display: 'inline-block'}} small>focus to me</Button>
-				<Scroller focusableScrollbar style={{height: '400px', width: '300px', display: 'inline-block'}}>
-					<Item>button1</Item>
-					<Item>button2</Item>
-					<Item>button3</Item>
-					<Item>button4</Item>
-					<Item>button5</Item>
-					<Item>button6</Item>
-					<Item>button7</Item>
-					<Item>button8</Item>
-					<Item>button9</Item>
-					<div>Test Test Test Test Test Test </div>
-				</Scroller>
-			</Panel>
-		);
-	}
-}
-
 storiesOf('Scroller', module)
 	.add(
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Scroller.js
```diff
@@ -126,6 +127,34 @@ class ScrollerWithLargeContainer extends React.Component {
 	}
 }
 
+class ScrollerWithFocusInContainer extends React.Component {
+	constructor (props) {
+		super(props);
+	}
+
+	render = () => {
+		const {...rest} = this.props;
```
webOS101: This can be simplified to just do `{...this.props}` in line 139.

packages/sampler/stories/qa/Scroller.js
```diff
@@ -126,6 +127,34 @@ class ScrollerWithLargeContainer extends React.Component {
 	}
 }
 
+class ScrollerWithFocusInContainer extends React.Component {
+	constructor (props) {
```
webOS101: Don't need to have an empty `constructor`, can remove these lines.

packages/sampler/stories/qa/Scroller.js
```diff
@@ -126,6 +127,34 @@ class ScrollerWithLargeContainer extends React.Component {
...
+		super(props);
+	}
+
+	render = () => {
```
webOS101: Also, I think we should not be using arrow functions for `render`.  This is simple enough we can convert this into an SFC.  We don't need any state here.

packages/sampler/stories/qa/Scroller.js
```diff
@@ -282,6 +311,12 @@ storiesOf('Scroller', module)
 			<ScrollerWithLargeContainer />
 		)
 	)
+	.add(
+		'With Focus in Container',
+		() => (
+			<ScrollerWithFocusInContainer />
```
Djspaceg: I believe based on the above new code, that this entire changeset could be reduced down to the following, and just replace this line (line 317) with the below code:
```
			<div>
				<Button style={{display: 'inline-block'}} small>focus to me</Button>
				<Scroller focusableScrollbar style={{height: '400px', width: '300px', display: 'inline-block'}}>
					<Item>button1</Item>
					<Item>button2</Item>
					<Item>button3</Item>
					<Item>button4</Item>
					<Item>button5</Item>
					<Item>button6</Item>
					<Item>button7</Item>
					<Item>button8</Item>
					<Item>button9</Item>
					<div>Test Test Test Test Test Test </div>
				</Scroller>
			</div>
```

PR #2396, opened 2019-07-02
PLAT-78973: Updated Icon to support arbitrary strings as icon names

Djspaceg:
### Checklist

Icon names longer than one character that are outside of the iconsList were being treated as URLs.

### Resolution
Updated the checker code to look for URL-like symbols to determine "URL-ness" of a string.

### DIFF
--- a/packages/ui/Icon/Icon.js
+++ b/packages/ui/Icon/Icon.js
@@ -49,18 +49,16 @@ const mergeStyle = function (style, src) {
 };
 
 /**
- * Tests if a character is a single printable character
+ * Tests if a string appears to be a URI/URL.
  *
  * @function
  * @param	{String}	c	Character to test
  *
- * @returns	{Boolean}		`true` if c is a single character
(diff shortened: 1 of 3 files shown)


packages/ui/Icon/Icon.js
```diff
@@ -188,8 +186,8 @@ const Icon = kind({
 					} else if (iconProp.indexOf('0x') === 0) {
 						// Converts a hex reference in string form
 						icon = String.fromCodePoint(iconProp);
-					} else if (isSingleCharacter(iconProp)) {
-						// A single character is assumed to be an explicit icon string
+					} else if (!isUri(iconProp)) {
```
ryanjduffy: Let's add some unit tests the valid `icon` values since we're updating this.

Djspaceg: Added several tests, which include testing this new approach.

Djspaceg:
Added a bunch of tests to validate the new code is working and catches edge-cases.

PR #2399, opened 2019-07-08
PLAT-80385: Utilize spatial-navigation event

jeonghee27:
### Checklist

The standardization of spatial navigation is progressing for the lightweight spotlight.
(https://drafts.csswg.org/css-nav-1/)

We need to check up the standard is appropriate to implement lightweight spotlight.
In the POC, We aim to apply the latest standard to spotlight and replace the spotlight code to spatial-navigation-polyfill.

There are two kind of spatial navigation event, `navnotarget` and `navbeforefocus` which is helpful for lightweight spotlight.

### Resolution

As the second step of this POC,  simplify some enact code with spatial navigation event.
In more details, the following are implemented:
1) Transfer handling of direction-key down to spatial-navigation from virtualList. Use `navnotarget`, `navbeforefocus` event for case.
2) Remove unsuport configuration in container (`straitOnly`, `straightMultiplier`, `straightOverlapThreshold`).

### Additional Considerations

### Links
PLAT-80385

### DIFF
--- a/packages/moonstone/VideoPlayer/VideoPlayer.js
+++ b/packages/moonstone/VideoPlayer/VideoPlayer.js
@@ -63,8 +63,7 @@ const RootContainer = SpotlightContainerDecorator(
 
 const ControlsContainer = SpotlightContainerDecorator(
 	{
-		enterTo: '',
-		straightOnly: true
+		enterTo: ''
 	},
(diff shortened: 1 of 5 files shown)


0x64:
LGTM

PR #2400, opened 2019-07-08
PLAT-80499: Transition to external iLib NPM module

JayCanuck:
### Checklist

Note: CLI alias mitigates the breaking effect, but in reality it is still breaking.

* We've been using an embedded iLib package in `@enact/i18n/ilib` however the actual software is on NPM and so it's better to source from upstream.

### Resolution
* Add `ilib` peer dependency on `@enact/i18n` and `@enact/moonstone`, along with a corresponding dev dependency. This way app devs can be the top-level control over what version of ilib is used, and apps can take advantage of the [NPM alias feature](https://github.com/npm/rfcs/blob/latest/implemented/0001-package-aliases.md) (from NPM >=6.9.0) to use iLib variants such as `ilib-webos-tv` without any need for sourcecode changes.
* Removed old local iLib from `@enact/i18n/ilib`
* All import/require references that use iLib have been updated for the `ilib` external package.
* Updated Sampler to add `ilib@^14.2.0` dependency.

### Additional Considerations
* Requires https://github.com/enactjs/cli/pull/192  The tests on the PR will continue failing until it's merged.
* This RP will need to be updated once https://github.com/enactjs/dev-utils/pull/44 is merged/released for the updated `@enact/dev-utils` dependency in Sampler.


ryanjduffy:
For safety, we should also rerun travis on this PR after the cli and dev tools PRs land but before we merge this one.

JayCanuck:
In theory `ilib/lib/ilib` >> `ilib` looks ok.  Though any places we import ilib from `@enact/i18n` should be retained, since that pulls main ilib objects with our special Enact glue applied.

ryanjduffy:
That's a good point. We should probably be using `@enact/i18n` instead. I hadn't considered that.

ryanjduffy:
Appears we only import the root `ilib` from `i18n`. `moonstone` and `analytics` already only use `@enact/i18n`.

JayCanuck:
Given ilib has been working previously, I imagine any existing 'ilib/lib/ilib' references were probably ok (and I imagine the average dev would probably access ilib directly anyway), but there were certain instances where `@enact/i18n` were explicitly use that we should make sure to retain :+1:

ryanjduffy:
I pushed a commit (2d242a0) for the root import and noticed a docs problem i updated in a separate commit (063abd4).

PR #2403, opened 2019-07-09
ENYO-6101: Set the aria-label of IconButton in IncrementSlider

changgilee:
### Checklist

There is no aria-label when the value is min or max in IncrementSlider

### Resolution
Because we support focus to disabled item, aria-label of IconButton in IncrementSlider should be set even the value is min or max.

### Additional Considerations

### Links
ENYO-6101

### Comments


Baekwoo-s:
@changgilee Could you check the conflicts?
Thank you.

Baekwoo-s:
LGTM

PR #2406, opened 2019-07-09
PLAT-81126: Fix VirtualList wrap knob configuration

germboy:
### Checklist

Since the storybook 5 update, addon knobs will convert string characters to html entities. This is causing problems in our `VirtualList` samples, where the value of the `wrap` prop set as `'noAnimation'` would be seen as `&#39;noAnimation&#39;`

This will ultimately lead to the `wrap` prop being passed an `undefined` value, causing seemingly incorrect component behavior.

### Resolution
Update the knob config to ensure the right value is passed to the component

### DIFF
--- a/packages/sampler/src/configure.js
+++ b/packages/sampler/src/configure.js
@@ -17,7 +17,6 @@ function config (stories, mod) {
 	// Set addon-knobs defaults
 	addDecorator(withKnobs({
 		// debounce: {wait: 500}, // Same as lodash debounce.
-		escapeHTML: false,
 		timestamps: true // Doesn't emit events while user is typing.
 	}));
 
--- a/packages/sampler/src/configure.js
+++ b/packages/sampler/src/configure.js
@@ -17,6 +17,7 @@ function config (stories, mod) {
 	// Set addon-knobs defaults
 	addDecorator(withKnobs({
 		// debounce: {wait: 500}, // Same as lodash debounce.
+		escapeHTML: false,
 		timestamps: true // Doesn't emit events while user is typing.
 	}));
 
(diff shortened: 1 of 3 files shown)


Djspaceg:
What if instead of changing the default Storybook behavior just updated these samples to match the default behavior? Are there cases in which we _want_ to be able to not escape HTML that this change is addressing? I suspect no, and think it may be a little more intuitive to not need to manually escape HTML in strings.

germboy:
I'm not opposed to changing the samples instead. My thinking was that I know the default storybook `escapeHTML` change negatively affects multiple samples, and reverting to the pre-5.x behavior doesn't negatively affect any of our samples (that I know of, correct me if I'm wrong). I'm not sure we need the default behavior currently - I suppose a use-case could pop up at any time though. This seemed like the most straight-forward fix, but I don't feel strongly either way. I'll concede to the one who feels most passionate about either direction

Djspaceg:
My personal opinion is that changing a default should be something we do only when it's necessary for supporting some feature that the default is in conflict with. If there's a way to make our samples "comply" with the prescribed default, that's the ideal solution for support and maintenance. I feel like it will make our samples slightly more durable and slightly less accident-prone. I'm in favor of updating the samples.

webOS101:
I think we already do a good job of escaping the HTML when we need to and, in cases where we might allow through HTML, it's because we wanted to do that anyhow.  I think this is a case of Storybook being overly protective.

Djspaceg:
Are there any cases where we _want_ to allow HTML to pass through? Should there be? My expectation is "no" to both.

webOS101:
It's not really HTML (as in tags) it's HTML escaping things like quotes and what-not.  We don't want text in children to be interpolated into escape sequences.

PR #2410, opened 2019-07-10
PLAT-80789: Add ui/Audio and ui/Video

teckliew:
### Checklist

Add ui/Audio and ui/Video

### Additional Considerations
Extracted the preloading logic from `Video` to `ui/Media`. It's not a fully flushed out merge but using `MediaBase` and `MediaDecorator` that'll handle preloading is good enough for now. Future work might be needed to have preloading built into `Media`.

### Links
PLAT-80789


ryanjduffy:
I (finally) sorted out all the stuff I broke trying to think/work through how this could play out. Let's talk through it more online.

https://github.com/enactjs/enact/compare/feature/PLAT-80789-ui-video-audio...feature/PLAT-80789-ryan?expand=1

packages/ui/Media/Media.js
```diff
@@ -312,7 +282,7 @@ class Media extends React.Component {
 				{...this.handledMediaEvents}
 				ref={this.mediaRef}
 			>
-				{source}
+				{children && children.type === 'source' ? children : source}
```
ryanjduffy: I think we need to revisit this logic. I'd suggest favoring `source` over `children` since it is more explicit in its purpose. Also, I'd suggest dropping the `type` check since it's possible that authors might want to abstract source selection behind a custom component (e.g. passing a generic path which the component resolves to the actual path and returns a `<source>`).

I've added a test for multiple sources which currently fails but I would pass with the following.

```
{source || children}
```

@teckliew Thoughts?

teckliew: Yep sounds good! Updated with `{source || children}`

PR #2413, opened 2019-07-11
PLAT-79873: Updated Global font to 3-languages font name(s)

Djspaceg:
### Checklist

The `LG Smart UI Global` collection of fonts has been removed (renamed) and replaced by `LG Smart UI AR HE TH` family of fonts.

### Resolution
These fonts have been updated to the new names.

### Additional Considerations
There is a change in font-weight definition. Previously we had `LG Smart UI-Light` and `LG Smart UI-Regular` yet we support at least 3 different weights: light, regular, and bold. These new fonts come in non-matching weights: `LG Smart UI AR HE TH Regular` and `LG Smart UI AR HE TH SemiBold`. Previously, we'd mapped light to light, regular to regular, and regular to bold. (regular and bold would look the same). Now, these new fonts are being mapped to regular to light, regular to regular, and semibold to bold. (light and regular will use the same font-weight, and bold will stand out as different, as light did in the past). This means that some text will change in apparent weight visually, even though no change has happened in the code.

I believe this change was intentional and we now are simply using the appropriate weight metadata for the thickness of the glyphs in each of the font files.

*Note:* I'm not sure the changelog entry is ideal. If someone could verify it or correct it, that would be super.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -10,7 +10,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 - `moonstone/Scroller` to properly move focus out of the container
 - `moonstone/VirtualList` to allow keydown events to bubble up when not handled by the component
 - `moonstone/IncrementSlider` to support aria-label when disabled
-- `Fonts` to use the updated names of global fonts available in the system
+- Fonts to use the updated names of global fonts available in the system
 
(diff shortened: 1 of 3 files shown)


Seonghyup:
@Djspaceg , I think we need to update the other `LG Display` fonts to `LG Smart UI` fonts. For example from `LG Display JP` to  `LG Smart UI JP`. As you can see, most `LG Display` fonts have been renamed in latest fonts version.

PR #2414, opened 2019-07-11
ENYO-6099: VirtualList to set focus correctly after an update

germboy:
### Checklist

When you call `.scrollTo({index, focus: true})` on a `VirtualList` after it has updated (after `componentDidUpdate`), the children may not have re-rendered yet, meaning the dom is not available and focus will not be set on the specified `index`, once the children are rendered and scrolled into view.

### Resolution
We can utilize the `this.handleUpdateItems` method of `VirtualList` in order to set focus after the list has been updated and the dom is available. We simply set the value of the index to be focused in the previous call to `this.focusByIndex`.


ybsung:
LGTM

PR #2417, opened 2019-07-12
Additional webOS platform detection, including a generic LG webOS fallback

JayCanuck:
Addendum to https://github.com/enactjs/enact/pull/2397

### Checklist

- webOS 5 and indeterminate versions of webOS would be ommitted as "webos" in platform detection

### Resolution
* Adds a webOS 5 detection entry as well as a generalized LG webOS catchall for indeterminate versions and versions with unexpected corresponding Chrome releases.

### DIFF
--- a/packages/core/platform/platform.js
+++ b/packages/core/platform/platform.js
@@ -55,6 +55,9 @@ const platforms = [
 	{platform: 'webos', regex: /Web0S;.*Safari\/538.2/, forceVersion: 2},
 	{platform: 'webos', regex: /Web0S;.*Chrome\/38/, forceVersion: 3},
 	{platform: 'webos', regex: /Web0S;.*Chrome\/53/, forceVersion: 4},
+	{platform: 'webos', regex: /Web0S;.*Chrome\/68/, forceVersion: 5},
+	// LG webOS of indeterminate version
+	{platform: 'webos', regex: /Web0S;/, forceVersion: -1},
 	// LuneOS
 	{platform: 'webos', regex: /LuneOS/, forceVersion: 0, extra: {luneos: 1}},
 	// Palm/HP/Open webOS
--- a/packages/core/platform/platform.js
+++ b/packages/core/platform/platform.js
@@ -59,12 +59,12 @@ const platforms = [
 	// LG webOS of indeterminate version
 	{platform: 'webos', regex: /Web0S;/, forceVersion: -1},
 	// LuneOS
-	{platform: 'webos', regex: /LuneOS/, forceVersion: 0, extra: {luneos: 1}},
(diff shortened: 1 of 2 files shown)


packages/core/platform/platform.js
```diff
@@ -55,6 +55,9 @@ const platforms = [
 	{platform: 'webos', regex: /Web0S;.*Safari\/538.2/, forceVersion: 2},
 	{platform: 'webos', regex: /Web0S;.*Chrome\/38/, forceVersion: 3},
 	{platform: 'webos', regex: /Web0S;.*Chrome\/53/, forceVersion: 4},
+	{platform: 'webos', regex: /Web0S;.*Chrome\/68/, forceVersion: 5},
```
webOS101: Since we saw a wild 'WebOS` (WHY CAPITAL W?!) should we do `/Web[0O]S....`?

Djspaceg: The zero is indeed odd. :donotwant:. `/[wW]eb[0O]S;...` Maybe we can talk to someone about this...

JayCanuck: The zero was to differentiate between old PalmHP webOS and LG webOS, so it makes sense historically (and out of our hands anyway).

`/[wW]eb[0O]S;...` does work, but I've since confirmed both 2020 TV and current PR/OSE uses `Web0S`, so I'm not sure where that single report came from.

If we really want to include support for `webOS` in LG to cover the use-case, I'd recommend the easier-to-read:
/(?:webOS|Web0S);.*Chrome\/68/` for webOS 5
/(?:webOS|Web0S);.*Chrome/` for LG webOS catchall

packages/core/platform/platform.js
```diff
@@ -55,6 +55,9 @@ const platforms = [
...
+	// LG webOS of indeterminate version
+	{platform: 'webos', regex: /Web0S;/, forceVersion: -1},
```
webOS101: Would `'unknown'` be better than `-1`?

JayCanuck: Depends.  We do want `platform.webos` to be truthy for conditional usage and whatnot. Introducing a string value seems odd given we've stuck to integers up til now (even using `0` for legacy webos),

webOS101: speaking of that, since `0` is falsy, it'll appear that the platform is 'legacy' but it't not totally clear that it's legacy webOS.  I wonder if `-1` works better in that case?

JayCanuck: I have no problem with that. Thoughts @ryanjduffy ?

ryanjduffy: I think the "truthy" check is important so should probably use -1 for any unknown or legacy version. We could add an "extras" field for the fallback if we felt it useful as we have for legacy.

PR #2418, opened 2019-07-12
PLAT-80194: Fix Scroller and VirtualList duplicate scrolling by page up/down

0x64:
### Checklist

VirtualList scrolls one more time on page up/down keys when an item is focused in pointer mode.

### Resolution
Since the focus disappears after scrolling by page keys, another handling logic for non-focused case in pointer mode performs once more. This makes a list scrolls twice far.
Basically, this happens `keydown` event is not consumed by the event handler which handles scroll by page. So, this PR adds simply `stopPropagation`.

### Links
PLAT-80194


ybsung:
LGTM

PR #2425, opened 2019-07-15
ENYO-6094: Reset spotlight accelerator when focus leaves VirtualList …

germboy:
…container

### Checklist

When focus leaves a `VirtualList` container via a 5way key press, the spotlight accelerator is not properly reset. This results in improper 5way behavior when the `VirtualList` regains focus. It will take 2 key presses before the list can respond to 5way key presses.

### Resolution
Reset the spotlight accelerator when focus leaves (or attempts to leave) the `VirtualList` container.


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -489,7 +490,7 @@ const VirtualListBaseFactory = (type) => {
 							const {dataSize, focusableScrollbar} = this.props;
 							const column = index % dimensionToExtent;
 							const row = (index - column) % dataSize / dimensionToExtent;
-							const isLeaving = directions.up && row === 0 ||
+							isLeaving = directions.up && row === 0 ||
```
ybsung: It seems to me that VirtualList worked well with `(directions.up || directions.right) && row === 0` condition. Spotlight could be on the 'x' close button when 5-way Right as well as 5-way Up is pressed on the last item.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -489,7 +490,7 @@ const VirtualListBaseFactory = (type) => {
...
 								directions.down && row === Math.floor((dataSize - 1) % dataSize / dimensionToExtent) ||
 								directions.left && column === 0 ||
 								directions.right && !focusableScrollbar && column === dimensionToExtent - 1;
```
ybsung: Even though `focusableScrollbar` is equal to `true`, Spotlight leaves from VirtualList if `verticalScrollbar` prop and `horizontalScrollbar` prop in `Scrollable` are `'hidden'` because there is no scrollbars in the list.

ybsung: Example
```
storiesOf('Moonstone', module)
	.add(
		'VirtualList.VirtualGridList',
		() => (
			<div style={{width: '100%', height: '100%'}}>
				<VirtualGridList
					dataSize={5}
					direction="horizontal"
					focusableScrollbar="true"
					itemRenderer={renderItem}
					itemSize={{
						minWidth: 180,
						minHeight: 50
					}}
					spacing={0}
					verticalScrollbar="hidden"
					horizontalScrollbar="hidden"
					style={{width: '100%', height: '50%'}}
				/>
				<GridListImageItem
					caption="test"
					subCaption="test"
					style={{position: 'static'}}
				/>
			</div>
		),
		{
			info: {
				text: 'Basic usage of VirtualGridList'
			}
		}
	);
```

packages/moonstone/VirtualList/VirtualListBase.js
germboy: Good point. I adjusted the behavior specified in the lines below (regarding considerations needed when `directions.right` is `true`), since there is more to consider under different scenarios.

packages/moonstone/VirtualList/VirtualListBase.js
germboy: Thanks, @ybsung - that's an excellent observation! It appears we've been overlooking whether or not the user explicitly hides the scrollbar :(
I've updated the PR to consider these scenarios. Can you review it again?

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -456,7 +456,7 @@ const VirtualListBaseFactory = (type) => {
 					ev.stopPropagation();
 				} else {
 					const {repeat} = ev;
-					const {dimensionToExtent, isPrimaryDirectionVertical} = this.uiRefCurrent;
+					const {dimensionToExtent, isPrimaryDirectionVertical, props: {isHorizontalScrollbarVisible, isVerticalScrollbarVisible}} = this.uiRefCurrent;
```
ryanjduffy: Is this a remnant from an early implementation? We're pulling these both in as props to this component now so accessing via the ref isn't necessary.

PR #2426, opened 2019-07-15
ENYO-6095: Fix VirtualGridList to not scroll unexpectedly when naviga…

germboy:
…ting to adjacent items

### Checklist

The scroller in a `VirtualGridList` can scroll unexpectedly when using 5way to navigate to an adjacent item while the previously focused item is still being scrolled into view.

### Resolution
When focus is being changed to the "next" item, we can check the row position of the current and next items and prevent programmatically initiating the scroll if they are the same. This applies for vertical and horizontal orientations.


ybsung:
LGTM. Travis warnings were already fixed in `develop` branch as I know.

PR #2430, opened 2019-07-16
ENYO-6109: Not try to scroll horizontally if the scroller can scroll only vertically

ybsung:
### Checklist

A Scroller jumped to the top when right key is pressed in the right most item of vertical scroller

### Resolution

If 5-way right is pressed on the right most item in the scroller and the scroller can scroll only vertically, the scroller should not jump. But he scroller tried to scroll horizontally due to no condition. Then the scroller jumped to the top because the `native` scroller do jump with wrong scroll x position. So I fixed the condition.

### Links

ENYO-6109

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


0x64:
@ybsung Could you please resolve conflicts?

ybsung:
@0x64 
I refreshed the branch.

PR #2431, opened 2019-07-16
Fix LabeledItem to pass marqueeOn prop

webOS101:

### Checklist

`LabeledItem` had no way to adjust when it would marquee. A use case existed where
it was used as a non-focusable component, so the default `marqueeOn` of `"focus"`
prevented it from being able to marquee.

### Resolution
Pass `marqueeOn` prop to label and contents.

### Additional Considerations
Another possible solution is to modify the component to detect `spotlightDisabled` and automatically set `marqueeOn` for the items.  This seemed a bit too magical and we'd need to make a similar change to all focusable components for consistency. This solution makes a very small change and gives developers the tools they need to accomplish the use case desired.

I decided to not pass along any additional marquee properties as they were not relevant to this request.

### Links
ENYO-6100

### DIFF
--- a/packages/ui/Marquee/MarqueeDecorator.js
+++ b/packages/ui/Marquee/MarqueeDecorator.js
@@ -226,9 +226,7 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			/**
 			 * Determines what triggers the marquee to start its animation.
 			 *
-			 * Valid values are `'focus'`, `'hover'` and `'render'`. The default is `'focus'`.
-			 *
-			 * @type {String}
+			 * @type {('focus'|'hover'|'render')}
 			 * @default 'focus'
 			 * @public
 			 */
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/LabeledItem` to pass `marqueeOn` prop to its contents
+
 ## [3.0.0-beta.1] - 2019-07-15
 
 ### Removed
--- a/packages/moonstone/LabeledItem/LabeledItem.js
+++ b/packages/moonstone/LabeledItem/LabeledItem.js
@@ -90,6 +90,15 @@ const LabeledItemBase = kind({
 		label: PropTypes.node,
 
 		/**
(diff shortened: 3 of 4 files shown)


packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -226,9 +226,7 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			/**
 			 * Determines what triggers the marquee to start its animation.
 			 *
-			 * Valid values are `'focus'`, `'hover'` and `'render'`. The default is `'focus'`.
-			 *
-			 * @type {String}
+			 * @type {('focus'|'hover'|'render')}
```
Djspaceg: Do we render out the possible values of typescript in our docs for cases like this? If so, neat!

webOS101: Something like that.  We do render the values into our typescript def. files and we do render out the array of possible values into the docs.  I think it's much more descriptive than just 'String'

Djspaceg: It is indeed much more descriptive. I'm just verifying that we're not _losing_ documentation/description on our website. Whether, for example, someone who doesn't use typescript would be able to read the list of acceptable values in some way.

webOS101: Nope, we don't lose anything by this.  The docs will list the acceptable values.

ryanjduffy: docs
image

.d.ts
image

codesandbox.io
image

Djspaceg: Cool. Thanks for sharing the pictures. They look different from what I was seeing.

ryanjduffy:
Did you investigate expanding support for `marqueeOn` in `MarqueeController`? I wonder if it should have API compatibility with `MarqueeDecorator` as it behaves as a proxy for multiple marquees.

I'm good with the current approach as is as expanding `MarqueeController` would be compatible with this API but curious if we should investigate further separately.

webOS101:
I did think about it briefly but there is already some weird coupling between the controller and the decorator and it seemed like teasing that out right now would possibly introduce a big chance of regression.  Marquee is already one of the biggest sources of bugs and simplifying the implementation would always be great... but right now didn't seem like the time.

webOS101:
By coupling, I meant overlapping concerns.  Of course they're coupled since Controller is supposed to control Decorator.

PR #2436, opened 2019-07-17
ENYO-6120: Req: hides border-bottom of the Header

Sangwook-Lee:
### Checklist

There're some cases that cannot use Header because of under line of that component.
Actually, GUI of some applications want to use Header without underline.

### Resolution
Added prop `hideLine` to hide `border-bottom` in the Header component.

### Links
ENYO-6120

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,11 +2,11 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
-## unreleased
+## [unreleased]
 
 ### Added
 
-- `moonstone/Header` prop `hideLine` to hide `border-bottom`
+- `moonstone/Panels.Header` prop `hideLine` to hide the bottom separator line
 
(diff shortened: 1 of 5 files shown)


packages/moonstone/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## unreleased
+
+### Added
+
+- `moonstone/Header` prop `hideLine` to hide `border-bottom`
```
webOS101: This is a bit too implementation-specific.  I'll fix it.

webOS101: Also, there is no `moonstone/Header` component.

Djspaceg: Nice change (collection of changes), @webOS101!

webOS101:
Oops, forgot root CHANGELOG.  Fixed.

PR #2437, opened 2019-07-17
PLAT-81341: Fix ProgressBar and Slider to match current designs

Djspaceg:
### Checklist

ProgressBar, Slider, and IncrementSlider need new colors.

### Resolution
New colors applied and cleaned up variables, abandoning old ones that didn't follow consistent naming conventions and removing unused variables. (Noted in the migration guide.)

### Additional Considerations
This includes a change to `ui/ProgressBar` which inserts a new DOM node in the stack, which decouples the styling of the bars from the knob. This allows the bars to be "grayed out" when disabled and leave the knob at full opacity, per the designs. I did experiment with several approaches that didn't necessitate adding a new node, but ultimately this seemed to be the cleanest.

### DIFF
--- a/packages/ui/ProgressBar/ProgressBar.js
+++ b/packages/ui/ProgressBar/ProgressBar.js
@@ -72,9 +72,8 @@ const ProgressBar = kind({
 		 * The following classes are supported:
 		 *
 		 * * `progressBar` - The root component class
-		 * * `bar`         - The background node (the empty part of the progressBar)
-		 * * `fill`        - The foreground of the progress bar (`progress`)
-		 * * `load`        - The `backgroundProgress` node
+		 * * `fill`        - The foreground node of the progress bar
+		 * * `load`        - The background node of the progress bar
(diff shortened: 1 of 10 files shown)


packages/ui/ProgressBar/ProgressBar.js
```diff
@@ -72,8 +72,9 @@ const ProgressBar = kind({
 		 * The following classes are supported:
 		 *
 		 * * `progressBar` - The root component class
-		 * * `fill`        - The foreground node of the progress bar
-		 * * `load`        - The background node of the progress bar
+		 * * `bar`         - The background node (the empty part of the progressBar)
```
ryanjduffy: We should note the addition of this in the change log

Djspaceg: I suppose we could. It didn't seem big enough for that.

ryanjduffy: As public API, I think it is important we manage these with the same rigor as other public APIs.

PR #2438, opened 2019-07-17
PLAT-81337: Remove Dropdown Button animation

teckliew:
### Checklist

Remove Dropdown Button animation

### Links
PLAT-81337

### DIFF
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -85,6 +85,7 @@ const ButtonBase = kind({
 		// `transparent` was intentionally excluded from the adove documented exported classes as it
 		// does not appear to provide value to the end-developer, but is needed by IconButton
 		// internally for its design guidelines, which differ from Button regarding `transparent`.
+		// Same for `pressed` which is used by Dropdown to nullify the key-press activate animation.
 		css: PropTypes.object,
 
 		/**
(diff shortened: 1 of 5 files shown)


ryanjduffy:
Often, when you have to expose something you want to be private as public, it points to some sort of underlying architectural or design issue (e.g. missing API or insufficient decomposition). What alternatives did you consider to resolve this?

Djspaceg:
We could add a "pressed" state HOC to Dropdown button which adds and duplicates the pressed state class, which adds runtime overhead. We could rebuild this Dropdown Button from a stateless to a stateful React.Component that stores the keypress state to apply a pressed class to get around the HOC overhead. These seemed like bad options.

Djspaceg:
For some background, when Moonstone was updated to use the `ui` architecture, we tried to only reveal the bare minimum of CSS classes necessary to get the job done. This was extremely restrictive; perhaps too restrictive. `Button` is a heavily used component with lots of use-cases and while it's possible to publicly expose classes for override, that may not serve the greater goal of having Moonstone be a structured opinionated UI library. With that, however, comes new design requirements, which despite our best efforts, sometimes need customizations on our strictly railed components.

Personally, I don't mind revealing the `.pressed` class as publicly accessible in Moonstone. That seems fine to me, but this semi-public approach seemed more acceptable.

PR #2439, opened 2019-07-18
PLAT-83080: Added a new type to Header: "dense"

Djspaceg:
### Checklist

* [x] At least one test case (sampler knob) is included for this feature or bug fix

A new type of `Header` is requested for `AlwaysViewing` types of panels.

### Resolution
A new type was added called "dense". This new type of header can be chosen using the
`<Header type="dense">` prop.

### DIFF
--- a/packages/moonstone/Panels/Header.module.less
+++ b/packages/moonstone/Panels/Header.module.less
@@ -61,13 +61,13 @@
 	// Standard Header
 	&.dense,
 	&.standard {
-		height: @moon-header-height-medium;
+		height: @moon-header-standard-height;
 
 		.title {
 			// The height is determined by the line-height below, but we want to maintain that as a
 			// relative unit (`em`) rather than dictate specific pixels to automatically accomodate
 			// font size changes.
-			line-height: @moon-medium-header-line-height;
+			line-height: @moon-header-standard-line-height;
 			margin-left: -3px; // Negative 3px here to allow the title text to horizontally align with the rest of the header text
 
 			.moon-locale-non-latin({line-height: @moon-non-latin-medium-header-line-height;});
@@ -102,7 +102,7 @@
 
 	// Compact Header
 	&.compact {
-		height: @moon-header-height-small;
+		height: @moon-header-compact-height;
 
 		.compactTitle {
 			margin-bottom: 3px;
@@ -111,7 +111,7 @@
 		.input,
 		.inputHighlight,
 		.title {
-			font-size: @moon-small-header-font-size;
+			font-size: @moon-header-compact-font-size;
 			display: inline;
 
 			.moon-locale-non-latin({line-height:  @moon-non-latin-small-header-line-height;});
--- a/packages/moonstone/Panels/Header.module.less
+++ b/packages/moonstone/Panels/Header.module.less
@@ -15,10 +15,6 @@
 	background-position: top left;
 	margin: 0 @moon-spotlight-outset;
 
-	// .titleAbove {
-	// 	.moon-super-header-text();
-	// }
-
 	// Remove the built-in h1 and h2 margins
 	.decorator,
 	.decorator .input,
@@ -43,11 +39,11 @@
 
 	.titleBelow,
 	.subTitleBelow {
-		.moon-sub-header-text();
+		.moon-header-title-below-text();
 	}
 
 	.subTitleBelow {
-		.moon-header-sub-title-below();
+		.moon-header-sub-title-below-text();
 	}
 
 	.headerComponents {
@@ -67,10 +63,10 @@
 			// The height is determined by the line-height below, but we want to maintain that as a
 			// relative unit (`em`) rather than dictate specific pixels to automatically accomodate
 			// font size changes.
-			line-height: @moon-header-standard-line-height;
+			line-height: @moon-header-standard-title-line-height;
 			margin-left: -3px; // Negative 3px here to allow the title text to horizontally align with the rest of the header text
 
-			.moon-locale-non-latin({line-height: @moon-non-latin-medium-header-line-height;});
+			.moon-locale-non-latin({line-height: @moon-non-latin-header-standard-title-line-height;});
 		}
 
 		.decorator,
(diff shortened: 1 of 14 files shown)


packages/moonstone/styles/mixins.less
```diff
@@ -270,9 +270,9 @@
 }
 
 .moon-sub-header-text() {
-	.moon-text-base(@moon-sub-header-font-size);
+	.moon-text-base(@moon-header-title-below-font-size);
 	font-weight: @moon-sub-header-font-weight;
-	font-size: @moon-sub-header-font-size;
+	font-size: @moon-header-title-below-font-size;
```
teckliew: Should we change the rest of `sub-header` to `header-title-below`?

packages/moonstone/styles/colors.less
```diff
@@ -145,6 +145,7 @@
 @moon-header-sub-title-color:         #a6a6a6;
 @moon-sub-header-text-color:          @moon-header-sub-title-color;
 @moon-sup-header-text-color:          @moon-text-color;
```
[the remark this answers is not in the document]
Djspaceg: This part of this file is heavily crufted. "sup" header (short for superscript) was removed ages ago, but parts still remain simply because we've not gone in to remove the old bits (low value), but in the interest of tightening this code up and having only meaningful code bearing on our maintenance cost, I think it's valid to remove these abandoned variables.

packages/moonstone/Panels/Header.module.less
```diff
@@ -88,9 +89,20 @@
...
 	}
 
+	// Dense Header
+	&.dense {
+		height: @moon-header-dense-height;
+
+		.title {
+			font-size: @moon-header-dense-title-font-size;
+			font-weight: @moon-header-dense-title-font-weight;
+			// line-height: @moon-header-dense-title-line-height;
```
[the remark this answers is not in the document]
Djspaceg: Oops nope! Little fella slipped through.

packages/moonstone/styles/mixins.less
Djspaceg: This would be wise to just unilaterally fix everywhere. It's labor-intensive but does raise the quality and consistency bar. I'll see what I can do in a reasonable amount of time.

teckliew: I could follow up with this work with the Dialog ticket too if you need it.

Djspaceg: This sort of cleanup would be great to get done now, before 3.0.0 finalizes. If you do remove any variables, or even rename any, be sure to include them in the 3.0 migration guide (referenced above).

I know you already know, but for historical record-keeping purposes, the current LESS variable naming convention is as follows:
```
@{theme}-{component}-[{state(s)}-]{cssProperty|abstractedName}
```
* The only allowed (known) safe abbreviation is for "background"->"bg". Otherwise, everything else should be spelled out.
* `color` rule can be conditionally referenced as `-text-color` if the commonly accepted content is indeed text. It may be shortened to simply its rule name, in cases like `Icon` where technically the thing you're adjusting is the text color, but you're actually modifying the color of a glyph and including "text" isn't particularly intuitive.
* States, if present fall between the component name and the CSS rule name.
   * If there are multiple, favoring design-time states first and runtime states second, with alphabetical being a tie-breaker. Ex:
```
@moon-button-bg-color;
@moon-button-focus-bg-color;
@moon-button-selected-bg-color;
@moon-button-selected-focus-bg-color;
@moon-button-disabled-bg-color;
@moon-button-disabled-focus-bg-color;
@moon-button-disabled-selected-bg-color;
@moon-button-disabled-selected-focus-bg-color;
```

Djspaceg:
I think I got all of the weird old inconsistent Header names.

Djspaceg:
Answers:
* [x] fullBleed text color is intended and didn't change.
* [x] Correct. Not a defined use-case. We could add a bunch of selectors to support this, but the value vs maintenance burden didn't seem worth it at this point.
* [x] Dense Header is indeed intentionally bold.
* [ ] Odd. It was focusable during development. I'll look into this.
* [x] This may indeed just be inevitable. With such low tolerances, we'll have to make some compromises until these issues can be discovered and raised back to GUI.
* [x] After looking over the locales in the dropdown, I believe all of them still relate to some aspect we would like to test or validate in the stories. The changes to the fonts regarding internationalization and non-English languages are primarily under-the-hood. `LG Display` which used to be a huge font file(s), incorporating almost all languages, was simply separated into multiple files, one for each supported language. With the plumbing underneath having been updated to support this change, there should be no significant difference between how things were vs how things are now.

Djspaceg:
What a convoluted bunch of nonsense this focusability issue was to track down.

TL;DR: info overlay lazily, sloppily injecting itself causes numerous problems. The close X is now visible while the info overlay is open/visible.

From the latest commit:
> This [custom CSS] was removed (but may be needed in the future) due to this causing layering ordering issues in the Panels.Header story. the info module injects code into the panel, which breaks the top-level expectations of Slottable for the Header to drop into the "header" slot (since the info overlay div is inserted at the top as a wrapper), which causes the Header render into the body (\<section\> tag), resulting in layering order problems.

The "[custom CSS]" referred to above targets the panel body to set it's `z-index` to a level that allows the info overlay to hide the close X button, by being on the same level. This also means that the body/content section, where the Header sample lives, because it can't follow the usual slottable rules (because of the info overlay injecting itself and levels of DOM) can't be assigned to in the correct node for it to behave as expected, like other the samples.

I've commented out the layering code with the above note which should resolve the issue in a compromise sort of way, due to the absence of the perfect solution.

ryanjduffy:
@Djspaceg What do you think of the spacing for non-latin locales? Seems off to me on my device (build 1777 atm) and in Chrome. Updating my device build now but I suspect the spacing will be the same.

ja-JP
image

th-TH
image

ryanjduffy:
The above is fixed by #2446

webOS101:
Regarding the font list, I wondered if we cover all the new possible fonts with the language choices.  If we now have more fonts, maybe we need to update the list to allow us to select at least one language from each.  Maybe we're already covered?

PR #2444, opened 2019-07-18
PLAT-83147: Fix Spotlight to restore focus through ancestor containers

ryanjduffy:
### Checklist

When the pointer times out on device, spotlight attempts to restore focus. If the last focused container and element cannot be focused (e.g. when you last focus an item in an ExpandableList that has closed), spotlight restores focus at the root which can end up in unexpected locations.

### Resolution
Change spotlight restoration logic to walk up the container stack from the last focused in order to find the "closest" target. In the case of Expandables, this will usually be the LabeledItem since the entire expandable is wrapped in a container.

### Additional Considerations
I also tried adding `setLastContainer` within `focusElement` so that when Spotlight is in pointer mode, the last container would also get updated. This resolved the issue as well and although it was a smaller diff it carries more risk from the broader impact.


packages/spotlight/src/spotlight.js
```diff
@@ -249,17 +250,20 @@ const Spotlight = (function () {
 
 	function restoreFocus () {
 		const lastContainerId = getLastContainer();
-		const next = [rootContainerId];
+		let next;
 
 		if (lastContainerId) {
-			next.unshift(lastContainerId);
+			// walk up the chain of containers from the last to attempt to find a target
+			next = getContainersForNode(getContainerNode(lastContainerId)).reverse();
```
webOS101: Should we test for an empty array here (or at the end of this) to set `next` to `[rootContainerId]`?

ryanjduffy: getContainersForNode will always contain the root container id.

webOS101: Argh, misread the source.  :nothingtodohere:

PR #2445, opened 2019-07-18
ENYO-6079: Fix LS2Request to return an error for null messages

ryanjduffy:
### Checklist

If an LS2 service returns `null` as the message, it will be treated as a success even though it isn't.

### Resolution
Update the failure logic to account for null and undefined messages.

### DIFF
--- a/packages/webos/LS2Request/LS2Request.js
+++ b/packages/webos/LS2Request/LS2Request.js
@@ -126,14 +126,23 @@ export default class LS2Request {
 		this.timeoutJob.stop();
 
 		let parsedMsg;
-		try {
-			parsedMsg = JSON.parse(msg);
-		} catch (e) {
+
+		if (msg == null) {
 			parsedMsg = {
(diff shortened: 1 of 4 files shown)


packages/webos/LS2Request/LS2Request.js
```diff
@@ -126,14 +126,23 @@ export default class LS2Request {
...
 		let parsedMsg;
-		try {
-			parsedMsg = JSON.parse(msg);
-		} catch (e) {
+
+		if (msg == null) {
 			parsedMsg = {
 				errorCode: -1,
-				errorText: msg,
+				errorText: 'No message returned',
```
webOS101: This error message seems a touch weak.  What about 'Invalid response' or something else that is more clear that an error occurred?  I worry that a developer might be confused about whether it was just an empty message vs. a response that's not expected.

ryanjduffy: works for me

PR #2446, opened 2019-07-19
PLAT-83171: Updated Header's design with better spaced titles-below

Djspaceg:
### Checklist

Out with the old.

### Resolution
In with the new (designs). Improved spacing between titles. Font size decreased slightly. Non-Latin changed as well, but due to an issue (described below) the old non-Latin sizes weren't applying.

### Additional Considerations
I discovered during this PR that it turns out our non-Latin rules have been terribly broken for more than a year. This was due to the mixin usage not being relocated into the header types classes, when they were added, so the rules didn't have the correct precedence. Furthermore, the line-height variables that needed to be applied simply didn't exist. Now they do, the mixins have been updated, the rules all apply in the correct order and places.

This is dependent on #2439 landing first.

### DIFF
--- a/docs/migration/enact/migrating-to-enact-3.md
+++ b/docs/migration/enact/migrating-to-enact-3.md
@@ -78,37 +78,64 @@ files, you may need to address these changes.
 * `@moon-radio-item-selected-spotlight-indicator-border-color` has been removed
 
 #### variables
-* `@moon-button-large-font-size` has been renamed to `@moon-button-font-size`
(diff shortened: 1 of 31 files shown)


webOS101:
Is there some way we could have caught the line heights not applying correctly before?

Djspaceg:
Screenshot testing or a watchful eye? I believe these line-heights actually broke a long time ago (prior to August 22, 2016), before our screenshot testing existed, and was masked by a hard-coded rule in the header file which explicitly set the line-height to 1.4em, which is odd for several reasons.

This is fine-tooth-comb level manual tracing of what and where each variable is used. I don't think it was catchable, practically.

PR #2447, opened 2019-07-19
PLAT-81336: Update Dialog for new designs

teckliew:
### Checklist

Update Dialog 

### Links
PLAT-81336


ryanjduffy:
The top/bottom padding around the root are reversed from the specs:

Specs:
image


PR:
image

Djspaceg:
That padding doesn't seem right. This is what I see.
DialogPadding
Was your screenshot taken from a freshly-loaded Storybook/App? HMR of Storybook doesn't load CSS rules properly sometimes.

ryanjduffy:
I deleted that comment. I had bad code running and it looks okay now.

ryanjduffy:
Looks like we need to update for large text mode.

LTR
image

RTL
image

packages/moonstone/Popup/Popup.js
```diff
@@ -171,7 +171,7 @@ const PopupBase = kind({
 	styles: {
 		css: componentCss,
 		className: 'popup',
-		publicClassNames: ['popup', 'reserveClose']
+		publicClassNames: ['closeButton', 'popup', 'reserveClose']
```
ryanjduffy: Let's document this either on `css` or inline with rationale for its exclusion.

packages/moonstone/styles/mixins.less
```diff
@@ -288,22 +288,8 @@
 	});
 }
 
-.moon-popup-header-text() {
```
ryanjduffy: We need to document that this was removed.

Djspaceg:
Aye. It looks like line 24 in Dialog needs an assist. The rule is applying correctly, but the value seems to be either old or just not based on the correct values.

packages/moonstone/styles/mixins.less
Djspaceg: This may be my fault. I had some difficulty resolving the merge conflict and I may have unintentionally dropped that line.

packages/moonstone/Popup/Popup.js
Djspaceg: It's possible this should be excluded from the publicClassNames, and the padding rules should just be applied to `Popup`, rather than explicitly on `Dialog`, so they match. We don't have specific specs for `Popup` in this regard, so this probably makes sense.

What do you two think?

teckliew: It would be nice if `Popup` shares the same padding rules, but It's probably safer to not assume that `Popup` will be the same as `Dialog`.

packages/moonstone/styles/variables.less
```diff
@@ -430,9 +433,14 @@
 
 // Dialog
 // ---------------------------------------
-@moon-dialog-top-padding: 24px;
+@moon-dialog-top-padding: 42px;
 @moon-dialog-h-padding: 42px;
-@moon-dialog-bottom-padding: 42px;
+@moon-dialog-bottom-padding: 24px;
+@moon-dialog-reserve-close-right-padding: (@moon-button-h-margin * 2) + @moon-icon-button-small-size + @moon-dialog-bottom-padding;
```
ryanjduffy: I think this sort of refactoring is immensely useful. Using other components variables as a base for calculations is great for visual consistency but often easy to overlook when changing them later. By consolidating those relationships into the variable files, I think those relationships are easier to identify. :+1: :+1:

ryanjduffy:
Notification is broken with the latest changes. I think the spirit of the proposed change makes sense but I'm concerned that it's growing more broad than we can safely take right now.

image

Djspaceg:
I will look into Notification real quick and see if it's going to cascade into a bigger deal than we want or if it's just a small oversight & fix.

packages/moonstone/Dialog/Dialog.module.less
```diff
@@ -18,21 +24,15 @@
...
-	.closeButton {
-		top: @moon-dialog-close-button-top;
-		right: @moon-dialog-close-button-right;
+	.closeContainer {
+		margin-top: @moon-popup-body-margin-top;
+		.margin-start-end(@moon-dialog-closecontainer-separation, auto);
 	}
 
 	.title {
-		.padding-start-end(0, @moon-dialog-title-right-padding);
```
teckliew: Should we add padding here for spacing between the title and the buttons?
spacing

Djspaceg: That's a good point. Think it should be the same as the Panels.Header title vs headerComponents spacing?

Wanna tackle that while I'm looking into and wrapping up the notification thing?

teckliew: Sure. I'll go with the `Header` specs

PR #2448, opened 2019-07-19
ENYO-6122: Add support for aria-owns in spotlight container nav

ryanjduffy:
### Checklist

* Unable to navigate to Panels controls (including close button) when the Header is in input mode.
* Focusing components within a Panel or the close button after focusing `controls` would trigger the panel title readout

### Resolution
* Add elements identified by "aria-owns" to a contains list of candidates
* Change the aria-owns of Panels to point to the new spotlight container around `controls` to fix the a11y readout

### DIFF
--- a/packages/moonstone/Panels/Controls.js
+++ b/packages/moonstone/Panels/Controls.js
@@ -8,10 +8,18 @@ import ApplicationCloseButton from './ApplicationCloseButton';
 
 import css from './Panels.module.less';
 
+/**
+ * Group of controls shared across Panel instances
+ *
+ * @class ControlsBase
+ * @memberof moonstone/Panels
+ * @ui
+ * @private
+ */
 const ControlsBase = kind({
 	name: 'Controls',
 
-	propTypes: {
+	propTypes: /** @lends moonstone/Panels.ControlsBase.prototype */ {
 		/**
 		 * Additional controls displayed before the close button.
 		 *
@@ -87,6 +95,22 @@ const ControlsBase = kind({
 	}
 });
 
+/**
+ * Group of controls shared across Panel instances.
+ *
+ * ```
+ * // remove the close button and use a star icon button
+ * <Controls noCloseButton>
+ *   <IconButton>star</IconButton>
+ * </Controls>
+ * ```
+ *
+ * @class Controls
+ * @mixes spotlight/SpotlightContainerDecorator
+ * @memberof moonstone/Panels
+ * @ui
+ * @private
+ */
 const Controls = ForwardRef(SpotlightContainerDecorator(ControlsBase));
 
 export default Controls;
--- a/packages/moonstone/Panels/Controls.js
+++ b/packages/moonstone/Panels/Controls.js
@@ -49,7 +49,7 @@ const ControlsBase = kind({
 		/**
 		 * Called with a reference to the root DOM node of this component.
 		 *
-		 * @type {Function}
+		 * @type {Function|Object}
 		 * @public
 		 */
 		forwardRef: PropTypes.oneOfType([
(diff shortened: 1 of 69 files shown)


ryanjduffy:
@germboy - I'd appreciate your review of the spotlight changes proposed here. In light of the spatial nav work aligning focus management with accessibility, I've extended spotlight containers to respect `aria-owns` in order to pull in related nodes that should be treated as part of the container even though they exist in another subtree.

There's some hackery going on in other parts of this PR to make this testable so don't worry too much about those bits. :)

packages/moonstone/Panels/Controls.js
```diff
@@ -0,0 +1,123 @@
...
+		 *
+		 * @type {String}
+		 * @public
+		 */
+		closeButtonBackgroundOpacity: PropTypes.oneOf(['translucent', 'lightTranslucent', 'transparent']),
+
+		/**
+		 * Called with a reference to the root DOM node of this component.
+		 *
+		 * @type {Function}
```
sugardave: Function | Object?

packages/moonstone/Panels/Panels.js
```diff
@@ -253,18 +231,23 @@ const PanelsBase = kind({
...
-		delete rest.onApplicationClose;
 		delete rest.onBack;
 
 		return (
 			<div {...rest} id={id}>
-				{controls}
+				<Controls
+					closeButtonAriaLabel={closeButtonAriaLabel}
+					closeButtonBackgroundOpacity={closeButtonBackgroundOpacity}
+					id={getControlsId(id)}
```
sugardave: If this value is used twice, can it be a computed prop or calculated at the top of the render function?  Would it save much performance-wise?

ryanjduffy: It's used both in a computed prop and the render so it itself can't be in computed since each of those is evaluated independent of each other by design. We could memoize it but I doubt we'd buy ourselves much in perf for string concatenation.

packages/moonstone/Panels/Controls.js
ryanjduffy: Yup!

ryanjduffy:
Fixed!

packages/moonstone/Panels/Panels.js
sugardave: I meant it's getting used twice in these props for `id` and `spotlightId`.  I have this thing about calling the function once :lol:

ryanjduffy: ah, right! I originally had some other code in here to change spotlight to use `id` as a fallback for `spotlightId` but decided to simplify this PR and remove that. I'll move this up to the top of the fn. Thanks for the catch and explanation.

packages/spotlight/src/container.js
```diff
@@ -272,6 +272,29 @@ const navigableFilter = (node, containerId) => {
...
+ * @returns {Node[]}               Array of owned nodes
+ * @memberof spotlight/container
+ * @private
+ */
+const getOwnedNodes = (node, selector) => {
+	// if node is document, it will not have getAttribute and therefore can't have aria-owns
+	const owns = node && node.getAttribute && node.getAttribute('aria-owns');
+	if (owns) {
+		const ids = owns.split(' ');
+		return ids
```
webOS101: I wonder if this could be simplified with `flatMap()`?  No need to actually do that, just wondering.

ryanjduffy: I don't _think_ so. `flatMap` handles a map that returns an array but we're only working with single elements here.

webOS101: yeah, it's a little messy because of the operations.  With `flatMap()` you could return empty array to drop the unmatched without the `filter(Boolean)` but it can't be easily done without a temp variable.

PR #2449, opened 2019-07-19
PLAT-83172: Space added for Heading with line, cleanup of divider cruft

Djspaceg:
### Checklist

Heading text is too close to the line. It gives people anxiety; like driving really close to the cement barrier on the highway.

### Resolution
Added some space (according to the spec) between the text and the line, when the line is present. The design doc indicates that the space should be `12px`, however, the 12px spacing is based on the bottom of Korean characters, which don't include descenders or a taller-than-what's-visible line-height. Therefore this PR sets a `9px` padding, which accounts for a roughly `3px` inherent gap added by the line-height rule respecting taller glyphs and descenders. The net visual effect should be 12px total.

### Additional Considerations
Finally doing house-keeping on the ancient divider mixin, which should no longer be necessary with the built-in customizability of the newer `Heading` component.

### DIFF
--- a/packages/moonstone/Heading/Heading.module.less
+++ b/packages/moonstone/Heading/Heading.module.less
@@ -7,12 +7,17 @@
 .heading {
 	border-bottom: @moon-heading-border-width solid transparent;
 	margin: 0 @moon-spotlight-outset (1.5 * @moon-spotlight-outset) @moon-spotlight-outset;
+	padding: @moon-heading-padding;
 	min-height: (1.5 * @moon-spotlight-outset);
 	letter-spacing: @moon-heading-letter-spacing;
 	line-height: @moon-heading-line-height;
 	.moon-font(@moon-heading-large-font-family, @moon-heading-non-latin-font-family, {
 		line-height: @moon-heading-non-latin-line-height;
 	});
+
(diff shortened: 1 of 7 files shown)


packages/moonstone/Heading/Heading.module.less
```diff
@@ -59,6 +59,7 @@
 
 		&.showLine {
 			border-bottom-color: @moon-heading-border-color;
+			padding-bottom: 9px;
```
teckliew: Should this be set outside of `applySkins` with its own variable?

Djspaceg: I suppose it could be, sure. I'll make that change real quick.

Djspaceg: Done.

ryanjduffy:
There are a few other `@moon-divider-` references that remain. I'm assuming they should be removed as well but perhaps they were left intetionally.

image

Djspaceg:
They were not left intentionally. I'll remove them and add them to the removed list after the merge conflict is resolved.

Djspaceg:
Branch updated. Merge conflict resolved, divider variables removed and documented.

PR #2450, opened 2019-07-19
PLAT-81342: Assigned proper colors to RadioItem dot

Djspaceg:
### Checklist

Who initially came up with these RadioItem dot colors?? J/k, it was likely me. New design colors have been delivered and need to be implemented.

### Resolution
Implemented new colors to match designs, which also bring dark and light skins inline with each other.


ryanjduffy:
@Djspaceg + @teckliew - Can we revise the high contrast bg color in the dark skin? It does not have enough contrast, imo.

image

Djspaceg:
Of course! I believe that light gray is actually supposed to be black in high-contrast. I'll make the change now.

Djspaceg:
My mistake. They are the way they're intended to be:
From the design document:
Screen Shot 2019-07-22 at 12 02 50 PM

ryanjduffy:
Okay!

PR #2451, opened 2019-07-22
PLAT-81468: Fix 5-way navigation for VirtualList with different height items

ybsung:
### Checklist

Needed to calculate the item position properly based on the different item size.

### Resolution

Calculate the item position properly with internal variables  having item position based on the different item size.

### Additional Considerations

Calculate the `firstVisibleIndex` and `lastVisibleIndex` to provide them when the list stops or starts scrolling.

### Links

PLAT-81468

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -175,6 +175,19 @@ const VirtualListBaseFactory = (type) => {
 			spotlightId: PropTypes.string,
 
 			/**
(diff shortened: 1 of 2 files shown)


0x64:
Just wondering whether it makes sense to make a newly added prop to a parameter of the factory function like native cases. I guess that no app will change the type on-demand in run time.

ybsung:
Let's think about it when cleaning up APIs. No opposite opinion to me.

0x64:
LGTM

PR #2452, opened 2019-07-22
PLAT-83146: Navigate from paging controls to items properly via 5way keys without focusableScrollbar

ybsung:
### Checklist

When pressing 5-way direction keys on scroll buttons without focus option (focusableScrollbar), Spotlight moves to incorrect item.

### Resolution

When pressing 5-way Up on next scroll button, 
1) Spotlight moves to out of VirtualList and 
2) Spotlight moves to the 5-way key direction.

For 1) case, it happened because scroll buttons could not get focus without focusableScrollbar prop. So I add the condition to skip manipulating spotlight by calling `moveFocusStraight`.

For 2) case, sometimes Spotlight should be just on the item next to the scroll button no matter which key is pressed. In this case, I called `preventDefault()` and `stopPropagation()`.

### Links

PLAT-83146

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -477,7 +477,7 @@ const VirtualListBaseFactory = (type) => {
 					ev.stopPropagation();
 				} else {
 					const {repeat} = ev;
-					const {isHorizontalScrollbarVisible, isVerticalScrollbarVisible, spotlightId} = this.props;
+					const {focusableScrollbar, isHorizontalScrollbarVisible, isVerticalScrollbarVisible, spotlightId} = this.props;
 					const {dimensionToExtent, isPrimaryDirectionVertical} = this.uiRefCurrent;
 					const targetIndex = target.dataset.index;
 					const isScrollButton = (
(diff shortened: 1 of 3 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -538,7 +538,7 @@ const VirtualListBaseFactory = (type) => {
 					} else if (
 						directions.right && repeat ||
 						directions.left && Spotlight.move(direction) ||
-						(directions.up || directions.down) && (repeat || moveFocusStraight({id: this.props.spotlightId, direction}) && currentTarget.contains(Spotlight.getCurrent()))
+						(directions.up || directions.down) && focusableScrollbar && (repeat || moveFocusStraight({id: this.props.spotlightId, direction}) && currentTarget.contains(Spotlight.getCurrent()))
```
germboy: I don't think that we can add this guard for `focusableScrollbar` quite like this, since it would mean we would never be able to prevent `onKeyDown` bubbling when the scroll button had focus via pointer and 5way up/down (on a vertical list) was pressed.
We want to consume `onKeyDown` events when focus remains within the list and bubble the event when focus would attempt to leave. This is the reason for the last conditional in this line: `currentTarget.contains(Spotlight.getCurrent())`, since focus remains in the container, we can prevent bubbling.

ybsung:
Closing in favor of https://github.com/enactjs/enact/pull/2484

PR #2453, opened 2019-07-22
ENYO-6131: Fix not to handle 5-way key twice on paging control in VirtualList

ybsung:
### Checklist

5-way key on paging control was handled twice in VirtualList and Spotlight. So Spotlight moved to incorrect item. 

### Resolution

Even though 5-way key was handled in VirtualList, `preventDefault()` and `stopPropagation() were not called when Spotlight moved out of the list. So those functions are forced to be called when the key is handled in VirtualList.

### Links

ENYO-6131

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -538,7 +538,7 @@ const VirtualListBaseFactory = (type) => {
 					} else if (
 						directions.right && repeat ||
 						directions.left && Spotlight.move(direction) ||
-						(directions.up || directions.down) && (repeat || moveFocusStraight({id: this.props.spotlightId, direction}) && currentTarget.contains(Spotlight.getCurrent()))
+						(directions.up || directions.down) && (repeat && currentTarget.contains(Spotlight.getCurrent()) || moveFocusStraight({id: this.props.spotlightId, direction}))
 					) {
 						ev.preventDefault();
 						ev.stopPropagation();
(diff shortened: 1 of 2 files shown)


0x64:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -538,7 +538,7 @@ const VirtualListBaseFactory = (type) => {
 					} else if (
 						directions.right && repeat ||
 						directions.left && Spotlight.move(direction) ||
-						(directions.up || directions.down) && (repeat || moveFocusStraight({id: this.props.spotlightId, direction}) && currentTarget.contains(Spotlight.getCurrent()))
+						(directions.up || directions.down) && (repeat && currentTarget.contains(Spotlight.getCurrent()) || moveFocusStraight({id: this.props.spotlightId, direction}))
```
germboy: I don't believe we can change the guard like this. At this point, we know that a scroll button has focus, so `currentTarget.contains(Spotlight.getCurrent())` will always be `true`. That specific guard was being used explicitly *after* the call to `moveFocusStraight()` to verify that after we attempt to move focus, that the container still had focus. If so, we would prevent event bubbling. This was done so that changing focus from one scroll button to the other would prevent `onKeyDown` event bubbling.
I see the problem now though. The call to `moveFocusStraight()` will always attempt to change focus, then if it moves focus outside the `VirtualList`, we don't prevent event bubbling so then spotlight moves focus a 2nd time. I believe we may have to verify *which* scroll button has focus and prevent bubbling when the 5way direction pressed would result in an attempt to leave the list - similarly to how we verify the value of `isLeaving` in order to perform specific actions when we know we'll attempt to leave.

On a side note, I can see a difference in event behavior between `VirtualList` and `VirtualListNative`. Event bubbling appears to be prevented more-or-less as intended (of course the issue in this use-case still exists) in `VirtualList`, but event bubbling does not appear to be prevented as intended in as many scenarios using `VirtualListNative`

germboy: I added an alternative solution. Could you give me your thoughts on #2457 ?

ybsung: @germboy 
Thanks for your comment. I was not familiar with the logic related with 5-way navigation in VirtualList. I'll review https://github.com/enactjs/enact/pull/2457/.

ybsung:
Closing in favor of https://github.com/enactjs/enact/pull/2457/.

PR #2455, opened 2019-07-22
ENYO-6130: Pass spotlightDisabled prop from list or scroller to scroll buttons

ybsung:
### Checklist

Even though a list or a scroller has a `spotlightDisabled` prop, the scroll buttons in it has spotlight when holding them.

### Resolution

Pass spotlightDisabled prop from a list or a scroller to scroll buttons

### Links

ENYO-6130

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -16,6 +16,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 - `moonstone/Spinner` to use the latest designs
 - `moonstone/Tooltip` layer order so it doesn't interfere with other positioned elements, like `ContextualPopup`
 - `moonstone/VirtualList.VirtualGridList` and `moonstone/VirtualList.VirtualList` to properly respond to 5way directional key presses
+- `moonstone/Scroller.Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList` scrollbar buttons not to get focus when spotlight deactivates
 
 ## [3.0.0-beta.1] - 2019-07-15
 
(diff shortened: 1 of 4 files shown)


0x64:
Maybe there would be a better name for the new prop, but I don't have any better idea. :)
So, looks good to me.

ryanjduffy:
@ybsung - Can you elaborate on the root cause and why this is the appropriate solution? Why are the scroll buttons focusable but items are not?

ybsung:
Hello @ryanjduffy 

> Can you elaborate on the root cause and why this is the appropriate solution? Why are the scroll buttons focusable but items are not?

I had to find the root cause. But it was very hard to me. According to my analysis, if `spotlightDisabled` prop was applied to Scroller, then Spottable.handleFocus() in the item was not called when clicking it. But Spottable.handleFocus() is the Scroll Button was called when clicking. So it would be the start point to debug more.

If I add Scrollbars as the content in Scroller, Spottable.handleFocus() in the Scroll Button of the scrollbars is not called. So there might no problem with item or scrollbars.

So I just passed `spotlightDisabled` prop to the Scroll button to blur in Spottable.handleFocus() to solve the problem simply.

ybsung:
Hello @germboy 
Do you have any hint for my comment above to know why Spottable.handleFocus() in the Scroll Button was called when clicking?

ryanjduffy:
I think the reason items are not focusable is due to the `onMouseDown` handler in `moonstone/Scrollable`:

https://github.com/enactjs/enact/blob/develop/packages/moonstone/Scrollable/Scrollable.js#L378-L382

For some reason, that handler is attached to the internal container and not the root node so it isn't able to suppress focus on the scroll buttons.

0x64:
Closing in favor of #2468

PR #2457, opened 2019-07-22
ENYO-6131: Fix VirtualList to properly prevent keydown events from sc…

germboy:
…roll buttons

### Checklist

When attempting to use 5way to leave a `VirtualList` while focus is on the scroll button, `onKeyDown` event bubbling is not prevented and/or allowed properly, causing a condition where focus can change twice for a single key press.

### Resolution
We can identify which scroll button has focus and depending on the 5way direction pressed (and whether the event is repeating), we can properly manage which events are bubbled - which will allow spotlight to handle the focus change rather than `VirtualList`.

We've been hesitant to import the `ScrollBar.module.less` into `VirtualListBase`, since we do not want to create more dependencies. So in order to identify which scroll button has focus, I opted to use `Spotlight.getSpottableDescendants(spotlightId);`, where the last spottable element returned is the last scroll button (bottom button in a vertical list, right button in a horizontal list, left button in a RTL horizontal list). The logic used applies to all these scenarios, so this change appears safe. I'l like to hear any thoughts or concerns.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -2,7 +2,7 @@ import {is} from '@enact/core/keymap';
 import Spotlight, {getDirection} from '@enact/spotlight';
 import Accelerator from '@enact/spotlight/Accelerator';
 import Pause from '@enact/spotlight/Pause';
-import Spottable from '@enact/spotlight/Spottable';
+import {Spottable, spottableClass} from '@enact/spotlight/Spottable';
 import {VirtualListBase as UiVirtualListBase, VirtualListBaseNative as UiVirtualListBaseNative} from '@enact/ui/VirtualList';
 import PropTypes from 'prop-types';
 import clamp from 'ramda/src/clamp';
@@ -27,7 +27,8 @@ const
 	Native = 'Native',
 	// using 'bitwise or' for string > number conversion based on performance: https://jsperf.com/convert-string-to-number-techniques/7
 	getNumberValue = (index) => index | 0,
-	nop = () => {};
+	nop = () => {},
+	spottableSelector = `.${spottableClass}`;
 
 /**
  * The base version of [VirtualListBase]{@link moonstone/VirtualList.VirtualListBase} and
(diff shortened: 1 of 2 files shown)


ybsung:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -535,15 +529,25 @@ const VirtualListBaseFactory = (type) => {
...
-					) {
-						ev.preventDefault();
-						ev.stopPropagation();
-					} else if (!repeat) {
-						isLeaving = true;
+					} else {
+						const spottables = Spotlight.getSpottableDescendants(spotlightId);
+						const spottablesTotal = spottables.length;
+						const firstScrollButton = spottables[spottablesTotal - 2];
+						const lastScrollButton = spottables[spottablesTotal - 1];
```
ybsung: if `focusableScrollbar` props is `false`, then the last list item will be assigned to `lastScrollButton`.

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -535,15 +529,25 @@ const VirtualListBaseFactory = (type) => {
...
+						const isLastScrollButton = target === lastScrollButton;
+
+						if (
+							directions.right && repeat ||
+							directions.left && Spotlight.move(direction) ||
+							(directions.up && !isLastScrollButton || directions.down && isLastScrollButton) && repeat ||
+							directions.down && !isLastScrollButton && Spotlight.focus(lastScrollButton) ||
+							directions.up && isLastScrollButton && Spotlight.focus(firstScrollButton)
```
ybsung: There is the same logic in `ScrollButtons` in https://github.com/enactjs/enact/blob/develop/packages/moonstone/Scrollable/ScrollButtons.js#L212. If 5-way navigation in scroll buttons is handled in VirtualList, we need to remove the logic in `ScrollButtons` and need to add the logic in Scroller, too.

germboy: I think the logic in `ScrollButtons` still applies. We only need to apply similar logic in `VirtualList`, since we are preventing the originating keydown events from reaching `Scroller`/`ScrollButtons`. It's perhaps less than ideal, but needed since we explicitly need to manage keydown event bubbling. I think keeping the original logic in `ScrollButtons` makes sense since we do not need to verify and prevent event bubbling there.

packages/moonstone/VirtualList/VirtualListBase.js
germboy: Good point. I've addressed this by:
- using a `querySelector` to verify scroll buttons, since they will be the only spottable components outside of the list wrapping container
- updating the focus logic to account for `focusableScrollbar`

ybsung:
LGTM

PR #2460, opened 2019-07-23
ENYO-6090: Change Scroller to support page up/down when scrollbars are hidden

germboy:
…r is hidden

### Checklist

A behavior change in needed for scrolling components (`Scroller`, `VirtualList`, etc) so that they can respond to and scroll when a remote paging key (channel up/down) is pressed when the scrollbar is hidden. Currently no action is taken.

### Resolution
We had specific guards against performing the requested scrolling action when scrollbars were hidden. We can remove this guard now. It should be noted that the existing behavior only affects vertical scrolling when a paging key is pressed.


ybsung:
LGTM

PR #2462, opened 2019-07-24
ENYO-6135: Not scroll horizontally via 5-way down in horizontal scroller

ybsung:
### Checklist

When 5-way Down on the first item in a horizontal scroller, the scroller scrolls horizontally.

### Resolution

If 5-way down is pressed and there is no more spottable items to the direction in a vertical scroller, the scroller scrolls vertically. But if it is a horizontal scroller, we don't have to scroll. So I guard this action with a condition not to scroll.

### Links

ENYO-6135

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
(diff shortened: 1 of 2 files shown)


0x64:
LGTM

ryanjduffy:
The reported bug was fixed incidentally by #2430 but this change prevents entering a function that shouldn't be called in this case so I'm merging as well!

PR #2463, opened 2019-07-24
ENYO-6115: Fix VideoPlayer more button tooltip rtl flip

teckliew:
### Checklist

Add `flip` prop to `Icon` and `IconButton`, for use with rtl and apply it as needed to `.moreButton`.

### Links
ENYO-6115


packages/moonstone/Icon/Icon.js
```diff
@@ -36,6 +37,14 @@ const IconBase = kind({
 	name: 'Icon',
 
 	propTypes: /** @lends moonstone/Icon.IconBase.prototype */ {
+		/**
+		 * Flip the icon horizontally or vertically.
+		 *
+		 * @type {('horizontal'|'vertical')}
+		 * @public
+		 */
+		flip: PropTypes.string,
```
ryanjduffy: I'd suggest we place this prop in `ui` as well for congruency with `ui/IconButton` adding it. I _think_ it makes sense to move the CSS there as well but I could be convinced that flipping should be theme-specific if you have an example in mind.

packages/moonstone/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
+### Added
+
+- `moonstone/Icon` prop `flip` to flip the icon horizontally or vertically
```
ryanjduffy: This would move to ui if the prop moves there.

packages/ui/Icon/Icon.js
```diff
@@ -141,11 +150,17 @@ const Icon = kind({
...
-			pressed
-		}, size),
+		className: ({children: icon, flip, iconList, pressed, size, styler}) => {
+			return styler.append(
+				{
+					// If the icon isn't in our known set, apply our custom font class
+					dingbat: !(icon in iconList),
+					pressed
+				},
+				flip ? `flip${cap(flip)}` : null,
```
webOS101: Curious why we don't just use `flip-${flip}` and update the class names.  I guess it's really not a big deal!

Djspaceg: Yeah, that one probably would have been good to catch. More in-line with our other class names.

Djspaceg: Nevermind! I got confused for a minute. We use camelcase CSS classes everywhere. This allows us to refer to them conveniently in JavaScript: e.g. `css.flipHorizontal` vs `css['flip-horizontal']` (since `css.flip-horizontal` isn't permissible).

LESS Mixins, on the other hand, were grandfathered into the hyphen naming convention, which works out because you _can't_ refer to those from within our JavaScript.

PR #2465, opened 2019-07-26
PLAT-83169: Major overhaul of localized fonts

Djspaceg:
### Checklist

Languages based on non-Latin glyphs can no longer be loaded using the existing font configuration of Moonstone and webOS. Languages were removed from the "global" font file and inserted into their own individual files. Moonstone defines fonts into one of two categories Latin and non-Latin, assigning those as-needed based on the current locale. The current system will only allow one language to be assigned to the desired font at a time, which doesn't work if you need 2 custom-font languages on the screen at the same time.

### Resolution
All languages' fonts have been consolidated into a single font family, easily referenceable via `font-family: "Moonstone";`. Each languages' scripts' glyphs can be referenced using the proposed method. This greatly simplifies the complexity of configuration, but also usage and predictability.

Custom fonts, like Japanese, can still override the primary LG Smart UI font when in the Japanese locale.

### Additional Considerations
This almost completely refactors the font generation rules into LESS from the fontGenerator javascript.

We should be sure to tell users about the replacement of "Moonstone Global" with "Moonstone", even though "Moonstone Global" is a mostly internal-only string (and always has been).


packages/ui/internal/localized-fonts/localized-fonts.js
```diff
@@ -2,19 +2,21 @@ const fontMap = {};
 let currentLocale = null;
 
 // Generate a single font-face rule
-const buildFont = function (inOptions) {
```
ryanjduffy: How will we know if this is input or output without the `in` prefix? ;)

webOS101: :lol:

Djspaceg: I originally wrote this back in Enyo and this was a naming convention that we used before Enyo 2.0 (more than 6 years ago), started by someone who had left before I started. It was a small callback to him/that, but it's really dated looking now and seems like a good time to change it.

_(This comment can act as that reference going forward.)_ ;)

ryanjduffy:
@Djspaceg Can you attach a couple screenshots to this to demonstrate the broken and fixed behaviors? That'll help communicating the impacts of the change.

Djspaceg:
Here's a repro case found in the QA sampler (with all of the system fonts installed).
Find the sample in `Text` -> `Languages`.
Fonts Language Mismatch-01

Notice how the top Oriya font doesn't change at all, even when switching to a different locale? That's the issue. Additionally, both strings should always be rendered in the same font, regardless of what locale we're in.

After the refactoring, all language glyphs render in their designated font regardless of the current locale (so the system is more multi-lingual robust). We still retain the ability to have a language font completely override everything based on the current locale. See Japanese locale. Japanese has its own custom font which actually includes Japanese, Latin, and Cyrillic glyphs, which replace other language-fonts when the current locale is set to the Japanese language or when you're using English in the region of Japan. This is intentional because the Japanese font is actually a different typeface, so having these Japanese glyphs next to other glyphs from other language fonts would look odd and out of place.

ryanjduffy:
Thanks, @Djspaceg! That's great documentation for the change.

packages/moonstone/MoonstoneDecorator/fontGenerator.js
```diff
@@ -4,78 +4,78 @@
...
-	'am': {
-		regular: 'LG Smart UI Amharic'
-	},
-	'bn': {
-		regular: 'LG Smart UI Bengali'
-	},
-	'gu': {
-		regular: 'LG Smart UI Gujarati'
-	},
+	// 'NonLatin': {
```
ryanjduffy: Can we remove these commented sections?

Djspaceg: Yes. I wasn't sure whether we wanted or needed them for later. I figured I'd save them just in case, but it's also completely fine to just remove them. I'll drop them now.

packages/moonstone/MoonstoneDecorator/fontGenerator.js
```diff
@@ -1,81 +1,42 @@
...
+ * This module loads Moonstone's locale specific fonts.
+ *
+ * _This is not intended to be directly included by external developers._ The purpose of this is to
+ * override the existing "Moonstone" font family with a new typeface, conditionally when the system
+ * locale matches the corrosponding locale for the font (defined in this component's code).
+ *
  */
 // eslint-disable-next-line no-var
-var {addLocalizedFont, generateFontRules, generateFontOverrideRules} = require('@enact/ui/internal/localized-fonts');
+let {addLocalizedFont, generateFontRules, generateFontOverrideRules} = require('@enact/ui/internal/localized-fonts');
```
Djspaceg: @JayCanuck I think my editor automatically converted this one. Do you think this will be ok? We have `const` mentioned and used below, so this seems like an ok alteration. Think it will impact precompiling/prerendering/etc?

JayCanuck: yea this seems fine :+1:

webOS101: I guess we should remove the eslint `no-var` line.

Djspaceg: @webOS101 You're right!

PR #2470, opened 2019-07-29
ENYO-6151: Fix focused, high contrast, and disabled appearance of expandables

teckliew:
### Checklist

Fix `moonstone/DatePicker`, `moonstone/DayPicker`, `moonstone/ExpandableList`, and `moonstone/TimePicker` disabled opacity to be consistent across high contrast skins.

Root cause:
When there are two layers in a component that receive `disabled` prop, both of these layers will also receive disabled rules from `.moonstone`/`.moonstone-light`; as a result it can be visibly seen that opacity disabled rule is applied twice.
To counter this, one layer has to explicitly unset the default disabled rules as seen on L49. This allows `ExpandableItem` to have full opacity while its child `LabeledItem` will be visibly disabled as normal.
The issue is reintroduced when `.highContrast` or when any skin is applied, and the default disabled opacity rule has a higher specificity as a result, overriding L49.
Wrapping `.disabled` rule with `Skinnable` will apply the opacity override across different skins.

### Links
ENYO-6151


ryanjduffy:
@teckliew Can you add some explanation to the description on the root cause of the problem and why the proposed fix resolves it? :awthanks:

teckliew:
:yougotitdude:

ryanjduffy:
@teckliew Thanks for the analysis on this. I think you're on the right track with a specificity fix. I searched through the code base for other instances of `.disabled()` with opacity and found that the other cases were wrapped with `.applySkins()` as well which fixes the specificity issue.

What do you think of changing ExpandableItem to wrap `.disabled()` with `.applySkins()` as well?

teckliew:
Yep that sounds good 👍

PR #2472, opened 2019-07-30
ENYO-6144: Prevent keydown event bubbling appropriately in scroll but…

germboy:
…tons

### Checklist

When using nested, vertically-oriented scrolling components (ex: `VirtualList` or `Scroller` within another `Scroller`), using the remote's paging keys (channel up/down) provides unexpected scroll/focus behavior.

### Resolution
We can prevent bubbling of the keydown event when paging controls have focus and the channel up/down buttons are pressed.


packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -265,13 +278,16 @@ class ScrollButtons extends Component {
 
 		if (isPageDown(keyCode)) {
 			if (focusableScrollButtons) {
+				preventDefault(ev);
 				Spotlight.setPointerMode(false);
 				Spotlight.focus(ReactDOM.findDOMNode(this.nextButtonRef.current)); // eslint-disable-line react/no-find-dom-node
 			} else if (!nextButtonDisabled) {
 				this.onClickNext(ev);
+				preventDefault(ev);
```
0x64: Is intended to call `preventDefault` _after_ calling back `onClickNext`?

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -144,6 +148,16 @@ class ScrollButtons extends Component {
 		this.prevButtonRef = React.createRef();
 	}
 
+	componentDidMount () {
+		this.nextButtonRef.current.addEventListener('keydown', this.onKeyDownNext);
```
0x64: It looks safe to add event listeners to scroll buttons on mounting only for our components, but I'm not sure it would be okay not to (re-)add listeners on updating in every single case. I expect there will be no DOM tree change on updating, but maybe we're able to add them again for sure since a redundant handler will be ignored. Do we need to do?

0x64:
It seems to be hard to keep synthetic events due to direct DOM event handlers in `Scrollable`. So, the approach looks good. And I don't expect any side effects.

packages/moonstone/Scrollable/ScrollButtons.js
germboy: No, no real reason. I will move this line up. 👍

packages/moonstone/Scrollable/ScrollButtons.js
germboy: Is there something that could potentially change in an update - like the ref/dom or something we would need to account for? I may just be unfamiliar with the life-cycle of this component, but I would expect if the dom isn't expected to change, there would be no need to register a dom event after every update.

0x64: > Is there something that could potentially change in an update - like the ref/dom or something we would need to account for?

No, I don't think either. I'm just worrying _what if later_. Let it go as is.

PR #2474, opened 2019-07-31
ENYO-6145: Fix Scrollable's focus target after using paging control

germboy:
### Checklist

`Scrollable` can set focus to an unexpected component when using the remote's paging buttons (channel up/down). After scrolling stops, it attempts to set focus to the control nearest to the previously focused control's position in the view, determined by reverse direction. However, this fails when spottable controls are spaced far apart (ex: 1 per scrolled view). This can result in the previously focused control (now off-screen) gaining focus unexpectedly, then the `Scroller` "bounces" back to the previously-scrolled position.

### Resolution
We need to determine if the "next" target is within (or intersects) the current viewport.

### Additional Considerations
I don't think we have rules around if *no* `Spottable` controls are within a newly scrolled view. At this point, we simply spot nothing. The `Scroller` container is still the active container, so pressing a 5way directional key will keep focus in the `Scroller`. However, since nothing could have focus in this scenario, the remote's paging buttons (channel up/down) will stop working.


0x64:
LGTM! 👍

PR #2477, opened 2019-08-05
PLAT-83423: Fix VirtualList not to scroll via channel up and down keys on horizontal paging controls

ybsung:
### Checklist

VirtualList should not scroll via channel up or down keys on horizontal paging controls.

### Resolution

If scrollbar is vertical, then skip to scroll via channel up and down keys.

### Additional Considerations

I added horizontal VirtualList qa-sampler to verify this issue.

### Links

PLAT-83423

### DIFF
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -277,6 +277,7 @@ class ScrollButtons extends Component {
 			{keyCode} = ev;
 
 		if (!vertical) {
+			ev.preventDefault();
 			return;
 		}
 
@@ -304,6 +305,7 @@ class ScrollButtons extends Component {
 			{keyCode} = ev;
 
 		if (!vertical) {
+			ev.preventDefault();
 			return;
 		}
 
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -276,24 +276,26 @@ class ScrollButtons extends Component {
 			{nextButtonDisabled, prevButtonDisabled} = this.state,
 			{keyCode} = ev;
 
-		if (!vertical) {
-			ev.preventDefault();
-			return;
-		}
-
-		if (isPageDown(keyCode)) {
-			if (focusableScrollButtons && !Spotlight.getPointerMode()) {
-				preventDefault(ev);
-				Spotlight.setPointerMode(false);
-				Spotlight.focus(ReactDOM.findDOMNode(this.nextButtonRef.current)); // eslint-disable-line react/no-find-dom-node
-			} else if (!nextButtonDisabled) {
-				preventDefault(ev);
-				this.onClickNext(ev);
+		if (isPageDown(keyCode) || isPageUp(keyCode)) {
+			if (!vertical) {
+				ev.preventDefault();
+				return;
 			}
-		} else if (isPageUp(keyCode)) {
-			if (!prevButtonDisabled) {
-				preventDefault(ev);
-				this.onClickPrev(ev);
+
+			if (isPageDown(keyCode)) {
+				if (focusableScrollButtons && !Spotlight.getPointerMode()) {
+					preventDefault(ev);
+					Spotlight.setPointerMode(false);
+					Spotlight.focus(ReactDOM.findDOMNode(this.nextButtonRef.current)); // eslint-disable-line react/no-find-dom-node
+				} else if (!nextButtonDisabled) {
+					preventDefault(ev);
+					this.onClickNext(ev);
+				}
+			} else {
+				if (!prevButtonDisabled) {
+					preventDefault(ev);
+					this.onClickPrev(ev);
+				}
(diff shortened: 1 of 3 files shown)


packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -272,10 +272,14 @@ class ScrollButtons extends Component {
 
 	onKeyDownPrev = (ev) => {
 		const
-			{focusableScrollButtons} = this.props,
+			{focusableScrollButtons, vertical} = this.props,
 			{nextButtonDisabled, prevButtonDisabled} = this.state,
 			{keyCode} = ev;
 
+		if (!vertical) {
+			return;
```
germboy: I don't think we should call `return` here. We will need to determine whether we should call `preventDefault(ev)` like below, which will prevent default browser scroll behavior and event bubbling. I think since we are not changing focus, we would want to prevent bubbling in these cases. 

Also, the browser can naturally scroll unexpectedly when attempting to use the paging controls. We will need to prevent that.

0x64: > Also, the browser can naturally scroll unexpectedly when attempting to use the paging controls. We will need to prevent that.

I also agree with @germboy. Just I want to leave two things for your information.

The first one is that, at least for Chrome, page up/down keys work only for a vertical scroller naturally. I think our components are relying on it. Maybe we need to check it up later if needed.

The second one is that `stopPropagation` and `preventDefault` will be called in `Scrollable` anyway. But that routine is not for scroll buttons but for a whole area. It looks better to prevent in buttons.

ybsung: Good Point! I missed considering event bubbling.

I called `ev.preventDefault()` instead of `preventDefault(ev)` because the Scroller having several horizontal VirtualLists should scroll when pressing a channel up or a down key when focus is on the scroll button of those VirtualLists. So `ev.stopPropagation()` should not be called to pass the event for the Scroller to scroll.

When pressing a channel up or a down key when focus is on the item of those VirtualLists, the Scroller does not scroll. We need to fix it after creating a new issue.

packages/sampler/stories/qa/VirtualList.js
```diff
@@ -123,14 +130,34 @@ const InPanels = ({className, title, ...rest}) => {
...
+					onKeyDown={action('onKeyDown')}
+					onScrollStart={action('onScrollStart')}
+					onScrollStop={action('onScrollStop')}
+					spacing={ri.scale(number('spacing', Config, 0))}
+					verticalScrollbar={select('verticalScrollbar', ['auto', 'hidden', 'visible'], Config, 'auto')}
+				/>
+			);
+		},
+		{propTables: [Config]}
+	)
```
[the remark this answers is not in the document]
ybsung: It is hard to consider whether we make a sampler covering all possible cases or just covering only practical cases. For my PR, I agree with your opinion for adding `horizontal scroll` sampler in VirtualList. So I removed the sampler.

But I'd like to add the 'horizontal scroll in Scroller' sampler which is more practical case for some apps. With the sampler, I could verify that horizontal VirtualList *DOES NO* scroll and the vertical Scroller including the VirtualList *DO* scroll when pressing a channel up or a down key when focus is on the scroll button of the VirtualLists. Please let me know your thought?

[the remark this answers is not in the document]
germboy: I like the new sample better, thanks!

packages/moonstone/Scrollable/ScrollButtons.js
webOS101: I felt like this might cause problems if we applied this for every key rather than just the Page-Up/Down, but it seems like it works correctly for all the scenarios I tried.

webOS101: Actually, I just tried some more things and it does cause problems (For example, try pressing the Print hotkey).  Can we adjust this to only preventDefault if it's a Page Up/Down?

packages/sampler/stories/qa/VirtualList.js
```diff
@@ -123,14 +139,44 @@ const InPanels = ({className, title, ...rest}) => {
 };
 
 storiesOf('VirtualList', module)
+	.add(
+		'horizontal scroll in Scroller',
```
webOS101: I noticed that this sample exposes a really bad behavior in horizontal VL.  Repeated pressing of the horizontal scroll buttons causes the list to scroll back before going forward.  I have a video showing this if needed.  Interestingly, that bounceback gets reset if you perform other actions before scrolling with the paging controls.

ybsung: As we discussed, this issue was related with ENYO-6110. So let's investigate it more in it.

packages/moonstone/Scrollable/ScrollButtons.js
ybsung: Yes. We should do. I wrapped the code with the condition to check if it's Page Up/Down.

PR #2479, opened 2019-08-05
ENYO-5944: Add additional icons to Moonstone

teckliew:
### Checklist

Add 37 new icons, and update 26 icons.
icons updated:

1. ellipsis
2. check
3. stop
4. play
5. forward
6. backward
7. jumpforward
8. jumpbackward
9. denselist
10. drawer
11. arrowsmallup
12. arrowsmalldown
13. search
14. exitfullscreen
15. fullscreen
16. trash
17. gear
18. playlist
19. resumeplay
20. audio
21. music
22. languages
23. cc
24. sub
25. repeat
26. info

### Comments
ENYO-5944


webOS101:
While you're in there, can you add a scroller to the icon sample so it's easy to scroll them on a small screen?

teckliew:
Updated!

ryanjduffy:
icons

ryanjduffy:
Since the icon names are public API, I'd like to get some consensus on names.

image

Currently `cycleone`, `cycleall`, and `cyclex`. Proposing `repeatone`, `repeatall`, and `repeatnone`. There's also a `cycle` which would change to `repeat`.

image
Currently `denselistdrawer`. We have `list`, `drawer`, and `denselist` so I propose we use `densedrawer` for this one.

image
Current `musicnoteplus` and `musicnoteplusminus`. These are accurate but literal names. I wonder if they are intended to be used for "add to playlist" and "edit playlist". If we knew, I would lean towards logic names but okay with what we have.

image
Currently `liverecord`, `liveplay`, `liveplayback`, `liveplaybackoff`. `liverecord` and `liveplayback` are very similar icons but the names convey different purposes. I'm not sure what these icons are trying to convey (which doesn't speak well of them imo) so I'm struggling to suggest new names. Maybe the first two are record and playback of a scheduled recording and the last two are on and off repeat scheduled recordings? So maybe `scheduledrecord`, `scheduledplay`, `scheduledrepeaton`, `scheduledrepeatoff`?

@teckliew , @webOS101 - thoughts? If we don't get them :100:, we can alias the better names later but we're stuck with these names until the next major release

ryanjduffy:
`fryingpan` is probably unlikely for a TV app ;) I'd suggest `searchfilled` or something along those lines but I kind of like `fryingpan` too. Maybe we leave it as a hidden alias :)

teckliew:
image
The original names `liveplayback` and `liveplaybackoff` were the names given with the icons, maybe we could go along with that naming convention and go with the `live*`? Open to other suggestions.

PR #2480, opened 2019-08-06
PLAT-84120: Add support for zh-Hans Font

ryanjduffy:
### Checklist

### Resolution
* Add support for zh-Hans font.
* Change the locale parser to handle the 4 character script code.

### Additional Considerations
Hasn't been tested on TV but verified to work in Chrome with the font installed locally.

### Links
https://stackoverflow.com/questions/18902072/what-standard-do-language-codes-of-the-form-zh-hans-belong-to

### DIFF
--- a/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
+++ b/packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
@@ -73,6 +73,7 @@ const locales = {
 	'th-TH - Thai, with tall characters':                   'th-TH',
 	'ar-SA - Arabic, RTL and standard font':                'ar-SA',
 	'ur-PK - Urdu, RTL and custom Urdu font':               'ur-PK',
+	'zh-Hans-HK - Simplified Chinese, custom Hans font':    'zh-Hans-HK',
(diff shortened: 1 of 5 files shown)


ryanjduffy:
zh-Hans-HK
image

zh-Hant-HK
image

packages/sampler/src/MoonstoneEnvironment/MoonstoneEnvironment.js
```diff
@@ -73,6 +73,7 @@ const locales = {
 	'th-TH - Thai, with tall characters':                   'th-TH',
 	'ar-SA - Arabic, RTL and standard font':                'ar-SA',
 	'ur-PK - Urdu, RTL and custom Urdu font':               'ur-PK',
+	'zh-Hans-HK - Simplified Chinese, custom Hans font':    'zh-Hans-HK',
```
Seonghyup: Hongkong(HK) and Taiwan(TW) use Traditional Chinese, China uses Simplified Chinese in webOS.
So, I think that the locale should be `zh-Hans-CN`.

ryanjduffy: Good to know, @seonghyup. Will fix this up.

PR #2481, opened 2019-08-06
ENYO-6088: Restore focus to elements at the pointer position for over…

germboy:
…flow containers when pointer hides

### Checklist

When the pointer hides while an overflow container is the active container (such as `VirtualList` components), spotlight will set focus to the last-focused element in the container. If the last-focused element has been scrolled out of the current container's view, then focus will be set unexpectedly - causing the list to scroll.

### Resolution
We can attempt to retrieve a navigable target from the pointer's last known position, when the active container is an overflow container.

### Additional Considerations
In spotlight, `restoreFocus()` is only called when an action takes place where we're in pointer-mode and the pointer will become hidden. I believe it's safe to perform the proposed change in logic there. Let me know if there are any concerns.


ybsung:
LGTM

packages/spotlight/src/spotlight.js
```diff
@@ -253,12 +254,16 @@ const Spotlight = (function () {
...
 		if (lastContainerId) {
+			const position = getLastPointerPosition();
+
 			// walk up the chain of containers from the last to attempt to find a target
 			next = getContainersForNode(getContainerNode(lastContainerId)).reverse();
 
 			// only prepend last focused if it exists so that Spotlight.focus() doesn't receive
 			// a falsy target
-			const lastFocused = getContainerLastFocusedElement(lastContainerId);
+			const lastFocused = getContainerConfig(lastContainerId).overflow && getNearestTargetFromPosition(position, lastContainerId) ||
```
ryanjduffy: This logic favors the "nearest" over the "last focused" independent of the visibility of the last focused. I suspect we'll get a report on this at some point but I think that is more edge case than the reported bug and fix so happy to move ahead with this.

PR #2483, opened 2019-08-07
PLAT-78618: Update disabled colors

teckliew:
### Checklist

Updated disabled focus and disabled opacity. 
Overall disabled opacity is changed from 30% to 40%.

Components that have some additional disabled focus changes:
- `Button` (disabled focus selected border color)
- `FormCheckboxItem` (disabled focus icon full opacity / content 40% opacity)
- `Input` (disabled focus ring full opacity)

### Additional Considerations
-  Added some `Input` disabled focus colors for the focus ring to be full opacity and also be consistent in light skin.
- `<input>` placeholder opacity cannot be separated from`<input>` text opacity afaik.
- Fixed `FormCheckboxItem` light skin normal icon bg color.

### Links
PLAT-78618

### DIFF
--- a/packages/moonstone/styles/colors-light.less
+++ b/packages/moonstone/styles/colors-light.less
@@ -101,7 +101,7 @@
 
 // FormCheckbox
 // ---------------------------------------
-@moon-formcheckbox-bg-color: #686868;
+@moon-formcheckbox-bg-color: @moon-white;
 @moon-formcheckbox-text-color: @moon-white;
 @moon-formcheckbox-border-color: @moon-dark-gray;
 @moon-formcheckbox-focus-bg-color: @moon-spotlight-color;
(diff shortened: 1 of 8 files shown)


packages/moonstone/styles/variables.less
```diff
@@ -256,8 +256,6 @@
 @moon-input-padding: 0 @moon-input-icon-spacing;
 @moon-input-font-weight: 600;
 @moon-input-text-indent: 0.2em;
-@moon-input-disabled-opacity: 0.3;
```
ryanjduffy: I'm hesitant to move these variables right now since it would be a potential breaking change which I'd like to avoid in an RC. I think we're safe to leave these here because the values in `colors(-light).less` will override them via `.applySkins()`.

packages/moonstone/styles/colors-light.less
```diff
@@ -101,7 +101,7 @@
 
 // FormCheckbox
 // ---------------------------------------
-@moon-formcheckbox-bg-color: #686868;
+@moon-formcheckbox-bg-color: @moon-white;
```
ryanjduffy: This change causes the checkbox to not be visible when unfocused in light skin because they are the same color. Not sure the background of this change but we either need to make the text color or the bg color different.

packages/moonstone/styles/colors-light.less
```diff
@@ -143,18 +143,19 @@
 
 // Input
 // ---------------------------------------
-@moon-input-text-color:               @moon-white;
```
ryanjduffy: Wanted to verify that the specs use black text for the input text color in light skin. I didn't see the specs in the ticket but assume that's what was showing in zeppelin.

PR #2484, opened 2019-08-07
PLAT-83146: Fix navigation from non-focusable paging controls

0x64:
### Checklist

If 5-way key is pressed while the pointer is hovering on non-focusable paging control buttons in pointer mode, spotlight moves to unexpected items in some cases. The current expected behavior is like below;
- If an outward key is pressed (e.g. press 'down' for horizontal lists),
  - spotlight moves to a spottable element if the spottable one exists on the direction
  - spotlight moves _into a nearest spottable item inside a list_ if no spottable one exists on the direction.
- Otherwise, spotlight moves on the key direction

### Resolution

To simplify roles and responsibilities between components, I moved logic from an internal handler of virtual lists to scroll buttons, then added some conditions for missing cases. Key events are consumed within a handler only if spotlight navigation is handled inside a handler.

### Additional Considerations

### Links
PLAT-83146


packages/moonstone/CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 - `moonstone/Picker` to avoid overlapping items on render
 - `moonstone/Scroller` and other scrolling components to properly scroll via remote page up/down buttons when nested within another scrolling component
 - `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList` to scroll via a page up or down key when focus is on any vertical paging control while in pointer mode
+- `moonstone/VirtualList.VirtualGridList` and `moonstone/VirtualList.VirtualList` to properly navigate from paging controls to items by 5-way key
```
ybsung: `focusableScrollbar` would be mentioned in the changelog.

0x64: Thanks, I updated it.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -236,36 +236,50 @@ class ScrollButtons extends Component {
...
 			{focusableScrollButtons, rtl, vertical} = this.props,
-			{target} = ev;
+			{target} = ev,
+			direction = getDirection(ev.keyCode),
+			isDown = direction === 'down',
+			isLeftMovement = direction === (rtl ? 'right' : 'left'),
+			isRightMovement = direction === (rtl ? 'left' : 'right'),
+			isUp = direction === 'up',
+			isNextButton = target === this.nextButtonRef.current,
+			isPrevButton = target === this.prevButtonRef.current;
```
ybsung: Before executing the following code, it seems to me that we need to guard the code with `direction` condition.
```
if (direction) {
...
}
```

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -236,36 +236,50 @@ class ScrollButtons extends Component {
...
+				// If is is horizontal `Scrollable`, move focus to the up
+				directionToContent = !vertical && 'up' || rtl && 'right' || 'left',
+				isDirectionToLeave =
+					vertical && (isRightMovement || isPrevButton && isUp || isNextButton && isDown) ||
+					!vertical && (isDown || isPrevButton && isLeftMovement || isNextButton && isRightMovement);
+
+			if (isDirectionToLeave) {
+				preventDefault(ev);
+				if (!Spotlight.move(direction)) {
+					if (Spotlight.getPointerMode()) {
```
ybsung: When I checked, `Spotlight.getPointerMode()` is always `false`. When could `Spotlight.getPointerMode()` value be `true`?

ybsung:
The most code that @0x64 tried to remove is related with ENYO-6131. But I verified ENYO-6131 with this PR and there was no issue. In addition, I checked the VirtualList Behavior with `spotlightDisabled` prop.

packages/moonstone/Scrollable/ScrollButtons.js
0x64: I thought it is safe for now since _onSpotlight[Down|Left|Right|Up]_ func props are called only for arrow keys. I will add guards here.

packages/moonstone/Scrollable/ScrollButtons.js
0x64: @ybsung Good catch! I should have checked the existing code first. In short, this routine will run always in 5-way mode. I will remove this routine.

Here is a short history. When this routine is born (#1776), we need to wait 5-way mode and put `setTimeout` here. Then, after another patch for navigation on disabled items (#1778), pointer mode is turned off within a keydown handler in VirtualListBase. Because of not only bubbling order but also the order between a DOM event handler (VirtualListBase) and React synthetic event handler (Spottable->ScrollButtons), ScrollButton's handler runs always on 5-way mode.

0x64:
@ybsung I've fixed up you pointed out.

packages/moonstone/Scrollable/ScrollButtons.js
ybsung: 👍

0x64:
@germboy I updated the PR.

> We want to prevent keydown events from bubbling in a VirtualList scenario unless the focus change could result in focus leaving the wrapped Scroller container.

I think I handled this requirement in `ScrollButtons'.

> Otherwise, we'll have inconsistent keydown events and this could affect apps that are expecting events when focus leaves the list.

When I test the code I updated, events are bubbled only when focus moves outside a list.

> This value should represent whether focus is leaving the scroller's container.

Updated for this also. But the approach is not so cool.

Could you please review again? Thank you!

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -286,46 +251,62 @@ class ScrollButtons extends Component {
 				if (focusableScrollButtons && !Spotlight.getPointerMode()) {
```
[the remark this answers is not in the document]
0x64: Oops, you're right. Thanks.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -286,46 +251,62 @@ class ScrollButtons extends Component {
...
+					// If is is horizontal `Scrollable`, move focus to the up
+					directionToContent = !vertical && 'up' || rtl && 'right' || 'left',
+					isDirectionToLeave =
+						vertical && (isRightMovement || isPrevButton && isUp || isNextButton && isDown) ||
+						!vertical && (isDown || isPrevButton && isLeftMovement || isNextButton && isRightMovement);
+
+				if (isDirectionToLeave) {
+					if (!getTargetByDirectionFromElement(direction, ev.target)) {
+						// move focus into contents and stop bubbling
+						preventDefault(ev);
```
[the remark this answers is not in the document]
0x64: > In this case, we wouldn't want to prevent event bubbling. Since we're moving in a "leaving" direction and there is no target in the direction pressed, that means focus is attempting to leave the scroller container area. In that case, an application should be able to receive onKeyDown event passed to VirtualList.

As far as I know, if there is no target to move when press 5-way key on non-focusable scrollbar buttons, focus should move into contents and it means the key input is consumed. (Well, I missed to check focusableScrollbar here. It is a bug. :() So, preventDefault() is intended here. Is this rule different between `Scroller` and `VirtualList`?

> One problem you will have with this implementation is that there should be a difference in whether we prevent onKeyDown events based on whether a Scroller is being used vs a VirtualList.

Thank you, you got the point about what is missing between us! I didn't know any requirements to bubble events to apps from `Scroller`, not like `VirtualList`.

0x64:
Removed a redundant call of `Spotlight.setPointerMode(false)` from `focusOnOppositeScrollButton`.

PR #2486, opened 2019-08-08
PLAT-84278: Add support to VirtualList for different item sizes

ybsung:
### Checklist

Needed to use different item size in VirtualList

### Resolution

By passing `itemSize` object with `minSize` and `size` properties into `VirtualList`, different size items could be used in the `VirtualList`.

- `size`: It is the property having the size information of the items as an array.
- `minSize`: It is the default item size. If `size` property with an index is not defined, the item with the index renders with the `minSize` width or height depending on the `direction` prop.

So if app developers don't know all item size when `VirtualList` renders at first, `size` property could be omitted. Then all items renders with the same size of the `minSize` property.

### Additional Considerations

For the APIs, we support different item size with different shape of `itemSize` prop in `VirtualList` instead of providing a new component because the APIs for the different item size is almost similar with the `VirtualList`.

### Links

PLAT-84278

### Comments

- `VirtualGridList` doesn't support different item size yet.


ybsung:

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -435,10 +451,40 @@ const VirtualListBaseFactory = (type) => {
...
+							) {
+								return true;
+							}
+						}
+
+						return false;
+					});
+					isCurrentItemInView = inView[0];
+					isNextItemInView = inView[1];
+					console.log(inView);
```
ryanjduffy: oops

packages/ui/Scrollable/Scrollable.js
```diff
@@ -420,7 +436,14 @@ class ScrollableBase extends Component {
 
 		this.resizeRegistry = Registry.create(this.handleResize.bind(this));
 
-		props.cbScrollTo(this.scrollTo);
+		if (props.type === 'VariableVirtualList') {
```
ryanjduffy: Can this condition be moved to VirtualList? I think it's out of place in Scrollable which should not be concerned with the constraints of VirtualList.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -673,6 +833,69 @@ const VirtualListBaseFactory = (type) => {
 			}
 		}
 
+		// For VariableVirtualList
+		adjustVariableGridPosition () {
```
ryanjduffy: I'm concerned this method is doing too much work. Appears that it is:

* calculating positions and caching
* applying positions to nodes
* updating the thresholds
* updating scroll bounds
* setting container size
* updating the moreInfo struct

We should at least decompose these a bit into separate functions for better readability and determine which should really be combined into a single method vs which should be called individually from other places.

packages/moonstone/VirtualList/VirtualListBase.js
ybsung: Oh~ I added while refreshing the branch. I fixed it.

packages/ui/Scrollable/Scrollable.js
ybsung: I moved it to `ScrollableVirtualList` in `VirtualListBase.js`. `ScrollableVirtualList` has a `Scrollable` and the `Scrollable` has `VirtualListBase`.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -21,15 +21,30 @@ const
...
 	minWidth: PropTypes.number.isRequired
 });
 
+/**
+ * The shape for the list different item size
+ * in a list for [VirtualList]{@link ui/VirtualList.VirtualList}.
+ *
+ * @typedef {Object} listDifferentItemSizeShape
+ * @memberof ui/VirtualList
+ * @property {Number} minWidth The minimum width of the grid list item.
```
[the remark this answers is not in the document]
ybsung: I forget to update it. I fixed it.

TOBE:
```
 * @property {Number} minSize The minimum size of the list item.
 * @property {Number[]} size An array of the list item size. If it is not defined, the list items will render with the `minSize` size.
```

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -21,15 +21,30 @@ const
...
+ * @property {Number[]} minHeight The minimum height of the grid list item.
+ * @public
+ */
+const listDifferentItemSizeShape = PropTypes.shape({
```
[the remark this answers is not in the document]
ybsung: I agree with your opinion.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -216,13 +231,37 @@ const VirtualListBaseFactory = (type) => {
...
+			 * Valid values are:
+			 * * `'VirtualList'`, and
+			 * * `'VirtualGridList'`, and
+			 * * `'VariableVirtualList'`.
+			 *
+			 * @type {String}
+			 * @default 'VirtualList'
+			 * @private
+			 */
+			type: PropTypes.oneOf(['VirtualList', 'VirtualGridList', 'VariableVirtualList']),
```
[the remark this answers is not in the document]
ybsung: Actually if an app passes `itemSize.minSize` to `VirtualList`, then it passes `VariableVirtualList` type  to `ScrollableVirtualList`. So the type is not defined app side and the type prop is only used in VirtualList inside.

So there is no problem to replace `this.props.type === 'VariableVirtualList'` with `this.props.itemSizes` after removing the type prop.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -216,13 +231,37 @@ const VirtualListBaseFactory = (type) => {
...
 			 * @private
 			 */
-			updateStatesAndBounds: PropTypes.func
+			updateStatesAndBounds: PropTypes.func,
+
+			/**
+			 * The array for variable item size.
+			 *
+			 * @type {String}
+			 * @default 'VirtualList'
```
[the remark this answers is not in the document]
ybsung: Absolutely I had to remove `@default` and update `@type`.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -216,13 +231,37 @@ const VirtualListBaseFactory = (type) => {
...
+			 * @private
+			 */
+			variableGridSizes: PropTypes.array
```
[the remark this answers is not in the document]
ybsung: I tried to remove `variable` prefix and `grid` in variables and props.
`variableGridPositions` -> `itemPositions`
`adjustVariableGridPosition` -> `adjustItemPositionWithItemSize`
`// For VariableVirtualList` -> `// For individually sized item`
`variableGridSize` -> `itemSize`

packages/ui/VirtualList/VirtualListBase.js
ybsung: First of all, I decomposed these into separate functions. But I hesitated to refactor these more because I didn't know how much I had to do more from now. Please let me know your thoughts.

ybsung:
> I'd like to see if there are ways to standardize around the idea of "item sizes" across fixed and variable lists so that fixed lists are treated as variable lists with the same values for each item.

The reason why I added `this.props.type === 'VariableVirtualList'` condition is..
I was not able to sure that we could guarantee the VirtualList performance and stability after supporting different item size feature. So I added the feature with the condition as the first step. Then I hope we could combine the code for the same item size and the different item size together for long term goal.

Please let me know your opinion.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -169,6 +184,14 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			getComponentProps: PropTypes.func,
 
+			/**
+			 * The array for individually sized items.
+			 *
+			 * @type {Number[]}
+			 * @private
+			 */
+			itemSizes: PropTypes.array,
```
ybsung: I forgot to push the commit replacing PropTypes.array with PropTypes.arrayOf(PropTypes.number).

packages/moonstone/VirtualList/VirtualList.js
```diff
@@ -40,16 +42,25 @@ const VirtualList = kind({
 		 * <VirtualList itemSize={ri.scale(72)} />
 		 * ```
 		 *
-		 * @type {Number}
+		 * @type {Number|ui/VirtualList.itemSizesShape}
 		 * @required
 		 * @public
 		 */
-		itemSize: PropTypes.number.isRequired
+		itemSize: PropTypes.oneOfType([PropTypes.number, itemSizesShape]).isRequired
```
ryanjduffy: This is an API change which makes this a difficult PR to accept right now. We may need to create a new private component export if we want to experiment with this now.

packages/moonstone/VirtualList/VirtualList.js
```diff
@@ -40,16 +42,25 @@ const VirtualList = kind({
 		 * <VirtualList itemSize={ri.scale(72)} />
 		 * ```
 		 *
-		 * @type {Number}
+		 * @type {Number|ui/VirtualList.itemSizesShape}
```
ryanjduffy: @webOS101 - do you know off hand if this will render correctly in our docs?

packages/ui/VirtualList/VirtualListBase.js
ryanjduffy: I updated it

ryanjduffy:
This PR is making public API changes with a new prop type export and the use of that prop type in moonstone. if we want to be strict, we can probably do some work to create new private VirtualList exports that use this new prop type and don’t change public API.

Practically, however, that doesn’t change anything and potentially makes things worse. apps (one app?) will start using the new API immediately so the API is effectively public. If we were to export new private APIs and then make the new APIs public in 3.1, we’d have to maintain the private-but-not-actually-private components until 4.0 thereby making more work for ourselves.

As much as I don’t like it, I’m inclined to merge the PR with the API changes since it makes the most sense even though it breaks the rules.

packages/ui/VirtualList/VirtualListBase.js
ryanjduffy: Looks much better! We can continue to refine as needed but you made some great progress over night. Thanks!!

Djspaceg:
I'm content _enough_ with this to let it pass. I don't like the idea of doing this often or in other components, but I'd rather have an easier to maintain codebase than one that is extremely strict but _technically_ correct.

packages/moonstone/VirtualList/VirtualList.js
webOS101: I'll test it.

webOS101: It works splendidly.

PR #2490, opened 2019-08-08
ENYO-6096: Fix VirtualList to retain proper scroll position after upd…

germboy:
…ating item size-spacing

### Checklist

After updating the item layout of a `VirtualList` (or `VirtualGridList`) via the `spacing` and/or `itemSize` props (where the amount of items are the same), the scroll position can appear to be incorrect, depending on where the scroller position was prior to the update. It will then "sync" back to its correct position once you begin to scroll the list contents. This can result in an instant and unexpected "jump" from item 0 to wherever the scroll position should be.

The scroll position is actually correct (as the scroll buttons and thumb reflect the correct position). It's the viewport that appears incorrect.

### Resolution
We can update the viewport to scroll to the scroll position after such an update, where the list content sizes have changed.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/VirtualList.VirtualGridList` and `ui/VirtualList.VirtualList` to retain the proper scroll position when updating the `itemSize` or `spacing` props
+
 ## [3.0.0-rc.1] - 2019-07-31
 
 ### Added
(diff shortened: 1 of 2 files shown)


ybsung:
I fixed the `CHANGELOG.md` conflict.

ybsung:
LGTM

But I'd like to note one thing.

When updating `spacing` or `itemSize`, we try to restore the previous scroll position in this PR. But when updating `dataSize`, `clientWidth` and `clientHeight`, we tried to reset the scroll position to 0. So there are two different policies to adjust the scroll position when changing the metrics of a list or items.

To reset the scroll position when changing a `spacing` or `itemSize` prop, VirtualList should request Scrollable to do it and it makes the code complex. Changing those props is not the common case in real apps. So I'm find with the current implementation.

PR #2491, opened 2019-08-08
ENYO-6169: Remove redundant disabled CSS rules

teckliew:
There were a lot of disabled overriding caused by applying `disabled` on the container div. Removed `disabled` on that root node and removed the redundant disabled rules in `ExpandableItem`.
We don't need disabled to be applied on the container because we want to keep its opacity when spotted and we can let the child components handle disabled styling.

Also cleaned up `LabeledItem` disabled rule because moonstone disabled rules will handle it.

### Additional Considerations
Not sure if `disabled` does anything else on the `ContainerDiv` of `ExpandableItem`. My guess is accessibility but that's already handled by `aria-disabled`.

This change affects `LabeledItem`, `ExpandableItem`, and all the components that uses `ExpandableItem` like `DatePicker`, `TimePicker`, `ExpandableInput`, etc.
Tested `LabeledItem`, `ExpandableItem` and also spot checked a few expandable components (the ones mentioned) disabled in dark/light skins and also high contrast modes for dark/light. 

### Links
ENYO-6169

### DIFF
--- a/packages/moonstone/ExpandableItem/ExpandableItem.js
+++ b/packages/moonstone/ExpandableItem/ExpandableItem.js
@@ -26,7 +26,6 @@ import last from 'ramda/src/last';
 import React from 'react';
 
 import LabeledItem from '../LabeledItem';
-import Skinnable from '../Skinnable';
 import {extractVoiceProps} from '../internal/util';
 
 import Expandable from './Expandable';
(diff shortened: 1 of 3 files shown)


webOS101:
We can check later what the other `[disabled] [disabled]` was for, maybe it was for Expandables explicitly and can now be removed.

PR #2494, opened 2019-08-12
ENYO-6123: Correctly scroll to spotted VirtualList control when focus…

germboy:
…ed via 5way

### Checklist

There are scenarios where you can use 5way to set focus to an "out-of-view" component within a `VirtualList` and the list will not scroll the focused control into view.

### Resolution
When quickly tapping 5way to change focus in a `VirtualList`, the successive taps can occur before the list is updated, thus you are not provided updated information about the list's potential scroll position (where the list would be when the momentum of the previously-initiated scroll - caused by a previous keydown - finishes).

When performing the steps outlined in the JIRA, the "next" target would be considered "in view" and the currently-focused item "not in view". We can update the guard used in deciding whether we simply set focus or scroll to the next item.

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -437,11 +437,12 @@ const VirtualListBaseFactory = (type) => {
 				const nextRow = (nextIndex - nextColumn) % dataSize / dimensionToExtent;
 				const numOfItemsInPage = Math.floor((clientSize + spacing) / gridSize) * dimensionToExtent;
 				const firstFullyVisibleIndex = Math.ceil(scrollPosition / gridSize) * dimensionToExtent;
+				const isCurrentItemInView = index >= firstFullyVisibleIndex && index < firstFullyVisibleIndex + numOfItemsInPage;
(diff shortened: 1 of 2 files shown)


ybsung:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -437,11 +437,12 @@ const VirtualListBaseFactory = (type) => {
 				const nextRow = (nextIndex - nextColumn) % dataSize / dimensionToExtent;
 				const numOfItemsInPage = Math.floor((clientSize + spacing) / gridSize) * dimensionToExtent;
 				const firstFullyVisibleIndex = Math.ceil(scrollPosition / gridSize) * dimensionToExtent;
+				const isCurrentItemInView = index >= firstFullyVisibleIndex && index < firstFullyVisibleIndex + numOfItemsInPage;
```
webOS101: It just finally clicked with me why I don't like these variable names...  They imply a question, so they make better _function_ names than _variable_ names.  `isCurrentItemInView` asks the question, `CurentItemIsInView` holds the answer.  I'm not suggesting a change, just making the observation.

PR #2499, opened 2019-08-14
ENYO-6143: Show overscroll effect properly via page up and down keys on paging controls

ybsung:
### Checklist

After passing `overscrollEffectOn` prop with {`pageKey`: true} in it to `VirtualList`, Overscroll Effect did not show properly via page up and down keys on paging controls.

### Resolution

`Scrollable` did not handle this case before. So I fixed this case via page up and down keys on paging controls

### Additional Considerations

- I added `with Overscroll Effect` sample in qa-sampler to adjuest `overscrollEffectOn` prop.
- `onKeyDown` was not called via page up and down on paging controls because the event handler was hooked in not ScrollableBase but VirtualListBase. So I fixed it.

### Links

ENYO-6143

### DIFF
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -192,6 +192,14 @@ class ScrollButtons extends Component {
 		return current === this.prevButtonRef.current || current === this.nextButtonRef.current;
 	}
 
+	checkAndApplyOverscrollEffect = (keyCode, checkAndApplyOverscrollEffect) => {
+		if (isPageUp(keyCode) && this.state.prevButtonDisabled) {
+			checkAndApplyOverscrollEffect('vertical', 'before');
+		} else if (isPageDown(keyCode) && this.state.nextButtonDisabled) {
+			checkAndApplyOverscrollEffect('vertical', 'after');
+		}
+	}
+
 	onDownPrev = () => {
 		if (this.announceRef.current.announce) {
 			const {rtl, vertical} = this.props;
--- a/packages/moonstone/Scrollable/ScrollButtons.js
+++ b/packages/moonstone/Scrollable/ScrollButtons.js
@@ -192,11 +192,11 @@ class ScrollButtons extends Component {
 		return current === this.prevButtonRef.current || current === this.nextButtonRef.current;
 	}
 
-	checkAndApplyOverscrollEffect = (keyCode, checkAndApplyOverscrollEffect) => {
+	checkAndApplyOverscrollEffect = (keyCode, checkAndApplyOverscrollEffect, type) => {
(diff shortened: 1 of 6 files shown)


ybsung:
I think we don't have to update CHANGELOG.md because `overscrollEffectOn` is a private prop.

packages/moonstone/Scrollable/ScrollButtons.js
```diff
@@ -192,6 +192,14 @@ class ScrollButtons extends Component {
 		return current === this.prevButtonRef.current || current === this.nextButtonRef.current;
 	}
 
+	checkAndApplyOverscrollEffect = (keyCode, checkAndApplyOverscrollEffect, type) => {
```
0x64: `checkAndApplyOverscrollEffectByDirection` method in `Scrollable` has a checking feature for scrollability and it looks better to use it rather than adding a new function into `ScrollButtons'. Is there any reason to add this?

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -595,11 +595,15 @@ class ScrollableBase extends Component {
...
-					if (overscrollEffectOn.pageKey) { /* if the spotlight focus will not move */
-						this.checkAndApplyOverscrollEffectByDirection(direction);
+				if (this.props.direction === 'vertical' || this.props.direction === 'both') {
+					if (this.isContent(target)) {
+						direction = isPageUp(keyCode) ? 'up' : 'down';
+						this.scrollByPage(direction);
+						if (overscrollEffectOn.pageKey) { /* if the spotlight focus will not move */
+							this.checkAndApplyOverscrollEffectByDirection(direction);
+						}
+					} else if (this.props.overscrollEffectOn.pageKey) {
```
0x64: If we will not check whether keypress happens on scrollbar or not, this clause works for all inputs when `this.isContent(target)` is false. If possible, let's change `isContent` into a proper condition to cover all area.

packages/moonstone/Scrollable/ScrollButtons.js
ybsung: To know the status of `this.state.prevButtonDisabled` and `this.state.nextButtonDisabled` in Scrollable,  we need to provide the `ScrollButtons` ref and pass it to `Scrollbar` ref so  that `Scrollable` could access is by `this.uiRef.current.verticalScrollbarRef.current.scrollButtonsRef.current.state.prevButtonDisabled`. Which one would be better?

0x64: I meant that we don't need to check if a button is disabled or not. We can simply check if _scrolling is possible to a given direction or not._

ybsung:
Closing in favor of https://github.com/enactjs/enact/pull/2499.

PR #2507, opened 2019-08-21
ENYO-6186: Fix EditableIntegerPicker to display value after editing

webOS101:

### Checklist

When editing `EditableIntegerPicker` an invalid value would cause the value in the picker to disappear

### Resolution
Regression from #2475 .  The input was reusing the `key` of the current value, so restoring the current value would not cause the old value to animate back in.  Solution: Use a different key for the edit mode.

### Additional Considerations

### Links
ENYO-6186


webOS101:
I think this is acceptable for this component.

PR #2508, opened 2019-08-21
ENYO-6176: Implement a more complete font definition system

Djspaceg:
### Checklist

Chinese fonts don't apply consistently.

### Resolution
The font generation mixins now, by default, assign the `font-weight: normal` rule, which motivates the browser to choose the correct font at the correct time. For Chinese specifically, each of the target weights had to be specified due to that font also including Latin characters, instead of just Chinese characters, like the other fonts. The primary mixins for generating rules are now significantly more robust, with advanced argument handling, in a hopefully easy to understand and read way. Font weight ranges and font stretching is now supported by the mixins which were added to support this change as well as LESS debugging tools which are intentionally included but comment out since they aren't necessary for code to function, but will be valuable next time we need to look into any font matter.

Chinese glyphs rendered in the "Moonstone" font will likely only apply the locale font in normal, 300, 600, and 700 font weights, with others relying on the system to supply the fallback. This is OK, however, due to the Moonstone design not leveraging those font weights.

### DIFF
--- a/packages/moonstone/styles/internal/fonts.less
+++ b/packages/moonstone/styles/internal/fonts.less
@@ -46,7 +46,7 @@
 	pa: local("LG Smart UI Gurmukhi");
 	ta: local("LG Smart UI Tamil");
 	te: local("LG Smart UI Telugu");
-	ur: local("LG Smart UI_Urdu") normal 300 600 700;
+	ur: local("LG Smart UI Urdu") normal 300 600 700;
 	zh-HK: local("LG Smart UI TC") normal 300 600 700;
 	zh-TW: local("LG Smart UI TC") normal 300 600 700; // same as `zh-HK`
 
--- a/packages/moonstone/styles/internal/fonts.less
+++ b/packages/moonstone/styles/internal/fonts.less
@@ -46,30 +46,30 @@
 	pa: local("LG Smart UI Gurmukhi");
 	ta: local("LG Smart UI Tamil");
 	te: local("LG Smart UI Telugu");
-	ur: local("LG Smart UI Urdu");
-	zh-HK: local("LG Smart UI TC"), local("LG Smart UI TC") 300, local("LG Smart UI TC") 600, local("LG Smart UI TC") 700;
-	zh-TW: local("LG Smart UI TC"), local("LG Smart UI TC") 300, local("LG Smart UI TC") 600, local("LG Smart UI TC") 700; // same as `zh-HK`
+	ur: local("LG Smart UI_Urdu") normal 300 600 700;
+	zh-HK: local("LG Smart UI TC") normal 300 600 700;
+	zh-TW: local("LG Smart UI TC") normal 300 600 700; // same as `zh-HK`
 
 	// DEV NOTE: This is the closest proximity to a collection of unit tests that we have to test this system.
-	// name:						local("name");
-	// nameWeight:					local("name") 100;
-	// nameWeightRange:			local("name") 100 900;
-	// nameWeightStyle:			local("name") 100 italic;
-	// nameWeightRangeStyle:		local("name") 100 900 italic;
-	// nameRepeat:					local("name"), local("name2");
-	// nameWeightRepeat:			local("name") 100, local("name2") 200;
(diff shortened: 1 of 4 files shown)


packages/moonstone/styles/internal/fonts.less
```diff
@@ -47,16 +47,29 @@
 	ta: local("LG Smart UI Tamil");
 	te: local("LG Smart UI Telugu");
 	ur: local("LG Smart UI Urdu");
-	zh-HK: local("LG Smart UI TC");
-	zh-TW: local("LG Smart UI TC"); // same as `zh-HK`
+	zh-HK: local("LG Smart UI TC"), local("LG Smart UI TC") 300, local("LG Smart UI TC") 600, local("LG Smart UI TC") 700;
```
ryanjduffy: is `zh` the only locale that should use the same fonts for all weights? I tested with Urdu text and locale and it exhibits the same behavior (rendering different fonts in different weights).

Djspaceg: `zh` is a little special because of the "LG Smart UI TC" font, which includes latin characters. Because of this they were clobbering the standard font when it was present. I did not exhaustively check every font for latin or other overlapping codepoints (other than Japanese). When testing Urdu, which font weights did you see multiple typefaces for and what text were you using? I'd like to repro locally and look into the numerous other questions I have.

Djspaceg:
I've (once again) updated the font generation mixin methods to hopefully be a bit easier to use, requiring less verbose input to get more useful output. It's now much simpler to generate font-face rules for multiple weights. While font-weight range support was removed (it was only added in the first commit of this PR) a more versatile system was implemented which accepts an arbitrary ruleset, where one could easily specify a font-weight range there if one were so inclined. 🙄 That being said, font-weight ranges don't behave in the browser(s) as predictably as they should, so their use is not recommended at this time anyway.

packages/moonstone/styles/internal/fonts.less
```diff
@@ -46,15 +46,28 @@
 	pa: local("LG Smart UI Gurmukhi");
 	ta: local("LG Smart UI Tamil");
 	te: local("LG Smart UI Telugu");
-	ur: local("LG Smart UI Urdu");
-	zh-HK: local("LG Smart UI TC");
-	zh-TW: local("LG Smart UI TC"); // same as `zh-HK`
+	ur: local("LG Smart UI Urdu") normal 300 600 700;
```
webOS101: Could you have done:
```
ur: local("LG Smart UI Urdu") normal 300 600 700; local("LG Smart UI_Urdu") normal 300 600 700;
```

Djspaceg: Yes, that would likely have worked. It would add 8 Mac-specific rules to our TV framework for a font that should only be installed on a TV or one of our framework development machines, in a language that none of us can read. If it's important to be able to notice the difference between standard Arabic LG Smart UI and Urdu LG Smart UI, then it may be valuable to add; however, it seemed like it would be fine to just have the TV version supported at this time. I'm open to other opinions though. I'm hoping that adding rules to fonts that don't exist doesn't mess up existing font-face definitions.

packages/moonstone/styles/internal/fonts.less
Seonghyup: Why don't we set weights in css for all single weight fonts like Urdu? We have still many single weight fonts
(https://wiki.lgsvl.com/pages/viewpage.action?title=Fonts+for+webOS+Jheri+Curl&spaceKey=webOSDocs).
I've tested this PR in rc4, but I can still see this bug in Tamil(`ta-IN`) too. We need to test all locale that use single weight font.

Djspaceg: We do set multiple font weights for Urdu. This is an "outdated" comment thread.

You probably shouldn't be linking to internal company web pages in a public place.

It's possible we need to add explicit font-weight assignments to other language fonts too, like Tamil, but I'd like to be deliberate about which ones we add it to, since the behavior doesn't seem to be consistent across all of the locale fonts. Comprehensive testing sounds like a good idea. We just need a time-efficient means of doing that. Open to suggestions.

PR #2509, opened 2019-08-21
ENYO-6188: Add overflow hidden property to FormCheckboxItem

xbinary:

### Checklist

Ellipses don't occur in FormCheckboxItem.
It is reproduced from rc.2.

### Resolution
Add `overflow: hidden` to `span` tag which contain `children`.

### Links
ENYO-6188

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 ### Fixed
 
 - `moonstone/ContextualPopupDecorator` arrow rendering issue in Chromium
+- `moonstone/FormCheckboxItem` to marquee its contents
 
 ## [3.0.0-rc.3] - 2019-08-15
 
--- a/packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
+++ b/packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
@@ -20,6 +20,10 @@
 
 // Interally theme our implementation using published classes from ToggleItem and ToggleIcon
 .toggleItem {
+	.content {
+		overflow: hidden;
+	}
+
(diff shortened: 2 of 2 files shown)


Djspaceg:
I'm concerned that this isn't the "right" approach. For example, if the problem is truly unique to FormCheckboxItem, then maybe this makes sense. However, that being said, FormCheckboxItem is based on the same thing that many other toggleable components are, and it's not unique in its visual representation of the toggle state (a checkmark), so it may be a bigger issue with Marquee or ToggleItem, and this would just be a superficial patch fix that we should address at a deeper level instead and fix all of the issues, rather than just the reported one.

Djspaceg:
To follow up with the comment above, an alternate approach that does not apply new rules would be to publish (add to publicClassNames) the `content` class name defined in ToggleItem, and remove the `span` tag from FormCheckboxItem entirely. This way, the content can still be styled for FormCheckboxItem (setting disabled opacity on content), we remove a DOM node, we don't duplicate CSS and intruduce potential bugs by creating unique scenarios for marquee, and leverage the existing class names and structure. Thoughts?

teckliew:
Opening up `content` class name would also simplify this as well:
https://github.com/enactjs/enact/pull/2483/files#diff-78698fc6d858b953a97382e93928f5ecR62

Djspaceg:
I think it's ok to accept this for this release and address this again, more holistically in the upcoming release.

What I was thinking was to refactor Marquee to expose its `.marquee` class through its hocs in our standard way (it's already public via the MarqueeBase), via the `css` prop, rather than exposing a design-time config key "className" which doesn't leverage our fancy CSS overrides mechanism. This will involve adding `css` prop support into the HOC layers of Marquee, but will be much more intuitive to use in the future. FormCheckboxItem and Button, as well as others, could immediately use the new approach and become simpler in the process.

PR #2516, opened 2019-08-27
PLAT-84674: Fix Dropdown performance with long lists

ryanjduffy:
### Checklist

Render performance of `Dropdown` is poor when using many options.

### Resolution
Adopt `VirtualList` within `Dropdown` instead of `Group` + `Scroller` to reduce the render load.

### Additional Considerations
* Because `VirtualList` requires `itemSize` to lay out the list, I had to hard code the height of `Item` and include `skinVariants` from `Skinnable` to scale for large text mode.
* Dropdown.js was getting long so I refactored the list component into a new file and converted the previous list component into a HOC so `Skinnable` could be moved to the decorator for consistency with other framework components.

### DIFF
--- a/packages/moonstone/Dropdown/DropdownList.js
+++ b/packages/moonstone/Dropdown/DropdownList.js
@@ -1,4 +1,5 @@
 import kind from '@enact/core/kind';
+import {forward} from '@enact/core/handle';
 import hoc from '@enact/core/hoc';
 import EnactPropTypes from '@enact/core/internal/prop-types';
 import Spotlight from '@enact/spotlight';
@@ -23,10 +24,10 @@ const DropdownListBase = kind({
 
 	propTypes: {
 		/*
-		* The selections for Dropdown
-		*
-		* @type {String[]|Array.<{key: (Number|String), children: (String|Component)}>}
-		*/
+		 * The selections for Dropdown
+		 *
+		 * @type {String[]|Array.<{key: (Number|String), children: (String|Component)}>}
+		 */
 		children: PropTypes.oneOfType([
 			PropTypes.arrayOf(PropTypes.string),
 			PropTypes.arrayOf(PropTypes.shape({
@@ -36,24 +37,24 @@ const DropdownListBase = kind({
 		]),
 
 		/*
-		* Called when an item is selected.
-		*
-		* @type {Function}
-		*/
+		 * Called when an item is selected.
+		 *
+		 * @type {Function}
+		 */
 		onSelect: PropTypes.func,
 
 		/*
-		* Callback function that will receive the scroller's scrollTo() method
-		*
-		* @type {Function}
-		*/
+		 * Callback function that will receive the scroller's scrollTo() method
+		 *
+		 * @type {Function}
+		 */
 		scrollTo: PropTypes.func,
 
 		/*
-		* Index of the selected item.
-		*
-		* @type {Number}
-		*/
+		 * Index of the selected item.
+		 *
+		 * @type {Number}
+		 */
 		selected: PropTypes.number,
 
 		/*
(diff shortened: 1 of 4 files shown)


packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -0,0 +1,193 @@
...
+
+		/*
+		* Callback function that will receive the scroller's scrollTo() method
+		*
+		* @type {Function}
+		*/
+		scrollTo: PropTypes.func,
+
+		/*
+		* Index of the selected item.
```
webOS101: weird indent!

packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -0,0 +1,193 @@
...
+				key: PropTypes.oneOfType([PropTypes.string, PropTypes.number]).isRequired
+			}))
+		]),
+
+		/*
+		* Called when an item is selected.
+		*
+		* @type {Function}
+		*/
+		onSelect: PropTypes.func,
```
webOS101: Required?  We don't validate presence below.

packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -0,0 +1,193 @@
...
+		}
+
+		getScrollTo = (scrollTo) => {
+			this.scrollTo = scrollTo;
+		}
+
+		scrollIntoView = () => {
+			if (!this.isScrolledIntoView) {
+				if (isSelectedValid(this.props)) {
+					const {nodeToFocus, nodeToScroll} = this.getNodeToFocus(this.props.selected);
```
webOS101: The selected is not scrolling into view when I test this.

packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -0,0 +1,193 @@
...
+	handlers: {
+		itemRenderer: ({index, ...rest}, {children, onSelect}) => {
+			let child = children[index];
+			if (typeof child === 'string') {
+				child = {children: child};
+			}
+
+			return (
+				<Item
+					{...rest}
```
webOS101: The selected class does not get applied to the selected item.

packages/moonstone/Dropdown/DropdownList.js
ryanjduffy: I'll add a guard.

ryanjduffy:
@webOS101 - I believe this is fixed up and ready for another review. Thanks for the feedback!

packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -145,26 +149,20 @@ const DropdownListSpotlightDecorator = hoc((config, Wrapped) => {
...
 		}
 
 		scrollIntoView = () => {
+			const {selected} = this.props;
+
 			if (!this.isScrolledIntoView) {
 				if (isSelectedValid(this.props)) {
-					const {nodeToFocus, nodeToScroll} = this.getNodeToFocus(this.props.selected);
+					const scrollIndex = selected > scrollOffset ? selected - scrollOffset : 0;
+					const selectedNode = this.node.querySelector(`[data-index='${selected}']`);
```
webOS101: Is this able to find one that's not currently in the DOM?  Should we make this a callback on the scroll, maybe?

ryanjduffy: good question. I'll give it a try with a bigger data set and see.

ryanjduffy: Fixed up (I hope!)

ryanjduffy: Fixed up (I hope!)

PR #2519, opened 2019-08-28
ENYO-6165: Fix horizontal scroll behavior in Scroller

germboy:
### Checklist

When using a `Scroller` that is expected to be scrolled vertically and horizontally - using `direction="both"`, the `Scroller` cannot scroll horizontally when navigating contents via 5way left/right.

The issue stems from how `Scroller` is checking its `direction` prop and calculating the next scroll position because of it. It's only accounting for the values of `'vertical'` and/or `'horizontal'`. If `'both'` is used instead, it defaults to the behavior of `'vertical'`.

### Resolution
Check for all valid values of `direction` and calculate the next scroll position accordingly.

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -439,8 +439,7 @@ class ScrollableBase extends Component {
 		const
 			spotItem = Spotlight.getCurrent(),
 			positionFn = this.childRef.current.calculatePositionOnFocus,
-			{containerRef} = this.uiRef.current.childRefCurrent,
-			containerNode = containerRef.current;
+			containerNode = this.uiRef.current.childRefCurrent.containerRef.current;
 
 		if (spotItem && positionFn && containerNode && containerNode.contains(spotItem)) {
 			const lastPos = this.lastScrollPositionOnFocus;
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -439,16 +439,27 @@ class ScrollableBase extends Component {
 		const
 			spotItem = Spotlight.getCurrent(),
 			positionFn = this.childRef.current.calculatePositionOnFocus,
-			{containerRef} = this.uiRef.current.childRefCurrent;
+			{containerRef} = this.uiRef.current.childRefCurrent,
+			containerNode = containerRef.current;
 
-		if (spotItem && positionFn && containerRef.current && containerRef.current.contains(spotItem)) {
+		if (spotItem && positionFn && containerNode && containerNode.contains(spotItem)) {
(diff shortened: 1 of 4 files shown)


ybsung:
LGTM

packages/moonstone/Scroller/Scroller.js
```diff
@@ -240,36 +280,24 @@ class ScrollerBase extends Component {
...
-
-			if (newItemLeft + itemWidth > (clientWidth + currentScrollLeft) && itemWidth < clientWidth) {
-				// If focus is moved to an element outside of view area (to the right), scroller will move
-				// to the right just enough to show the current `focusedItem`. This does not apply to
-				// `focusedItem` that has a width that is bigger than `this.scrollBounds.clientWidth`.
-				this.uiRefCurrent.scrollPos.left += rtlDirection * ((newItemLeft + itemWidth) - (clientWidth + currentScrollLeft));
-			} else if (newItemLeft < currentScrollLeft) {
-				// If focus is outside of the view area to the left, move scroller to the left accordingly.
-				this.uiRefCurrent.scrollPos.left += rtlDirection * (newItemLeft - currentScrollLeft);
+		if (!contains(containerRect, itemRect)) {
```
0x64: Is this condition is necessary? It looks performing some redundant comparison with following two `if` statements.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -439,16 +439,27 @@ class ScrollableBase extends Component {
 		const
 			spotItem = Spotlight.getCurrent(),
 			positionFn = this.childRef.current.calculatePositionOnFocus,
-			{containerRef} = this.uiRef.current.childRefCurrent;
+			{containerRef} = this.uiRef.current.childRefCurrent,
+			containerNode = containerRef.current;
```
0x64: This is the only one line in this function to use `containerRef`. :)

packages/moonstone/Scrollable/ScrollableNative.js
```diff
@@ -509,7 +509,8 @@ class ScrollableBaseNative extends Component {
 		const
 			spotItem = Spotlight.getCurrent(),
 			positionFn = this.childRef.current.calculatePositionOnFocus,
-			{containerRef} = this.uiRef.current.childRefCurrent;
+			{containerRef} = this.uiRef.current.childRefCurrent,
+			containerNode = containerRef.current;
 
 		if (spotItem && positionFn && containerRef.current && containerRef.current.contains(spotItem)) {
```
0x64: We can use `containerNode` here.

packages/moonstone/Scroller/Scroller.js
germboy: Good catch. That was a redundant guard based on my initial implementation.

PR #2520, opened 2019-08-29
PLAT-83223: An item scrolls into view when the item expands or shrinks dynamically

ybsung:
### Checklist

When an item in VirtualList with different item size expands or shrinks, the part of item or the item hides.

### Resolution

When changing the item size, the item sticks to the top when being shrunk and sticks to the bottom when being expanded.

### Additional Considerations

As mentioned in JIRA description, the requirement of the expanded or shrunk item behavior is not defined yet because there is no app using it. So I implemented the behavior reasonably.

### Links

PLAT-83223

### DIFF
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -327,6 +327,9 @@ const VirtualListBaseFactory = (type) => {
 				this.emitUpdateItems();
 			}
 
+			// When an item expands or shrinks,
+			// we need to calculate the item position again and
+			// the item needs to scroll into view if the item does not show fully.
 			if (this.props.itemSizes) {
 				if (this.itemPositions.length > this.props.itemSizes.length) {
 					// The item with `this.props.itemSizes.length` index is not rendered yet.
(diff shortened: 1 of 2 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -329,9 +329,40 @@ const VirtualListBaseFactory = (type) => {
 
 			if (this.props.itemSizes) {
 				if (this.itemPositions.length > this.props.itemSizes.length) {
+					// The item with `this.props.itemSizes.length` index is not rendered yet.
+					// So the item could scroll into view after it rendered.
+					// To do it, `this.props.itemSizes.length` value is cached in `this.indexToScrollIntoView`.
+					this.indexToScrollIntoView = this.props.itemSizes.length;
```
0x64: I need some explanation why the last index of `itemSizes` array should be the index to show. And, `indexToScrollIntoView` will be the last index + 1 with this code if I read understand correctly.

seunghoh:
Reviewed in pair offline. LGTM

PR #2522, opened 2019-08-29
ENYO-6152: Fix IncrementSlider qa sample

ryanjduffy:

### Checklist

Using a knob to set the value makes the `IncrementSlider` "controlled" and therefore unresponsive to internal state management.

### Resolution
Add some additional prop tracking to allow setting from both the knob and via interaction

### Additional Considerations
The knob doesn't refresh with the value change which is a bit weird but probably okay for the purposes of the QA sample. If someone disagrees, we can discuss more.


webOS101:
We shouldn't really have a sample that has both live updates and a knob.  That's silly!

packages/sampler/stories/qa/IncrementSlider.js
```diff
@@ -88,13 +89,27 @@ class IncrementSliderWithContextualPopup extends React.Component {
...
 	}
 
+	static getDerivedStateFromProps (props, state) {
+		if (props.value !== state.propValue) {
+			return {
+				propValue: props.value,
+				value: props.value
+			};
+		}
+	}
```
webOS101: Should return `null` here, I believe.  Will update.

PR #2523, opened 2019-08-30
PLAT-85170: Apply `will-change` CSS property to the proper node in lists

0x64:
### Checklist

`ui/VirtualList` and `ui/VirtualGridList` have `will-change: transform` CSS property which is added mainly for scroller in #1857, and the CSS property was applied to an unintended node in lists due to the difference of node structure between lists and scrollers.

### Resolution
This PR moves `will-change` CSS property to the proper node in `ui/VirtualList` and `ui/VirtualGridList`.

### Links
PLAT-85170

### DIFF
--- a/packages/ui/Scroller/Scroller.module.less
+++ b/packages/ui/Scroller/Scroller.module.less
@@ -1,3 +1,7 @@
+.willChange {
+	will-change: transform;
+}
+
 .hideNativeScrollbar {
 	&::-webkit-scrollbar {
 		display: none;
--- a/packages/ui/Scroller/Scroller.module.less
+++ b/packages/ui/Scroller/Scroller.module.less
@@ -1,8 +1,7 @@
-.willChange {
+.scroller {
 	will-change: transform;
-}
 
-.hideNativeScrollbar {
+	/* hide native scrollbar */
 	&::-webkit-scrollbar {
 		display: none;
 	}
--- a/packages/ui/VirtualList/VirtualList.module.less
+++ b/packages/ui/VirtualList/VirtualList.module.less
@@ -18,7 +18,8 @@
 
 		will-change: transform;
 	}
-	& > .willChange {
+	
+	.content {
 		will-change: transform;
 	}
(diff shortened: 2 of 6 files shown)


packages/ui/Scroller/Scroller.module.less
```diff
@@ -1,5 +1,9 @@
...
-}
+.willChange {
+	will-change: transform;
+}
+
+.hideNativeScrollbar {
+	&::-webkit-scrollbar {
+		display: none;
+	}
+}
```
ybsung: As we discussed , let's combine two CSS class selector above into one selector like `.scroller`.

ybsung:
LGTM

ybsung:
NOTE:
With this PR,
- Scroller and ScrollerNative still retain `will-change: transform` at the same DOM element because of https://github.com/enactjs/enact/pull/1857.
- `will-change: transform` is applied to the proper DOM element in VirtualList
- `will-change: transform` is removed in VirtualListNative.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -1058,21 +1058,26 @@ const VirtualListBaseFactory = (type) => {
 
 		// render
 
-		mergeClasses = (className) => {
+		makeContainerClasses (className) {
```
ryanjduffy: Changed this and `getContentClasses` to prototype functions instead of member arrow functions. I tend to prefer prototype functions unless we are passing them around since member functions are transpiled to be created at construction which incurs unnecessary overhead.

packages/ui/VirtualList/VirtualList.module.less
```diff
@@ -18,6 +18,10 @@
 
 		will-change: transform;
 	}
+	
+	.content {
```
ryanjduffy: * Renamed to `.content` to use semantic names rather than behavioral names
* removed the direct descendant selector (`>`) which wasn't necessary

PR #2528, opened 2019-09-02
ENYO-6227: FormCheckboxItem: there's no opacity when disabled and focused

Sangwook-Lee:
### Checklist

`Itemicon` is displayed with white color when it's disabled and focused.
Because there's no opacity when disabled and focused

### Resolution
fixed FormChckboxItem to apply opacity when disabled and focused

### Links
ENYO-6227

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,10 +2,11 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
-## unreleased
+## [unreleased]
 
 ### Fixed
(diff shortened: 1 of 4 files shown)


Sangwook-Lee:
@ryanjduffy ,
Thank you for the review.
Should we add style in `moonstone/SlotItem.module.less` to apply `opacity` ?

`ItemIcon` doesn't have a own class. So, I'm wondering how can it access to `Icon` in SlotItem like below.
`	.focus({
		.disabled({
			.slot {
				&.before,
				&.after {
					.icon {
						opacity: @moon-disabled-opacity;
					}
				}
			}
		})
	});
`

PR #2533, opened 2019-09-04
Adjust VirtualList keydown event consumption when leaving via ScrollB…

germboy:
…utton

### Checklist

Re-adding the `keydown` consumption - which was being done previously - only I'm adding a guard for when to consume it. Without it, the original issue in the ticket can be reproduced when using a `VirtualList`. Pressing 5way down on a vertical list's "down" scroll button could result in the item outside the viewport becoming focused.


0x64:
:lgtm:

PR #2537, opened 2019-09-06
PLAT-85351: Fix lists and scrollers spotlight behavior after page up/down

0x64:
### Checklist

We have got a new requirement to move spotlight to the last item (especially on the same column for grid lists) when a list or a scroller is expected that no more scroll by page up/down keys is possible on a given direction. This behavior will be helpful to end users to indicate whether further scrolling is possible or not.

### Resolution
Added logic to lists and scrollers to move spotlight to the last item on the same column when a list/scroller is about to touch the edge as a result of scroll by page toward a given direction.

### Additional Considerations
This behavior is applied only to 5-way mode. For pointer mode, there is no change.
Since scroll by page works only for vertical scrolling, this behavior is invalid for horizontal scrolling.

### Links
PLAT-85351

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -4,10 +4,6 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ## [unreleased]
 
-### Changed
-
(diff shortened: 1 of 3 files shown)


0x64:
Good point! No objection. Thank you @germboy!

PR #2539, opened 2019-09-06
ENYO-6227: Fix FormCheckboxItem styling of `itemIcon`

Djspaceg:
### Checklist

Items change size when an icon is added

### Resolution
Remove the margin of child icons which bump the item to be taller than intended.
Also sets the block sizing so the new span element has the correct dimensions.

### DIFF
--- a/packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
+++ b/packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
@@ -24,6 +24,15 @@
 		overflow: hidden;
 	}
 
+	.itemIcon {
+		display: inline-block;
+
+		* {
+			margin-top: 0;
+			margin-bottom: 0;
+		}
+	}
+
 	// Skin colors
 	.applySkins({
 		.focus({
--- a/packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
+++ b/packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
@@ -24,27 +24,36 @@
 		overflow: hidden;
 	}
 
-	.itemIcon {
-		display: inline-block;
-
-		* {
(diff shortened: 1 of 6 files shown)


packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
```diff
@@ -24,6 +24,15 @@
 		overflow: hidden;
 	}
 
+	.itemIcon {
+		display: inline-block;
+
+		* {
```
ryanjduffy: Good catch! Can we be more conservative and only affect children with `> *` as is done in SlotItem?

Djspaceg: I was concerned about nested elements, say you have a DIV or a SPAN with icons in it. Those icons wouldn't get trimmed and you wouldn't know why they don't fit. DIV and SPAN may still be commonly used if developers are unaware of `React.createFragment`.

packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
```diff
@@ -24,6 +24,15 @@
 		overflow: hidden;
 	}
 
+	.itemIcon {
+		display: inline-block;
```
ryanjduffy: Might want to use `display: contents` here. Noticing that the left margins collapse in 3.0.0 but not any longer with the current code and this fix.

nebula:

image

this:

image

packages/moonstone/FormCheckboxItem/FormCheckboxItem.module.less
ryanjduffy: Similarly, if devs put non-`moonstone/Icon`s there, every node would lose its margins which would be equally strange.

PR #2542, opened 2019-09-09
Add focus restoring for container in voice control

changgilee:
### Checklist

When scroll position is changed with voice control, focus has to restored.
But in case of newly target is container, focus is not restored.

### Resolution
Add focus restoring logic in case of container

### Additional Considerations
N/A

### Links
https://jira2.lgsvl.com/browse/ENYO-6252

### Comments

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Scroller` to restore focus properly when newly target is container in voice control
+- `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList` and `moonstone/VirtualList.VirtualList` to restore focus properly when newly target is container in voice control
 - `moonstone/Button` text alignment when `color` is set
 - `moonstone/FormCheckboxItem` opacity of `itemIcon` value when focused and disabled
 - `moonstone/Notification` to shrink to fit small content
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Scroller` to restore focus properly when newly target is container in voice control
 - `moonstone/Button` text alignment when `color` is set
 - `moonstone/FormCheckboxItem` opacity of `itemIcon` value when focused and disabled
 - `moonstone/Notification` to shrink to fit small content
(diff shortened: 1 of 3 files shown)


packages/moonstone/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Scroller` to restore focus properly when newly target is container in voice control
```
ybsung: This fix seems to affect lists as well. If yes, please replace `moonstone/Scroller` with `moonstone/Scroller`, `moonstone/VirtualList.VirtualGridList`, and `moonstone/VirtualList.VirtualList`.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -853,7 +853,12 @@ class ScrollableBase extends Component {
 				for (let i = 0; i < nodes.length; i++) {
 					const nodeBounds = nodes[i].getBoundingClientRect();
 					if (nodeBounds[first] > viewportBounds[first] && nodeBounds[last] < viewportBounds[last]) {
-						Spotlight.focus(nodes[i]);
+						const node = nodes[i];
+						if (node && node.dataset && 'spotlightContainer' in node.dataset) {
```
ybsung: Is there reason to use 'spotlightContainer' in node.dataset instead of node.dataset.spotlightContainer?

packages/moonstone/CHANGELOG.md
changgilee: I did fix the CHANGELOG.md. Thanks~

packages/moonstone/Scrollable/Scrollable.js
changgilee: I think we can use node.dataset.spotlightContainer, but just want to be consistence with isContainerNode() in container.js

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -853,7 +853,12 @@ class ScrollableBase extends Component {
...
+							Spotlight.focus(`@${node.dataset.spotlightId}`);
```
ybsung: I'm not sure that an app or a component should check if a node is a DOM element or a Spotlight container. If can, it would be better to fix Spotlight side. But I don't have much information about Spotlight. So I hope that someone else could check it.

If a node is a Spotlight container and Spotlight.focus(node) is called, then node.focus() in https://github.com/enactjs/enact/blob/develop/packages/spotlight/src/spotlight.js#L231 is called instead of the child node of the node.

ybsung:
LGTM

packages/moonstone/Scrollable/Scrollable.js
ryanjduffy: I've had a similar thought. Officially, `Spotlight.focus()` doesn't support passing a node but it's an oft used method within the framework.

I think I'd be in favor of adding a branch within `Spotlight.focus()` to resolve the container ID when the node is passed.

Should be something like the following added to `Spotlight.focus()`:
 ```
			} else if (isContainer(elem)) {
				target = getTargetByContainer(getContainerId(elem));
			}
```

ryanjduffy:
Closing in favor of #2554

PR #2543, opened 2019-09-09
ENYO-6253: font-size on dense type of Header

Sangwook-Lee:
### Checklist

The font-sizes of titleBelow, subTitleBelow on dense type of Header are different from GUI spec.

### Resolution
fixed font-size on dense type of Header

### Additional Considerations

### Links

### DIFF
--- a/packages/moonstone/Panels/Header.module.less
+++ b/packages/moonstone/Panels/Header.module.less
@@ -84,7 +84,7 @@
 		}
 
 		.titlesCell {
-			margin-bottom: 9px;
+			margin-bottom: 15px;
 		}
 	}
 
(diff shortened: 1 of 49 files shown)


Sangwook-Lee:
1) There're duplicated code lines from 101 line to 109 line in `Header.module.less`.
2) The font-sizes of titleBelow, subTitleBelow on `dense` type of Header are different from GUI spec. That should be `24px`.

packages/moonstone/Panels/Header.module.less
```diff
@@ -84,7 +84,7 @@
 		}
 
 		.titlesCell {
-			margin-bottom: 9px;
+			margin-bottom: 15px;
```
webOS101: Should we be changing this for a standard header, too?  Is this something we can calculate from some other value?

Djspaceg: It could technically be calculated from other values but the equation is unintuitive and harder to maintain than just this value.

Oops! Nope, that shouldn't be affecting standard, just dense. I'll update this to only affect dense.

PR #2544, opened 2019-09-09
PLAT-85480: Move Routable to ui

ryanjduffy:
### Checklist

Routable is useful beyond moonstone

### Resolution
* Move `Routable` to `ui`
* Move `IdProvider` to `ui/internal` mostly because I'm not convinced it's the "best practice" as a public API but serves our particular needs.


webOS101:
Why don't we re-export `Routable` from `moonstone` to prevent a breaking API change?

ryanjduffy:
They are still exported by adding this import from ui rather than the local import.

https://github.com/enactjs/enact/pull/2544/files?file-filters%5B%5D=.js#diff-8d9952dc5ce298ffb29c8c1cb93a9b28R16

I do still need to add docs for them as we have for `Marquee` so they appear in our online docs.

webOS101:
:nothingtodohere:

packages/ui/Routable/Link.js
```diff
@@ -0,0 +1,69 @@
...
+const Linkable = hoc({navigate: 'onClick'}, (config, Wrapped) => {
+	const {navigate} = config;
+
+	return kind({
+		name: 'Linkable',
+
+		contextType: RouteContext,
+
+		propTypes: {
+			path: PropTypes.string.isRequired
```
Djspaceg: For some reason, this reqire specifically, and no other, is firing. Removing `isRequired` here resolves the issue with no additional warnings or errors.

ryanjduffy: Did you see those prop-type warnings in settings? If so, it's because `MenuItem` is wrapped by `Routable` by not all of them have a path (because we haven't build those views out) so it warns.

packages/moonstone/Panels/index.js
```diff
@@ -32,6 +32,30 @@ export {
...
+	 * @hoc
+	 * @name Routable
+	 * @extends ui/Routable.Routable
+	 * @memberof moonstone/Panels
+	 * @public
+	 */
 	Routable,
+
+	/**
+	 * Used with {@link ui/Routable.Routable} to define the `path` segment and the
```
webOS101: Update this to `moonstone/Routable.Routable`?

packages/ui/Routable/Routable.js
```diff
@@ -1,39 +1,43 @@
 import hoc from '@enact/core/hoc';
```
webOS101: This needs a module header to generate docs properly.

PR #2548, opened 2019-09-10
PLAT-85321: Add EventCacheable to cache event

ybsung:
### Checklist

Needed to cache the event when `onScrollStop` event is bubbled.

### Resolution

 I added `EventCacheable` component and the sample using it in VirtualList for `onScrollStop`.

### Links

PLAT-85321

### Comments

~I'll update CHANGELOG.md and JSDocs after 1st reivew.~

### DIFF
--- a/packages/ui/EventCacheable/EventCacheable.js
+++ b/packages/ui/EventCacheable/EventCacheable.js
@@ -2,7 +2,7 @@ import React from 'react';
 import hoc from '@enact/core/hoc';
 
 const EventCacheable = hoc((config = {defaultEvent: null, type: null}, Wrapped) => {
-    const {defaultEvent, type} = config;
+	const {defaultEvent, type} = config;
 
 	return class EventParamCacheableBase extends React.Component {
 		cachedEvent = defaultEvent || null
--- a/packages/ui/EventCacheable/EventCacheable.js
+++ b/packages/ui/EventCacheable/EventCacheable.js
@@ -1,13 +1,71 @@
+/**
+ * A higher-order component to cache an event.
+ *
+ * @module ui/EventCacheable
+ * @exports EventCacheable
+ */
+
 import {forward, handle, call} from '@enact/core/handle';
 import hoc from '@enact/core/hoc';
+import PropTypes from 'prop-types';
 import React from 'react';
 
-const defaultConfig = {defaultEvent: null, type: null};
+/**
+ * Default config for {@link ui/EventCacheable.EventCacheable}.
+ *
+ * @memberof ui/EventCacheable.EventCacheable
+ * @hocconfig
+ */
+const defaultConfig = {
+	/**
+	 * The default event before caching an event.
+	 *
+	 * @type {*}
+	 * @default null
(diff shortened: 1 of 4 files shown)


packages/ui/EventCacheable/EventCacheable.js
```diff
@@ -0,0 +1,33 @@
...
+		setEvent = (event) => {
+			this.cachedEvent = event;
+		}
+
+		eventHandler = (ev) => {
+			this.cachedEvent = ev;
+		}
+
+		render () {
+			const eventHandler = type ? {[type]: this.eventHandler} : null;
```
ryanjduffy: Minor optimization but this code creates 4 new objects:
* This object
* clone of props: `{...this.props}`
* clone of this object: `{...eventHandler}`
* Merged props of `this.props` clone and `eventHandler` clone

An equally readable (imo) but more efficient alternative is:

```
const props = {...this.props};
if (type) props[type] = this.eventHandler;
```

packages/ui/EventCacheable/EventCacheable.js
```diff
@@ -0,0 +1,33 @@
+import React from 'react';
+import hoc from '@enact/core/hoc';
+
+const EventCacheable = hoc((config = {defaultEvent: null, type: null}, Wrapped) => {
```
ryanjduffy: This use of default parameters doesn't take advantage of the config merging the `hoc` provides. It doesn't make much difference in this case since there are no default values but, for consistency, I'd suggest using the following form which allows the author to pass partial configs which `hoc` will populate with defaults.

```
const defaultConfig = {...};
hoc(defaultConfig, (config, Wrapped) => { ... });
```

packages/ui/EventCacheable/EventCacheable.js
```diff
@@ -0,0 +1,33 @@
...
+	return class EventParamCacheableBase extends React.Component {
+		cachedEvent = defaultEvent || null
+
+		getEvent = () => (this.cachedEvent)
+
+		setEvent = (event) => {
+			this.cachedEvent = event;
+		}
+
+		eventHandler = (ev) => {
```
ryanjduffy: You'd probably want to forward this on rather than capture the event.

```js
import {forward, handle, call} from '@enact/core/handle'
eventHandler = handle(
  forward(type), // forward to consumers
  call('setEvent'), // invoke this.setEvent(ev)
).bindAs(this, 'eventHandler')
```

ybsung:
I reopened this PR because of following up the @ryanjduffy's comments. I think that this PR was closed by mistake. If not, please correct me.

ybsung:
Hello @ryanjduffy 

Thank you so much for your review. I had to concrete my PR more before requesting review. I updated the PR as you mentioned and added CHANGELOG.md & JSDocs. Could you review them?

ryanjduffy:
Thanks, @ybsung, I should have been more clear in my original review. I closed it because this feature isn't a pattern I think we should support within the framework. The code seems sound and likely suitable to be used within applications that would want access to an imperative API.

Glad you created the PR to spur this discussion and because it can be pointed to as an example for interested apps.

PR #2549, opened 2019-09-10
ENYO-6260: Fix Dropdown to include `data` prop in `onSelect` handler

Sangwook-Lee:
### Checklist

There's no `data` prop in `onSelect` handler of `Dropdown` component.

### Resolution
Added `data` into the `onSelect` handler.

### Additional Considerations

### Links

### DIFF
--- a/packages/moonstone/Dropdown/DropdownList.js
+++ b/packages/moonstone/Dropdown/DropdownList.js
@@ -84,10 +84,10 @@ const DropdownListBase = kind({
 			const {children, selected} = props;
 
 			let child = children[index];
-			const data = child;
 			if (typeof child === 'string') {
 				child = {children: child};
 			}
+			const data = child.children;
 
 			return (
 				<Item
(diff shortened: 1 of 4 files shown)


packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -87,14 +87,15 @@ const DropdownListBase = kind({
 			if (typeof child === 'string') {
 				child = {children: child};
 			}
+			const data = child[children];
```
webOS101: This should be `child.children`, I will update it.

ryanjduffy:
Thanks for catching this, @Sangwook-Lee! I missed that `Group` included `data` in its payload. To avoid future regressions, I pushed a couple commits to:

* Document the payload of `onSelect`
* Add unit tests for `onSelect`

I had also pushed a commit to change `data` to include the entire object when objects are used as children but discovered that `ui/Group` doesn't (like I thought it did). So, while I think it's probably the right approach, I'm going to leave it as is to favor consistency.

PR #2550, opened 2019-09-10
PLAT-85398: Fix accessibility read out of the Dropdown components.

Baekwoo-s:
### Checklist

### Issue Resolved
The TV should read "X of Y" when focusing on items in Dropdown component.
Expected > "Option 1 + Checked + X of Y"
Actual > "Option 1 + Checked"

### Resolution
- To propagate `aria-posinset` and `aria-setsize` to the Item component to include the X of Y information of the item.
- For the expected behavior above, the `role` of VirtualList must be `group`, not `list`. So I modified the `role` prop given by the Dropdown component to apply to the VirtualList.

### Links
PLAT-85398

### DIFF
--- a/packages/moonstone/Dropdown/DropdownList.js
+++ b/packages/moonstone/Dropdown/DropdownList.js
@@ -81,9 +81,7 @@ const DropdownListBase = kind({
 
 	handlers: {
 		itemRenderer: ({index, ...rest}, props) => {
-			const
-				{children, selected} = props,
-				posinset = index + 1;
+			const {children, selected} = props;
 
 			let child = children[index];
 			if (typeof child === 'string') {
@@ -94,8 +92,6 @@ const DropdownListBase = kind({
 				<Item
 					{...rest}
 					{...child}
-					aria-posinset={posinset}
-					aria-setsize={children.length}
 					data-selected={index === selected}
 					// eslint-disable-next-line react/jsx-no-bind
 					onClick={() => forward('onSelect', {selected: index}, props)}
--- a/packages/moonstone/Dropdown/DropdownList.js
+++ b/packages/moonstone/Dropdown/DropdownList.js
@@ -81,7 +81,9 @@ const DropdownListBase = kind({
 
 	handlers: {
 		itemRenderer: ({index, ...rest}, props) => {
-			const {children, selected} = props;
+			const
+				{children, selected} = props,
+				posinset = index + 1;
 
 			let child = children[index];
 			if (typeof child === 'string') {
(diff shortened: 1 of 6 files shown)


packages/moonstone/Dropdown/DropdownList.js
```diff
@@ -92,6 +94,8 @@ const DropdownListBase = kind({
 				<Item
 					{...rest}
 					{...child}
+					aria-posinset={posinset}
```
ryanjduffy: Let's move these to `Dropdown` where there other aria props are set

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -153,6 +153,16 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			pageScroll: PropTypes.bool,
 
+			/**
+			 * The role for the list.
+			 * If unset, it defaults to the `list` of `role`
+			 *
+			 * @type {String}
+			 * @default 'list'
+			 * @publie
```
ryanjduffy: Spelling: `@public`

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -153,6 +153,16 @@ const VirtualListBaseFactory = (type) => {
 			 */
 			pageScroll: PropTypes.bool,
 
+			/**
+			 * The role for the list.
+			 * If unset, it defaults to the `list` of `role`
```
ryanjduffy: We can remove this comment. The default is documented below so this is redundant.

Baekwoo-s: > We can remove this comment. The default is documented below so this is redundant.

If you mean to remove the role comment below, a lint error occurs.


```
/**
 * The role for the list.
 * If unset, it defaults to the `list` of `role`
 *
 * @type {String}
 * @default 'list'
 * @public
 */
 role: PropTypes.string,
```

`npm run lint` result:
```
@enact/moonstone@3.0.1 lint /Users/baekwoo.jung/work/enact/packages/moonstone
enact lint --strict

/Users/baekwoo.jung/work/enact/packages/moonstone/VirtualList/VirtualListBase.js
  799:35  warning  'role' is missing in props validation  enact/prop-types
```

packages/moonstone/VirtualList/VirtualListBase.js
Baekwoo-s: It was my mistake. I'll fix it.

seunghoh:
Slightly modified change log.
I am not sure if it makes sense to mention added prop`role` in the change log
@ryanjduffy please feel free to remove if not

ryanjduffy:
The change log entry makes sense to me since `role` previously would not have updated the list's role and instead set it on the outer scrollable node.

PR #2551, opened 2019-09-10
ENYO-6177: Fix scrollers to handle paging keys correctly when nested

0x64:
### Checklist

When page key is pressed on a horizontal list/scroller which is nested inside a vertical list/scroller, the outer vertical one needs to scroll but not the inner horizontal one. Currently, nothing scrolls.

### Resolution
The cause is propagation blocking of `keydown` event between the inner and the outer scrollable components. In #2418, we added `stopPropagation` in the handler to avoid incorrect scroll distance after applying a new UX concept for page key inputs and it is the direct cause of this issue. #2429 is merged after the original cause and also related to event propagation. So we must resolve the original issues of #2418 and #2429 at the same time.

This issue requires not to block propagation when the inner component does not scroll by page. The issue of #2418 is about duplicated scroll by page for 5-way key mode and pointer mode, and it requires to block propagation for scroll by page on 5-way key mode.
The issue of #2429 is to allow propagation for page key input on scrollbar buttons.

So, I rolled back related code to the status before #2418 and implemented to stop propagation only when a component actually scrolls in 5-way key mode by page key input which is not from scrollbar area.

### Additional Considerations

### Links
ENYO-6177
#2418, #2429


0x64:
> I see inconsistent results with this. If an item doesn't have focus then the page up/down doesn't propagate properly so sometimes you can page down and then you can't page up again. The sample indicates that the key code is being forwarded to `onKeyDown`, it's just not properly bubbling out to the scroller, I guess?

Thank you for catching missing conditions! The case you pointed out is pressing page key on a scrollable area without focus in pointer mode. In this case, scrollable components can not catch `keydown` event directly since key events only delivered to a _focused_ element. We are catching `keydown` event on _document (using so-called global handler)_ for this case and it means that there is no actual _bubbling_ between nested scrollable components.

So, I updated logic to check whether `keydown` is consumed or not in a scrollable component in _mounting order_. I think this approach is able to cover missing cases.

webOS101:
If the `onkeyDown` is coming from the document, why is it being caught by the action logger?  Is it forwarded out because it's theoretically consumed by that component?  Will the 'fixed' logic show it firing on the outer scroller now?  I guess I can test and see!

0x64:
> If the `onkeyDown` is coming from the document

To be clear, we are using a handler on `document` only for the case of non-focused in pointer mode. (Please see `pageKeyHandler` in `Scrollable`) If something is focused inside scrollable components, we catch the event in normal way. (`onKeyDown` in `Scrollable`)

> why is it being caught by the action logger? Is it forwarded out because it's theoretically consumed by that component?

From #1913, we are `forward`ing `onKeyDown` to the outer React element. And it looks that the same code is copied into the code for pointer mode.
I have a little concern about _when_ we need to _forward_.

> Will the 'fixed' logic show it firing on the outer scroller now?

If press page keys on an inner scroller, there will be two logs in the action logger; the first one from the inner scroller and the second one from the outer scroller.

PR #2552, opened 2019-09-10
ENYO-6244: Fix scrollWith or scrollHeight with different item size in VirtualList

ybsung:
### Checklist

Depending on the `minSize` in VirtualList `itemSize` prop and the number of items in VirtualList, the item getting focus does not scroll into view.

### Resolution

The virtual scroll dimension was calculated without considering the different item size. So I fixed the virtual scroll dimension based on the this.props.itemSizes for the different item size.

### Links

ENYO-6244


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -1061,11 +1061,15 @@ const VirtualListBaseFactory = (type) => {
 		getScrollWidth = () => (this.isPrimaryDirectionVertical ? this.scrollBounds.clientWidth : this.getVirtualScrollDimension())
 
 		getVirtualScrollDimension = () => {
-			const
-				{dimensionToExtent, primary, curDataSize} = this,
-				{spacing} = this.props;
+			if (this.props.itemSizes) {
+				return this.props.itemSizes.reduce((total, size) => total + size);
```
0x64: We assumed that an app gives an array of item size, so it looks simple and good way to calculate the total size. Just one case I concern is that there is actually no item at all. In this case, `itemSizes` could be an empty array and `TypeError` will occur in this line. So it looks better to specify the initial value.

ybsung: @0x64 
For an empty array, I added the initial value of 0.
For null or undefined, `this.props.itemSizes` is false.
For other types, `itemSizes` was defined as `PropTypes.arrayOf(PropTypes.number)`. So other types can't be acceptable.

ybsung:
When calculating scrollWith or scrollHeight, I forgot to add the sum of `spacing` size. So I pushed one more commit.

PR #2553, opened 2019-09-10
ENYO-6249: Fix to show items properly when reducing data size

ybsung:
### Checklist

When reducing `dataSize` prop in `VirtualList, some items do not show properly.

### Resolution

If the `dataSize` prop is reduced and the scroll position is needed to be updated together, the first index of the item is calculated incorrectly because the first index is updated when the `dataSize` prop is updated and the first index is updated again when scrolling and items are needed to swap.

So the first calculation for the first index was deferred by calling `setTimeout`.

### Additional Considerations

My solution is calling `setState` asynchronously with `setTimeout`. VirtuaList lifecycle is very complex now. So currently, I could not find any other way to remove the `setTimeout`.

1) Updating the `firstIndex` by scrolling was deferred in the commit - https://github.com/enactjs/enact/commit/9eeeed8194f909bd1b1958703179ef9680e0e155#diff-cd0137f683e41402c5d8c0e084142452R1316. If the commit is reverted, then there is no problem.

2) If https://github.com/enactjs/enact/pull/2529 is reverted, then there is also no problem.

So those two patches generate this issue. But currently I also could not find any other patches for those issues.

### Links

ENYO-6249

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/VirtualList.VirtualGridList` and `ui/VirtualList.VirtualList` to show items properly when reducing data size
+
 ## [3.0.1] - 2019-09-09
 
 No significant changes.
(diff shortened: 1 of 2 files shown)


ybsung:
Closing in favor of https://github.com/enactjs/enact/pull/2558.

PR #2555, opened 2019-09-10
PLAT-85398: Fix Dropdown option read out behavior

ryanjduffy:
### Checklist

Dropdown did not include the index and set size ("1 of 3") in its read out. This is a regression from #2516 which replaced `Group` (which included this support) with `VirtualList` (which doesn't).

### Resolution
* Add `aria-posinset` and `aria-setsize` to items
* Add support for overriding `role` of VirtualList


ryanjduffy:
I should pay better attention ... @Baekwoo-s already fixed this in #2550 and I missed it. :nothingtodohere:

PR #2556, opened 2019-09-11
ENYO-6259: Fix font selection in some locales 

Seonghyup:
### Checklist

Some India fonts is changed to `Lohit` fonts when weight is changed, even if `Moonstone` is assigned to font style. In addition, in some non-latin locales(`ar`, `he`, `th`), `LG Display-Regular` is still applied.

### Resolution
Compose font-face by weight to prevent using another fonts for the currently set locale .
Remove non-latin font variables and assign language-specific font for each locales to apply `LG Smart UI AR HE TH`.


packages/moonstone/MoonstoneDecorator/fontGenerator.js
```diff
@@ -30,11 +30,13 @@ const fontName = 'Moonstone';
 // };
 const fonts = {
 	'ja': {
-		regular: 'LG Smart UI JP',
-		bold: 'LG Display_JP_Bold'
+		regular: 'LG Smart UI JP'
```
Seonghyup: According to Japan office, bold JP font is made for subtitle and it is not recommended to use for UI.

packages/moonstone/styles/internal/fonts.less
```diff
@@ -33,20 +34,20 @@
 // In the case of glyphs existing in two fonts, the later glyphs will be used.
 @fonts: {
 	am: local("LG Smart UI Amharic");
-	as: local("LG Smart UI Bengali"); // same as `bn`
-	bn: local("LG Smart UI Bengali");
-	en-JP: local("LG Smart UI JP"), local("LG Display_JP_Bold") 700; // same as `ja`
```
Djspaceg: Please do not remove `en-JP`. It should match `ja` below.

packages/moonstone/styles/internal/fonts.less
```diff
@@ -33,20 +34,20 @@
...
-	gu: local("LG Smart UI Gujarati");
+	ar: local("LG Smart UI AR HE TH") normal 300 600 700;
+	as: local("LG Smart UI Bengali") normal 300 600 700; // same as `bn`
+	bn: local("LG Smart UI Bengali") normal 300 600 700;
+	gu: local("LG Smart UI Gujarati") normal 300 600 700;
+	he: local("LG Smart UI AR HE TH") normal 300 600 700;
 	hi: local("LG Smart UI Devanagari");
-	ja: local("LG Smart UI JP"), local("LG Display_JP_Bold") 700;
```
Djspaceg: Please do not remove `ja`. It should be updated to read:
```
ja: local("LG Smart UI JP") normal 300 600 700;
```

packages/moonstone/styles/internal/fonts.less
```diff
@@ -33,20 +34,20 @@
...
-	ml: local("LG Smart UI Malayalam");
+	ml: local("LG Smart UI Malayalam") normal 300 600 700;
 	mr: local("LG Smart UI Devanagari"); // same as `hi`
 	or: local("LG Smart UI Oriya");
-	pa: local("LG Smart UI Gurmukhi");
-	ta: local("LG Smart UI Tamil");
+	pa: local("LG Smart UI Gurmukhi") normal 300 600 700;
+	ta: local("LG Smart UI Tamil") normal 300 600 700;
 	te: local("LG Smart UI Telugu");
-	ur: local("LG Smart UI Urdu") normal 300 600 700;
```
Djspaceg: Please do not remove `ur`.

packages/moonstone/styles/internal/fonts.less
```diff
@@ -99,13 +100,6 @@
 .buildFontFace(@font-family-base-name; @font-system-src-moonstone-condensed-600; { font-stretch: condensed; font-weight: 600; });
 .buildFontFace(@font-family-base-name; @font-system-src-moonstone-condensed-700; { font-stretch: condensed; font-weight: 700; });
 
-
-/* ----- Manual Moonstone Locale Fonts (Stacking onto "Moonstone") ------ */
```
Djspaceg: Please do not remove this override.

packages/moonstone/styles/internal/fonts.less
Seonghyup: I think it is repeated code because `en-JP` is handled in `fontGenerator.js` and Engilsh and Japanese should be printed with `LG Smart UI` except in Japan locales(`en-JP`, `ja-JP`). Could you please explain to me why we should keep this line?

packages/moonstone/styles/internal/fonts.less
Seonghyup: I think it is repeated code because `ja-JP` is handled in `fontGenerator.js` and Engilsh and Japanese should be printed with `LG Smart UI` except in Japan locales(`en-JP`, `ja-JP`). Could you please explain to me why we should keep this line?

packages/moonstone/styles/internal/fonts.less
Seonghyup: If `ar` is handled in this language block for applying `LG Smart UI AR HE TH`, `ur` locales need to be handled in `fontGenerator.js` and in that case, this line is useless.

packages/moonstone/styles/internal/fonts.less
Djspaceg: It's not repeated. The two sets of code do different things with the same language key. These establish the font-face rules. fontGenerator provides locale-specific (vs language-specific) overrides, conditional based on the current locale setting.

If the requirements are changing and Japanese should be shown in a different font in the non-Japan case, these will need to be refactored.

packages/moonstone/styles/internal/fonts.less
Seonghyup: If `ar`, `he` and `th` locales are handled in this language block for applying `LG Smart UI AR HE TH`, this lines are useless.

packages/moonstone/styles/internal/fonts.less
Djspaceg: `ar` should not be handled in this language definition block.

packages/moonstone/styles/internal/fonts.less
Djspaceg: Please refer to the earlier comment on the PR regarding {{AR HE TH}}.

packages/moonstone/styles/internal/fonts.less
Seonghyup: `non-latin` is ambiguous and it has just three locales (`ar`, `he`, and `th`). Why do you think they should not be handled in this block?

packages/moonstone/styles/internal/fonts.less
Seonghyup: The requirements are not changed it is same as before.

packages/moonstone/styles/internal/fonts.less
Djspaceg: Correct, "non-latin" is no longer representative of what is inside that font. But these are just the variable names; the names have no impact on *how* they are used in this file.

I've explained the reasoning in the main level of this PR's comments. https://github.com/enactjs/enact/pull/2556#pullrequestreview-289452667

Seonghyup: So what about the case of `as`, `bn` and `zh-HK`, `zh-TW`. These are similarly repeated. Because "non-latin" is no longer representative of what is inside that font, I think precedence is not important things.

packages/moonstone/styles/internal/fonts.less
Djspaceg: This is not repeated code. This establishes the language font into the base "Moonstone" font and adds a "Moonstone en-JP" font family in case an app needs to refer to a specific locale's font family. `fontGenerator` only creates fonts at runtime linked to a locale-switching event.

Historically, all Japanese text should use this font. That's why this code exists. The addition of a more complete `LG Smart UI` font means that these requirements have changed. This can be removed if and only if the following are both true:
* We do not want apps to be able to reference the Japanese font directly (via `font-family: "Moonstone en-JP";`)
* We do not want these glyphs being inserted into the main/common font: "Moonstone".

If we do want the first bullet above, this code must be retained. The `LG Smart UI` font being applied on top of these will override any previous existing glyph with its own, so if that font does indeed have all of the same Japanese and english codepoints covered, this will be fully overwritten and everything will look correct. Also, the "Moonstone en-JP" font will still exist for any app currently referencing it.

packages/moonstone/styles/internal/fonts.less
Djspaceg: This is not repeated code. This establishes the language font into the base "Moonstone" font and adds a "Moonstone en-JP" font family in case an app needs to refer to a specific locale's font family. `fontGenerator` only creates fonts at runtime linked to a locale-switching event.

Historically, all Japanese text should use this font. That's why this code exists. The addition of a more complete `LG Smart UI` font means that these requirements have changed. This can be removed if and only if the following are both true:
* We do not want apps to be able to reference the Japanese font directly (via `font-family: "Moonstone en-JP";`)
* We do not want these glyphs being inserted into the main/common font: "Moonstone".

If we do want the first bullet above, this code must be retained. The `LG Smart UI` font being applied on top of these will override any previous existing glyph with its own, so if that font does indeed have all of the same Japanese and english codepoints covered, this will be fully overwritten and everything will look correct. Also, the "Moonstone en-JP" font will still exist for any app currently referencing it.

packages/moonstone/styles/internal/fonts.less
Djspaceg: Precedence in CSS is always important. It's how the language works and how everything is structured. Based on the very definition of the word precedence, these rules literally precede the later rules. They establish a base for this font to be applied correctly.

The `as`, `bn`, `zh-HK`, and `zh-TW` lines are not repeated because this block of code does multiple jobs:
* It creates locale/language-specific font-families for apps to refer to directly
* Adds the glyphs of the font to the stack of glyphs that comprise the "Moonstone" font-family.

`ar`, `he`, and `th` are **not** included here because the font that contains those glyphs includes **multiple languages** and isn't specific to that language. Therefore it gets treated differently. The way Moonstone treats these cumulative fonts is by referring to them in the "non-Latin" category and `@font-system-src-non-latin` variable. Please do not get confused by the fact that the variable name includes the words "non-latin". This is a legacy variable name and can't change now. We just have to deal with the fact that it has those words in the its name, however, this is completely irrelevant to *how* the variable is being used. It is establishing a basis for the multi-lingual "Moonstone" font-family in the rules you've removed below. Those rules must be restored.

Seonghyup: ar, he, and th do same thing that you commented about as, bn and zhs doing(creates font-faces and add the glyphs).
If multiple languages(arabic, thai, hebrew) is printed same time, it's OK because they use same font.
Strictly speaking, LG Smart UI JP has multiple language, too.
Yes, precedence in CSS and lagacy code are always important. But you've already changed the legacy codes through your overhaul PR. In this state, you never get LG Smart UI AR HE TH in ar locale. In all ar locales, LG Smart UI UR is printed now. If you would like to keep your font theme in Moonstone, it's OK, but I would like to you fix these bugs ASAP. If you do so, I will close this PR without merging.

Djspaceg: > ar, he, and th do same thing that you commented about as, bn and zhs doing(creates font-faces and add the glyphs).

Yes, however `LG Smart UI AR HE TH` is a cumulitive font that includes multiple languages. I'm content with retaining the ar, he, and th locales in this collection. Having them generate a standalone font-family is fine, even if it contains other languages, because at least it will cover the target language too.

>But you've already changed the legacy codes through your overhaul PR.

This was before we released 3.0.0, when breaking changes were acceptable. Now they're not, until 4.0.0. So we must live with this variable name until then.

>In this state, you never get LG Smart UI AR HE TH in ar locale. In all ar locales, LG Smart UI Urdu is printed now.

This is because the Urdu font has conflicting characters with the `LG Smart UI AR HE TH` font. The last one applied gets to stay. The Urdu font is likely displaying because the "legacy" rules don't specifically call-out "normal" for their font weight, so this rule takes precedence over the one below it, since it's more specific. Adding "normal" to the legacy ruleset will resolve this and will mean that we don't break an app that refers to `font-family: "Moonstone ur";`.

> ...I would like to you fix these bugs ASAP. If you do so, I will close this PR without merging.

I am interested in keeping and using this PR as it does include good changes and fixes. I was hoping, through discussion, that we could make changes/updates to parts of it and I could simply approve it, rather than redoing the work all on my own.

Seonghyup: >> This is because the Urdu font has conflicting characters with the LG Smart UI AR HE TH font. The last one applied gets to stay. The Urdu font is likely displaying because the "legacy" rules don't specifically call-out "normal" for their font weight, so this rule takes precedence over the one below it, since it's more specific. Adding "normal" to the legacy ruleset will resolve this and will mean that we don't break an app that refers to font-family: "Moonstone ur";.

Does you mean like this?
```
@font-system-src-non-latin-400:  local("LG Smart UI AR HE TH Regular") normal, local("LGSmartUIARHETH-Regular") normal;

Comment (Djspaceg) [reply packages/moonstone/styles/internal/fonts.less]: Technically this won't work since this variable just holds multiple font file references, not weights. We'll need to add the `normal` keyword down below, in the first line of this section:

```
/* ----- Manual Moonstone Locale Fonts (Stacking onto "Moonstone") ------ */
.buildFontFace(@font-family-base-name; @font-system-src-non-latin-400, normal);
.buildFontFace(@font-family-base-name; @font-system-src-non-latin-300; 300);
.buildFontFace(@font-family-base-name; @font-system-src-non-latin-600; 600);
.buildFontFace(@font-family-base-name; @font-system-src-non-latin-700; 700);
```

File: packages/moonstone/MoonstoneDecorator/fontGenerator.js [hunk 6]
```diff
@@ -32,11 +32,11 @@ const fonts = {
 	'ja': {
 		regular: 'LG Smart UI JP'
 	},
+	'ur': {
+		regular: ['LG Smart UI_Urdu', 'LGSmartUIUrdu'] // This needs 2 references because the "full name" differs from the "family name". To target this font file directly in all OSs we must also include the "postscript name" in addition to the "full name".
```
Seonghyup: I doesn't work. We need to remove `_` (`LG Smart UI Urdu`)

packages/moonstone/styles/internal/fonts.less
```diff
@@ -41,13 +40,15 @@
...
 	kn: local("LG Smart UI Kannada");
+	ja: local("LG Smart UI JP") normal 300 600 700;
 	ml: local("LG Smart UI Malayalam") normal 300 600 700;
 	mr: local("LG Smart UI Devanagari"); // same as `hi`
 	or: local("LG Smart UI Oriya");
 	pa: local("LG Smart UI Gurmukhi") normal 300 600 700;
 	ta: local("LG Smart UI Tamil") normal 300 600 700;
 	te: local("LG Smart UI Telugu");
 	th: local("LG Smart UI AR HE TH") normal 300 600 700;
+	ur: local("LG Smart UI_Urdu") normal 300 600 700;
```
Seonghyup: It doesn't work. We need to remove `_` (`LG Smart UI Urdu`)

Djspaceg: I believe you must have a different set of fonts from me. 

From my PR comment:
> * The Urdu font name was updated as it didn't match the metadata in the font, which is unique in the font collection, as its family name (LG Smart UI Urdu), full name (LG Smart UI_Urdu), and postscript names (LGSmartUIUrdu) are all different. This requires special handling to be supported universally. Surprisingly this and the AR HE TH font is the only case of mismatch in the font collection.

I'll download and flash the latest TV build and extract the fonts from it so I'll be guaranteed to be working with the latest set of fonts.

Djspaceg: I've done some extensive testing and it appears the current version of blink on the Reference Platform build (72) no longer uses the full-name but now prefers the family-name in the metadata. I've removed the `_`, as you suggested, and it works as expected now on the TV. Also, I've removed the support for the `"Moonstone ur"` font-family, since it was causing the Urdu font to be used with all Arabic text, which is not what we want. Urdu font should only be used on Urdu/Arabic text when in the Urdu locale. It, like Japanese is technically a different typeface, so it shouldn't build onto the "LG Smart UI" -> "Moonstone" font-family.

I also removed the `normal` keyword from the non-Latin area as it didn't seem to actually be necessary for the fonts to apply, once this collision was removed.

PR has been updated.

packages/moonstone/MoonstoneDecorator/fontGenerator.js
Djspaceg: Removed `_` to match the family name instead of the full name.

packages/moonstone/styles/internal/fonts.less
Seonghyup: Good, please test on actual environment before comment.

PR #2558, opened 2019-09-11
ENYO-6249: Fix to show items properly when reducing data size

ybsung:
### Checklist

When reducing `dataSize` prop in `VirtualList, some items do not show properly.

### Resolution

If the `dataSize` prop is reduced and the scroll position is needed to be updated together, the first index of the item is calculated incorrectly because the first index is updated when the `dataSize` prop is updated and the first index is updated again with a wrong previous first index when scrolling and items are needed to swap.

So the second calculation for the first index was deferred until the first index updates.

### Links

ENYO-6249


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -391,16 +393,28 @@ const VirtualListBaseFactory = (type) => {
...
 			}
+
+			const maxPos = this.isPrimaryDirectionVertical ? this.scrollBounds.maxTop : this.scrollBounds.maxLeft;
+
+			if (!this.deferScrollTo && this.scrollPosition > maxPos) {
+				this.props.cbScrollTo({position: (this.isPrimaryDirectionVertical) ? {y: maxPos} : {x: maxPos}, animate: false});
+			}
 		}
 
+		deferScrollTo = false
```
ryanjduffy: I don't _think_ we need this to be a class member since it's set and only read within `componentDidUpdate`. Am I reading that correctly?

ybsung: Yes. I'll fix it. Thanks.

webOS101: You need to fix your clock, it's too early to be up!

ybsung:
@ryanjduffy 
I fixed it and verified it. Thanks.

PR #2563, opened 2019-09-13
ENYO-6118: Add QA sample to Popup to test focus resumes after open state

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/qa/Popup.js
+++ b/packages/sampler/stories/qa/Popup.js
@@ -56,9 +56,7 @@ class PopupResumeFocusAfterOpenState extends React.Component {
 				<div>Popup will open and dismiss immediately, ensure spotlight still functional.</div>
 				<Button onClick={this.handlePopup}>Open popup</Button>
 				<Popup open={this.state.isPopup}>
-					<buttons>
-						<Button>close</Button>
-					</buttons>
+					<Button>close</Button>
 				</Popup>
 			</div>
 		);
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Popup.js
```diff
@@ -30,6 +30,59 @@ const PopupFromSelfOnlyContainer = Toggleable(
...
+		this.setState({
+			isNotification: false
+		});
+	}
+
+	render () {
+		return (
+			<div>
+				<Button onClick={this.handlePopup}>open</Button>
+				<Button>hello</Button>
```
webOS101: Is this button supposed to open the popup?  If so, let's rename the buttons to clarify which does what.  Otherwise, there's no way to open the `Popup`

packages/sampler/stories/qa/Popup.js
```diff
@@ -30,6 +30,59 @@ const PopupFromSelfOnlyContainer = Toggleable(
...
+	handleNotification = () => {
+		this.setState({
+			isNotification: false
+		});
+	}
+
+	render () {
+		return (
+			<div>
+				<Button onClick={this.handlePopup}>open</Button>
```
webOS101: Do we need to add a paragraph with some text describing what should happen here?

packages/sampler/stories/qa/Popup.js
LucieRoy: It was based on ENYO-6074. The Popup opens quickly and closes quickly. Then we switch to 5-way and 5-way Spot 'hello'.

packages/sampler/stories/qa/Popup.js
LucieRoy: It is a good idea. I was going to write a tc for it anyway.  How about:
"Click on "open" so the Popup appears and disappears quickly. Then 5-way Spot 'hello'.
? 
Please modify as needed.

PR #2565, opened 2019-09-13
ENYO-6200: Add position prop to ProgressBarTooltip

ryanjduffy:
### Checklist

Authors are unable to control the position of `ProgressBarTooltip` and `SliderTooltip` effectively with the current API. For example, it is not possible to force the tooltip to always display to one side.

### Resolution
* Add `position` prop with broader set of values to replace `side`
* Deprecate `side`

### Additional Considerations
* Reordered knobs in `ProgressBar` and `Slider` stories to force the component tab to appear before the tooltip tab
* Added an internal `deprecated` prop type to simplify deprecating props
* Added a custom prop type validate for `position` to warn when using invalid values with `orientation`

### DIFF
--- a/packages/moonstone/ProgressBar/ProgressBarTooltip.js
+++ b/packages/moonstone/ProgressBar/ProgressBarTooltip.js
@@ -13,6 +13,7 @@ import Tooltip from '../TooltipDecorator/Tooltip';
 import css from './ProgressBarTooltip.module.less';
 
 // prop-type validator that warns on invalid orientation + position
+/* istanbul ignore next */
 const validatePosition = (base) => (props, key, componentName, location, propFullName, ...rest) => {
 	const {position} = props;
 	let result = base(props, key, componentName, location, propFullName, ...rest);
@@ -263,12 +264,12 @@ const ProgressBarTooltipBase = kind({
 			return styler.append(
 				orientation,
 				{
-					above: v === 'above',
-					below: v === 'below',
-					before: h === 'before',
-					after: h === 'after',
-					left: h === 'left' || (h === 'auto' && proportion <= 0.5),
-					right: h === 'right' || (h === 'auto' && proportion > 0.5)
+					above: (v === 'above'),
+					below: (v === 'below'),
+					before: (h === 'before'),
+					after: (h === 'after'),
+					left: (h === 'left' || (h === 'auto' && proportion <= 0.5)),
+					right: (h === 'right' || (h === 'auto' && proportion > 0.5))
 				}
 			);
 		},
--- a/packages/moonstone/ProgressBar/ProgressBarTooltip.js
+++ b/packages/moonstone/ProgressBar/ProgressBarTooltip.js
@@ -86,7 +86,7 @@ const getSide = (orientation, side, position) => {
 
 		if (orientation === 'horizontal') {
 			// Testing for 'after' so if side === left or right, we default to "above"
-			return [side === 'after' ? 'below' : 'above', 'auto'];
+			return ['auto', side === 'after' ? 'below' : 'above'];
 		}
 
 		return [side, 'auto'];
--- a/packages/moonstone/ProgressBar/ProgressBarTooltip.js
+++ b/packages/moonstone/ProgressBar/ProgressBarTooltip.js
@@ -11,6 +11,7 @@ import Tooltip from '../TooltipDecorator/Tooltip';
 
 import css from './ProgressBarTooltip.module.less';
 
+// prop-type validator that warns on invalid orientation + position
 const validatePosition = (base) => (props, key, componentName, location, propFullName, ...rest) => {
 	const {position} = props;
 	let result = base(props, key, componentName, location, propFullName, ...rest);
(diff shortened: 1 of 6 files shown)


ryanjduffy:
Oof. Should add some unit tests to this.

packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -11,22 +12,84 @@ import Tooltip from '../TooltipDecorator/Tooltip';
...
-	return valid ? side : 'before';
+		if (orientation === 'horizontal') {
+			switch (position) {
+				case 'above':
+				case 'below':
+					return ['auto', position];
+				case 'after':
+				case 'before':
+				case 'left':
+				case 'right':
```
webOS101: As I mentioned before, I super-prefer to see the defaults in `default:`.  I guess there are more than twice as many valid ones as these 4 invalid ones.  Could possibly get all the 'above' and 'below' options on one line each?

packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -97,10 +213,18 @@ const ProgressBarTooltipBase = kind({
 		 * * `'right'` renders to the right of a `vertical` ProgressBar/Slider regardless of locale
 		 *
 		 * @type {String}
-		 * @default 'before'
+		 * @deprecated Deprecated since 3.1 until 4.0 in favor of `position`
```
webOS101: I think this duplicates deprecated.  Will verify.

packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -135,39 +258,54 @@ const ProgressBarTooltipBase = kind({
...
+		arrowAnchor: ({orientation, position, proportion, rtl}) => {
 			if (orientation === 'vertical') return 'middle';
-			return proportion > 0.5 ? 'left' : 'right';
+
+			const [h] = getSide(orientation, false, position);
+			switch (h) {
+				case 'auto':
+					return proportion > 0.5 ? 'left' : 'right';
+				case 'before':
+					return rtl ? 'right' : 'left';
```
webOS101: We'll need to communicate this new RTL ability in the release notes.  It should probably also be called out above in the description (at least, when we suppress the default behavior).

packages/sampler/stories/default/ProgressBar.js
```diff
@@ -15,20 +15,25 @@ storiesOf('Moonstone', module)
 	.add(
 		'ProgressBar',
 		() => {
-			const side = select('side', ['after', 'before', 'left', 'right'], ProgressBarTooltipConfig, 'before');
+			// added here to force Storybook to put the ProgressBar tab first
+			const disabled = boolean('disabled', ProgressBarConfig);
```
webOS101: This is a bummer we need to do this but I guess it makes sense.  We can fix other stories (if any) later that have a similar problem

packages/moonstone/ProgressBar/ProgressBarTooltip.js
webOS101: Nope, this is fine!

packages/moonstone/ProgressBar/ProgressBarTooltip.js
ryanjduffy: okay

packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -65,6 +128,59 @@ const ProgressBarTooltipBase = kind({
 		 */
 		percent: PropTypes.bool,
 
+		/**
+		 * Position of the tooltip with respect to the progress bar.
+		 *
+		 * * For `orientation="horizontal"` progress bars, the default value is `'above'`.
+		 * * For `orientation="vertical"` progress bars, the default value is `'above'`.
```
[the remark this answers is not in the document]
ryanjduffy: yep

packages/moonstone/ProgressBar/ProgressBarTooltip.js
ryanjduffy: It is vaguely noted but agree it should be expanded.

packages/moonstone/ProgressBar/ProgressBarTooltip.js
```diff
@@ -11,22 +12,84 @@ import Tooltip from '../TooltipDecorator/Tooltip';
 
 import css from './ProgressBarTooltip.module.less';
 
+// prop-type validator that warns on invalid orientation + position
+/* istanbul ignore next */
+const validatePosition = (base) => (props, key, componentName, location, propFullName, ...rest) => {
```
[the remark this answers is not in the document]
ryanjduffy: Fair point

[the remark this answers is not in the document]
webOS101: To clarify my point, I meant only in propTypes we don't cross-validate.  We do cross-validate in other spots.

PR #2566, opened 2019-09-16
ENYO-6257: RadioItem: background-color doesn't match GUI spec

Sangwook-Lee:
### Checklist

As the GUI specification has been updated.

### Resolution
I've updated the colors.

### Links
ENYO-6257

### DIFF
--- a/packages/moonstone/styles/colors-light.less
+++ b/packages/moonstone/styles/colors-light.less
@@ -199,8 +199,8 @@
 
 // RadioItem
 // ---------------------------------------
-@moon-radio-item-indicator-color: @moon-dark-gray;
-@moon-radio-item-indicator-bg-color: @moon-white;
+@moon-radio-item-indicator-color: #333333;
+@moon-radio-item-indicator-bg-color: fade(#888888, 50%);
 @moon-radio-item-focus-indicator-color: @moon-spotlight-text-color;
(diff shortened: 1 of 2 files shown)


packages/moonstone/styles/colors-light.less
```diff
@@ -199,8 +199,8 @@
 
 // RadioItem
 // ---------------------------------------
-@moon-radio-item-indicator-color: @moon-dark-gray;
-@moon-radio-item-indicator-bg-color: @moon-white;
+@moon-radio-item-indicator-color: #333333;
+@moon-radio-item-indicator-bg-color: fade(#888888, 50%);
 @moon-radio-item-focus-indicator-color: @moon-spotlight-text-color;
 @moon-radio-item-focus-indicator-bg-color: fade(#a6a6a6, 50%);
```
ryanjduffy: The specs seems to use the same bg color for focus as well so I think we should update this to `#888888` as well.

Sangwook-Lee: You meant `@moon-radio-item-focus-indicator-bg-color: fade(#888888, 50%);`  ?

ryanjduffy: Yes

Sangwook-Lee: Done!

PR #2567, opened 2019-09-16
ENYO-6268: Resolve Sampler build errors

JayCanuck:
### Checklist

* Newly-release Storybook 5.2.0 uses updated dependencies (css-loader, etc.) which break the current configuration.

### Resolution
* Lock to Storybook 5.1.9 to avoid issue until all configuration issues can be resolved


ryanjduffy:
@JayCanuck Did you open an issue with Storybook on this by chance? Seems like we should communicate to them that they can't make major version dependency changes of public-facing dependencies in a minor version.

JayCanuck:
I haven't.  Looking through their changelog it looks like they regularly do breaking changes in minor releases.  From now on, I'd opt for us to use `~` rather than `^` for Storybook packages.

ryanjduffy:
:(

PR #2569, opened 2019-09-17
ENYO-6272: Fix ScrollerNative generating console error when pressing page down after 5-way down several times in ScrollerNative

ybsung:
### Checklist

If pressing page down when Spotlight is on the item out of viewport while 5-way Down several times, Spotlight suddenly hides. There is the following error in console.
```
Cannot read property 'closest' of null
```

### Resolution

When pressing page down, then Spotlight stored the current focus point, then tried to restore focus from the point. But the point is out of viewport sometimes. So the point is clamped with the viewport.

### Links

ENYO-6272

### Comments

This issue should be fixed in https://github.com/enactjs/enact/pull/2532 properly. But I forgot to update ScrollableNative part.


ybsung:
We don't need to update CHANGELOG.md because ScrollerNative is a private component.

PR #2570, opened 2019-09-19
ENYO-6278: Fix moontone internationalization resource loading

JayCanuck:
### Checklist

* When we switch from `@enact/i18n/ilib` v13.2.0 to an external `ilib` v14.2.0 package, it introduced an internal iLib change in how iLib cache ResBundles. It appears that now iLib only allow 1 of each kind of file, and caches it.  So when the app loads the app's `['strings.json', 'en/strings.json', 'en/US/strings.json']` from the app's `./resources` directory, those files/contents are cached to their respective objects on `ilib.data`.
    
    Moonstone (and agate/other themes) have their own embedded localizations. When iLib then attempts to load the ResBundle strings from `./node_modules/@enact/moonstone/resources`, it's seeing the `'*strings.json'` files cached already in `ilib.data` and it uses that (the app's localization data) instead, to avoid what it thinks would be a repetitious XHR. As a result, Moonstone localizations are never correctly applied or even loaded.

### Resolution
* This workaround sets up a local moonstone-contained cache object which it swaps in and out to `ilib.data` to avoid app resource cache interfering with moonstone resource cache.

### Additional Considerations
* Since moonstone's `$L` is guaranteed sync, we don't need to worry about any async issues (cross/parallel requests could cause resource cache to be saved to the wrong object). If/when the moonstone `$L` is updated to support async, an alternate solution may be required.

### Comments
* Additionally, I think it will be worthwhile to get in touch with the iLib folks to see if we can get them to support differing iLib cache objects depending on `loadParams` object.

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone` internationalization resource loading
 - `moonstone/VirtualList.VirtualList` to scroll properly when an item gets focus in VirtualList with different item size
 
 ## [3.1.0] - 2019-09-16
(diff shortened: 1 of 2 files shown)


seunghoh:
@gouniLee 
I would like to mention Edwin here but I can't. Please share this issue to Edwin

Sangwook-Lee:
I built verification image(epk) on this PR and verified this fix.

PR #2572, opened 2019-09-19
PLAT-85707: Forward drag events from Scrollable

germboy:
### Checklist

`ui/Scrollable` did not forward drag events.

### Resolution
If the appropriate drag events are present (`onDrag`, `onDragStart`, `onDragEnd`), we will forward them.

### Additional Considerations
- These event props are private, so I did not include a changelog. I also removed a reference to a non-existent `this.onTouchStart` class method.

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -676,7 +676,7 @@ class ScrollableBase extends Component {
 		);
 
 		if (this.props.onFlick) {
-			forward('onFlick', ev, this.props);
+			forward('onFlick', {...ev, flickTarget: this.flickTarget}, this.props);
 		}
 	}
 
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -676,7 +676,7 @@ class ScrollableBase extends Component {
 		);
 
 		if (this.props.onFlick) {
-			forward('onFlick', {...ev, flickTarget: this.flickTarget}, this.props);
+			forward('onFlick', ev, this.props);
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


packages/ui/Scrollable/Scrollable.js
```diff
@@ -589,6 +613,10 @@ class ScrollableBase extends Component {
 		this.isDragging = true;
 		this.dragStartX = this.scrollLeft + this.getRtlX(ev.x);
 		this.dragStartY = this.scrollTop + ev.y;
+
+		if (this.props.onDragStart) {
```
ryanjduffy: These checks aren't required as `forward` will safely bail if the event doesn't exist on `props`.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -589,6 +613,10 @@ class ScrollableBase extends Component {
...
+			forward('onDragStart', ev, this.props);
```
ryanjduffy: We _usually_ put the `forward` at the top of the function to ensure we always forward (unless there are occasions in which we stop propagation). Do you know of a reason why we need them at the end?

germboy: No real reason that I'm aware of. The placement at the end of the handler here was more-or-less consistency with the other events within the file, although I'm not certain those have a specific purpose or could/should be moved to the top either. I can change these, but the other handlers (`onMouseDown`, `onFlick`) could be verified as well later.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -204,6 +204,30 @@ class ScrollableBase extends Component {
 		 */
 		noScrollByWheel: PropTypes.bool,
 
+		/**
+		 * Called when trigerring a drag event.
+		 *
+		 * @type {Function}
+		 * @private
```
webOS101: I was going to remove the `@private` since it's redundant and, if we opened this, would want it public, I think... but...  I'd have to update all the other props, too.  So, I didn't do it now.

PR #2573, opened 2019-09-20
PLAT-86061: Fix TypeScript types for handlers and computed functions in kind

RenukaAtale:
Update: ComputedFunction and HandlerFunction in kind make event, props and context
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/core/kind/kind.js
+++ b/packages/core/kind/kind.js
@@ -21,17 +21,17 @@ import styles from './styles';
 /**
  * @callback ComputedPropFunction
  * @memberof core/kind
- * @param {Object<string, any>} [props]
- * @param {Object<string, any>} [context]
+ * @param {Object<string, any>} props
+ * @param {Object<string, any>} context
  * @returns any
  */
 
 /**
  * @callback HandlerFunction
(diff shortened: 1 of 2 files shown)


packages/core/kind/kind.js
```diff
@@ -21,17 +21,17 @@ import styles from './styles';
 /**
  * @callback ComputedPropFunction
  * @memberof core/kind
- * @param {Object<string, any>} [props]
- * @param {Object<string, any>} [context]
+ * @param {Object<string, any>} props
+ * @param {Object<string, any>} context
```
webOS101: I think we should leave `context` optional as it will be undefined if context is not applied, I believe.  @ryanjduffy ?

ryanjduffy: You may be right. I'll have to check again

ryanjduffy: Looks like context is set to an empty object when `contextType` is unset or set wrong so the proposed change is correct.

webOS101: :themoreyouknow:

PR #2575, opened 2019-09-20
ENYO-6248: Fix VirtualList with different item sizes to scroll properly

0x64:
### Checklist

VirtualList with Different item Size could not reach the end via 5-way key or fast wheeling.
- Sometimes, a list was scrolled _backward_ during scroll down via 5-way key
- (Found after fixing the issue above) Sometimes, items at the end of a list did not display via fast wheeling

### Resolution
The first issue (backward scrolling) occurs due to not-yet-calculated position info of item to be rendered. A list already has all item sizes (from apps) but it doesn't calculate all item positions at the beginning to spread out computation cost at mounting. So, in some cases, position info could be not calculated yet. This is fixed by adding calculation logic for the case.

The second issue occurs due to a simple bug on updating threshold.max value which is used to reuse virtualized slots.

### Links
ENYO-6248


ybsung:
LGTM

ryanjduffy:
Pushed a commit with some minor refactoring.

PR #2578, opened 2019-09-23
PLAT-86061 : Fix docs for event handlers and adaptEvent

RenukaAtale:
…ction

### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -676,10 +676,10 @@ const call = function (method) {
  * ```
  *
  * @method   adaptEvent
- * @param    {EventAdapter}  adapter  Function to adapt the event payload
- * @param    {Function}  handler  Handler to call with the new event payload
+ * @param    {EventAdapter}     adapter  Function to adapt the event payload
+ * @param    {HandlerFunction}  handler  Handler to call with the handler function
  *
- * @returns  {HandlerFunction}    New event payload
+ * @returns  {HandlerFunction}           Returns value of the handler function
  * @curried
  * @memberof core/handle
  * @public
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -679,7 +679,7 @@ const call = function (method) {
  * @param    {EventAdapter}     adapter  Function to adapt the event payload
  * @param    {HandlerFunction}  handler  Handler to call with the handler function
  *
- * @returns  {HandlerFunction}           Returns value of the handler function
+ * @returns  {HandlerFunction}           Returns an [event handler]{@link core/handle.HandlerFunction} (suitable for passing to handle) that returns the result of handler
  * @curried
  * @memberof core/handle
  * @public
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -91,7 +91,7 @@ import {is} from '../keymap';
 /**
  * @callback HandlerFunction
  * @memberof core/handle
- * @param {Object<string, any>} event
+ * @param {any} event
  * @param {Object<string, any>} props
  * @param {Object<string, any>} context
  */
@@ -99,7 +99,7 @@ import {is} from '../keymap';
  /**
  * @callback EventAdapter
  * @memberof core/handle
- * @param {Object<string, any>} event
+ * @param {any} event
  * @param {Object<string, any>} props
  * @param {Object<string, any>} context
  * @returns {Object<string, any>}
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -102,7 +102,7 @@ import {is} from '../keymap';
  * @param {any} event
  * @param {Object<string, any>} props
  * @param {Object<string, any>} context
- * @returns {Object<string, any>}
+ * @returns {any}
  */
 
 
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -92,9 +92,19 @@ import {is} from '../keymap';
  * @callback HandlerFunction
  * @memberof core/handle
  * @param {Object<string, any>} event
- * @param {any} props
+ * @param {Object<string, any>} props
  * @param {Object<string, any>} context
  */
+
+ /**
+ * @callback EventAdapter
+ * @memberof core/handle
+ * @param {Object<string, any>} event
+ * @param {Object<string, any>} props
+ * @param {Object<string, any>} context
+ * @returns {Object<string, any>}
+ */
+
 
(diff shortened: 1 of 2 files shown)


packages/core/handle/handle.js
```diff
@@ -92,10 +92,20 @@ import {is} from '../keymap';
  * @callback HandlerFunction
  * @memberof core/handle
  * @param {Object<string, any>} event
- * @param {any} props
+ * @param {Object<string, any>} props
  * @param {Object<string, any>} context
```
webOS101: @ryanjduffy Should we update the typedef for this to set the return type?  Seems to be working fine without it, but it seems a touch odd to me.

packages/core/handle/handle.js
```diff
@@ -666,9 +676,8 @@ const call = function (method) {
  * ```
  *
  * @method   adaptEvent
- * @param    {Function}  adapter  Function to adapt the event payload
+ * @param    {EventAdapter}  adapter  Function to adapt the event payload
  * @param    {Function}  handler  Handler to call with the new event payload
- * @param    {...*}      [args]   Additional args passed to both `adapter` and `handler`
  *
  * @returns  {HandlerFunction}    New event payload
```
webOS101: We should update the description to describe that it returns a handler function.  Possibly worth saying that it returns the return value of the handler when invoked.

packages/core/handle/handle.js
```diff
@@ -666,9 +676,8 @@ const call = function (method) {
  * ```
  *
  * @method   adaptEvent
- * @param    {Function}  adapter  Function to adapt the event payload
+ * @param    {EventAdapter}  adapter  Function to adapt the event payload
  * @param    {Function}  handler  Handler to call with the new event payload
```
webOS101: This should be updated to `HandlerFunction`.

packages/core/handle/handle.js
```diff
@@ -676,10 +676,10 @@ const call = function (method) {
  * ```
  *
  * @method   adaptEvent
- * @param    {EventAdapter}  adapter  Function to adapt the event payload
- * @param    {Function}  handler  Handler to call with the new event payload
+ * @param    {EventAdapter}     adapter  Function to adapt the event payload
+ * @param    {HandlerFunction}  handler  Handler to call with the handler function
  *
- * @returns  {HandlerFunction}    New event payload
+ * @returns  {HandlerFunction}           Returns value of the handler function
```
webOS101: Maybe the description should be closer to the description of the return from `oneOf`.  Maybe something like:
Returns a function that accepts an event (to be modified) and props and returns the return value of `handler`.
Or
Returns an [event handler]{@link core/handle.HandlerFunction} (suitable for passing to `handle`) that returns the result of `handler`
Something like that?

ryanjduffy:
@RenukaAtale + @webOS101 - what is the "problem" this change is resolving? I couldn't tell from the ticket or the description above.

webOS101:
TS errors when attempting to supply functions for `adapter` and `handler` as well as incorrect return type of `adaptEvent` so it couldn't be passed to `handle` without error.

webOS101:
test code:  https://github.com/enactjs/samples/pull/85/files#diff-42711af9925732f88724711572f75fa0R18-R27

packages/core/handle/handle.js
ryanjduffy: If we wanted to be strict, we could document the return as `Boolean | undefined`.

packages/core/handle/handle.js
```diff
@@ -92,10 +92,20 @@ import {is} from '../keymap';
...
  * @param {Object<string, any>} event
- * @param {any} props
+ * @param {Object<string, any>} props
  * @param {Object<string, any>} context
  */
 
+ /**
+ * @callback EventAdapter
+ * @memberof core/handle
+ * @param {Object<string, any>} event
```
ryanjduffy: I'm concerned this will be more restrictive than we ultimately want. While all the native events and all our events fulfill this type, we can't guarantee that all callbacks will which would force some type gymnastics on consumers for little value.

packages/core/kind/kind.js
```diff
@@ -29,7 +29,7 @@ import styles from './styles';
 /**
  * @callback HandlerFunction
  * @memberof core/kind
- * @param {any} event
+ * @param {Object<string, any>} event
```
ryanjduffy: Same here on restrictions.

ryanjduffy:
For a little context, my comments about the event arg stem from seeing other code bases that don't adhere to "events are objects" as we do and instead pass back other values. In some cases, they pass multiple args back which will be incompatible with `handle` but there's little we can do about that universally.

ryanjduffy:
Need to fix the linting error as well :)

packages/core/handle/handle.js
webOS101: I suspect, though, changing to `any` would mean that 100% of code will have to have a type override to access properties on events.  Would developers have to anyhow?

ryanjduffy: They would have to anyway. TS will complain about implicit `any`s by default.

https://www.typescriptlang.org/play/index.html#code/FASwdgLgpgTgZgQwMZQAQFEBuVIAkFgAmANrKgN7CrWoAUVNjUmAXKgQJ4A0Dj1ADjAD2-AM5sASlCRCYhADyiIMcAHMu7MBwB8PPjRmQoADwiTpshUpVh1mnb1QBKNgCMhQ0gVQAfVJiEQQmAAX2BgQmliBBg0OABXMCQIECEwVAALAhI0WgA6AqyiUhhxDGw8bJKAbQBdF1R3TyhvPwCggG5whKSUtNRCIQBlCHi4ODpmNgBZIXjRKCwcCA1BETKpGTlFZTUNTm0nCkcQCdpmPNd4iAh+gF4H1ABGI8p9akNRZrziIVVaNZiPIpCCkJxdRhhMLAHrJVLpADuIAgGQAggA5eIAW1cZFoMkSZlQYGxuJgq2EYnMWysu1s+y0h2OjFOdAJkFQDzuz1ejkYn2+v3+gNEwORYIhNCh3UScP6BH6+LmkAp62plh2NjsB15LLO7IgnMeL2Z7wFpB+fwBlNFIIljmlwE+huMnMyVSgtAVYA0gxGYzgGiRKIxpNg4OAQA

If you toggle off the `noImplicitAny` option, the above will be free of errors.

webOS101: I guess it makes sense.  Particularly when talking about the `MouseEvent` case.  We aren't really taking advantage of the proper event types in any of the places where we could.

PR #2580, opened 2019-09-24
ENYO 6287: Add QA Header sample for RTL with long string

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/qa/Header.js
+++ b/packages/sampler/stories/qa/Header.js
@@ -16,13 +16,13 @@ const inputData = {
 	shortTitle: 'Enact',
 	shortTitleBelow: 'An app framework',
 	shortSubTitleBelow: 'Built atop React',
-	shortRtlTitle: 'כתוביות למט ',
+	shortRtlTitle: 'כתוביות למט',
 	shortRtlTitleBelow: 'כתוביות למט',
 
 	longTitle: 'Core, The building blocks of an Enact application. Moonstone, our TV-centric UI library.',
 	longTitleBelow: 'An app development framework built atop React that’s easy to use, performant and customizable. The goal of Enact is to provide the building blocks for creating robust and maintainable applications.',
 	longSubTitleBelow: 'With over 50 components to choose from, Moonstone provides a solid base for creating applications designed for large screens. The Enact team welcomes contributions from anyone motivated to help out.',
-	longRtlTitle: 'כתוביות למט כתוביות למט סביבה לימודית מתוקשבת מבית מטח שינויים מרחיקי לכת": כנס מט"ח מציג את המהפכות בעולם החינוך'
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/qa/Header.js
```diff
@@ -16,12 +16,13 @@ const inputData = {
 	shortTitle: 'Enact',
 	shortTitleBelow: 'An app framework',
 	shortSubTitleBelow: 'Built atop React',
-	shortRtlTitle: 'כתוביות למט',
+	shortRtlTitle: 'כתוביות למט ',
```
Djspaceg: Is this space addition intentional?

packages/sampler/stories/qa/Header.js
```diff
@@ -16,12 +16,13 @@ const inputData = {
...
 	shortRtlTitleBelow: 'כתוביות למט',
 
 	longTitle: 'Core, The building blocks of an Enact application. Moonstone, our TV-centric UI library.',
 	longTitleBelow: 'An app development framework built atop React that’s easy to use, performant and customizable. The goal of Enact is to provide the building blocks for creating robust and maintainable applications.',
-	longSubTitleBelow: 'With over 50 components to choose from, Moonstone provides a solid base for creating applications designed for large screens. The Enact team welcomes contributions from anyone motivated to help out.'
+	longSubTitleBelow: 'With over 50 components to choose from, Moonstone provides a solid base for creating applications designed for large screens. The Enact team welcomes contributions from anyone motivated to help out.',
+	longRtlTitle: 'כתוביות למט כתוביות למט סביבה לימודית מתוקשבת מבית מטח שינויים מרחיקי לכת": כנס מט"ח מציג את המהפכות בעולם החינוך'
```
Djspaceg: Where did this text come from? Is it a Hebrew translation of the `longTitle` text?

Our `locale` knob in the Global Knobs refers to `ar-SA` (Arabic in Saudi Arabia) or `ur-PK` (Urdu in Pakistan) for RTL text tests, so the language in each RTL test should probably be one of those unless it's necessary for it to be otherwise. I'd lean toward Arabic only because Urdu also uses a custom language font, but either is ok really.

I guess the shortRtl text above does already refer to Hebrew. Oops. Maybe we can correct that in this PR too, for consistency? Just switch everything to Arabic? Thoughts?

packages/sampler/stories/qa/Header.js
LucieRoy: Extra space was not intended. Thank you for catching it. Corrected.

packages/sampler/stories/qa/Header.js
LucieRoy: Arabic is fine too. It is usually the standard for RTL. I thought there was a reason for Hebrew. Updated the string, long and short.
from http://generator.lorem-ipsum.info/_arabic

Djspaceg: For future reference, typically what I do is write out the test string in English, then go to Google Translate™ and translate it to the target language. That way, a native reader of that language could at least a tell what the string says.

PR #2581, opened 2019-09-25
ENYO-6285: VirtualList scrolls speedfully when wheeling

ybsung:
### Checklist

When wheeling over a VirtualList, the VirtualList scrolls smoothly.

### Resolution

Due to the patch for ENYO-6230, VirtualList rendered again whenever the list scroll position updated. So I reverted the patch and create a new patch for ENYO-6230.

If firstIndex state is not updated, setState() in didScroll() is usually not called as before except the case when setState() is called in other functions so that the state could be always overriden when an app wants to scroll.

### Links

ENYO-6285

### DIFF
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -319,8 +319,9 @@ const VirtualListBaseFactory = (type) => {
 
 		componentDidUpdate (prevProps, prevState) {
 			let deferScrollTo = false;
-
 			const {firstIndex, numOfItems} = this.state;
+
+			this.isNeededToUpdateBounds = false;
 
 			// TODO: remove `this.hasDataSizeChanged` and fix ui/Scrollable*
 			this.hasDataSizeChanged = (prevProps.dataSize !== this.props.dataSize);
@@ -1179,8 +1180,6 @@ const VirtualListBaseFactory = (type) => {
 			if (primary) {
 				this.positionItems();
 			}
-
-			this.isNeededToUpdateBounds = false;
 
 			return (
 				<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} ref={this.containerRef} style={style}>
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -321,7 +321,7 @@ const VirtualListBaseFactory = (type) => {
 			let deferScrollTo = false;
 			const {firstIndex, numOfItems} = this.state;
 
-			this.isNeededToUpdateBounds = false;
+			this.shouldUpdateBounds = false;
 
 			// TODO: remove `this.hasDataSizeChanged` and fix ui/Scrollable*
 			this.hasDataSizeChanged = (prevProps.dataSize !== this.props.dataSize);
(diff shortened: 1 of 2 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -1170,6 +1180,8 @@ const VirtualListBaseFactory = (type) => {
 				this.positionItems();
 			}
 
+			this.isNeededToUpdateBounds = false;
```
germboy: I'm slightly concerned by setting an internal state variable in a render method, where the variable isn't used. Could we instead make this change in `componentDidUpdate`? It seems more appropriate there, since that's where the value is used (or changed when appropriate). Initially, I preferred if this value could be stored in `this.state` (where the value could be updated via `getDerivedStateFromProps`, and then we could prevent updates in the case when wheeling - but I'm not sure how possible this is. Do you have thoughts on that?

Also, I think I would prefer naming it `shouldUpdateBounds`, which is less verbose.

ybsung: > Could we instead make this change in componentDidUpdate

**Short answer**
Do you mean we make a local variable in componentDidUpdate()?
If yes, we can't because we need to decide whether calling setState() or not in didScroll() after executing the componentDidUpdate().

**Long answer**
The way the ENYO-6230 issue reproduces is...
- Start executing ui/VirtualList/VirtualListBase componentDidUpdate()
- Will update firstIndex state by calling setState() (Not update it directly yet)
- Stop executing ui/VirtualList/VirtualListBase componentDidUpdate()
- Start executing App componentDidUpdate()
- Call cbScroll() then ui/VirtualList/didScroll()
- DID NOT call setState() to update firstIndex because firstIndex is not updated yet and previous firstIndex and nextFirstIndex are same. setState() should be called because the firstIndex that will be changed and the nextFirstIndex are different. So that's the reason why I used `isNeededToUpdateBounds` flag.
- Stop executing App componentDidUpdate()

Like the call steps above, we need to check if we have to call setState() between several functions calling setState(). So we need a member variable to check it with `isNeededToUpdateBounds`.

> Initially, I preferred if this value could be stored in this.state (where the value could be updated via getDerivedStateFromProps, and then we could prevent updates in the case when wheeling - but I'm not sure how possible this is. Do you have thoughts on that?

**Short answer**
It seems to me that it is almost impossible because we need to move all the code updating state to `getDerivedStateFromProps` and to remove all `this` reference in the `getDerivedStateFromProps` like https://github.com/enactjs/enact/pull/2094.

**Long answer**
If adding `getDerivedStateFromProps`, then the call stacks are..
- getDerivedStateFromProps()
- componentDidUpdate()
- didScroll()

If calling setState() in `componentDidUpdate()`, we allow calling setState() in didScroll(). To do it, I used the `isNeededToUpdateBounds`. If we want to handle those code in getDerivedStateFromProps(), we need to move all the code calling setState() to `getDerivedStateFromProps`. That means we don't use `this` in it. So we need to move all member variables to state like https://github.com/enactjs/enact/pull/2094 (This PR was not merged because of too much modification).

ybsung: > Could we instead make this change in componentDidUpdate

Oh do you mean moving `this.isNeededToUpdateBounds = false;` to `componentDidUpdate()`? Yes. It looks better. When verifying the issues after applying it, there was no issue. So I fixed it. Thanks.

germboy: Yes, that's what I meant - thanks!
I agree that it may be too impractical (or not possible) to store the value of `this.isNeededToUpdateBounds` in `this.state` and explicitly preventing re-renders based on it, due to the nature of how this component can be updated. So, I think this looks good. I'd still suggest renaming `this.isNeededToUpdateBounds`, but will defer to anyone else who has an opinion on that. Nice job, thanks!

PR #2585, opened 2019-09-27
PLAT-85817: Remove maxDuration from Scroller flick config

germboy:
### Checklist

`ui/Scroller` does not properly scroll with inertia after flick events are fired.

### Resolution
By default, `Touchable` configures a `maxDuration` time value where after `x` amount of time after a drag has begun, then the drag stops being tracked. At that point, flick events do not provide flick target data. In the case of `Scroller`, this means we're unable to scroll after flick events have occurred when a drag has continued for over `250ms` (the default config value). We need to continually track a drag, so I've configured a `null` value for `Scroller`'s `maxDuration` config and added some validation in `Touchable` to not cancel tracking unless a value is provided.

### DIFF
--- a/packages/ui/Touchable/Flick.js
+++ b/packages/ui/Touchable/Flick.js
@@ -11,11 +11,13 @@ class Flick {
 		return this.tracking;
 	}
 
-	begin = (config, {onFlick}, coords) => {
-		this.minVelocity = config.minVelocity;
-		this.maxMoves = config.maxMoves;
+	begin = ({maxDuration, maxMoves, minVelocity}, {onFlick}, coords) => {
+		this.minVelocity = minVelocity;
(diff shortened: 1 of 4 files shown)


packages/ui/Touchable/Flick.js
```diff
@@ -11,11 +11,13 @@ class Flick {
 		return this.tracking;
 	}
 
-	begin = (config, {onFlick}, coords) => {
-		this.minVelocity = config.minVelocity;
-		this.maxMoves = config.maxMoves;
+	begin = ({maxDuration, maxMoves, minVelocity}, {onFlick}, coords) => {
```
webOS101: I'd really love to document all these config values someday!

PR #2586, opened 2019-09-30
ENYO-6286: Add new marquee overflow style

ryanjduffy:
### Checklist

Adds support for the new "fade" style of marquee

### Additional Considerations

* Doesn't yet implement the "fade" visual which should show only on sides with more content
  * hide on "before" side when not animating
  * hide on "after" side when animating completes but hasn't reset

### DIFF
--- a/packages/ui/Marquee/Marquee.module.less
+++ b/packages/ui/Marquee/Marquee.module.less
@@ -17,40 +17,18 @@
 
 	&.fade {
 		// fade left side only
-		-webkit-mask-image: -webkit-gradient(
-			linear,
-			left top,
-			right top,
-			from(rgba(0,0,0,1)),
-			color-stop(0.9, rgba(0,0,0,1)),
-			to(rgba(0,0,0,0))
-		);
+		mask-image: linear-gradient(to right, transparent, black 48px);
 
 		&.rtl {
 			// fade right side only
-			-webkit-mask-image: -webkit-gradient(
-				linear,
-				left top,
-				right top,
-				from(rgba(0,0,0,0)),
-				color-stop(0.1, rgba(0,0,0,1)),
-				to(rgba(0,0,0,1))
-			);
+			mask-image: linear-gradient(to left, transparent, black 48px);
 		}
 
 		// using willAnimate allows the fade to remain during the reset pause since the GPU layer isn't
 		// removed until marquee is cancelled
 		&.willAnimate {
 			// fade both sides
-			-webkit-mask-image: -webkit-gradient(
-				linear,
-				left top,
-				right top,
-				from(rgba(0,0,0,0)),
-				color-stop(0.1, rgba(0,0,0,1)),
-				color-stop(0.9, rgba(0,0,0,1)),
-				to(rgba(0,0,0,0))
-			);
+			mask-image: linear-gradient(to right, transparent 0, black 48px, black calc(100% - 48px), transparent 100%);
 		}
 	}
 
--- a/packages/ui/Marquee/Marquee.module.less
+++ b/packages/ui/Marquee/Marquee.module.less
@@ -14,8 +14,32 @@
 	white-space: pre !important;
 	overflow: hidden;
 	text-align: left;
+	-webkit-mask-image: -webkit-gradient(
+		linear,
+		left top,
+		right top,
+		from(rgba(0,0,0,1)),
+		color-stop(0.9, rgba(0,0,0,1)),
+		to(rgba(0,0,0,0))
+	);
+
+	// using willAnimate allows the fade to remain during the reset pause since the GPU layer isn't
+	// removed until marquee is cancelled
+	&.willAnimate {
+		-webkit-mask-image: -webkit-gradient(
+			linear,
+			left top,
+			right top,
+			from(rgba(0,0,0,0)),
+			color-stop(0.1, rgba(0,0,0,1)),
+			color-stop(0.9, rgba(0,0,0,1)),
+			to(rgba(0,0,0,0))
+		);
+	}
 
 	.text {
+		--ui-marquee-padding: 0;
+
 		pointer-events: none;
 		text-overflow: clip;
 		overflow: hidden;
@@ -31,6 +55,11 @@
 
 		&.animate {
 			overflow: visible;
+		}
+
+		.padding {
+			display: inline-block;
+			width: calc(1px * var(--ui-marquee-padding, 0));
 		}
 	}
 
--- a/packages/ui/Marquee/Marquee.module.less
+++ b/packages/ui/Marquee/Marquee.module.less
@@ -14,27 +14,44 @@
 	white-space: pre !important;
 	overflow: hidden;
 	text-align: left;
-	-webkit-mask-image: -webkit-gradient(
-		linear,
-		left top,
-		right top,
-		from(rgba(0,0,0,1)),
-		color-stop(0.9, rgba(0,0,0,1)),
-		to(rgba(0,0,0,0))
-	);
 
-	// using willAnimate allows the fade to remain during the reset pause since the GPU layer isn't
-	// removed until marquee is cancelled
-	&.willAnimate {
+	&.fade {
+		// fade left side only
 		-webkit-mask-image: -webkit-gradient(
 			linear,
 			left top,
 			right top,
-			from(rgba(0,0,0,0)),
-			color-stop(0.1, rgba(0,0,0,1)),
+			from(rgba(0,0,0,1)),
 			color-stop(0.9, rgba(0,0,0,1)),
 			to(rgba(0,0,0,0))
 		);
+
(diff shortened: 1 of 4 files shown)


ryanjduffy:
Just noticed when looking at the reference video that my version scrolls too far and should instead only scroll to when the copy of the text hits the start which would preclude us from removing the "after" fade.

ryanjduffy:
Fixed the distance calculation to match the reference video.

ryanjduffy:
Noting that in this example, the first marquee has `marqueeOverflow` set to "ellipsis" which is why it shows an ellipsis and the other lines do not.

ENYO6248-v1

packages/ui/Marquee/MarqueeBase.js
```diff
@@ -99,11 +99,20 @@ const MarqueeBase = kind({
...
+		 * @type {('clip','ellipsis')}
 		 * @public
 		 */
 		overflow: PropTypes.oneOf(['clip', 'ellipsis']),
 
+		/**
+		 * Amount of padding, in pixels, between the instances of the content
+		 *
+		 * @type {Number}
+		 * @eefault 0
```
webOS101: Keyboard strikes again.

packages/ui/Marquee/MarqueeBase.js
```diff
@@ -203,6 +214,12 @@ const MarqueeBase = kind({
 					onTransitionEnd={onMarqueeComplete}
 				>
 					{children}
+					{duplicate ? (
+						<React.Fragment>
+							<div style={{display: 'inline-block', width: padding}} />
```
webOS101: Do we need units on this if it'll be non-zero?

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -478,42 +498,46 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+			if (typeof marqueePadding === 'string') {
+				if (/^\d+(.\d+)?%$/.test(marqueePadding)) {
+					return width * Number.parseFloat(marqueePadding) / 100;
+				}
+
+				// warning for invalid string value;
+				return 0;
 			}
+
+			return scale(marqueePadding);
```
webOS101: Should the docs indicate the padding will be scaled?

packages/ui/Marquee/MarqueeBase.js
ryanjduffy: Assumed to be pixels by React when its numeric.

packages/ui/Marquee/MarqueeBase.js
ryanjduffy: Everyone is using `@eefault` these days I hear

Djspaceg:
How do you feel about abstracting the marqueePadding out to CSS. Evaluate whether the marquee needs to be applied on the subject, then add the class/css-prop/etc to apply a padding to the marquee child node in CSS, then animate the marquee.

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -478,42 +498,46 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
 
-				const overflow = this.calculateTextOverflow(this.distance);
-				this.setState(state => {
-					return state.overflow === overflow ? null : {overflow};
-				});
+		getPadding (width) {
+			const {marqueePadding} = this.props;
+
+			if (typeof marqueePadding === 'string') {
+				if (/^\d+(.\d+)?%$/.test(marqueePadding)) {
```
Djspaceg: Is this `/^\d+(.\d+)?%$/` supposed to be `/^\d+(\.\d+)?%$/` instead?
Start of string; 1 or more decimal characters; start capture; any single character; 1 or more decimal characters; end capture; flag the capture as optional; literal `%` character; end of string.

ryanjduffy: Indeed!

packages/ui/Marquee/MarqueeDecorator.js
ryanjduffy: Updated

ryanjduffy:
> Can you add a change to the Marquee sampler to be able to set either a string or number for marqueePadding to make it easy to test?

Added knobs for both `marqueePadding` and `marqueeOverflow`.

packages/ui/Marquee/MarqueeBase.js
```diff
@@ -165,20 +166,22 @@ const MarqueeBase = kind({
...
 			// If the components content directionality doesn't match the context, we need to set it
 			// inline
 			const direction = rtl ? 'rtl' : 'ltr';
 			const sideProperty = rtl ? 'left' : 'right';
 			const style = {
 				direction,
 				textAlign: alignment,
-				textOverflow: overflow
+				textOverflow: overflow,
+				'--ui-marquee-padding': padding
```
webOS101: Would this prevent developers from having different padding values for different marquees?

ryanjduffy:
I've updated this to use `marqueeOverflow` as a means to select the marquee style. The default, "fade", is the new UX (duplicated content with padding). The other option, "ellipsis", is the previous marquee (text-overflow: ellipsis with only one copy of the content).

Definitely open for discussion on this but I think there might be some value in this approach both because marquee is in UI which calls for more flexibility and because we might wish we had a fallback if there are scenarios in which the duplicate content causes problems.

webOS101:
Should the left fade disappear after the marquee stops?

ryanjduffy:
@webOS101 - can be either way. It seemed odd for it to disappear and then reappear when the marquee restarts but I think this way is a bit odd too.

Also noticed that the fade isn't applying initially since we've deferred measuring. Need to look into that more.

webOS101:
Yep, I didn't mention that because measuring at render will introduce a performance regression.  We can use a delayed calculation, but... there are often a _lot_ of marquee instances on the page.  Is there a pure CSS approach that could apply the class?

packages/ui/Marquee/Marquee.module.less
```diff
@@ -15,9 +15,50 @@
 	overflow: hidden;
 	text-align: left;
 
+	&.fade {
+		// fade left side only
+		-webkit-mask-image: -webkit-gradient(
```
Djspaceg: Do you mind if we simplify these?
```
	&.fade {
		&.onLeft {
			-webkit-mask-image: linear-gradient(to right, transparent, black 48px);
		}
		&.onRight {
			-webkit-mask-image: linear-gradient(to left, transparent, black 48px);
		}
		&.onLeft.onRight {
			-webkit-mask-image: linear-gradient(to right, transparent 0, black 48px, black calc(100% - 48px), transparent 100%);
		}
	}
```
Also, the current implementation uses percentages for the fade, when they likely should be fixed length.

Not tieing this to RTL as a concept is probably safer as well.

Djspaceg:
> Yep, I didn't mention that because measuring at render will introduce a performance regression. We can use a delayed calculation, but... there are often a _lot_ of marquee instances on the page. Is there a pure CSS approach that could apply the class?
> https://stackoverflow.com/questions/9333379/check-if-an-elements-content-is-overflowing

This effect appears to work because of the following combination of things (which I don't think is available to us based on our requirements)
1. Scroller (marquee would need to be refactored into a scroller and animated)
2. Opaque background (we need to support a translucent component with unpredictable background, so we must use a mask rather than a simpler gradient overlay
3. `background-attachment: local;` (cool! Not an inhibitor to our implementation, but not relevant. It just seemed worth calling this out as a neat little-known thing. `local` affixes the background to the scroller area, so as it scrolls the background does too. In the linked implementation they just stick a gradient on the background of the scroller, fix it in place and boom, automatic scroller.) https://developer.mozilla.org/en-US/docs/Web/CSS/background-attachment

packages/ui/Marquee/Marquee.module.less
ryanjduffy: I've updated the gradient syntax. Thanks for the suggestion.

I want to think a bit more on the RTL question. Generally, I agree with your sentiment. In this case, `MarqueeBase` accepts an `rtl` prop and tying the fade to the text directionality makes sense to me given the context.

We'd also need a way to control the size of the fade from a theme. Couple options that come to mind:

* Don't implement the fade in UI at all and leave it to the theme
* Implement the fade in UI but require that themes override it if they want a different configuration
* Use a CSS variable which can be overridden by a theme
* Use a prop on `MarqueeBase` which would likely feed into a CSS variable.

Djspaceg: Options 2 and 3 sound the best to me. There's also potentially a fifth option, which would be a customizable LESS variable, but with our packaging scheme, that's not settable by most users.
A sixth option could be to export a mixin from ui which generates these gradients and a theme can just use it to help options 1 and 2.

PR #2590, opened 2019-10-01
PLAT-86483: Fix Button color bar height

ryanjduffy:
### Checklist

Height of the color bar should be 9px for both small and large buttons.

### Resolution
* Added `client` class to `publicClassNames` in order to override it in `IconButton` which has different requirements for the size of the color bar.
* Added/updated LESS variables for the sizes

### Additional Considerations
I included `client` in our "not for public consumption" note since we've avoided exposing it in the past. I feel like we've expanded the list enough that we're nearing the point where exposing everything might be useful since the component is reused so broadly. Alternatively, we might make an `internal/Button` that is more open and inline it in `moonstone/Button` with a more limited API.

### Links

### DIFF
--- a/packages/moonstone/Button/Button.js
+++ b/packages/moonstone/Button/Button.js
@@ -82,9 +82,9 @@ const ButtonBase = kind({
 		 * @type {Object}
 		 * @public
 		 */
-		// `transparent` was intentionally excluded from the adove documented exported classes as it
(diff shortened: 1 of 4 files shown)


Djspaceg:
I'm curious if by opening `.client` if there are other awkward workarounds we've implemented in the past to get around the fact that `.client` wasn't available that we could now simplify with this exposure.

PR #2593, opened 2019-10-02
ENYO-6286: Add new marquee fade style

ryanjduffy:
Experiment using IntersectionObserver to add the fade class without a re-render.

**Note:** Required a fair amount of CSS reworking to calc correctly which is currently affecting layout. Just an experiment to aid perf investigations.

marquee


ryanjduffy:
Note: Need to remove fade after first block of text leaves so it doesn't fade the second block.

PR #1502, opened 2018-03-15
PLAT-50128: Add touch-dragging support to ui/Scrollable

0x64:
### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
Implemented of touch-dragging support in `Scrollable` components.

### Resolution
[//]: # (Does the code work as
[//]: # (What is the impact of this change and *why* was it made?)
Made `Scrollable` components to handle dragging and flicking through `Touchable`.
- Updated `Touchable` to provide coordinates via `onDragStart`
- Updated `Scrollable` to perform dragging/flicking by `onDragStart`, `onDrag`, `onDragEnd`, and `onFlick` through `Touchable`

### Links
[//]: # (Related issues, references)
PLAT-50128

### Comments
Enact-DCO-1.0-Signed-off-by: Seungcheon Baek (sc.baek@lge.com)

PR #2598, opened 2019-10-07
ENYO-6299: Call stop() when clicking not on but out of paging controls

ybsung:
### Checklist

If clicking on a paging control several times, the Scroller including the paging control did not scroll smoothly.

### Resolution

Historically to stop scrolling when pressing a mouse down on items while flicking, stop() in Scrollable was called in https://github.com/enactjs/enact/pull/1502.
But the DOM element having that mouse down event handler was changed from the container having items to the container having not only items but also paging controls in https://github.com/enactjs/enact/pull/2468.
Then whenever clicking on the paging controls, the stop() was called and the Scroller did not scroll smoothly.

So if clicking on paging controls, the stop() is now not called.

### Links

ENYO-6299

### Comments

I tried to make any unit test for this issue and spent several time. But it failed. I hope that we could use TC instead of unit test case for this issue.

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -604,8 +604,11 @@ class ScrollableBase extends Component {
 	getRtlX = (x) => (this.props.rtl ? -x : x)
 
 	onMouseDown = (ev) => {
-		this.stop();
-		forward('onMouseDown', ev, this.props);
+		if (this.props.onMouseDown) {
+			forward('onMouseDown', ev, this.props);
+		} else {
+			this.stop();
+		}
 	}
 
 	onDragStart = (ev) => {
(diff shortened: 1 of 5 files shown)


packages/ui/Scrollable/Scrollable.js
```diff
@@ -604,8 +604,11 @@ class ScrollableBase extends Component {
 	getRtlX = (x) => (this.props.rtl ? -x : x)
 
 	onMouseDown = (ev) => {
-		this.stop();
-		forward('onMouseDown', ev, this.props);
+		if (this.props.onMouseDown) {
```
webOS101: `forward` doesn't need to be gated like this so we could simplify this a bit.  Perhaps a better approach is to use `forwardWithPrevent()` and call `preventDefault` inside `moonstone/Scrollable`?

ybsung:
Hello @webOS101 
With `forwardWithPrevent`, the code became much elegant. I'm happy to know how to use `forwardWithPrevent` brightly.

webOS101:
:Awesome:

PR #2600, opened 2019-10-07
PLAT-84676: Update Moonstone Icons

teckliew:
### Checklist

Add new icons

newicons1007


teckliew:
@ryanjduffy I can bring back `liverecord`. As for `searchfilled`, my assumption is that was an unfinished icon that was meant to be `zoomin` and `zoomout` and we could just have `zooomin`/`zoomout` replace `searchfilled`. WDYT?

webOS101:
we could just alias 'liverecord' to 'liverecordone', too.

teckliew:
Actually, on closer inspection, it is highly likely that `liverecord` is the updated version of `liveplayback`. Because these were unnamed and there were no indication of which ones were duplicates or updated versions of old icons, we thought these two were different icons.

But looking at it closer now, I think it's likely Crane replaced  the icon with the arrow to the one without:
icon with the arrow
old
icon without the arrow
new

ryanjduffy:
I don't think we can remove anything that might have been used by apps but we can duplicate code points under multiple names.

If we wanted to "hide" values that shouldn't be used (e.g. search filled), we could probably make them non-enumerable on the icon list.

```
const icons = {
  plus: ...
};

Object.defineProperty(icons, 'searchfilled', {
  value: theCodePoint,
  enumerable: false
});
```

Djspaceg:
> @ryanjduffy I can bring back `liverecord`. As for `searchfilled`, my assumption is that it was an unfinished icon that was meant to be `zoomin` and `zoomout` and we could just have `zooomin`/`zoomout` replace `searchfilled`. WDYT?
> searchfilled
> zoominzoomout

I disagree. I believe these represent different visual concepts, and that each of these should be included, not replacing or removing anything. A hollow-search, a filled-search, a zoom-in, and a zoom-out all need to coexist.

Regarding `liveplayback` and `liverecord` these too are separate concepts, and need different icons. One indicates starting to record live tv, the other resuming playback of said recorded live TV.

> I don't think we can remove anything that might have been used by apps

Correct, we cannot delete any icons, we can only add and update (when it's clear and obvious, or we've been explicitly instructed that a new icon replaces an old icon).

teckliew:
Sounds good. I added back in the two icons that were missing from the latest `Moonstone Icons`.

ryanjduffy:
These are a bit odd. Different styling and different naming. I read that the two different styles (left pair vs right pair) were intentional but the left pair do not match each other.

image

Djspaceg:
> These are a bit odd. Different styling and different naming. I read that the two different styles (left pair vs right pair) were intentional but the left pair do not match each other.
> 
> image

Looking at those, my assumption would be that the right pair is actually a replacement for the left two. They seem to be reflecting a new stylistic approach to that same concept, a filled-arrow pointing at an equivalently sized vertical line. But, as much as I feel that way, I can't say for 100% dead certainty that they are not supposed to exist together; I'm only reasonably certain about that.

Additionally, as you said, the left two definitely don't match each other.

ryanjduffy:
Confirmed the icons are as intended by the design team.

PR #2601, opened 2019-10-08
ENYO-6286: Update Marquee to new repeated content style

ryanjduffy:
### Checklist

* Marquee style has changed from single copy of text to duplicated text with configurable padding.
* Text continues to be truncated with an ellipsis when it exceeds the width of its container.

### Resolution
* Add `marqueeSpacing` to configure padding either relative to the container (e.g. `'50%'`) or absolutely (e.g. `200`) which is scaled for the current resolution
* Add `spacing` class to CSS overrides to customize the styling of the padding node
* Update `MarqueeDecorator` to calculate the new distance and spacing
* Update `MarqueeBase` to duplicate the content just before animating when distance requires it.


Djspaceg:
Can we change "padding" to "spacing" to avoid using the name of a CSS property?

germboy:
This looks good to me. I didn't notice any real behavior or performance changes that would concern me too much. When the marquee completes (and the overflowing text changes to ellipsis), it appears that the text itself shifts to the right by a few pixels. I don't know if this is something hat can be prevented or would concern anyone. Could you confirm this behavior? Otherwise, I'm good to approve this PR.

ryanjduffy:
I think there are two issues at play here.

* There's a rounding error when using `scrollWidth` because it only reports whole pixel values. There isn't a whole lot we can do about this as far as I can tell but generally isn't very disruptive.
* `moonstone/Panels.Header` includes a `text-indent` on the `.title` which messes with the alignment of the duplicated content since it isn't indented.

I've pushed a fix for the latter which required touching other bits of Marquee and Header to get everything landing where it should.

cc: @Djspaceg

ryanjduffy:
Realized after I walked away that I don't want to forward `css` through MarqueeDecorator but the effect will be the same for the purposes of the result in Header.

ryanjduffy:
Pushed a fix for the improper Marquee css override approach

Djspaceg:
Should the `.spacing` class be renamed since it's a really common term and is now being exported for override? `.marqueeSpacing` at the simplest.

ryanjduffy:
@Djspaceg I think the concern over naming is valid -- we've had cases where we've inadvertently overridden classes. Do you think this class is more common than others and therefore should be changed?

Djspaceg:
> @Djspaceg I think the concern over naming is valid -- we've had cases where we've inadvertently overridden classes. Do you think this class is more common than others and therefore should be changed?

Well, `.text`, `.animate`, and `.willAnimate` seem likely to also exist organically, but those are pre-existing names, so we can't do much about those; but we _could_ do something about the new `.spacing`. **Alternatively** having to be concerned about what external things are doing is pretty lame and makes modular CSS less awesomely isolated; so maybe NBD. I'm sort of torn now.

PR #2603, opened 2019-10-09
ENYO-6309: Properly attach click handler to FloatingLayer

germboy:
### Checklist

`ui/FloatingLayer` does not attach a click handler when it is mounted while opened. This results in not being able to dismiss the component via clicks.

### Resolution
Attach the click handler during `componentDidMount` when appropriate.


ryanjduffy:
Reworded the changelog to be more author-centric than implementation-centric

PR #2610, opened 2019-10-16
PLAT-87182: Update Enact to use updated ResBundle basepath format

JayCanuck:
### Checklist

* iLib 14.4.0 introduces new handling to ResBundle pathing in its cache. Specifically, added support for `basePath` ResBundle parameter as well as passing a root path argument in the Loader functions.

### Resolution
* Adds support for new iLib format, while still providing backward support for older copies of iLib temporarily.

### DIFF
--- a/packages/moonstone/internal/$L/$L.js
+++ b/packages/moonstone/internal/$L/$L.js
@@ -32,8 +32,10 @@ function createResBundle (options) {
 	if (typeof ILIB_MOONSTONE_PATH !== 'undefined') {
 		opts = {
 			loadParams: {
+				// Deprecated; to be removed in future
 				root: ILIB_MOONSTONE_PATH
 			},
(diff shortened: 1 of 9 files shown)


packages/moonstone/internal/$L/$L.js
```diff
@@ -32,8 +32,10 @@ function createResBundle (options) {
 	if (typeof ILIB_MOONSTONE_PATH !== 'undefined') {
 		opts = {
 			loadParams: {
+				// Deprecated; to be removed in future
 				root: ILIB_MOONSTONE_PATH
```
ryanjduffy: Why are we retaining this if it's deprecated?

JayCanuck: Backward compatibility with iLib 14.0.0 through 14.3.0.

PR #2612, opened 2019-10-17
ENYO-6318: Fix mediaTitle to adjust horizontal alignment

xbinary:

### Checklist

There is horizontal alignment issue between title and playing time.
The title has `bottom: -5px`.
I think that It is intended to be horizontally aligned with the baseline of the font.
But there is a problem in non latin languages.

### Resolution
title has bottom: 5px in non latin languages

### Additional Considerations

### Links
[ENYO-6318]

### DIFF
--- a/packages/moonstone/VideoPlayer/VideoPlayer.module.less
+++ b/packages/moonstone/VideoPlayer/VideoPlayer.module.less
@@ -104,8 +104,9 @@
 						}
 
 						.moon-locale-non-latin({
-							bottom: 5px;
-							line-height: @moon-non-latin-header-line-height;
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Thanks for the find and the good analysis, @xbinary. I agree that the `bottom` rule is used to align the title text with the time. I believe this is necessary because the font sizes of each is different in latin locales so some adjustment is required to retain appropriate line heights but align the baselines.

In non-latin locales, the font sizes are the same and the current code has a smaller the line-height than the new default added by #2016. That, coupled with the `bottom` rule are the cause of pushing the text below the baseline.

PR #2614, opened 2019-10-17
ENYO-6316: Fix Marquee to start on focus when disabled

ryanjduffy:
### Checklist

Focusing a disabled marquee element does not start marquee

### Resolution
Remove the `!disabled` check in `MarqueeDecorator`

### Additional Considerations
This matches `MarqueeController` which already allowed marqueeing when it was disabled. This regression is a hold over from the change to allow focusing disabled elements.

### DIFF
--- a/packages/ui/Marquee/MarqueeDecorator.js
+++ b/packages/ui/Marquee/MarqueeDecorator.js
@@ -178,18 +178,6 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			children: PropTypes.node,
 
 			/**
-			 * Passed through to the wrapped component.
-			 *
-			 * Does not affect Marquee behavior except that components that are `marqueeOn="focus"`
-			 * will be treated as if they were `marqueeOn="hover"`, to allow disabled (and thus,
-			 * unfocusable) components to marquee.
-			 *
-			 * @type {Boolean}
-			 * @public
-			 */
-			disabled: PropTypes.bool,
-
-			/**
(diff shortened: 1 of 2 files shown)


teckliew:
Not sure if we want to address this but when adding more than 1 line to `title` overflows out side of the dropdown button:
twolinetitle

ryanjduffy:
Pulled your commit in, @teck, and added another to clean up some other `disabled` references that were no longer relevant.

Of note, I removed `disabled` from `MarqueeDecorator` altogether because it is no longer used. This will cause it to appear as an API diff in our tooling but isn't since the prop is still forwarded on as noted in the prior docs. If anyone objects to this, we can add it back.

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -377,7 +365,7 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				prevProps.forceDirection !== forceDirection
 			) {
 				this.cancelAnimation();
-			} else if (disabled && this.isHovered && marqueeOn === 'focus' && this.sync) {
+			} else if (this.isHovered && marqueeOn === 'focus' && this.sync) {
 				this.context.enter(this);
```
teckliew: We might have to keep `disabled` for this `MarqueeController` sync logic, or there might be some logic needed for `handleLeave` in `MarqueeController` .

With `disabled` remove from here, `enter` is registered into the controller, but `leave` is never called when blurred.

PR #2629, opened 2019-10-30
ENYO-6325: Fix Marquee to show ellipsis when content changes

ryanjduffy:
### Checklist

When changing the content of a Marquee when it had been promoted, the text would not be truncated with an ellipsis.

### Resolution
The logic was close before in that we called `cancelAnimation()` when the content changed which would start the demote job. However, in cases where we immediately restart marquee on update (render, hover when hovered, focus when focused) the demote job is stopped by the promote job so `promoted` was never reset.

The proposed fix resets `promoted` to `false` when invalidating metrics to ensure it's in the correct state. It will then be re-promoted later just before the animation starts as designed.

### Additional Considerations
The underlying issue existed before but manifested now due to the change in Marquee to maintain the clipped text when restarting marquee. Previously, the layer would have remained promoted (and therefore add the `willAnimate` class) but the text would still have been truncated with an ellipsis due to the prior CSS.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/Marquee` to display ellipsis when its content changes and overflows its bounds
+- `ui/Marquee` to display an ellipsis when its content changes and overflows its bounds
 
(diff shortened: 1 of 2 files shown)


ryanjduffy:
I based the branch on `3.2.2` but targeted to PR against `develop`. Assume we'll be creating a release branch off of `3.2.2` and we'll retarget this against that branch then.

PR #2630, opened 2019-10-31
PLAT-86868: Fixed Routable.Linkable to properly support event forwarding

Djspaceg:
### Checklist

`Linkable` doesn't support adding your own event handler, as it's clobbered by Linkable's own event handler.

### Resolution
Consumer-provided events are now _also_ fired, as expected, like in the rest of the framework.

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -7,7 +7,6 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Marquee` to display an ellipsis when its content changes and overflows its bounds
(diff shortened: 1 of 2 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Marquee` to display an ellipsis when its content changes and overflows its bounds
+- `ui/Routable.Linkable` to properly support event forwarding
```
webOS101: This is currently a private API so we should not document this change.

PR #2636, opened 2019-11-05
ENYO-6347: Sinhala text overlaps in Notification popup

Sangwook-Lee:
### Checklist

I found this problem while testing translated text.
Notification popup use line-height of BodyText, not Popup's.
Thus, the line-height is calculated as `px` (rem). I think it should be changed in 'em' units.
When I changed the locale to a non-Latin language, the line-height did not change in 'em' units.

### Resolution
I fixed CSS `line-height` to use `em` unit for non-Latin locales.

### Links
ENYO-6347

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact moonstone module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `moonstone/Notification` CSS `line-height` for non-latin locales
+
 ## [3.2.3] - 2019-11-01
 
 ### Changed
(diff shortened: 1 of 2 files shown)


Sangwook-Lee:
Is it a good way to fix it?

Djspaceg:
We need to be careful when switching a line-height from `px` to `em`. We always prefer `em` for measuring the height of text, which means that if it's set to a `px` value, it's very likely that was done for a specific reason. We should look into why it was set to `px` in the first place and evaluate whether that logic is still important to us (if a reason can be determined). As an example, Item's line-height was/is set to a `px` value so the rendered result is always the correct pixel size for lists and layouts. Changing this to an `em` value could have detrimental effects in several scenarios.

PR #2637, opened 2019-11-05
ENYO-6348: Fix Notification to show all buttons on one line

Sangwook-Lee:
### Checklist

If Notification currently has more than three buttons, it adds wide CSS.
But currently it has two long buttons but the buttons overflow.
We need to improve the root cause and consider whether or not to add wide CSS when it has a three buttons.

### Resolution
I fixed `max-width` of Notification by `max-width` of Button and changed to add `wide` css if length of the buttons is greater than two.

### Links
ENYO-6348

### DIFF
--- a/packages/moonstone/Notification/Notification.js
+++ b/packages/moonstone/Notification/Notification.js
@@ -135,8 +135,9 @@ const NotificationBase = kind({
 	},
 
 	computed: {
-		className: ({buttons, styler}) =>
-			styler.append({wide: (buttons && buttons.filter(Boolean).length > 2)}),
+		className: ({buttons, styler}) => styler.append({
+			wide: (buttons && React.Children.toArray(buttons).filter(Boolean).length > 2)
+		}),
 		buttons: ({buttons}) => React.Children.map(buttons, (button) => {
 			if (button && button.props && !button.props.small) {
 				return React.cloneElement(button, {size: 'small'});
--- a/packages/moonstone/Notification/Notification.js
+++ b/packages/moonstone/Notification/Notification.js
@@ -135,13 +135,8 @@ const NotificationBase = kind({
 	},
 
 	computed: {
-		className: ({className, buttons, styler}) => {
-			if (buttons && buttons.length > 2) {
-				return styler.append({wide: true});
-			} else {
-				return className;
-			}
-		},
+		className: ({buttons, styler}) =>
+			styler.append({wide: (buttons && buttons.filter(Boolean).length > 2)}),
 		buttons: ({buttons}) => React.Children.map(buttons, (button) => {
 			if (button && button.props && !button.props.small) {
 				return React.cloneElement(button, {size: 'small'});
--- a/packages/moonstone/Notification/Notification.js
+++ b/packages/moonstone/Notification/Notification.js
@@ -136,7 +136,7 @@ const NotificationBase = kind({
 
 	computed: {
 		className: ({className, buttons, styler}) => {
-			if (buttons && buttons.length > 3) {
+			if (buttons && buttons.length > 2) {
 				return styler.append({wide: true});
 			} else {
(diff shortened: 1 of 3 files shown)


packages/moonstone/styles/variables.less
```diff
@@ -470,7 +470,7 @@
 @moon-notification-padding-left-right: 39px;
 @moon-notification-padding-top-bottom: 24px;
 @moon-notification-min-width: 300px;
-@moon-notification-max-width: 978px;
+@moon-notification-max-width: 1017px;
```
webOS101: Can we calculate this from the max button width?

Djspaceg: There are other values involved. Ex: The previous value was not `300px * 3`, it's `978px`. A math based solution would be nice, but we need to understand the other sizing influencers.

packages/moonstone/styles/variables.less
```diff
@@ -470,7 +470,7 @@
 @moon-notification-padding-left-right: 39px;
 @moon-notification-padding-top-bottom: 24px;
 @moon-notification-min-width: 300px;
-@moon-notification-max-width: 978px;
+@moon-notification-max-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (2 * ((@moon-button-small-h-margin * 2) + @moon-button-max-width)));  // Attempt to calculate the width of all of the buttons and the pieces of Notification that influence the width to get the maximum size
```
ryanjduffy: The math checks out but I'd suggest we use either the "small" variables or the "large" variables for button. My inclination would be the "large" values to ensure that two large buttons would fit (even though small buttons are the default).

packages/moonstone/Notification/Notification.js
```diff
@@ -135,13 +135,8 @@ const NotificationBase = kind({
...
 	computed: {
-		className: ({className, buttons, styler}) => {
-			if (buttons && buttons.length > 3) {
-				return styler.append({wide: true});
-			} else {
-				return className;
-			}
-		},
+		className: ({buttons, styler}) =>
+			styler.append({wide: (buttons && buttons.filter(Boolean).length > 2)}),
```
ryanjduffy: This is a good structural improvement. I did notice that our original logic was wrong in that it didn't correctly check for a single button. In that case, `length` would be undefined since it doesn't exist on a React Element object.

In the updated code, I believe this will throw an error since we'd be invoking an undefined method. Using `React.Children.toArray(buttons)` will insulate us from managing that condition since it returns an array for a single child.

Stylistically, I'd suggest moving `styler` to the previous line and line break the object which is our typical style and is a bit more friendly to diffs.

packages/moonstone/Notification/Notification.js
```diff
@@ -135,13 +135,9 @@ const NotificationBase = kind({
...
-			if (buttons && buttons.length > 3) {
-				return styler.append({wide: true});
-			} else {
-				return className;
-			}
-		},
+		className: ({buttons, styler}) => styler.append({
+			wide: (buttons && React.Children.toArray(buttons).filter(Boolean).length > 2)
+		}),
 		buttons: ({buttons}) => React.Children.map(buttons, (button) => {
```
Djspaceg: @ryanjduffy think this should also include `toArray`, or is `map` acceptable directly on this one?

ryanjduffy: This one should be map since it needs to alter the children.

packages/moonstone/styles/variables.less
webOS101: Does this account for large text mode, too?

ryanjduffy: IIRC, large text doesn't affect the max width of buttons so we don't need to account for that here.

PR #2641, opened 2019-11-05
PLAT-90766: Add support for minimized Header

teckliew:
### Checklist

Add `minimized` prop to hide the title, only showing the header components, subtitles, and title below.
Using `Transition` to animate and then resize the title cell.

minimizedheader

### DIFF
--- a/packages/moonstone/Panels/Header.js
+++ b/packages/moonstone/Panels/Header.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types';
 import {isRtlText} from '@enact/i18n/util';
 import {Layout, Cell} from '@enact/ui/Layout';
 import Slottable from '@enact/ui/Slottable';
+import Transition from '@enact/ui/Transition';
 import ComponentOverride from '@enact/ui/ComponentOverride';
 
 import {MarqueeDecorator, MarqueeBase} from '../Marquee';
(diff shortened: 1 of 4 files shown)


packages/moonstone/Panels/Header.js
```diff
@@ -258,7 +270,9 @@ const HeaderBase = kind({
 			case 'dense':
 			case 'standard': return (
 				<Layout component="header" aria-label={title} {...rest} orientation="vertical">
-					{titleOrInput}
+					<Cell component={Transition} type="slide" visible={!minimized} size={minimized ? '0' : titleHeight}>
```
Djspaceg: Why "slide" and not "clip"? Won't that mean that the Header will only slide out of the way, not changing the DOM dimensions? You may need to add children to verify they move during/after the transition.

teckliew: Using "slide" for the animation and the dimension is set by setting size in Cell

packages/moonstone/Panels/Header.js
```diff
@@ -217,24 +229,24 @@ const HeaderBase = kind({
...
-	render: ({children, direction, marqueeOn, subTitleBelowComponent, title, titleOrInput, /* titleAbove, */titleBelowComponent, type, ...rest}) => {
+	render: ({children, minimized, direction, marqueeOn, subTitleBelowComponent, title, titleOrInput, /* titleAbove, */titleBelowComponent, type, ...rest}) => {
 		delete rest.centered;
 		delete rest.fullBleed;
 		delete rest.headerInput;
 		delete rest.hideLine;
 		delete rest.subTitleBelow;
 		delete rest.titleBelow;
 
+		const titleHeight = type === 'dense' ? '69px' : '89px';
```
Djspaceg: Need to update these to RI scaled values:
```
const titleHeight = ri.unit(ri.scale((type === 'dense') ? 69 : 89), 'rem');
```

teckliew:
The line doesn't move is the indication that the height didn't change for Header when its parent is relative or absolute positioned.
Looking at it, looks like Layout is set at 100% height, so unsetting it seems to work.

PR #2642, opened 2019-11-06
Alternate approach using "clip"

Djspaceg:
Setting the title height in the JS is not ideal.
The CSS should be informed when minimized is true.
"slide" Transition type doesn't adjust layout, which we need this to affect.

### Resolution
Switched to "clip" (same transition `Expandable` uses) to facilitate the layout change.
Moved all measurements into the CSS.
Added an inertial delay animation of the content to more closely match the model video.
Added a `.minimized` class to inform the CSS of our current state.

### DIFF
--- a/packages/moonstone/Panels/Header.module.less
+++ b/packages/moonstone/Panels/Header.module.less
@@ -65,7 +65,8 @@
 	// Standard and Dense Header
 	&.dense,
 	&.standard {
-		max-height: @moon-header-standard-height;
+		transition: height 600ms ease-in-out;  // Intentionally slightly longer than the transition for an inertial effect
+		will-change: height;
 
 		.title {
 			// The height is determined by the line-height below, but we want to maintain that as a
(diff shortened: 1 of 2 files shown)


packages/moonstone/Panels/Header.module.less
```diff
@@ -86,19 +87,24 @@
...
 	// Standard Header
 	&.standard {
+		height: @moon-header-standard-height;
+
 		.titlesCell {
 			margin-bottom: 9px;
 		}
+
+		&.minimized {
+			height: (@moon-header-standard-height - 89px);  // Must subtract the height of the header element. Considering changing the `em` line-height to a px value to make this more straightforward.
```
webOS101: Is `89px` something special?  Why is this not a multiple of 3?

Djspaceg: It's vestigial, from the previous commit. I'll update it to 90. 89 likely came from the browser's measurement of the node, since the node's size is based on the font size and `em`.

PR #2643, opened 2019-11-06
Uses simpler scaling and styling

Djspaceg:
Simplified the sizing even further (forgot that Layout automatically adjusts to rem calculation if you just supply a number, rather than a size string).
Removed unnecessary `css` prop
Removed `style` prop since it would clobber external values and could be assigned from within the module's LESS file.
Added `.minimized` class to inform CSS of our state.

### Additional Considerations
* Using this approach means that subtile changes in line-height can affect the size of the Header component (since it's now auto-sizing), which may have detremental effects on app layout if they weren't expecting the body area to change dimensions.
* Setting `auto` or `unset` height means that we can no longer smoothly transition, since animating from `x` to `auto` isn't a valid change. Transitions must always be to known values.

### DIFF
--- a/packages/moonstone/Panels/Header.js
+++ b/packages/moonstone/Panels/Header.js
@@ -294,7 +294,7 @@ const HeaderBase = kind({
 			case 'standard': return (
 				<Layout component="header" aria-label={title} {...rest} orientation="vertical">
 					{titleOrInput}
-					<Cell shrink size={96}>
+					<Cell shrink size={96} className={css.info}>
 						<Layout align="end">
 							<Cell className={css.titlesCell}>
 								{titleBelowComponent}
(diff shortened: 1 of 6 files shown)


packages/moonstone/Panels/Header.js
```diff
@@ -246,7 +246,7 @@ const HeaderBase = kind({
 		delete rest.subTitleBelow;
 		delete rest.titleBelow;
 
-		const titleHeight = ri.unit(ri.scale((type === 'dense') ? 69 : 90), 'rem');
+		const titleHeight = (type === 'dense') ? 69 : 90;
```
webOS101: Would it be worth documenting this as almost all other measurements _aren't_ in rem?

Djspaceg: It was documented in `ui/Layout.Cell#size`, but I can't find that documentation now. `size` supports strings, which are applied literally to the size rule, or as a number, which is converted safely to `rem` inside `Cell`. There are even examples of this in the docs, so I'm not sure where the docs went.

PR #2646, opened 2019-11-08
ENYO-6352: Fix Notification to support 3 full-width buttons in one line

Djspaceg:
### Checklist

3 max-width buttons don't fit in a wide `Notification`.

### Resolution
The width of `Notification`is now based on the size of 3 max-width buttons. (It's wider now.)

### DIFF
--- a/packages/moonstone/styles/variables.less
+++ b/packages/moonstone/styles/variables.less
@@ -470,8 +470,8 @@
 @moon-notification-padding-left-right: 39px;
 @moon-notification-padding-top-bottom: 24px;
 @moon-notification-min-width: 300px;
-@moon-notification-max-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (2 * ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width)));  // Attempt to calculate the width of all of the buttons and the pieces of Notification that influence the width to get the maximum size
-@moon-notification-wide-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (3 * ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width)));
+@moon-notification-max-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (2 * ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width)) + 12px);  // Attempt to calculate the width of all of the buttons and the pieces of Notification that influence the width to get the maximum size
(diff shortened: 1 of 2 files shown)


packages/moonstone/styles/variables.less
```diff
@@ -470,8 +470,8 @@
 @moon-notification-padding-left-right: 39px;
 @moon-notification-padding-top-bottom: 24px;
 @moon-notification-min-width: 300px;
-@moon-notification-max-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (2 * ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width)));  // Attempt to calculate the width of all of the buttons and the pieces of Notification that influence the width to get the maximum size
-@moon-notification-wide-width: 1287px;
+@moon-notification-max-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (2 * ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width)) + 12px);  // Attempt to calculate the width of all of the buttons and the pieces of Notification that influence the width to get the maximum size
+@moon-notification-wide-width: ((@moon-notification-border-width * 2) + (@moon-notification-padding-left-right * 2) + (3 * ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width)) + 12px);
```
webOS101: Could we make this something like:
```
@moon-notification-wide-width: @moon-notification-max-width + ((@moon-button-small-h-margin * 2) + @moon-button-small-max-width))
```
(Or, perhaps even collapse `(@moon-button-small-h-margin * 2) + @moon-button-small-max-width)` into a `@moon-button-small-full-width`)

Djspaceg: I'd like to discorage the creation (and maintinence) of specialized variables that aren't directly used by the components' LESS files.

webOS101: That makes sense.  What about the meatier part of the comment?

Djspaceg: I'm less excited about that approach as it assumes that the metrics for normal vs wide are the same. With this approach it's currently very clear that they're identical with the only difference being the amount of buttons. That approach also abstracts what's going on which means you need to understand what's going on in the normal version before cloning part of that into the wide. Imagine in the future we need to change the amount of buttons. Now that line must change drasticly by adding in an additional calculation. The current approach allows you to just change 1 number. Similar to what would happen if wide mode suddenly had smaller paddings. The current approach you'd just update the padding and be done. With the composite approach you'd have to reverse engineer where those numbers are different and figure out how to get it all right again.

**tl;dr**: It's better for maintainability if we don't do that.

webOS101: Works for me!

PR #2648, opened 2019-11-11
PLAT-94532: Prevent double read of marqueeing text

webOS101:

### Checklist

A marqueeing item that needed to read again would read the contents twice

### Resolution
Set an `aria-label` if one was not supplied, the contents (`children`) is plain text and the marquee will animate

### Additional Considerations
If the marquee does not contain 'plain' children, the readout could be  unexpected. Developers can override by setting `aria-label` directly in these cases.

### Links
PLAT-94352

### DIFF
--- a/packages/ui/Marquee/MarqueeBase.js
+++ b/packages/ui/Marquee/MarqueeBase.js
@@ -45,6 +45,15 @@ const MarqueeBase = kind({
 		animating: PropTypes.bool,
 
 		/**
+		 * Sets the value of the `aria-label` attribute for the wrapped component.
+		 *
+		 * @memberof ui/Marquee.MarqueeBase.prototype
+		 * @type {String}
+		 * @public
+		 */
+		'aria-label': PropTypes.string,
+
+		/**
 		 * The text or a set of components that should be marqueed
 		 *
 		 * This prop may be empty in some cases, which is OK.
@@ -182,6 +191,8 @@ const MarqueeBase = kind({
 	},
 
 	computed: {
+		'aria-label': ({'aria-label': aria, children}) => aria == null &&
+			typeof children === 'string' ? children : aria,
 		clientClassName: ({animating, willAnimate, styler}) => styler.join({
 			animate: animating,
 			text: true,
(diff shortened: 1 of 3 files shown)


packages/ui/Marquee/MarqueeBase.js
```diff
@@ -182,6 +191,8 @@ const MarqueeBase = kind({
 	},
 
 	computed: {
+		'aria-label': ({'aria-label': aria, children}) => aria == null &&
+			typeof children === 'string' ? children : aria,
```
ryanjduffy: I slightly more robust version of this would map over the children and join all children. This would be necessary when evaluating content within children:

```
// children.length would be 3 but all are strings
this {count === 0 ? 'does not have' : 'has'} children
```

https://codesandbox.io/s/inspiring-williams-j3rmv

webOS101: I briefly considered this, but do you have any examples of this pattern?  If there's one, I would definitely suggest we robustify this.  Otherwise, I think you could get carried away trying to iterate through everything.  On the other hand, is potentially omitting some data and having it repeated once better than repeating the full string twice?

ryanjduffy: I don't have product app examples but I think the one above is a reasonable pattern and likely how I'd have written the code.

webOS101: Do you have a response for the last question there?

webOS101: Updated and now will only apply label if absolutely necessary.

packages/ui/Marquee/MarqueeBase.js
```diff
@@ -182,6 +191,15 @@ const MarqueeBase = kind({
 	},
 
 	computed: {
+		'aria-label': ({'aria-label': aria, children, distance, willAnimate}) => {
+			if (aria == null && willAnimate && distance > 0) {
+				return React.Children.map(children, c => typeof c === 'string' && c)
+					.filter(Boolean)
+					.join(' ') || aria;
```
ryanjduffy: Why fallback to `aria` here? It reads a bit odd to me since the purpose of this branch is to override `aria`.

webOS101: I could have gone with `void 0` but that looks :yuck:

webOS101: Added some tests to validate the intentions of this.

PR #2649, opened 2019-11-13
ENYO-6336: Sort icons in sampler knobs

LucieRoy:
### Checklist

### Resolution

### Additional Considerations

### Links

### DIFF
--- a/packages/sampler/stories/default/ExpandableInput.js
+++ b/packages/sampler/stories/default/ExpandableInput.js
@@ -1,11 +1,11 @@
 import ExpandableInput, {ExpandableInputBase} from '@enact/moonstone/ExpandableInput';
-import iconNames from './icons';
+import icons from './icons';
 import React from 'react';
 import {storiesOf} from '@storybook/react';
 import {boolean, select, text} from '../../src/enact-knobs';
 import {action, mergeComponentMetadata} from '../../src/utils';
 
-const sortedIconNames = ['', ...iconNames.sort()];
+const iconNames = ['', icons];
 
 const Config = mergeComponentMetadata('ExpandableInput', ExpandableInputBase, ExpandableInput);
 ExpandableInput.displayName = 'ExpandableInput';
@@ -21,8 +21,8 @@ storiesOf('Moonstone', module)
 		() => (
 			<ExpandableInput
 				disabled={boolean('disabled', Config)}
-				iconAfter={select('iconAfter', sortedIconNames, Config)}
-				iconBefore={select('iconBefore', sortedIconNames, Config)}
+				iconAfter={select('iconAfter', iconNames, Config)}
+				iconBefore={select('iconBefore', iconNames, Config)}
 				noneText={text('noneText', Config, 'noneText')}
 				onChange={action('onChange')}
 				onClose={action('onClose')}
--- a/packages/sampler/stories/default/ExpandableInput.js
+++ b/packages/sampler/stories/default/ExpandableInput.js
@@ -1,12 +1,11 @@
 import ExpandableInput, {ExpandableInputBase} from '@enact/moonstone/ExpandableInput';
-import {icons} from '@enact/moonstone/Icon';
+import iconNames from './icons';
 import React from 'react';
 import {storiesOf} from '@storybook/react';
-
 import {boolean, select, text} from '../../src/enact-knobs';
 import {action, mergeComponentMetadata} from '../../src/utils';
 
-const iconNames = ['', ...Object.keys(icons)];
+const sortedIconNames = ['', ...iconNames.sort()];
 
 const Config = mergeComponentMetadata('ExpandableInput', ExpandableInputBase, ExpandableInput);
 ExpandableInput.displayName = 'ExpandableInput';
@@ -22,8 +21,8 @@ storiesOf('Moonstone', module)
 		() => (
 			<ExpandableInput
 				disabled={boolean('disabled', Config)}
-				iconAfter={select('iconAfter', iconNames, Config)}
-				iconBefore={select('iconBefore', iconNames, Config)}
+				iconAfter={select('iconAfter', sortedIconNames, Config)}
+				iconBefore={select('iconBefore', sortedIconNames, Config)}
 				noneText={text('noneText', Config, 'noneText')}
 				onChange={action('onChange')}
 				onClose={action('onClose')}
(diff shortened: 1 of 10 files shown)


packages/sampler/stories/default/ExpandableInput.js
```diff
@@ -1,12 +1,11 @@
 import ExpandableInput, {ExpandableInputBase} from '@enact/moonstone/ExpandableInput';
-import {icons} from '@enact/moonstone/Icon';
+import iconNames from './icons';
```
webOS101: Should this same change be applied in other files that import `icons` from `@enact/moonstone/Icon`?

packages/sampler/stories/qa/ExpandableInput.js
```diff
@@ -9,15 +9,16 @@ import {action} from '../../src/utils';
 ExpandableInput.displayName = 'ExpandableInput';
 
 const iconNames = ['', ...Object.keys(icons)];
+const sortedIconNames = ['', ...iconNames.sort()];
```
webOS101: This doubles the empty string.  Can simplify this change to just:
```
const iconNames = ['', ...Object.keys(icons).sort()];
```

And then you won't have to change the name used below.

packages/sampler/stories/qa/LabeledIcon.js
```diff
@@ -70,7 +71,7 @@ storiesOf('LabeledIcon', module)
 			const labelPosition = select('labelPosition', ['above', 'after', 'before', 'below', 'left', 'right'], Config);
 			return (
 				<LabeledIcon
-					icon={select('icon', ['', ...iconNames], Config, 'fullscreen')}
+					icon={select('icon', ['', ...sortedIconNames], Config, 'fullscreen')}
```
webOS101: This would also double up the empty string here.  Can either add it as above or use as you do in other places.

packages/sampler/stories/default/ExpandableInput.js
webOS101: It might be clearer for me to say: Should we re-use the icons file from the default sampler in the qa sampler?

packages/sampler/stories/qa/ExpandableInput.js
Djspaceg: Alternatively: `iconNames.sort();`

packages/sampler/stories/qa/LabeledIcon.js
Djspaceg: I'd recommend doing the following for these, to reduce the amount of code duplication:
```
icon={select('icon', sortedIconNames, Config, 'fullscreen')}
```
Clearer to read, less duplication, no spreading, no extra spaces.

packages/sampler/stories/default/ExpandableInput.js
LucieRoy: "Can't resolve './icons' in '/Users/lucie.roy/Work/enact/packages/sampler/stories/qa'"
It was one of the reason I left it as it was ... although would have like the code to be similar across. Any suggestion please?

LucieRoy: tried to adjust the path but it is not fixing the error either.

Djspaceg: I think Roy meant it should refer to the `icons.js` file in the default folder, so: `'../default/icons'`

webOS101: It would probably require moving the `icons.js` file into `../../utils/` and then including it from there.

webOS101: In theory, you could import from `../default/icons.js` but that seems lazy.

Djspaceg: Personally, I don't feel it's "correct" to move Moonstone's icon list into the general util directory. Meanwhile, it's more appropriate to have Moonstone's QA folder refer to stuff in Moonstone's "default" folder.

webOS101: Well...  OK.

LucieRoy: Thank you to both... the path fixed it. I was using the incorrect path and the incorrect file but it was the idea (Blake fixed the path). It works for at least one. Will test the others.

webOS101: If we're always sorting it, we could export the sorted list from `icons.js`.

Djspaceg: That seems like an ok approach too.

```
const sortedIcons = Object.keys(icons).sort();

export default Object.keys(icons);
export {decrementIcons, incrementIcons, listIcons, mediaIcons, arrowIcons, starIcons, sortedIcons};
```

packages/sampler/stories/qa/ExpandableInput.js
LucieRoy: fixed with:
```const sortedIconNames = ['', ...iconNames.sort()];```
else the blank selection does not show with:
```const sortedIconNames = iconNames.sort();```

Comment (Djspaceg) [reply packages/sampler/stories/qa/ExpandableInput.js]: Is the blank just sorted to the end of the list? So now there are 2 blanks, one at the top and one at the bottom?

Comment (LucieRoy) [reply packages/sampler/stories/qa/ExpandableInput.js]: Blank is at the top only in the knob in the dropdown

Comment (Djspaceg) [reply packages/sampler/stories/qa/ExpandableInput.js]: It must be getting pruned out elsewhere in the code. Because line 11 adds a blank (empty string) and then this line, 12, adds another blank (empty string).

File: packages/sampler/stories/qa/LabeledIcon.js (hunk 3 above)
Comment (LucieRoy) [reply packages/sampler/stories/qa/LabeledIcon.js]: Removed one of the 2 options and used 
```icon={select('icon', sortedIconNames, Config, 'fullscreen')}```

File: packages/sampler/stories/default/ExpandableInput.js (hunk 1 above)
Comment (LucieRoy) [reply packages/sampler/stories/default/ExpandableInput.js]: I will submit what I have and you can decide for next steps

File: packages/sampler/stories/qa/Picker.js [hunk 4]
```diff
@@ -1,12 +1,12 @@
...
-import {icons} from '@enact/moonstone/Icon';
 import PickerAddRemove from './components/PickerAddRemove';
 import PickerRTL from './components/PickerRTL';
 import React from 'react';
 import {storiesOf} from '@storybook/react';
 
-import {boolean, select} from '../../src/enact-knobs';
 import {action} from '../../src/utils';
+import {boolean, select} from '../../src/enact-knobs';
+import iconNames from '../default/icons';
```
ryanjduffy: In the future, we might consider moving the icons file into a shared folder rather than referencing `default` from `qa`.

PR #2652, opened 2019-11-18
PLAT-86361: Fix VirtualList with different item sizes to render properly when dataSize is 0

0x64:
### Checklist

There is a bug to access a member property of `undefined` if `VirtualList` has no item when `itemSizes` prop is given. This PR adds conditions to avoid invalid access.
Even if there is no item inside a list, a hidden placeholder item for spotlight management gets focus and a list try to get/calculate the position of the item has index 0. This makes access to an invalid `undefined` object.
Also, a new potential risk is found in a case that `itemSIzes` does not have sizing info for a target index.

### Resolution
Added conditions to avoid invalid access.

### Additional Considerations
The fix assumes that `itemSizes` has no empty element between the first and the last elements.

### Links
PLAT-86361

### DIFF
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -471,7 +471,7 @@ const VirtualListBaseFactory = (type) => {
 			const extent = Math.floor(index / dimensionToExtent);
 			let primaryPosition;
 
-			if (this.props.dataSize > index && this.props.itemSizes && typeof this.props.itemSizes[index] !== 'undefined') {
+			if (this.props.itemSizes && typeof this.props.itemSizes[index] !== 'undefined' && this.props.dataSize > index) {
 				const firstIndexInExtent = extent * dimensionToExtent;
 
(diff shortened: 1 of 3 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -471,7 +471,7 @@ const VirtualListBaseFactory = (type) => {
 			const extent = Math.floor(index / dimensionToExtent);
 			let primaryPosition;
 
-			if (this.props.itemSizes) {
+			if (this.props.dataSize > index && this.props.itemSizes && typeof this.props.itemSizes[index] !== 'undefined') {
```
ybsung: Is it better to check `this.props.itemSizes` first because `VirtualList` usually does not include the `itemSize` prop, then the condition would be `false` quickly?

0x64: 👌

ybsung:
LGTM

PR #2653, opened 2019-11-19
PLAT-90768: Add ScrollToTop Button

teckliew:
### Checklist

Add scroll to top button that appears at the edge of the screen.

### DIFF
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -25,6 +25,7 @@ import $L from '../internal/$L';
 import {SharedState} from '../internal/SharedStateDecorator';
 
 import Scrollbar from './Scrollbar';
+import Button from '../Button';
 import Skinnable from '../Skinnable';
 
 import overscrollCss from './OverscrollEffect.module.less';
@@ -458,6 +459,10 @@ class ScrollableBase extends Component {
 				this.lastScrollPositionOnFocus = pos;
 			}
 		}
+	}
+
+	onScrollToTopButtonClick = () => {
+		this.uiRef.current.start({targetX: 0, targetY: 0});
 	}
 
 	calculateAndScrollTo = () => {
@@ -1036,6 +1041,17 @@ class ScrollableBase extends Component {
 								null
 							}
 						</div>
+						<Button
+							icon="arrowlargeup"
+							onTap={this.onScrollToTopButtonClick}
+							style={{
+								position: 'absolute',
+								right: '0',
+								bottom: '0'
+							}}
+						>
+							Back to top
+						</Button>
 						{isHorizontalScrollbarVisible ?
 							<Scrollbar
 								{...horizontalScrollbarProps}
--- a/packages/moonstone/Scrollable/Scrollable.js
+++ b/packages/moonstone/Scrollable/Scrollable.js
@@ -28,6 +28,7 @@ import Scrollbar from './Scrollbar';
 import Button from '../Button';
 import Skinnable from '../Skinnable';
 
+import css from './Scrollable.module.less';
 import overscrollCss from './OverscrollEffect.module.less';
 import scrollbarCss from './Scrollbar.module.less';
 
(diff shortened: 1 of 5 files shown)


packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -460,6 +475,10 @@ class ScrollableBase extends Component {
 		}
 	}
 
+	onScrollToTopButtonClick = () => {
+		this.uiRef.current.start({targetX: 0, targetY: 0});
```
ryanjduffy: Expect we'll want to focus the first item along with this which I think means we'll want to use `scrollTo` rather than this (more) private method. @germboy or @ybsung can likely help on this.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -316,6 +327,10 @@ class ScrollableBase extends Component {
 		this.childRef = React.createRef();
 		this.uiRef = React.createRef();
 
+		this.state = {
```
ryanjduffy: We might want to use imperative APIs to only update the state at the button rather than on the scrollable since this will trigger re-renders for everything within the scroller which can be expensive.

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -1036,6 +1083,19 @@ class ScrollableBase extends Component {
...
 						</div>
+						<Button
+							className={classNames(
+								css.scrollToTopButton,
+								this.state.shouldShowScrollToTopButton ? css.showScrollToTopButton : null,
+								isHorizontalScrollbarVisible ? css.horizontalScrollbar : null,
+								isVerticalScrollbarVisible ? css.verticalScrollbar : null,
+								rtl ? css.rtlScrollToTopButton : null
+							)}
+							icon="arrowlargeup"
```
ryanjduffy: Could choose alternate icons for horizontal and horizontal + rtl too. We'll sort that out when the final UX lands.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1051,7 +1051,7 @@ class ScrollableBase extends Component {
 		}
 
 		this.childRefCurrent.setScrollPosition(this.scrollLeft, this.scrollTop, this.props.rtl, ...rest);
-		this.forwardScrollEvent('onScroll');
+		this.forwardScrollEvent('onScroll', this.getReachedEdgeInfo());
```
ryanjduffy: `forwardScrollEvent` might need to be updated to always include this object. I'm not clear on why only some events include it. Also, the docs for these events seem to be missing that this data is included as well.

cc: @ybsung

packages/moonstone/Scrollable/Scrollable.js
```diff
@@ -837,6 +856,31 @@ class ScrollableBase extends Component {
...
+		const isRtl = this.uiRef.current.props.rtl;
+		const hasReachedEdge = (
+			bottom && (direction === 'both' || direction === 'vertical') ||
+			!isRtl && (right && (direction === 'both' || direction === 'horizontal')) ||
+			isRtl && (left && (direction === 'both' || direction === 'horizontal'))
+		);
+
+		if (showScrollToTopButton && hasReachedEdge) {
+			this.showScrollToTopButton();
+		} else if (!hasReachedEdge) {
```
ryanjduffy: I suspect we'll want to show this button sooner than when we reach the end as it would be useful much earlier for long scrollers. We might need a prop to configure this either absolutely (e.g. after 200px) or relatively (e.g. after 20%). Or maybe it'll be fixed for moonstone since it's opionated. Yet to be decided. :)

packages/moonstone/Scrollable/Scrollable.js
ybsung: As @ryanjduffy mentioned, calling `scrollTo` would be more safer than calling `start` directly because the states in the `VirtualList` or the `Scrollable` could be conflict while swapping items. In this case, calling `start` would be deferred internally in the `scrollTo` function.

In addition, we usually check several points before jumping to another index.
In the `onAcceleratedKeyDown` of the `VirtualList`,
1) if the next item (the 0 index item in this case) is in a current page, we just focus it.
2) if the next item is out of a current page, we just blur the current focused item and call the `scrollTo` with an index (the 0 in this case).

PR #2657, opened 2019-12-04
ENYO-6364: Upgraded ui/Button to leverage ComponentOverride for iconComponent

Djspaceg:
### Checklist

Customization the `iconComponent` of `ui/Button` don't behave as expected.

Example error:
```
Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: object.
```

### Resolution
When using the `iconComponent` prop of `ui/Button`, the expectation is that it behaves the same way (supports the same features and capabilities) as other overridable components. `ui/Button` used a legacy approach to supporting custom component overrides and this proved to be problematic for customization.

The new code not only updates the approach to the modern Enact standard (`ComponentOverride`) but also now supports simpler usage of HTML elements, as now, you're no longer required to prune out unsupported non-HTML-safe props off your component.

### DIFF
--- a/packages/ui/Button/Button.js
+++ b/packages/ui/Button/Button.js
@@ -94,10 +94,12 @@ const ButtonBase = kind({
 		/**
 		 * The component used to render the [icon]{@link ui/Button.ButtonBase.icon}.
 		 *
-		 * This component will receive the `size` and `spriteCount` properties and should them.
-		 * This coomponent will also receive the `icon` class to customize its styling.
-		 * If [icon]{@link ui/Button.ButtonBase.icon} is not assigned, this component will not be
-		 * rendered.
+		 * This component will receive the `icon` class to customize its styling.
+		 * If [icon]{@link ui/Button.ButtonBase.icon} is not assigned or is false, this component
+		 * will not be rendered.
+		 *
+		 * If this is a component rather than an HTML element string, this component will also
+		 * receive the `size` property and should be configured to handle it.
 		 *
 		 * @type {Component}
 		 * @public
@@ -175,32 +177,25 @@ const ButtonBase = kind({
 		}, size),
 		icon: ({css, icon, iconComponent, size}) => {
 			if (icon == null || icon === false) return;
-			// If the iconComponent is a string, we exclude the obviously unsupported props
-			// (non-HTML) from the component to simplify the usage of plain HTML elements as the
-			// component. Ex: 'div', 'span', etc.
-			if (typeof iconComponent === 'string') {
-				return (
-					<ComponentOverride
-						component={iconComponent}
-						className={css.icon}
-					>
-						{icon}
-					</ComponentOverride>
-				);
-			} else {
-				// Relay the full compliment of props to the iconComponent. Full Enact components
-				// should be responsible for handling these additional props themselves.
-				// console.log('Rendering %s as a %s icon:', icon, size);
-				return (
-					<ComponentOverride
-						component={iconComponent}
-						size={size}
-						className={css.icon}
-					>
-						{icon}
-					</ComponentOverride>
-				);
+
+			// Establish the base collection of props for the moost basic `iconComponent` type, an
+			// HTML element string.
+			const props = {
+				className: css.icon,
+				component: iconComponent
+			};
+
+			// Add in additional props that any Component supplied to `iconComponent` should be
+			// configured to handle.
+			if (typeof iconComponent !== 'string') {
+				props.size = size;
 			}
+
+			return (
+				<ComponentOverride {...props}>
+					{icon}
+				</ComponentOverride>
+			);
 		}
 	},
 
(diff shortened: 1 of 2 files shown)


packages/ui/Button/Button.js
```diff
@@ -81,17 +82,22 @@ const ButtonBase = kind({
...
+		icon: PropTypes.oneOfType([PropTypes.node, PropTypes.bool]),
 
 		/**
 		 * The component used to render the [icon]{@link ui/Button.ButtonBase.icon}.
 		 *
-		 * This component will receive the `size` property set on the Button as well as the `icon`
-		 * class to customize its styling. If [icon]{@link ui/Button.ButtonBase.icon} is not a
-		 * string, this property is not used.
+		 * This component will receive the `size` and `spriteCount` properties and should them.
+		 * This coomponent will also receive the `icon` class to customize its styling.
```
sugardave: 97: What should it do with the `size` and `spriteCount` props?

98: `coomponent` -> `component`

packages/ui/Button/Button.js
```diff
@@ -167,10 +173,34 @@ const ButtonBase = kind({
...
+				return (
+					<ComponentOverride
+						component={iconComponent}
+						size={size}
+						className={css.icon}
+					>
+						{icon}
+					</ComponentOverride>
+				);
+			}
```
sugardave: These returns are very similar, just one prop difference?  What about something like:
```
const props = {
    className: css.icon,
    component: iconComponent,
    size
};

if (typeof iconComponent === 'string') {
    delete props.size;
}

return (
    <ComponentOverride {...props}>
        {icon}
    </ComponentOverride>
);
```

Could also add `children: icon` to those props for a one-line return.

packages/ui/Button/Button.js
Djspaceg: It should ^handle^ them. How it handles them, is the "coooooomponent's" deal.

packages/ui/Button/Button.js
Djspaceg: Yeah, this is a better approach. Originally, during the development of this, there were more props, so it was more appropriate earlier. I'll update.

Djspaceg:
Cleaned up, should be g2g now.

packages/ui/Button/Button.js
```diff
@@ -167,10 +175,27 @@ const ButtonBase = kind({
...
+			// Establish the base collection of props for the moost basic `iconComponent` type, an
+			// HTML element string.
+			const props = {
+				className: css.icon,
+				component: iconComponent
+			};
+
+			// Add in additional props that any Component supplied to `iconComponent` should be
+			// configured to handle.
+			if (typeof iconComponent !== 'string') {
```
ryanjduffy: @Djspaceg - Would like to talk through this change a bit more together. We're creating an exception here -- `iconComponent` can be a string and thereby opts out of receiving the custom props -- and I want to understand the conditions for the exception better.

packages/ui/Button/Button.js
```diff
@@ -81,17 +82,24 @@ const ButtonBase = kind({
...
-		 * class to customize its styling. If [icon]{@link ui/Button.ButtonBase.icon} is not a
-		 * string, this property is not used.
+		 * This component will receive the `icon` class to customize its styling.
+		 * If [icon]{@link ui/Button.ButtonBase.icon} is not assigned or is false, this component
+		 * will not be rendered.
+		 *
+		 * If this is a component rather than an HTML element string, this component will also
+		 * receive the `size` property and should be configured to handle it.
 		 *
 		 * @type {Component}
```
ryanjduffy: This and the prop type below should be updated to be `Component | Node` and `componentOverride`, respectively.

Djspaceg: Cool. Didn't know about that. Thanks!

packages/ui/Button/Button.js
Djspaceg: ComponentOverride automatically excludes props that are in all lowercase? I didn't realize it did any props augmentation. That's really cool. We can then refactor this whole conditional then.

ryanjduffy: > ComponentOverride automatically excludes props that are in all lowercase?

No. Sorry for the confusion. `ComponentOverride` passes everything through. My point was that this is the only place where we attempt to process props differently when overriding. Everywhere else we always pass all the documented props through.

ryanjduffy: > `iconComponent` can be a string and thereby opts out of receiving the custom props

I was calling out the exception and not stating the current behavior. I see how that was confusing!

Djspaceg: Ah ok. This seemed like a good value-add, since this code could be really buried in the stack. It didn't seem intuitive to me to have `size` added to a `div`. If a developer creates their own component, and it is just a plain `div`, it will actually be a React component and will get the `size` prop, and because it's a component, I'll have the ability to prune it off, if I don't want to act on it.

PR #2659, opened 2019-12-10
Update a list item style for horizontal lists

0x64:
### Checklist

A virtual list sets a CSS class for its item internally and the CSS covers a vertical list only.

### Resolution
Update CSS to cover an item of a horizontal list also.

### Additional Considerations
A list sets a selector of two CSS classes for an item currently and it blocks an app side single CSS class for an item by higher specificity. We are using modular CSS now, so we actually don't need to put such a complex CSS selector for a list item. To make apps easy to override item sizes, needs to be flattened.

### Links
PLAT-86356

### DIFF
--- a/packages/sampler/stories/default/VirtualList.js
+++ b/packages/sampler/stories/default/VirtualList.js
@@ -17,17 +17,26 @@ const
 		'&quot;noAnimation&quot;': 'noAnimation'
 	},
 	prop = {
+		direction: {horizontal: 'horizontal', vertical: 'vertical'},
 		scrollbarOption: ['auto', 'hidden', 'visible']
 	},
 	items = [],
 	defaultDataSize = 1000,
 	// eslint-disable-next-line enact/prop-types, enact/display-name
-	uiRenderItem = (size) => ({index, ...rest}) => {
-		const itemStyle = {
-			borderBottom: ri.unit(3, 'rem') + ' solid #202328',
-			boxSizing: 'border-box',
-			height: size + 'px'
-		};
+	uiRenderItem = (direction, size) => ({index, ...rest}) => {
+		const
+			sizePx = size + 'px',
+			itemStyle = {
+				borderBottom: ri.unit(3, 'rem') + ' solid #202328',
+				boxSizing: 'border-box'
+			};
+
+		if (direction !== 'horizontal') {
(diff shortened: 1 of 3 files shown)


packages/sampler/stories/default/VirtualList.js
```diff
@@ -17,17 +17,26 @@ const
 		'&quot;noAnimation&quot;': 'noAnimation'
 	},
 	prop = {
+		direction: {horizontal: 'horizontal', vertical: 'vertical'},
```
Djspaceg: This can be simplified to just an array: `direction: ['horizontal', 'vertical']`

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -1145,13 +1145,12 @@ const VirtualListBaseFactory = (type) => {
...
-
-			if (type === Native) {
-				containerClass = this.isPrimaryDirectionVertical ? css.vertical : css.horizontal;
-			}
-
-			return classNames(css.virtualList, containerClass, className);
+			return classNames(
+				css.virtualList,
+				this.isPrimaryDirectionVertical ? css.vertical : css.horizontal,
+				(type === Native) ? css.native : null,
```
Djspaceg: The `classNames` module accepts several convenient formats, including using key/value pairs to represent classname/truthyness, so this line could also be represented more easily as:
```
{[css.native]: (type === Native)}
```
Admittedly this looks a little complicated, but that's just because there's only one key/value pair in this object. Any number of keys can be in this object and it's the basis of how we structure adding styles in `kind()`. Also, when we accept one of a specific set of keywords as a prop, ex: `orientation: PropTypes.oneOf(['vertical', 'horizontal'])`, we can then easily dump that directly into the class list and use it as a class name:
```
	computed: {
		className: ({orientation, styler}) => styler.append(orientation),
	}
```
The propTypes limits the names allowed to one of our known class names, and we only have styles written for the known class names, so it all just works out.

0x64: Good!

packages/sampler/stories/default/VirtualList.js
0x64: Oops!

PR #2660, opened 2019-12-13
PLAT-87249: Add noLockPointer config to InputSpotlightDecorator

jeongran1230:
### Checklist

1) One of `moonstone/Input` requirements is that users should not be able to click on other components while focusing on an input field. It caused confusing user interaction on touchable devices by a user have to tap twice to click any components.

### Resolution
By adding `unlockedPointer` config to InputSpotlightDecorator, we provide an optional way to avoid a pointer locked which block click and tap actions when the input has focus.

### Additional Considerations
We set the default value of `unlockedPointer` config to follow the UX requirement. The application developer can apply the config to the exceptional cases selectively.

### Links
 PLAT-87249

### DIFF
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,7 +6,6 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
-- `moonstone/Input.InputSpotlightDecorator` to add unLockedPointer config
 - `moonstone/VirtualList.VirtualList` to render properly without error when `itemSizes` is given and `dataSize` is 0
 
 ## [3.2.5] - 2019-11-14
--- a/packages/moonstone/CHANGELOG.md
+++ b/packages/moonstone/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Input.InputSpotlightDecorator` to add unLockedPointer config
 - `moonstone/VirtualList.VirtualList` to render properly without error when `itemSizes` is given and `dataSize` is 0
 
 ## [3.2.5] - 2019-11-14
--- a/packages/moonstone/Input/InputSpotlightDecorator.js
+++ b/packages/moonstone/Input/InputSpotlightDecorator.js
@@ -31,7 +31,7 @@ const handleKeyDown = handle(
 /**
  * Default config for [InputSpotlightDecorator]{@link moonstone/Input.InputSpotlightDecorator}
  *
- * @memberof moonstone/Input.InputSpotlightDecorator
+ * @memberof moonstone/Input/InputSpotlightDecorator.InputSpotlightDecorator
  * @hocconfig
  */
 const defaultConfig = {
@@ -40,7 +40,7 @@ const defaultConfig = {
 	 *
 	 * @type {Boolean}
 	 * @default false
-	 * @memberof moonstone/Input.InputSpotlightDecorator.defaultConfig
+	 * @memberof moonstone/Input/InputSpotlightDecorator.InputSpotlightDecorator.defaultConfig
 	*/
 	unlockedPointer: false
 };
@@ -50,7 +50,7 @@ const defaultConfig = {
  * spotlight behavior for an {@link moonstone/Input.Input}
  *
  * @class InputSpotlightDecorator
- * @memberof moonstone/Input.InputSpotlightDecorator
+ * @memberof moonstone/Input/InputSpotlightDecorator
  * @hoc
  * @private
  */
--- a/packages/moonstone/Input/InputSpotlightDecorator.js
+++ b/packages/moonstone/Input/InputSpotlightDecorator.js
@@ -30,8 +30,9 @@ const handleKeyDown = handle(
 
 /**
  * Default config for {@link moonstone/Input.InputSpotlightDecorator}
- * @hocconfig
+ *
  * @memberof moonstone/Input.InputSpotlightDecorator
+ * @hocconfig
  */
 const defaultConfig = {
 	/**
--- a/packages/moonstone/Input/InputSpotlightDecorator.js
+++ b/packages/moonstone/Input/InputSpotlightDecorator.js
@@ -29,7 +29,7 @@ const handleKeyDown = handle(
 
 
 /**
- * Default config for {@link moonstone/Input.InputSpotlightDecorator}
+ * Default config for [InputSpotlightDecorator]{@link moonstone/Input.InputSpotlightDecorator}
  *
  * @memberof moonstone/Input.InputSpotlightDecorator
  * @hocconfig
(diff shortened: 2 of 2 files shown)


packages/moonstone/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact moonstone module, newest
 
 ### Fixed
 
+- `moonstone/Input.InputSpotlightDecorator` to add unLockedPointer config
```
ybsung: I think that I would be better to move this CHANGELOG from Fixed to Added with the format like the following changelog.

```
### Added

- `moonstone/Heading` prop `spacing` with default value `'small'`
```

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -27,6 +27,24 @@ const handleKeyDown = handle(
 	call('onKeyDown')
 );
 
+
+/**
+ * Default config for {@link moonstone/Input.InputSpotlightDecorator}
+ * @hocconfig
```
ybsung: It would be better to move this line after the next line.

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -27,6 +27,24 @@ const handleKeyDown = handle(
 	call('onKeyDown')
 );
 
+
+/**
+ * Default config for {@link moonstone/Input.InputSpotlightDecorator}
```
ybsung: Please add one empty line after this line.

ybsung: Not `Default config for {@link moonstone/Input.InputSpotlightDecorator}`
but `Default config for [`InputSpotlightDecorator`]{@link moonstone/Input/InputSpotlightDecorator.InputSpotlightDecorator}`

The format is [`config name`]{@link path/moduleName.componentName}.

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -27,6 +27,24 @@ const handleKeyDown = handle(
...
+ * @hocconfig
+ * @memberof moonstone/Input.InputSpotlightDecorator
```
ybsung: Same above.

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -27,6 +27,24 @@ const handleKeyDown = handle(
...
+ */
+const defaultConfig = {
+	/**
+	 * Suppress the pointer lock behavior of moonstone input
+	 *
+	 * @type {Boolean}
+	 * default false
```
ybsung: Not `default` but `@default`

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -155,11 +176,15 @@ const InputSpotlightDecorator = hoc((config, Wrapped) => {
 			if (focusChanged) {
 				if (this.state.focused === 'input') {
 					forward('onActivate', {type: 'onActivate'}, this.props);
-					lockPointer(this.state.node);
+					if (!unLockedPointer) {
+						lockPointer(this.state.node);
+					}
```
ybsung: If we have to wrap `lockPointer` and `releasePointer` with `unLockedPointer` condition, why don't you define new functions for them in which `unLockedPointer` condition is checked. Then we could avoid making a mistake to forget wrapping them with the condition.

webOS101:
We should not add CHANGELOG entries for private components.  The config should be private as the hoc itself is private.

webOS101:
We should not add CHANGELOG entries for private components.  The config should be private as the hoc itself is private.

packages/moonstone/CHANGELOG.md
jeongran1230: The change written in CHANGELOG is removed. (Roy mentioned the reason below)

packages/moonstone/Input/InputSpotlightDecorator.js
jeongran1230: Fixed

jeongran1230: Fixed

packages/moonstone/Input/InputSpotlightDecorator.js
jeongran1230: I checked it comparing to other components. I think it looks fine the way it is.

packages/moonstone/Input/InputSpotlightDecorator.js
jeongran1230: Fixed

packages/moonstone/Input/InputSpotlightDecorator.js
jeongran1230: It's a good idea but  `if` condition is used only three times in total. I think this code will be refactoring next year when we are thinking of handling touchable events. I don't know which is better.

ybsung: OK. Let's consider it later.

ybsung:
LGTM

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -27,6 +27,24 @@ const handleKeyDown = handle(
 	call('onKeyDown')
 );
 
+
+/**
+ * Default config for [InputSpotlightDecorator]{@link moonstone/Input.InputSpotlightDecorator}
+ *
+ * @memberof moonstone/Input/InputSpotlightDecorator.InputSpotlightDecorator
```
ryanjduffy: There were some doc parsing errors due to the pathing change above not matched everywhere in the module. While I think what you had `moonstone/Input.InputSpotlightDecorator` is probably "more right", I erred on the side of fewer changes and aligned the new docs with the existing naming.

packages/moonstone/Input/InputSpotlightDecorator.js
```diff
@@ -27,6 +27,24 @@ const handleKeyDown = handle(
...
+ */
+const defaultConfig = {
+	/**
+	 * Suppress the pointer lock behavior of moonstone input
+	 *
+	 * @type {Boolean}
+	 * @default false
+	 * @memberof moonstone/Input/InputSpotlightDecorator.InputSpotlightDecorator.defaultConfig
+	*/
+	noLockPointer: false
```
ryanjduffy: @ybsung, @jeongran1230 - I changed this name to `noLockPointer` to follow our usual convention for disabling default features. See `noAutoFocus` in `SpotlightRootDecorator` for some precedent on this. I know there's some dependent work on this so let me know if this change is a problem and we can adapt. It's not critical for a private API like this but want to strive for consistency when possible.

packages/moonstone/Input/InputSpotlightDecorator.js
webOS101: I would prefer to leave it the correct way.  It errors because the decorator is private, so this should be, too.  If we do that, the warnings should go away.

webOS101: Ugh, I see.  Well, just leave it as-is, then.

packages/moonstone/Input/InputSpotlightDecorator.js
jeongran1230: @ryanjduffy - Okey, I updated the title of this PR.

PR #2669, opened 2020-01-13
PLAT-98572: Convert scrollbar and scroll thumb components into functional components

0x64:
### Checklist

Convert scrollbar and scroll thumb components into functional components

### Links
PLAT-98572

### Comments
Note that this PR may not work until all related decoupling works get together.
Note: the target branch/repo may be changed later. This PR is for review for now.

### DIFF
--- a/packages/moonstone/Scrollable/ScrollThumb.js
+++ b/packages/moonstone/Scrollable/ScrollThumb.js
@@ -1,45 +1,44 @@
 import {ScrollThumb as UiScrollThumb} from '@enact/ui/Scrollable/Scrollbar';
 import PropTypes from 'prop-types';
-import React, {Component} from 'react';
+import React, {useEffect, forwardRef} from 'react';
 
 const nop = () => {};
 
 /**
  * A Moonstone-styled scroll thumb with moonstone behavior
  *
- * @class ScrollThumb
+ * @function ScrollThumb
  * @memberof moonstone/Scrollable
  * @extends ui/Scrollable/ScrollThumb
  * @ui
  * @private
  */
-class ScrollThumb extends Component {
-	static propTypes = /** @lends moonstone/Scrollable.ScrollThumb.prototype */ {
-		/**
-		 * Called when [ScrollThumb]{@link moonstone/Scrollable.ScrollThumb} is updated.
-		 *
-		 * @type {Function}
-		 * @private
-		 */
-		cbAlertThumb: PropTypes.func
-	}
+const ScrollThumb = forwardRef((props, ref) => {
+	const rest = Object.assign({}, props);
+	delete rest.cbAlertThumb;
 
-	static defaultProps = {
-		cbAlertThumb: nop
-	}
+	useEffect (() => {
+		props.cbAlertThumb();
+	});
 
-	componentDidUpdate () {
-		this.props.cbAlertThumb();
-	}
+	return <UiScrollThumb {...rest} ref={ref} />;
+});
 
-	render () {
-		const props = Object.assign({}, this.props);
+ScrollThumb.displayName = 'ScrollThumb';
 
-		delete props.cbAlertThumb;
+ScrollThumb.propTypes = /** @lends moonstone/Scrollable.ScrollThumb.prototype */ {
+	/**
+	 * Called when [ScrollThumb]{@link moonstone/Scrollable.ScrollThumb} is updated.
+	 *
+	 * @type {Function}
+	 * @private
+	 */
+	cbAlertThumb: PropTypes.func
+};
 
-		return <UiScrollThumb {...props} />;
-	}
-}
+ScrollThumb.defaultProps = {
+	cbAlertThumb: nop
+};
 
 export default ScrollThumb;
 export {
--- a/packages/moonstone/Scrollable/ScrollThumb.js
+++ b/packages/moonstone/Scrollable/ScrollThumb.js
@@ -13,12 +13,9 @@ const nop = () => {};
  * @ui
  * @private
  */
-const ScrollThumb = forwardRef((props, ref) => {
-	const rest = Object.assign({}, props);
-	delete rest.cbAlertThumb;
-
+const ScrollThumb = forwardRef(({cbAlertThumb, ...rest}, ref) => {
 	useEffect (() => {
-		props.cbAlertThumb();
+		cbAlertThumb();
 	});
 
 	return <UiScrollThumb {...rest} ref={ref} />;
--- a/packages/moonstone/Scrollable/ScrollThumb.js
+++ b/packages/moonstone/Scrollable/ScrollThumb.js
@@ -1,6 +1,6 @@
 import {ScrollThumb as UiScrollThumb} from '@enact/ui/Scrollable/Scrollbar';
 import PropTypes from 'prop-types';
-import React, {useEffect, forwardRef} from 'react';
+import React, {forwardRef, useEffect} from 'react';
 
 const nop = () => {};
 
--- a/packages/moonstone/Scrollable/Scrollbar.js
+++ b/packages/moonstone/Scrollable/Scrollbar.js
@@ -1,7 +1,7 @@
 // import ApiDecorator from '@enact/core/internal/ApiDecorator';
 import {ScrollbarBase as UiScrollbarBase} from '@enact/ui/Scrollable/Scrollbar';
 import PropTypes from 'prop-types';
-import React, {forwardRef, useImperativeHandle, useRef} from 'react';
+import React, {forwardRef, memo, useImperativeHandle, useRef} from 'react';
 
 import ScrollButtons from './ScrollButtons';
 import ScrollThumb from './ScrollThumb';
@@ -18,7 +18,7 @@ import componentCss from './Scrollbar.module.less';
  * @ui
  * @private
  */
-const ScrollbarBase = forwardRef((props, ref) => {
+const ScrollbarBase = memo(forwardRef((props, ref) => {
 	// Refs
 	const scrollbarRef = useRef();
 	const scrollButtonsRef = useRef();
@@ -66,7 +66,7 @@ const ScrollbarBase = forwardRef((props, ref) => {
 			)}
 		/>
 	);
-});
+}));
 
 ScrollbarBase.displayName = 'ScrollbarBase';
 
--- a/packages/moonstone/Scrollable/Scrollbar.js
+++ b/packages/moonstone/Scrollable/Scrollbar.js
@@ -19,8 +19,11 @@ import componentCss from './Scrollbar.module.less';
  * @private
  */
 const ScrollbarBase = forwardRef((props, ref) => {
+	// Refs
 	const scrollbarRef = useRef();
 	const scrollButtonsRef = useRef();
+	// render
+	const {cbAlertThumb, clientSize, corner, vertical, ...rest} = props;
 
 	useImperativeHandle(ref, () => {
 		const {getContainerRef, showThumb, startHidingThumb, update: uiUpdate} = scrollbarRef.current;
@@ -37,9 +40,7 @@ const ScrollbarBase = forwardRef((props, ref) => {
 			},
 			focusOnButton
 		};
-	}, []);
-
-	const {cbAlertThumb, clientSize, corner, vertical, ...rest} = props;
+	}, [scrollbarRef, scrollButtonsRef]);
 
 	return (
 		<UiScrollbarBase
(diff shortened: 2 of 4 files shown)


packages/moonstone/Scrollable/ScrollThumb.js
```diff
@@ -1,45 +1,44 @@
...
-
-	render () {
-		const props = Object.assign({}, this.props);
-
-		delete props.cbAlertThumb;
-
-		return <UiScrollThumb {...props} />;
-	}
-}
+const ScrollThumb = forwardRef((props, ref) => {
```
ybsung: What about `{cbAlertThumb, ...rest}` instead of the `props` above? Then we could remove the following two lines.

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -29,165 +44,130 @@ const setCSSVariable = (element, variable, value) => {
...
-	hideThumb = () => {
-		ReactDOM.findDOMNode(this.thumbRef.current).classList.remove(this.props.css.thumbShown); // eslint-disable-line react/no-find-dom-node
-	}
-
-	hideThumbJob = new Job(this.hideThumb.bind(this), thumbHidingDelay);
-
-	calculateMetrics = () => {
-		const primaryDimenstion = this.props.vertical ? 'clientHeight' : 'clientWidth';
-		let trackSize;
+const ScrollbarBase = forwardRef((props, ref) => {
```
ybsung: For `PureComponent`, please replace it with `React.memo`.

Ref: https://www.peterbe.com/plog/memo-instead-of-purecomponent

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -14,6 +14,21 @@ const
 	minThumbSize = 18, // Size in pixels
 	thumbHidingDelay = 400; // in milliseconds
 
+function useHidingThumbJob (callback, timeout) {
+	const [state] = React.useState({});
+	state.job = state.job || new Job(callback, timeout);
```
ybsung: My recommendation is...
```
const hideThumbJob = useRef(new Job(callback, timeout));
```
It would be better to store the instance  in not `useState` but `useRef` because we don't change it.

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -29,165 +44,130 @@ const setCSSVariable = (element, variable, value) => {
...
+
+	function hideThumb () {
+		addRemoveClass(thumbRef.current, css.thumbShown, 'remove');
+	}
+
+	useImperativeHandle(ref, () => ({
+		getContainerRef: () => (containerRef),
+		showThumb: () => {
+			hideThumbJob.stop();
+			addRemoveClass(thumbRef.current, css.thumbShown, 'add');
```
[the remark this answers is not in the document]
ybsung: I have the same opinion with @jeonghee27 .
Anyway, we already have the plan to remove the `findDOMNode`. So let's refactor this line later.

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -29,165 +44,130 @@ const setCSSVariable = (element, variable, value) => {
...
+		update: (bounds) => {
+			const
+				{clientSize} = props,
+				primaryDimenstion = vertical ? 'clientHeight' : 'clientWidth',
+				trackSize = clientSize ? clientSize[primaryDimenstion] : containerRef.current[primaryDimenstion],
+				scrollViewSize = vertical ? bounds.clientHeight : bounds.clientWidth,
+				scrollContentSize = vertical ? bounds.scrollHeight : bounds.scrollWidth,
+				scrollOrigin = vertical ? bounds.scrollTop : bounds.scrollLeft,
+				thumbSizeRatioBase = (scrollViewSize / scrollContentSize),
+				scrollThumbPositionRatio = (scrollOrigin / (scrollContentSize - scrollViewSize)),
```
ybsung: `minThumbSizeRatio = ri.scale(minThumbSize) / trackSize;` code might be needed here before using it.

ybsung: You might miss it.

ybsung: Oh I misunderstood it. Please never mind my comments above.

By the way, could we use `React.useMemo` for the complex calculation above?
Ref: https://reactjs.org/docs/hooks-reference.html#usememo

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -29,165 +44,130 @@ const setCSSVariable = (element, variable, value) => {
...
+			setCSSVariable(thumbRef.current, '--scrollbar-size-ratio', scrollThumbSizeRatio);
+			setCSSVariable(thumbRef.current, '--scrollbar-progress-ratio', scrollThumbPositionRatio);
 		}
+	}));
+
+	return (
+		<div {...rest} className={containerClassName} ref={containerRef}>
+			{childRenderer({
+				thumbRef
+			})}
```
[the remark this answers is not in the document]
0x64: I will fix this.

packages/ui/Scrollable/Scrollbar.js
0x64: Good point. If this API returns _already computed_ values frequently, `useMemo` can help this to be optimized.
And, IMO, this API is not so good to use `useMemo` with following reasons;
- This API is called on every scrolling by a browser to update thumb, and the scroll position can be *any number* between 0 and the maximum scroll position. Since it looks that there is no limitation to memoize values for now, memoized values can be a set of thousands or more.
- Even if there is a limitation, we cannot expect optimization in general except some frequently positioned cases like the beginning, the ending, and some scroll positions by 5-way spotlight moving.
- This API is called only the scroll position is changed in Scrollables. When just showing and hiding a thumb, this API is not called at all.

packages/ui/Scrollable/Scrollbar.js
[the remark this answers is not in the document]
0x64: I thought it could be better for refactoring later to remove `findDOMNode`. I will revert this part. :)

packages/ui/Scrollable/Scrollbar.js
0x64: Okay. (not to replace but to add)

packages/ui/Scrollable/Scrollbar.js
0x64: I don't think we are safe to assume that `callback` will not be changed ever. Also, `Job` related code may be refined later also later.

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -1,7 +1,7 @@
 import classNames from 'classnames';
 import {Job} from '@enact/core/util';
 import PropTypes from 'prop-types';
-import React, {PureComponent, Component} from 'react';
+import React, {forwardRef, useRef, useImperativeHandle} from 'react';
```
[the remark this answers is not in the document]
0x64: 👍

packages/moonstone/Scrollable/ScrollThumb.js
0x64: Oh, right.

packages/ui/Scrollable/Scrollbar.js
0x64: Add)
When I tested, `callback` is not changed actually. Also, this hook is not so reusable nor helpful for the separation of concerns. So I  did put this back into Scrollbar with `useRef`.

0x64:
@jeonghee27, I updated the PR.

0x64:
**This PR is ONLY for review for now.**

packages/moonstone/Scrollable/Scrollbar.js
```diff
@@ -18,131 +18,121 @@ import componentCss from './Scrollbar.module.less';
...
+			getContainerRef,
+			showThumb,
+			startHidingThumb,
+			uiUpdate,
+			isOneOfScrollButtonsFocused,
+			update: (bounds) => {
+				updateButtons(bounds);
+				uiUpdate(bounds);
+			},
+			focusOnButton
```
ybsung: When creating a new PR, it would be better to order the properties alphabetically.

packages/moonstone/Scrollable/Scrollbar.js
```diff
@@ -18,131 +18,121 @@ import componentCss from './Scrollbar.module.less';
...
+const ScrollbarBase = memo(forwardRef((props, ref) => {
+	// Refs
+	const scrollbarRef = useRef();
+	const scrollButtonsRef = useRef();
+	// render
+	const {cbAlertThumb, clientSize, corner, vertical, ...rest} = props;
+
+	useImperativeHandle(ref, () => {
+		const {getContainerRef, showThumb, startHidingThumb, update: uiUpdate} = scrollbarRef.current;
+		const {isOneOfScrollButtonsFocused, updateButtons, focusOnButton} = scrollButtonsRef.current;
```
ybsung: Alphabetically later.

packages/moonstone/Scrollable/Scrollbar.js
```diff
@@ -18,131 +18,121 @@ import componentCss from './Scrollbar.module.less';
...
+
+	return (
+		<UiScrollbarBase
+			corner={corner}
+			clientSize={clientSize}
+			css={componentCss}
+			ref={scrollbarRef}
+			vertical={vertical}
+			childRenderer={({thumbRef}) => ( // eslint-disable-line react/jsx-no-bind
+				<ScrollButtons
```
ybsung: I think it would be better `() => { return ....;}` instead of `() => ()` to debug.
It is not related with this PR. But I hope we could do it this time.

packages/moonstone/Scrollable/Scrollbar.js
```diff
@@ -18,131 +18,121 @@ import componentCss from './Scrollbar.module.less';
...
+					{...rest}
+					ref={scrollButtonsRef}
+					vertical={vertical}
+					thumbRenderer={() => ( // eslint-disable-line react/jsx-no-bind
+						<ScrollThumb
```
ybsung: Same above.

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -29,165 +37,136 @@ const setCSSVariable = (element, variable, value) => {
...
+	function hideThumb () {
+		removeClass(thumbRef.current, css.thumbShown);
+	}
+
+	useEffect(() => {
+		const job = hideThumbJob.current;
+		return () => {
+			job.stop();
+		};
+	}, []);
```
ybsung: Where do we use the `job` variable?

packages/ui/Scrollable/Scrollbar.js
```diff
@@ -197,53 +176,53 @@ class ScrollbarBase extends PureComponent {
...
-
-			this.getContainerRef = getContainerRef;
-			this.showThumb = showThumb;
-			this.startHidingThumb = startHidingThumb;
-			this.update = uiUpdate;
+const Scrollbar = forwardRef((props, ref) => {
+	const scrollbarBaseRef = useRef(null);
+
+	useImperativeHandle(ref, () => {
+		if (scrollbarBaseRef.current) {
```
ybsung: If `scrollbarBaseRef.current` is equal to `null`, then what about returning {} object? Because the `current` of the scrollbar ref could be accessed.

packages/moonstone/Scrollable/Scrollbar.js
0x64: 👌

packages/moonstone/Scrollable/Scrollbar.js
0x64: 👌

packages/moonstone/Scrollable/Scrollbar.js
0x64: 👌

packages/moonstone/Scrollable/Scrollbar.js
0x64: 👌

packages/ui/Scrollable/Scrollbar.js
0x64: I will remove this. (a missing piece from my clean up).

packages/ui/Scrollable/Scrollbar.js
0x64: Maybe we don't need to test its validity at all. I will remove _if_ statement for now. If we need to put it back, then we can consider returning an empty object.

0x64:
Closed in favor of #2676(for moonstone) and #2677(for ui).

PR #2670, opened 2020-01-14
PLAT-98204: Resolve dependency problem between Moonstone Scroller React components

jeonghee27:
### Checklist

Resolve a problem about component method accessing using ref.
This problem cause by PLAT-98201.

### Resolution

We have converted 4 moonstone React components to functional component on PLAT-98201.
But there was many dependency on class instance. 
(ex, Scrollable call shouldPreventScrollByFocus() in  virtualListBase class using childRef.current.shouldPreventScrollByFocus())
It cause many console error, and some operations did not work.
I  used `useImperativeHandle` to access to functional component method on Moonstone functional component.

1. Use `useImperativeHandle`
2. Solve binding problem on `Scrollable`. 
3. Avoid a error of empty ref on initial loading. 

### Additional Considerations

I avoid initial empty ref problem by checking null on Scrollable.
Operation seems there is no error now, but I left a annotation using TODO.

### Links

### DIFF
--- a/packages/moonstone/VirtualList/VirtualListBase.js
+++ b/packages/moonstone/VirtualList/VirtualListBase.js
@@ -7,7 +7,7 @@ import {Spottable, spottableClass} from '@enact/spotlight/Spottable';
 import {VirtualListBase as UiVirtualListBase, VirtualListBaseNative as UiVirtualListBaseNative} from '@enact/ui/VirtualList';
 import PropTypes from 'prop-types';
 import clamp from 'ramda/src/clamp';
-import React, {useEffect, useRef} from 'react';
+import React, {useEffect, useImperativeHandle, useRef} from 'react';
 import warning from 'warning';
 
 import {Scrollable, dataIndexAttribute} from '../Scrollable';
@@ -44,7 +44,7 @@ const
 const VirtualListBaseFactory = (type) => {
 	const UiBase = (type === JS) ? UiVirtualListBase : UiVirtualListBaseNative;
 
-	const VirtualListCore = (props) => {
+	let VirtualListCore = (props, refernce) => {
 		/* No displayName here. We set displayName to returned components of this factory function. */
 
 		// Instance variables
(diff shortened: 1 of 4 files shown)


packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -44,7 +44,7 @@ const
 const VirtualListBaseFactory = (type) => {
 	const UiBase = (type === JS) ? UiVirtualListBase : UiVirtualListBaseNative;
 
-	const VirtualListCore = (props) => {
+	let VirtualListCore = (props, refernce) => {
```
MikyungKim: typo.. ?
I assume reference..?

jeonghee27: it's typo. I'll fix.

0x64:
About `scrollByPageOnPointerMode`...
This function is called in the global scope to figure out which `Scrollable` should handle a page up/down key input in pointer mode. To do this, every `Scrollable` instance was registered to the global `Map` named `Scrollable` with an instance reference of itself as a key and a content area element as a value. So, in a `useEffect` of `ScrollableBase` component, we actually don't need to put something replacement for `this` as a key but need something referable object that contains `scrollByPageOnPointerMode` function. Since the referable object is also required to delete an entry from `scrollable` global `Map` in a return function of `useEffect`, I recommend defining const object variable inside `useEffect`.

0x64:
For warnings related `createOverscrollJob`, what about put the function into `useEffect`?

0x64:
For warnings on a missing dependency of `props`,  we may update argument of `configureSpotlightContainer` function with really needed properties.

0x64:
Note: `calculatePositionOnFocus` function of `Scroller` should be exposed to `Scrollable`.

jeonghee27:
@0x64 I fixed all lint warning as possible you mentioned. please review again.

0x64:
LGTM

packages/moonstone/VirtualList/VirtualListBase.js
```diff
@@ -64,6 +64,17 @@ const VirtualListBaseFactory = (type) => {
...
+		const pause = variables.current.pause;
+
+		useImperativeHandle(reference, () => ({
+			calculatePositionOnFocus,
+			focusByIndex,
+			shouldPreventScrollByFocus,
+			shouldPreventOverscrollEffect,
+			setLastFocusedNode,
+			getScrollBounds,
+			setContainerDisabled
```
ybsung: `focusOnNode` might be needed here.

ybsung:
@jeonghee27 
Some changed function names would be conflicted in my PR. I'll approve your PR. Then let's talk about the naming in my PR.

PR #2672, opened 2020-01-15
PLAT-98876: Change `ui/Button` of ARIA structure to support touch accessibility

Baekwoo-s:
### Checklist

- Touching the content area of the button ignores the button role and only reads the content.
- Clicking the button (double tap for GMRSI VVO) reads the button's contents.

### Resolution
- Put aria-hidden in the button's content area, and put the content in aria-label. If there is a react element other than string in the content like Moonstone, it is excluded(Line 165).

### Additional Considerations

### DIFF
--- a/packages/ui/Button/Button.js
+++ b/packages/ui/Button/Button.js
@@ -161,6 +161,9 @@ const ButtonBase = kind({
 	},
 
 	computed: {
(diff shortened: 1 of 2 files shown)


seunghoh:
Close this PR as we don't need this workaround. Web engine fixed.

PR #2673, opened 2020-01-15
PLAT-98974: Remove moonstone from enact

germboy:
### Checklist

This PR:
- removes moonstone from enact
- removes moonstone samples from the sampler
- removes qa-sampler (since all were moonstone samples)

### Additional Considerations
We'll want to migrate the moonstone samples to the moonstone repo. This will be done in a separate ticket.

I've removed the moonstone components from sampler, but have left the ui components. We used `MoonstoneDecorator` to wrap the sampler by default, so I have removed this. This results in a lack of proper styling for the sampler. Also, as the ui components are unstyled, I'm not sure the sampler is too terribly helpful here. I feel we'd want to use sampler for enact, but I'd like some feedback regarding whether we should keep the sampler (and ui/platform samples) and how we'd like to style it (and the UI components).


teckliew:
Ran the sampler and it looks like the background should be reset to white? Maybe could style storybook a bit to make things more readable.

PR #2678, opened 2020-01-21
PLAT-98506: [VirtualList and Scroller] Convert from UI React component to functional component

jeonghee27:
### Checklist

We want to migrate HoC and render props to hooks for all VirtualList and Scroller.
As the first step, I converted 4 UI scroller components defined by class to functional component.
And then, resolved a problem about component method accessing using ref.

### Resolution

4 UI scroller components have been converted to functional component.
* Target component
UI/VirtualListBase
UI/Scrollable
UI/ScollableNative
UI/Scroller

1) Change `class NameOfComponent extends Component` to `const NameOfComponent => (props){`
2) Change `defaultProp`, `propTypes` to `NameOfComponent.defaultProp`,  `NameOfComponent.propTypes`. 
3) Remove the constructor
4) Remove the render() method, keep the return
5) Convert all methods to `function` for hoisting.
6) Use `useRef` for all instance variables. 
7) Replace all lifecycle functnion to `useEffect`. 
8) Use `useRef` or `useContext` As needed.
9) Remove all references to ‘this’ throughout the component
10) Set initial state with `useState()`. ie…
11) Declare some undeclared member variables. (this.variable)
12) Adjust duplicated variable names.
13) Resolve dependency problem between UI Scroller React components ( `useImperativeHandle`)

### Additional Considerations
* When I use useEffect,  I got lint warning from exhaustive-deps ESLint rule
(https://github.com/facebook/react/issues/14920).
I checked the operation, but lint-warning is occurring.
More discussion is needed to solve this problem.

### Links
PLAT-98506
PLAT-98204

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -15,7 +15,7 @@ import Registry from '@enact/core/internal/Registry';
 import {Job} from '@enact/core/util';
 import PropTypes from 'prop-types';
 import clamp from 'ramda/src/clamp';
-import React, {Component} from 'react';
+import React, {useContext, useState, useReducer, useRef, useEffect} from 'react';
 
 import ForwardRef from '../ForwardRef';
 import {ResizeContext} from '../Resizable';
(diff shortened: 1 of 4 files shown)


packages/ui/Scroller/Scroller.js
```diff
@@ -27,128 +27,95 @@ import css from './Scroller.module.less';
...
-
-	componentDidMount () {
-		this.calculateMetrics();
-	}
-
-	componentDidUpdate (prevProps) {
-		this.calculateMetrics();
-		if (this.props.isVerticalScrollbarVisible && !prevProps.isVerticalScrollbarVisible) {
-			this.forceUpdate();
+let ScrollerBase = (props, reference) => {
```
0x64: What about applying `forwardRef` here rather than re-allocate `ScrollerBase`?

packages/ui/Scroller/Scroller.js
```diff
@@ -27,128 +27,95 @@ import css from './Scroller.module.less';
...
+	useEffect(() => {
+		// componentDidUpdate
+		forceUpdate();
+	}, [props.isVerticalScrollbarVisible]);
+
+	useEffect(() => {
+		// componentDidUpdate
+		calculateMetrics();
+	});
+	// }
```
0x64: Clean up please.

packages/ui/Scroller/Scroller.js
```diff
@@ -27,128 +27,95 @@ import css from './Scroller.module.less';
...
-		if (this.props.isVerticalScrollbarVisible && !prevProps.isVerticalScrollbarVisible) {
-			this.forceUpdate();
+let ScrollerBase = (props, reference) => {
+	// constructor (props) {
+	const containerRef = useRef();
+	const [, forceUpdate] = useReducer(x => x + 1, 0);
+
+	useEffect(() => {
+		// componentDidMount
+		calculateMetrics();
```
0x64: We have already `useEffect` without dependency, so this effect is redundant.

packages/ui/Scroller/Scroller.js
```diff
@@ -27,128 +27,95 @@ import css from './Scroller.module.less';
...
+	}, []);
+
+	useEffect(() => {
+		// componentDidUpdate
+		forceUpdate();
```
0x64: This code introduced from #1618 and there is a possibility to cover the original issue in another way now. Until we have an alternative solution, let's keep this for now.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -48,446 +47,322 @@ const itemSizesShape = PropTypes.shape({
...
-
-		static defaultProps = {
-			cbScrollTo: nop,
-			dataSize: 0,
-			direction: 'vertical',
-			overhang: 3,
-			pageScroll: false,
-			spacing: 0
-		}
+		let nextState = null;
```
0x64: It looks better to move this definition into `if` statement.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -48,446 +47,322 @@ const itemSizesShape = PropTypes.shape({
...
+		}));
+
+		// getDerivedStateFromProps
+		useLayoutEffect(() => {
 			const
 				shouldInvalidate = (
-					state.prevFirstIndex === state.firstIndex ||
-					state.prevChildProps !== props.childProps
+					prevFirstIndex === firstIndex ||
+					prevChildProps !== props.childProps
```
0x64: This is not a bug of this PR, but let's reconsider whether this comparison is actually meaningful. In a normal render from Scrollable, `childProps` is re-generated.
And, it looks that `prevFirstIndex` and `prevChildProps` have a similar pattern; those values are accessed only in this effect. We can put those values into a single state object if we have a better outcome. Please consider this.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -48,446 +47,322 @@ const itemSizesShape = PropTypes.shape({
...
-		constructor (props) {
-			let nextState = null;
+		const containerRef = useRef();
+		const contentRef = useRef();
+		const itemContainerRef = useRef();
 
-			super(props);
+		const [prevChildProps, setPrevChildProps] = useState(null);
+		const [prevFirstIndex, setPrevFirstIndex] = useState(0);
+		const [updateFrom, setUpdateFrom] = useState(0);
```
0x64: `updateFrom` and `updateTo` are a pair of values. They are updated and consumed always at the same time. If we want to reduce a number of states, we can put them in a single state.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -48,446 +47,322 @@ const itemSizesShape = PropTypes.shape({
...
-				numOfItems: 0,
-				prevChildProps: null,
-				prevFirstIndex: 0,
-				updateFrom: 0,
-				updateTo: 0,
-				...nextState
-			};
+		if (props.clientSize) {
+			calculateMetrics(props);
+			nextState = getStatesAndUpdateBounds();
```
0x64: I'm just curious why the name of this variable is not `statesAndUpdateBounds` since you named like that in other functions.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -48,446 +47,322 @@ const itemSizesShape = PropTypes.shape({
...
+		}, [props, firstIndex, numOfItems]);
 
 		// Calculate metrics for VirtualList after the 1st render to know client W/H.
-		componentDidMount () {
-			if (!this.props.clientSize) {
-				this.calculateMetrics(this.props);
+		useEffect(() => {
+			// componentDidMount
+			if (!props.clientSize) {
+				calculateMetrics(props);
```
0x64: These four lines are repeating three more times. :)

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -48,446 +47,322 @@ const itemSizesShape = PropTypes.shape({
...
+			state: {
+				numOfItems
+			},
+			get scrollPosition () {
+				return variables.current.scrollPosition;
+			},
+			scrollToPosition,
+			setScrollPosition,
+			syncClientSize
+		}));
```
0x64: Is it helpful to set up dependencies here to avoid re-creation?

packages/ui/Scrollable/Scrollable.js
```diff
@@ -64,710 +64,499 @@ const TouchableDiv = ForwardRef({prop: 'ref'}, Touchable('div'));
...
+	// state member.
+	const enqueueForceUpdate = useCallback(() => {
+		variables.current.childRefCurrent.calculateMetrics(variables.current.childRefCurrent.props);
+		forceUpdate();
+	}, []);
+
+	const handleResizeWindow = useCallback(() => {
+		// `handleSize` in `ui/resolution.ResolutionDecorator` should be executed first.
+		setTimeout(() => {
+			if (handleResizeWindow) {
```
0x64: In the original code, there are two `handleResizeWindow` functions. One is from props and the other one is a member function of a component. We need to check the existence and call `props.handleResizeWindow` here. (We need to put it into dependency array, too)

packages/ui/Scrollable/Scrollable.js
```diff
@@ -64,710 +64,499 @@ const TouchableDiv = ForwardRef({prop: 'ref'}, Touchable('div'));
...
 			ev.preventDefault();
 			ev.stopPropagation();
 		} else {
 			const
-				{verticalScrollbarRef, horizontalScrollbarRef} = this,
-				bounds = this.getScrollBounds(),
-				canScrollHorizontally = this.canScrollHorizontally(bounds),
-				canScrollVertically = this.canScrollVertically(bounds),
+				bounds = getScrollBounds(),
+				bCanScrollHorizontally = canScrollHorizontally(bounds),
```
0x64: I don't think the naming convention is familiar to members.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -64,710 +64,499 @@ const TouchableDiv = ForwardRef({prop: 'ref'}, Touchable('div'));
...
+		variables.current.lastInputType = 'pageKey';
 
-		this.scrollToAccumulatedTarget(pageDistance, canScrollVertically, this.props.overscrollEffectOn.pageKey);
+		scrollToAccumulatedTarget(pageDistance, bCanScrollVertically, props.overscrollEffectOn.pageKey);
 	}
 
-	onKeyDown = (ev) => {
-		if (this.props.onKeyDown) {
-			forward('onKeyDown', ev, this.props);
+	const onKeyDown = useCallback((ev) => {
```
0x64: No dependency means `props.onKeyDown` will not be changed ever since the callback function is memoized. I'm wondering this is intended.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -984,146 +772,132 @@ class ScrollableBase extends Component {
...
-			this.scroll(targetX, targetY, targetX, targetY);
-			this.stop();
-		}
+		};
 	}
 
-	scroll = (left, top, ...rest) => {
-		if (left !== this.scrollLeft) {
-			this.setScrollLeft(left);
+	function scroll (left, top, ...restParam) {
```
0x64: What about `restParams`? :)

packages/ui/Scrollable/Scrollable.js
```diff
@@ -64,710 +64,499 @@ const TouchableDiv = ForwardRef({prop: 'ref'}, Touchable('div'));
...
-		maxTop: 0,
-		maxLeft: 0
-	}
+	});
 
-	// wheel/drag/flick info
-	wheelDirection = 0
-	isDragging = false
+	const onMouseDown = useCallback((ev) => {
+		forwardWithPrevent('onMouseDown', ev, props);
```
0x64: This callback is converted from `handle` chains, so you must check the return value from `forwardWithPrevent` to determine the execution flow should be stopped or not.

packages/ui/Scrollable/ScrollableNative.js
```diff
@@ -50,603 +50,401 @@ const TouchableDiv = Touchable('div');
...
+		childRefCurrent.scrollToPosition(variables.current.scrollLeft + 0.1, variables.current.scrollTop + 0.1);
+		childContainerRef.current.style.scrollBehavior = 'smooth';
+	});
 
-	onMouseDown = handle(
-		forwardWithPrevent('onMouseDown'),
-		this.stop
-	).bindAs(this, 'onMouseDown')
+	const onMouseDown = useCallback((ev) => {
+		forwardWithPrevent('onMouseDown', ev, props);
```
0x64: Also, the return value should be checked before proceeding.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -64,710 +64,499 @@ const TouchableDiv = ForwardRef({prop: 'ref'}, Touchable('div'));
...
 		} else if (deltaMode === 2) { // page
 			delta = delta < 0 ? -maxPixel : maxPixel;
 		}
 
 		return delta;
 	}
 
-	onWheel = (ev) => {
-		if (this.isDragging) {
+	const onWheel = useCallback((ev) => {
```
0x64: I found `props` inside this callback and there is no dependency for `props`.

packages/ui/Scroller/Scroller.js
jeonghee27: I just added description here. I'll check later this code is needed still or not.

packages/ui/VirtualList/VirtualListBase.js
jeonghee27: I pursued short name. you're right. I fixed.

packages/ui/VirtualList/VirtualListBase.js
jeonghee27: I changed it to a object `updateFromTo`.

packages/ui/VirtualList/VirtualListBase.js
jeonghee27: That's right.
But I need to add all the function names to dep. This array could be very long.
I thought it was easier to read and safer to leave empty.

packages/ui/Scroller/Scroller.js
jeonghee27: I fixed `scroller`, `scrollable`, `scrollableNative`.
But, in case of `VirtuaListBase`, `VirtualListCore function` is located in `VirtualListBaseFactory function`.
So I cannot fix the `VirtuaListBase` due to lint error  "React Hook \"useRef\" cannot be called inside a callback. React Hooks must be called in a React function component or a custom React Hook function.", 

I think we need to use curry instead of  `VirtualListBaseFactory function` .
I hope we could do that on PLAT-98957

packages/ui/VirtualList/VirtualListBase.js
jeonghee27: > let's reconsider whether this comparison is actually meaningful.
I couldn't trim that logic currently. so I just left description on this code. 

>  it looks that prevFirstIndex and prevChildProps have a similar pattern; ...
I don't think these variables have to be state. I moved these variable to `variables.current`.

packages/ui/Scrollable/Scrollable.js
jeonghee27: It was my mistake. I fixed it. Thank you.

packages/ui/Scrollable/Scrollable.js
jeonghee27: As far as I know, No dependency means `props.onKeyDown` will be changed everytimes. it is same with normal function. It is meanless useCallback.

I hope to fill the dep parameter properly. But, I need to set all functions that called inside.
and the inside function should be useCallback function too.

and there are many connections in the dep connection.
I hope to solve this part onrestructuring with `// TODO: Handle exhaustive-deps`

packages/ui/Scrollable/Scrollable.js
jeonghee27: Good point. I fixed it.

packages/ui/Scrollable/Scrollable.js
jeonghee27: I added `props` for dependency. Is this what you mean?

ybsung:
LGTM

PR #2679, opened 2020-01-21
PLAT-97858: Migrate ui/Toggleable to use hooks

ryanjduffy:
### Checklist

First step in migrating our HOCs to composable hooks to reduce component layers and improve readability through less "prop magic" from the HOC stacks.

### Resolution
* Add `useControlledState` hook to standardize controlled state mgmt
* Add `useToggle` hook, adopting `useControlledState`, to support toggleable behavior
* Add `core/handle.not` to simply complementary conditions (e.g. prop is falsy checks)
* Adopt `ui/Toggleable` to use `useToggle`

### Additional Considerations
Need to determine documentation approach for hooks. For the time being, it is private API so we can defer that for a little while if we choose.


packages/ui/Toggleable/tests/Toggleable-specs.js
```diff
@@ -290,7 +290,7 @@ describe('Toggleable', () => {
 				<Component defaultSelected />
 			);
 
-			subject.find(DivComponent).prop('onToggle')();
+			subject.find(DivComponent).invoke('onToggle')();
```
ryanjduffy: `invoke` is a newer method that is required when calling a method that produces a state change via hooks. We need to update CLI to use a more recent enzyme and react adapter for this feature.

packages/ui/Toggleable/tests/Toggleable-specs.js
```diff
@@ -149,11 +149,11 @@ describe('Toggleable', () => {
 		);
 
 		test(
-			'should use selected prop when selected changed from truthy to undefined',
+			'should use defaultSelected prop when selected changed from truthy to undefined',
```
ryanjduffy: I've changed the behavior intentionally so it matches the test above in which `Toggleable` uses `defaultSelected` when `selected` is `null` at mount time.

webOS101: This seems like a pretty significant change.  Do we want to allow toggling from controlled to uncontrolled?

ryanjduffy: That's not really what is happening here. It would still be considered controlled because the initial state included the `selected` prop. The change is only that for Toggleable (per the test above), `null` is treated as "unset" so it reverts to use the default value in that case.

webOS101: It seems like that's exactly what's happening here.  The component, already mounted, reverts to `defaultSelected` which should not apply for a controlled component, after `selected` becomes `undefined`.  Am I misreading the test?

ryanjduffy: This is my mistake. I commented on the wrong test initially. It was intended for the change above.

To your point, this is a notable change that we should discuss further. Sorry for the confusion I caused!

packages/ui/Toggleable/Toggleable.js
```diff
@@ -133,145 +135,116 @@ const ToggleableHOC = hoc(defaultConfig, (config, Wrapped) => {
...
+	const forwardActivate = forwardWithEventProps(activate);
+	const forwardDeactivate = forwardWithEventProps(deactivate);
+	const forwardToggle = forwardWithEventProps(toggle);
+	const forwardToggleProp = forwardWithEventProps(toggleProp);
+
+	function Toggleable (props) {
+		const updated = {...props};
+		const hook = useToggle({
+			disabled: props.disabled,
+			// FIX: Current behavior for Toggleable treats `null` as undefined so we coerce it here
```
Djspaceg: _Should_ this behavior change? In terms of something that toggles, I'd imagine that `false`, `null`, and `undefined` would all mean the same thing. What are your thoughts on this that didn't fit into this code-comment?

ryanjduffy: I believe it should but can't within the confines of 3.x. We can discuss separately to what degree we want to maintain backward compatibility in order to deliver this work within the current major release.

Djspaceg: Let's definitely discuss soon. Something to think about between now and then, and something I'm really curious about, is how you'd approach an omission of a prop in JSX if this behavior were to change. For example
`<Check />` or `<Button selected={undefined} />`. Would this no longer be allowed?

webOS101:
Using `handle.not()` obviates the need for an explicit `false`.  We could remove that `defaultProps` entry, I believe.

packages/ui/Toggleable/useToggle.js
```diff
@@ -0,0 +1,77 @@
...
+
+	toggle.setContext(...state);
+
+	return {
+		activate: toggle.handleActivate,
+		deactivate: toggle.handleDeactivate,
+		toggle: toggle.handleToggle,
+		selected: toggle.value
+	};
+}
```
ybsung: Hello @ryanjduffy 

Have you ever considered using useEvent in 'react-dom'?

The useEvent is in the master branch.
https://github.com/facebook/react/blob/master/packages/react-dom/src/client/ReactDOM.js#L217
But it is not included in any release yet. So it could be too early to apply it to our libraries.

I haven't executed any code with the useEvent. But according to the useEvent API, it seems to me that we don't have to return any event handlers in a custom hook with the useEvent to pass them to a DOM element if we could receive the ref for the DOM element.
So the `useToggle` could be

```
function useToggle (ref, config = {}) {
	const toggle = useClass(Toggle, ref, config);
	const state = useControlledState(
		config.defaultSelected,
		config.selected,
		typeof config.selected !== 'undefined'
	);

	toggle.setContext(...state);

	// no return values
}
```

Then the `toggle` could be
```
class Toggle {
	constructor (ref, config) {
		const {activate, deactivate, eventProps, prop, toggle, toggleProp} = config;

		// remapping to props for better compatibility with core/handle and binding
		this.props = config;
		this.context = {};

		if (config.activate) {
			const forwardActivate = forwardWithEventProps(config.activate);
			const activate = ReactDOM.unstable_useEvent(activate);

			React.useEffect(() => {
				activate.setListener(ref.current, (ev) => {
					if (this.activate()) forwardActivate(ev, this.props);
				});
			});
		}
		...
	}
}
```

Please fix me If I understand incorrectly.

packages/ui/Toggleable/useToggle.js
```diff
@@ -0,0 +1,77 @@
...
+ * Creates one instance of the class, `Ctor` with the provided `args`, for the life of the
+ * component.
+ *
+ * @param {Function} Ctor Class constructor
+ * @param  {...any} args Arguments to pass to the constructor
+ * @private
+ */
+function useClass (Ctor, ...args) {
+	const ref = React.useRef(null);
+	ref.current = ref.current || new Ctor(...args);
```
ybsung: Could we combine 16 and 17 lines into `const ref = React.useRef(new Ctor(...args));`?

ryanjduffy: The reason for the separation is to avoid creating new classes. Functionally, I believe the result will be the same due to how `useRef` behaves but in your code sample above, a new instance of `Ctor` would be created on each render and then discarded.

packages/ui/Toggleable/useToggle.js
ryanjduffy: I'm not familiar with that API so I can't say for sure. It seems to only work for native events (due to the need for a ref) which would be challenging for our use cases (e.g. toggling on `onOpen`/`onClose`). It could be useful for managing internal event listeners for a component though so we'll have to keep an eye on it. Thanks for sharing!

webOS101: I agree with Ryan.  We don't know enough about consumers and aren't consistent enough with refs that we could be sure the events end up on the right element.  And, as Ryan says, many of the current use cases involve non-native events.

ybsung: Yes. It looks this API works with native events. Thank you for your opinions. Let's keep an eye on it.

PR #2685, opened 2020-02-06
PLAT-99881: Support item DOM element wrapper

MikyungKim:
### Checklist

In this PR, we'd like to support a DOM element wrapper for each list item for the following reasons.
There has been a desire to have an easier way to set the style of list items for both vertical and horizontal VirtualList from app developer side.
We need an approach that doesn't interfere with internal VirtualList logic, such as setting transform style to each list item.
Also, we, as VirtualList developers, had a concern that each app has a duty to pass props which are mandatory to make VirtualList work, such as `key`.

### Resolution
Among several feasible approaches mentioned in PLAT-86356, we've decided to add a DOM element wrapper to fulfill the above consideration.
We can explicitly hide internal prop and logic from app developers and also they can define styles more intuitively.

### Additional Considerations
While implementing, I've also considered un-nest the classes. So I moved `listItem` class out of `virtualList`. I couldn't un-nest others because others are used in the child of the DOM which has `virtualList`.
Also, I've considered rename `childProps` to `itemProps` for clarity.

### Links
PLAT-99881


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -316,7 +315,7 @@ class VirtualListBase extends Component {
 
 		return {
 			...nextUpdateFromAndTo,
-			prevChildProps: props.childProps,
+			prevItemProps: props.itemProps,
```
ybsung: `prevItemProps` would be after `prevFirstIndex` prop.

packages/ui/VirtualList/VirtualList.module.less
```diff
@@ -12,18 +12,18 @@
...
 		will-change: transform;
 	}
 }
 
+.listItem {
+	overflow: hidden;
+	width: 100%;
+
+	will-change: transform;
+}
```
ybsung: Why don't we add the `position: absolute` in the `listItem` class instead of the following inline style.

```
		composeStyle (width, height, primaryPosition, secondaryPosition) {
			const
				{x, y} = this.getXY(primaryPosition, secondaryPosition),
				style = {
					position: 'absolute',
```

MikyungKim: Good point!

ybsung:
LGTM

PR #2690, opened 2020-02-10
PLAT-100227: Fix VirtualList unit test fails

ybsung:
### Checklist

1> Enable VirtualList-specs.js test suits by replacing `test.skip()` with `test()`

2> Fix `enzyme` adapter issue

When executing ui/VirtualList unit tests, there were the following errors. So we called test.skip() instead test() in the `VirtualList-spec.js` before.

```
Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:
    1. You might have mismatching versions of React and the renderer (such as React DOM)
    2. You might be breaking the Rules of Hooks
    3. You might have more than one copy of React in the same app
```

By using the latest `enzyme` and `enzyme-adapter-react-16` and
```
"enzyme": "^3.11.0",
"enzyme-adapter-react-16": "^1.15.2"
```

executing the following command in the `VirtualList-specs.js`, those errors don't occur.
```
configure({adapter: new Adapter()});
```

3> Fix the null `ref` reference error

Sometimes `null` was passed to the parameter of the ui/Scrollable.utilEvent.addEventListener(). So this case is guarded.

### Additional Considerations

It seems that enact cli tool already used the `enzyme-adapter-react-16` module.(https://github.com/enactjs/cli/blob/develop/config/jest/setupTests.js#L43)
But the above error occurred. @ryanjduffy , @JayCanuck, are we missing something?

**Additional Fix**
1> Fix the case that some functions in `uiScrollableAdapter` used some cached mutable variables

So the `uiScrollableAdapter` is now refreshed  when ui/useScrollBase updates.

2> Provide the setter for `isDragging` in `uiScrollableAdapter`, which will be set in other modules in future.

### Links

PLAT-100227


0x64:
LGTM

JayCanuck:
I suspect the Enzyme issue may have been related to https://github.com/enactjs/cli/pull/203  I tried `develop` branch of CLI and the VirtualList unit tests pass successfully without the adapter setup on this PR. May want to try `develop` CLI branch.  Was waiting on the one last CLI PR before tagging, but given its a low priority one, I may just release/publish the updates today to get the pending fixes distributed.

PR #2691, opened 2020-02-11
PLAT-100231: Fix not to receive `onScroll` event if type is equal to `JS` in Scroller

ybsung:
### Checklist

When scrolling to the boundaries of the Scroller, the Scroller vibrates and it seems that the Scroller keeps scrolling.

### Resolution

Even though the type of the Scroller was 'JS', the Scroller received the 'onScroll' event and tried to scroll again like having the 'Native' type. So the Scroller is fixed not to receive the 'onScroll' event like before.

It's the side effect while unifying ui/Scrollable.js and ui/ScrollableNative.js into one custom hook.

### Additional Considerations

Clean up some code related with the JS and Native type in Scroller

### Links

PLAT-100231

### DIFF
--- a/packages/ui/Scrollable/Scrollable.js
+++ b/packages/ui/Scrollable/Scrollable.js
@@ -31,7 +31,10 @@ import utilEvent from './utilEvent';
 
 import css from './Scrollable.module.less';
 
-const nop = () => {};
+const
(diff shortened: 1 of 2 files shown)


0x64:
LGTM

PR #2693, opened 2020-02-11
PLAT-99880: Pass props not being used in component to the outermost div element

ybsung:
### Checklist

**This is an API breaking change**
: Any VirtualList or Scroller props which are not used in them, will be passed to the outer div element.

**Pass custom props to the outer div element by default**

Any custom props are passed to not the outer div element but the inner div element. So we handled the `style` prop being passed to the outer div element by force. From the perspective of a component, it would be better that any props not being handled in the component, would be passed to the outer div element.

### Additional Considerations

**1> Destructuring props once**

Some props are destructured from the `props` more than twice. It would be better to unify them and to destructure some props from the `props` once. By doing it, we could reduce the time executing `delete rest[propName]` from the `rest` variable.

https://github.com/enactjs/enact/blob/develop/packages/ui/Scrollable/Scrollable.js#L457
```
	const {
		decorateChildProps,
		horizontalScrollbarRef,
		setUiChildAdapter,
		type,
		uiChildAdapter,
		uiChildContainerRef,
		uiScrollContainerRef,
		verticalScrollbarRef
	} = props;
```

https://github.com/enactjs/enact/blob/develop/packages/ui/Scrollable/Scrollable.js#L641
```
	const
		{className, noScrollByDrag, rtl, style, ...rest} = props,
```

```
const {direction} = props;
```

**2> Use the destructuring assignment**

If possible, use the destructuring assignment instead of any prop with the `props`
AS-IS:
```
if (props.setUiScrollAdapter) {
...
}
```

TO-BE:
```
const {
   ...
   setUiScrollAdapter
} = props;

if (setUiScrollAdapter) {
```

NOTE: For the following props, I didn't destructure them from the `props` because there are the same naming functions and I don't want to rename them.

```
addEventListeners
applyOverscrollEffect
cbScrollTo
clearOverscrollEffect
clientSize
handleResizeWindow
onFlick
onKeyDown
onMouseDown
onScroll
onScrollStart
onScrollStop
onWheel
removeEventListeners
scrollStopOnScroll
scrollTo
start
stop
```
For example
```
const {
   ...
   onKeyDown: onKeyDownProp,
   ...
} = props;

	function onKeyDown (ev) {
		if (type === 'JS') {
			if (onKeyDownProp) { // instead of if (props.onKeyDown) {
```
: `props.onKeyDown` is more clear to understand the code. If it would be better to use the `onKeyDownProp` or another better name, please let me know.

### Links

PLAT-99880


MikyungKim:
Please check virtuallist-horizontal-scroll-in-scroller sample in qa-sampler.
There are some prop bleeding caused by this change.

MikyungKim:
Could we change the subject of this PR to something related to the original issue?
like, "Pass props not being used in component to the outermost div element"

MikyungKim:
We are still bleeding `clientSize` prop. Please check.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1834,10 +1832,27 @@ const useScrollBase = (props) => {
...
+			wrap
+		} :
+		{children};
+
 	decorateChildProps('childProps', {
-		...rest,
+		...childProps,
 		cbScrollTo: scrollTo,
 		className: [css.scrollFill],
+		clientSize,
```
MikyungKim: I thought you will move this to `childProps` for VirtualList.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -1834,10 +1832,27 @@ const useScrollBase = (props) => {
...
+
```
MikyungKim: Why do we have these 2 extra empty rows?

packages/ui/Scroller/Scroller.js
```diff
@@ -198,6 +198,7 @@ class ScrollerBase extends Component {
 			});
 
 		delete rest.cbScrollTo;
+		delete rest.clientSize;
```
MikyungKim: `Scroller` doesn't need to have this prop.

packages/ui/Scrollable/Scrollable.js
```diff
@@ -445,25 +445,70 @@ class ScrollableBase extends Component {
...
+		scrollClasses = classNames(css.scroll, className);
+
+	// The following props are the one having the same naming function in this scope.
+	// So it is better to use props[propName]
+	// instead of extracting it from the `props` and renaming it
+	delete rest.addEventListeners;
+	delete rest.applyOverscrollEffect;
+	delete rest.cbScrollTo;
+	delete rest.clearOverscrollEffect;
+	delete rest.clientSize;
```
MikyungKim: I think this line is redundant.

MikyungKim:
LGTM, but we need to sort out props/default props better of VL and Scroller.
Let's discuss separately.

PR #2694, opened 2020-02-12
PLAT-98240: Support to specify the item size of VirtualList

MikyungKim:
### Checklist

There were no default item size for horizontal VirtualList.

### Resolution
To support default item sizes for each direction, added `vertical` and `horizontal` classes to both JS and native list.
To specify css attributes only for native list, added `native` class.

### Additional Considerations
Moved out `vertical` and `horizontal` classes of `virtualList`.

### Links
PLAT-98240

### DIFF
--- a/packages/ui/VirtualList/VirtualList.module.less
+++ b/packages/ui/VirtualList/VirtualList.module.less
@@ -4,23 +4,32 @@
 	width: 100%;
 	overflow: hidden;
 
-	&.vertical {
-		overflow-y: scroll;
-	}
-
-	&.horizontal {
-		overflow-x: scroll;
-	}
-
 	.content {
 		will-change: transform;
+	}
+}
+
+.vertical {
+	&.native {
+		overflow-y: scroll;
+	}
+	.listItem {
+		width: 100%;
+	}
+}
+
+.horizontal {
+	&.native {
+		overflow-x: scroll;
+	}
+	.listItem {
+		height: 100%;
 	}
 }
 
 .listItem {
 	position: absolute;
 	overflow: hidden;
-	width: 100%;
 
 	will-change: transform;
 }
--- a/packages/ui/VirtualList/VirtualListBase.js
+++ b/packages/ui/VirtualList/VirtualListBase.js
@@ -1155,26 +1155,12 @@ class VirtualListBase extends Component {
 
 	// render
 
-	getContainerClasses (className) {
-		let containerClass = null;
-
-		if (this.props.type === 'Native') {
-			containerClass = this.isPrimaryDirectionVertical ? css.vertical : css.horizontal;
-		}
-
-		return classNames(css.virtualList, containerClass, className);
-	}
-
-	getContentClasses () {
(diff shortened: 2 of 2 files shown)


packages/ui/VirtualList/VirtualList.module.less
```diff
@@ -4,23 +4,32 @@
...
+		height: 100%;
 	}
 }
 
 .listItem {
 	position: absolute;
 	overflow: hidden;
-	width: 100%;
 
 	will-change: transform;
```
ybsung: I'm not sure that it is in the scope of this PR, but is the `will-change` needed for native list?

ybsung: Let's discus it in a separate issue.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -1155,26 +1155,12 @@ class VirtualListBase extends Component {
...
 	render () {
 		const
-			{className, 'data-webos-voice-focused': voiceFocused, 'data-webos-voice-group-label': voiceGroupLabel, 'data-webos-voice-disabled': voiceDisabled, itemsRenderer, style, ...rest} = this.props,
-			{cc, itemContainerRef, primary} = this,
-			containerClasses = this.getContainerClasses(className),
-			contentClasses = this.getContentClasses();
+			{className, 'data-webos-voice-focused': voiceFocused, 'data-webos-voice-group-label': voiceGroupLabel, 'data-webos-voice-disabled': voiceDisabled, itemsRenderer, style, type, ...rest} = this.props,
+			{cc, isPrimaryDirectionVertical, itemContainerRef, primary} = this,
+			containerClasses = classNames(css.virtualList, isPrimaryDirectionVertical ? css.vertical : css.horizontal, type === 'Native' ? css.native : null, className),
+			contentClasses = type === 'Native' ? null : css.content;
```
ybsung: `css.content` is only applied in 'JS' type VirtualList. So why don't we just add `css.native` or `css.js` depending on the type?

```
.vertical {
	&.native {
		overflow-y: scroll;
	}
	&.js > div {
		will-change: transform;
	}
	.listItem {
		width: 100%;
	}
}

containerClasses = classNames(css.virtualList, isPrimaryDirectionVertical ? css.vertical : css.horizontal, type === 'Native' ? css.native : css.js, className),
contentClasses = css.content;
```

In addition, we could also just remove the `contentClasses` variable.

packages/ui/VirtualList/VirtualList.module.less
MikyungKim: okay

packages/ui/VirtualList/VirtualListBase.js
MikyungKim: I don't like this idea because things are getting more ambiguous to readers.
Also, we need to declare `will-change` part to `horizontal` case too.

ybsung:
LGTM

PR #2695, opened 2020-02-13
PLAT-99657: Support padding in the scroll area in Scroller

ybsung:
### Checklist

According to the new GUI guideline, it is needed to add paddings between the scroll boundaries and the content in a VirtualList or a Scroller. In addition, if scrolling, then the content is also shown in the padding area.

Green area: padding
Blue area: content
image

### Resolution

- The content in the Scroller is wrapped with `<div className={css.contentWrapper}>` to display the `100%` width content properly.
- If the `padding` prop is passed to the VirtualList, then the item width is 100% - padding size.

### Additional Considerations

1. Update the JSDoc description of setUiChildAdapter and uiChildContainerRef.

2. We don't need the getter of horizontalScrollbarRef and verticalScrollbarRef because they are object and only their current property is used. I just pass themselves.
AS-IS
```
	const {
		...
	} = useScrollBase({
		...
		get horizontalScrollbarRef () {
			return horizontalScrollbarRef;
		},
		...
	};
```
TO-BE
```
	const {
		...
	} = useScrollBase({
		...
		horizontalScrollbarRef,
		...
	};
```

### Links

PLAT-99657

### Comments

- The scrollbar is clipped. But it will be fine after applying new GUI in the scrollbar.
- Item layout is broken. This issue will be handled in a separate issue.

### DIFF
--- a/packages/ui/VirtualList/VirtualList.module.less
+++ b/packages/ui/VirtualList/VirtualList.module.less
@@ -1,6 +1,4 @@
 .virtualList {
-	background: green;
-
 	position: relative;
 	height: 100%;
 	width: 100%;
@@ -44,8 +42,6 @@
 }
 
 .listItem {
-	background: red;
-
 	position: absolute;
 	overflow: hidden;
 
--- a/packages/ui/VirtualList/VirtualList.module.less
+++ b/packages/ui/VirtualList/VirtualList.module.less
@@ -15,7 +15,7 @@
 	}
 
 	.listItem {
-		width: calc(100% - 96px);
+		width: 100%;
 	}
 
 	&.padding {
@@ -31,7 +31,7 @@
 	}
 
 	.listItem {
-		height: calc(100% - 96px);
+		height: 100%;
 	}
 
 	&.padding {
--- a/packages/ui/VirtualList/VirtualList.module.less
+++ b/packages/ui/VirtualList/VirtualList.module.less
@@ -15,8 +15,15 @@
 	&.native {
 		overflow-y: scroll;
 	}
+
 	.listItem {
-		width: 100%;
+		width: calc(100% - 96px);
+	}
+
+	&.padding {
+		.listItem {
+			width: calc(100% - 96px);
+		}
 	}
 }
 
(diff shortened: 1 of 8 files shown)


packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -1191,7 +1211,14 @@ class VirtualListBase extends Component {
 		}
 
 		return (
-			<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} data-webos-voice-disabled={voiceDisabled} ref={this.props.uiChildContainerRef} style={style}>
+			<div
+				className={containerClasses}
+				data-webos-voice-focused={voiceFocused}
```
MikyungKim: Please re-order these props.

packages/ui/VirtualList/VirtualList.module.less
```diff
@@ -13,18 +13,32 @@
...
 		overflow-y: scroll;
 	}
+
 	.listItem {
 		width: 100%;
 	}
+
+	&.padding {
+		.listItem {
+			width: calc(100% - 96px);
```
0x64: ui lib is not a themed lib, so we need to avoid a specific design. Let's find another way to add paddings.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -213,6 +213,14 @@ class VirtualListBase extends Component {
 		 */
 		overhang: PropTypes.number,
 
+		/**
+		 * Support additional padding size
+		 *
+		 * @type {Boolean}
+		 * @private
+		 */
+		padded: PropTypes.bool,
```
0x64: Don't we need some 'padding' prop rather than 'padded'? Theme libs require different padding values from each other.

packages/ui/VirtualList/VirtualListBase.js
```diff
@@ -245,12 +256,15 @@ class VirtualListBase extends Component {
 		spacing: PropTypes.number,
 
 		/**
-		 * TBD
+		 * JS or Native type
```
0x64: I don't have a better idea yet, but I believe we should refine this explanation later. :) Let's think about this together.

ybsung:
Closing in favor of https://github.com/enyojs/sandstone/pull/29.

PR #2699, opened 2020-02-14
PLAT-99528: Add radial orientation to ProgressBar

germboy:
### Checklist

Adds additional `orientation` option `'radial'` and associated public classes to `ProgressBar`. Basic expected "radial-style" styles are included, requiring only color and specific sizing styles to be applied when used.


packages/ui/ProgressBar/ProgressBar.js
```diff
@@ -90,6 +91,7 @@ const ProgressBar = kind({
 		 *
 		 * * `'horizontal'` - A left and right orientation
 		 * * `'vertical'` - An up and down orientation
+		 * * `'radial'` - A circular orientation
 		 *
 		 * @type {String}
```
ryanjduffy, line 96: I think we should consider specifying the allowed values in `@type` and via `PropTypes.oneOf`. This seems different that `size` which should be unbounded but it doesn't make sense to allow other values here (except maybe `null` to suppress any default handling). Thoughts @germboy, @webOS101? Not a blocking issue but wanted to ask.

webOS101: I considered that.  But, if we had done that, we wouldn't have been able to implement this in Sandstone without altering `ui`.  Because a theme could invent their own, we should leave this as-is.

Djspaceg: Allowing a freeform string in UI means that UI is capable of supporting other orientations not listed here. In those cases, the theme would be fully responsible for styling it, but would get the behavioral capabilities and CSS values provided by UI to aid in that effort.

Technically, we need a little extra functionality related to the CSS override capabilities here which involve changing how we define the publicClassNames, but that's a different discussion.

ryanjduffy: I think passing `null` would have allowed it to work. That aside, if we want this unbounded, the docs should reflect that. I'd go a step further to say that if it's unbounded we shouldn't blindly pass it to styler since unknown class names will land as unscoped classes on the component which could have unexpected side effects.

ryanjduffy: This seems like a bigger discussion that is warranted for this fix. Let's handle separately.

webOS101: I agree that filtering the names might be worthwhile.  I also don't object to requiring consumers to pass down `null` when not using a supported prop.  Let's discuss this further in another ticket.

Djspaceg: > I think passing `null` would have allowed it to work. That aside, if we want this unbounded, the docs should reflect that. I'd go a step further to say that if it's unbounded we shouldn't blindly pass it to styler since unknown class names will land as unscoped classes on the component which could have unexpected side effects.

I agree with this approach. That seems like a safe plan. I'm content with that, and what @webOS101 said. I don't have anything further to add to a discussion related to this prop specifically. Let's fix it now! :D

Djspaceg: ok it's merged. Nevermind.

ryanjduffy: Adding that restriction would be an API change we should probably wait on anyway :/

ryanjduffy: Didn't occur to me when I brought it up.

PR #2708, opened 2020-02-26
PLAT-100713: Fix item layout and an item being scrolled into view properly

0x64:
### Checklist

Items' layout is broken after structural changes of `VirtualList` so items on right bottom edges are not rendered fit into a content area of a list.

### Resolution
Updated internal calculation logic to handle items' position and size correctly.

### Additional Considerations

### Links
PLAT-100713

### DIFF
--- a/packages/ui/VirtualList/VirtualList.module.less
+++ b/packages/ui/VirtualList/VirtualList.module.less
@@ -12,8 +12,12 @@
 .vertical {
 	&.native {
 		overflow-y: scroll;
+
+		.listItem {
+			width: calc(100% - 2 * var(--scroll-fadeout-size));
+		}
 	}
-	.listItem {
+	&:not(.native) .listItem {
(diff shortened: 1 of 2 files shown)


packages/ui/VirtualList/VirtualList.module.less
```diff
@@ -12,17 +12,25 @@
 .vertical {
 	&.native {
 		overflow-y: scroll;
+
+		.listItem {
+			width: calc(100% - 2 * var(--scroll-fadeout-size));
```
ybsung: This might work incorrectly with moonstone/VirtualList.

ybsung: Oh I misunderstood it. moonstone/VirtualList doesn't use this less.

0x64:
This PR is reverted due to unexpected side effects.

PR #2711, opened 2020-02-27
PLAT-99533: Refine VirtualList and Scroller JS Docs

MikyungKim:
### Checklist

We've refined some VirtualList and Scroller JS Docs

### Resolution

### Additional Considerations

### Links
PLAT-99533


ybsung:
Do we need to replace the `Scrollable` with the `useScroll` here?
https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/Scroller/ScrollerBasic.js#L26

ybsung:
`// scrollMode 'translate'` instead of `// Scrollable`
`// scrollMode 'native'` instead of `// ScrollableNative`
We might have to fix it before.

https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/Scroller/ScrollerBasic.js#L120

ybsung:
Please check this line. The `Scrollable` was used here.
```
 * An unstyled base component for a scroll bar. It is used in [Scrollable]{@link ui/Scrollable.Scrollable}.
```
https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/useScroll/Scrollbar.js#L36

packages/ui/useScroll/useScroll.js
```diff
@@ -1,3 +1,16 @@
+/**
+ * Unstyled scrollable hook and behaviors to be customized by a theme or application.
+ *
+ * @module ui/useScroll
+ * @exports constants
+ * @exports Scrollable
```
ybsung: If we don't need, we could not export the `Scrollable`.

packages/ui/useScroll/useScroll.js
```diff
@@ -1,3 +1,16 @@
...
+ * @exports ScrollableBasic
```
ybsung: Same above. There is no Component and apps using the `ScrollableBasic` now.

packages/ui/useScroll/useScroll.js
ybsung: Let's remove the ui/useScroll.Scrollable.js.

ybsung:
Please check this line.

```
* Normally, [Scrollable]{@link ui/Scrollable.Scrollable} should set this value.
```

https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/VirtualList/VirtualListBasic.js#L105

ybsung:
The `ScrollAnimator` should be belong to the ui/uiScroll.

```
* @memberof ui/Scrollable
```

https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/useScroll/ScrollAnimator.js#L42

packages/ui/useScroll/useScroll.js
MikyungKim: I'll do after removing the references from other theme libs.

packages/ui/useScroll/useScroll.js
```diff
@@ -29,10 +39,10 @@ const
 		isPageDown: is('pageDown'),
 		isPageUp: is('pageUp'),
 		nop: () => {},
-		overscrollTypeNone: 0,
+		overscrollTypeDone: 9,
```
MikyungKim: @0x64 do you agree with this change?

0x64: Honestly, No.

MikyungKim: I agree

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -69,9 +69,9 @@ class VirtualListBasic extends Component {
 		 * ```
```
ybsung: I forgot now, but do you remember why we added the `delete rest.data;` above?

ybsung:
@MikyungKim 
Please remove ui/useScroll/Scrollable.js file if it is not used.

ybsung:
Please check this line including `Scrollable`.
https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/VirtualList/VirtualListBasic.js#L357

ybsung:
Please fix the following lint errors. The https://github.com/enyojs/sandstone/pull/75 travis passed. Please check that PR.
```
/home/travis/build/enactjs/enact/packages/ui/VirtualList/VirtualList.js
  112:2  warning  defaultProp "direction" has no corresponding propTypes declaration  react/default-props-match-prop-types
  204:2  warning  defaultProp "direction" has no corresponding propTypes declaration  react/default-props-match-prop-types
`

**Comment by ybsung** [comment]
I'm not sure which one would be better, but...which one would be better between {Scroller as UiScroller} or {Scroller} from '@enact/ui/Scroller' in samplers?

```
import {Scroller as UiScroller} from '@enact/ui/Scroller';
import {Scroller} from '@enact/ui/Scroller';
```

**Comment by ybsung** [comment]
There is no ScrollerNative from now.

https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/Scroller/Scroller.js#L8

File: packages/ui/VirtualList/VirtualListBasic.js (hunk 4 above)
Comment (MikyungKim) [reply packages/ui/VirtualList/VirtualListBasic.js]: I think we were passing `data` in `renderItem` in the past. I'll remove it.

File: packages/ui/Scroller/Scroller.js [hunk 5]
```diff
@@ -66,41 +67,21 @@ const Scroller = (props) => {
...
+	noScrollByWheel: false,
+	onScroll: nop,
+	onScrollStart: nop,
+	onScrollStop: nop,
+	overscrollEffectOn: {
+		drag: false,
+		pageKey: false,
+		wheel: true
+	},
+	scrollMode: 'translate',
```
ybsung: Why is the default value for the `scrollMode` is not `native` but `translate`?

MikyungKim: Because it's ui..

packages/ui/Scroller/ScrollerBasic.js
```diff
@@ -340,6 +353,20 @@ class ScrollerBasic extends Component {
...
 
+/**
+ * Specifies how to scroll.
+ *
+ * Valid values are:
+ * * `'translate'`,
+ * * `'native'`.
+ *
+ * @name scrollMode
+ * @memberof ui/Scroller.ScrollerBasic.prototype
```
ybsung: Why is the `scrollMode` defined here not in ui/Scroller? The `scrollMode` is not used in the ScrollerBasic.

packages/ui/Scroller/ScrollerBasic.js
```diff
@@ -340,6 +353,20 @@ class ScrollerBasic extends Component {
  * @public
  */
```
ybsung: Please check where is the right place for the documentation above.

ybsung:
The `ui/useScroll.Scrollbar.prototype`'s vertical prop is documented twice.

```
ScrollbarBase.propTypes = /** @lends ui/useScroll.Scrollbar.prototype */ {
	...

	/**
	 * If `true`, the scrollbar will be oriented vertically.
	 *
	 * @type {Boolean}
	 * @default true
	 * @public
	 */
	vertical: PropTypes.bool
};
```

```
Scrollbar.propTypes = /** @lends ui/useScroll.Scrollbar.prototype */ {
	/**
	 * If `true`, the scrollbar will be oriented vertically.
	 *
	 * @type {Boolean}
	 * @default true
	 * @public
	 */
	vertical: PropTypes.bool
};
```

ybsung:
Do we need to define the vertifcal's default value because it is already defined in the ScrollbarBase?

https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/useScroll/Scrollbar.js#L228

```
Scrollbar.defaultProps = {
	vertical: true
};
```

packages/ui/useScroll/useScroll.js
```diff
@@ -1,3 +1,14 @@
+/**
+ * Unstyled scrollable hook and behaviors to be customized by a theme or application.
+ *
+ * @module ui/useScroll
+ * @exports assignPropertiesOf
+ * @exports constants
+ * @exports useScroll
+ * @exports useScrollBase,
```
ybsung: We could omit `,`.

packages/ui/Scroller/ScrollerBasic.js
MikyungKim: That's because to make things easier when extends ScrollerBasic from other theme libs.
`scrollMode` is for `useScroll` but then we should extend both ScrollerBasic and useScroll from theme libs.
So I copied those props from useScroll to *Basic series.

ybsung:
We also need to export the `itemSizesShape`.

https://github.com/enactjs/enact/blob/feature/PLAT-99533/packages/ui/VirtualList/VirtualList.js#L5

packages/ui/Scroller/ScrollerBasic.js
MikyungKim: Yes for the same reason above. I should remove `moreInfo` though.

packages/ui/VirtualList/VirtualList.js
```diff
@@ -175,26 +345,48 @@ VirtualGridList.displaytName = 'ui:VirtualGridList';
...
+	 * Once received, the `scrollTo` method can be called as an imperative interface.
 	 *
-	 * * Valid value is an object that has `minWidth` and `minHeight` as properties.
+	 * The `scrollTo` function accepts the following parameters:
+	 * - {position: {x, y}} - Pixel value for x and/or y position
+	 * - {align} - Where the scroll area should be aligned. Values are:
+	 *   `'left'`, `'right'`, `'top'`, `'bottom'`,
+	 *   `'topleft'`, `'topright'`, `'bottomleft'`, and `'bottomright'`.
+	 * - {index} - Index of specific item. (`0` or positive integer)
+	 *   This option is available for only `VirtualList` kind.
```
ybsung: > for only `VirtualList`

"for `VirtualList` and `VirtualGridList`" would be better to me.

packages/ui/VirtualList/VirtualList.js
```diff
@@ -211,8 +403,137 @@ VirtualGridList.propTypes = /** @lends ui/VirtualList.VirtualGridList.prototype
...
+	 * @param {Object} event.moreInfo The object including `firstVisibleIndex` and `lastVisibleIndex` properties.
+	 * @public
+	 */
+	onScroll: PropTypes.func,
+
+	/**
+	 * Called when scroll starts.
+	 *
+	 * Passes `scrollLeft`, `scrollTop`, and `moreInfo`.
+	 * You can get firstVisibleIndex and lastVisibleIndex from VirtualList with `moreInfo`.
```
ybsung: How about `VirtualGridList` instead of `VirtualList`?

packages/ui/VirtualList/VirtualList.js
```diff
@@ -211,8 +403,137 @@ VirtualGridList.propTypes = /** @lends ui/VirtualList.VirtualGridList.prototype
...
+	 *
+	 * Example:
+	 * ```
+	 * onScrollStart = ({scrollLeft, scrollTop, moreInfo}) => {
+	 *     const {firstVisibleIndex, lastVisibleIndex} = moreInfo;
+	 *     // do something with firstVisibleIndex and lastVisibleIndex
+	 * }
+	 *
+	 * render = () => (
+	 *     <VirtualList
```
ybsung: Same above

packages/ui/VirtualList/VirtualList.js
```diff
@@ -211,8 +403,137 @@ VirtualGridList.propTypes = /** @lends ui/VirtualList.VirtualGridList.prototype
...
+	 *
+	 * Example:
+	 * ```
+	 * onScrollStop = ({scrollLeft, scrollTop, moreInfo}) => {
+	 *     const {firstVisibleIndex, lastVisibleIndex} = moreInfo;
+	 *     // do something with firstVisibleIndex and lastVisibleIndex
+	 * }
+	 *
+	 * render = () => (
+	 *     <VirtualList
```
ybsung: Same above

packages/ui/VirtualList/VirtualList.js
MikyungKim: I think with kind is better.

packages/ui/VirtualList/VirtualList.js
MikyungKim: 👍

packages/ui/VirtualList/VirtualList.js
MikyungKim: 👍

packages/ui/VirtualList/VirtualList.js
MikyungKim: 👍

packages/sampler/stories/default/VirtualList.js
```diff
@@ -45,23 +44,23 @@ const updateDataSize = (dataSize) => {
 
 updateDataSize(defaultDataSize);
 
-const UiVirtualListConfig = mergeComponentMetadata('VirtualList', UiVirtualListBase, UiScrollableBase);
+const VirtualListConfig = mergeComponentMetadata('VirtualList', VirtualList, VirtualListBasic);
```
ybsung: The default props for the `VirtualListBasic` overrides the default props for the `VirtualList`.
Currently, all the default props are same between them. As the aspect of the maintenance point, it would be better to switch them.

Actually, in theme samplers, there might be problems if defining the `VirtualList` first.

The `VirtualListBasic` should be in front of the `VirtualList`.

ybsung:
LGTM

PR #2712, opened 2020-02-28
PLAT-100713: Fix item layout and an item being scrolled into view properly

0x64:
### Checklist

Items' layout is broken after structural changes of `VirtualList` so items on right bottom edges are not rendered fit into a content area of a list.

### Resolution
Updated internal calculation logic to handle items' position and size correctly.

### Additional Considerations
The former approach (#2708) is reverted (#2710) due to side effect by `getComputedStyle`. In this PR, ui lib does not handle padding directly but ask theme lib to correct along with the case of theme lib.

### Links
PLAT-100713


ybsung:
Please fix conflicting files.

0x64:
> Please fix conflicting files.

I will rebase the PR.

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -554,10 +571,11 @@ class VirtualListBasic extends Component {
 
 	getXY = (primaryPosition, secondaryPosition) => (this.isPrimaryDirectionVertical ? {x: secondaryPosition, y: primaryPosition} : {x: primaryPosition, y: secondaryPosition})
 
-	getClientSize = (node) => ({
-		clientWidth: node.clientWidth,
-		clientHeight: node.clientHeight
-	})
+	getClientSize = (props) => {
+		const clientSize = props.clientSize || props.scrollContentRef.current;
```
ybsung: I'm not sure that it is the right way for app developers to pass the `clientSize` prop to the sandstone/VirtualList.

It's right to do it in moonstone/VirtualList because the `width` and the `height` are same with the `clientWidth` and the `clientHeight` if no scrollbars. But for sandstone, it is hard for app developers to calculate the width, height of the sandstone/VirtualList because they  have to add proper some size of padding and margin for fadeout area and scrollbar area.

For now, I think that it would be better to pass not `clientSize` but `width` and `height` to the sandstone/VirtualList. Please let me know your thought.

0x64: I'm sorry but I'm very confused. Is your point about `prop` for app developers or additional post-processing for each themed library?

0x64: Note: @ybsung pointed out the prop `clientSize` may not be comfortable to our app devs on other themed libraries as I discussed off-line. This topic is valid but we don't want to discuss it here since it is a topic of API. So, we will discuss it separately if needed.

ybsung:
LGTM

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -593,22 +611,22 @@ class VirtualListBasic extends Component {
 
 	calculateMetrics (props) {
 		const
-			{clientSize, direction, itemSize, overhang, spacing} = props,
-			node = this.props.scrollContentRef.current;
+			{direction, itemSize, overhang, spacing} = props,
+			contentSizes = this.getContentSize(props);
```
MikyungKim: Why it's Sizes?

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -576,10 +593,11 @@ class VirtualListBasic extends Component {
...
 	getXY = (primaryPosition, secondaryPosition) => (this.isPrimaryDirectionVertical ? {x: secondaryPosition, y: primaryPosition} : {x: primaryPosition, y: secondaryPosition})
 
-	getClientSize = (node) => ({
-		clientWidth: node.clientWidth,
-		clientHeight: node.clientHeight
-	})
+	getContentSize = (props) => {
+		const contentSizes = props.clientSize || props.scrollContentRef.current;
+
+		return contentSizes && props.getContentSize ? props.getContentSize(contentSizes) : contentSizes;
```
MikyungKim: Don't we need a proptypes for `getContentSize`?

packages/ui/VirtualList/VirtualListBasic.js
0x64: I named it since it has two values: clientHeight and clientWidth. Just 'size' would be good to me also.

packages/ui/VirtualList/VirtualListBasic.js
0x64: No lint warning/error so I missed it.

PR #2713, opened 2020-03-03
PLAT-101582: Cache item refs in VirtualList

ybsung:
### Checklist

The `querySelector` are used a lot in the VirtualList to access item DOM elements easily. We might avoid using DOM APIs directly.

### Resolution

Cache item refs which are passed as the VirtualList's prop - `itemRefs` and use it when needed. If there is no `itemRefs` prop, then the item refs are not cached.

### Additional Considerations

### Links

PLAT-101582


packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1044,20 +1052,35 @@ class VirtualListBasic extends Component {
...
-			{itemProps, itemRenderer, getComponentProps} = this.props,
-			key = index % this.state.numOfItems,
-			componentProps = getComponentProps && getComponentProps(index) || {};
+			{itemRefs, itemProps, itemRenderer, getComponentProps} = this.props,
+			{numOfItems} = this.state,
+			key = index % numOfItems,
+			componentProps = getComponentProps && getComponentProps(index) || {},
+			itemContainerRef = (ref) => {
+				if (itemRefs) {
+					itemRefs.current[index % numOfItems] = (ref === null) ? null : ref;
```
MikyungKim: Could we replace the `(ref === null) ? null : ref;` to 'ref ? ref : null` ?

MikyungKim: Isn't `index % numOfItems` `key`?

0x64: It looks the same as just `ref`.

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1044,20 +1052,35 @@ class VirtualListBasic extends Component {
...
 		);
 	}
 
 	applyStyleToHideNode = (index) => {
-		const key = index % this.state.numOfItems;
-		this.cc[key] = <div key={key} style={{display: 'none'}} />;
+		const
+			{itemRefs} = this.props,
+			{numOfItems} = this.state,
+			key = index % this.state.numOfItems,
```
MikyungKim: You mean `numOfItems` here?

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1044,20 +1052,35 @@ class VirtualListBasic extends Component {
...
+			itemContainerRef = () => {
+				if (itemRefs) {
+					itemRefs.current[index % numOfItems] = null;
```
MikyungKim: maybe `key`?

packages/ui/VirtualList/VirtualListBasic.js
ybsung: > Could we replace the `(ref === null) ? null : ref;` to `ref ? ref : null` ?

Is `ref || null` fine?

MikyungKim: yes

ybsung: > Isn't `index % numOfItems` `key`?

Yes.... I replaced that part with the key.

ybsung: > It looks the same as just `ref`.

OK.

ybsung:
I fixed all that @MikyungKim  and @0x64 mentioned.

packages/ui/Scroller/ScrollerBasic.js
```diff
@@ -187,6 +187,7 @@ class ScrollerBasic extends Component {
 		delete rest.cbScrollTo;
 		delete rest.scrollContainerContainsDangerously;
 		delete rest.direction;
+		delete rest.itemRefs;
```
MikyungKim: Can we assign this prop only for VL?

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1194,7 +1211,7 @@ class VirtualListBasic extends Component {
 		return (
 			<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} data-webos-voice-disabled={voiceDisabled} ref={this.props.scrollContentRef} style={style}>
 				<div {...rest} className={contentClasses} ref={this.contentRef}>
-					{itemsRenderer({cc, itemContainerRef, primary})}
+					{itemsRenderer({cc, primary})}
```
MikyungKim: I think we are referring to `itemContainerRef` in sandstone VL.
Please make a change in sandstone along with this.

ybsung: Fix it in the sandstone PR and in the agate PR.

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -993,7 +1000,7 @@ class VirtualListBasic extends Component {
 
 	// For individually sized item
 	adjustItemPositionWithItemSize () {
-		if (this.itemContainerRef.current) {
+		if (this.contentRef.current) {
```
MikyungKim: I think this one is not identical to the old condition.
Maybe `cc.length` would be better.

PR #2714, opened 2020-03-03
PLAT-101792: Fix not to scroll abnormally while pressing 5-way keys after wheeling

ybsung:
### Checklist

1. After wheeling in native VirtualList,
2. if pressing 5-way keys,
3. then the list suddenly scrolls.
4. So the spotted item is stick to the viewport.

### Resolution

While wheeling, the scroll position was not cached properly. So if pressing 5-way keys, then the wrong scroll position was referenced. Therefore that scroll position is now cached properly while wheeling.

### Additional Considerations

### Links

PLAT-101792

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/VirtualList.VirtualList` and `ui/VirtualList.VirtualGridList` not to scroll while pressing 5-way keys after wheeling.
+
 ## [3.3.0-alpha.1] - 2020-02-26
 
 ### Added
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/VirtualList.VirtualList` and `ui/VirtualList.VirtualGridList` not to scroll while pressing 5-way keys after wheeling.
+- `ui/VirtualList.VirtualList` and `ui/VirtualList.VirtualGridList` not to suddenly jump when pressing directional keys after wheeling.
 
 ## [3.3.0-alpha.1] - 2020-02-26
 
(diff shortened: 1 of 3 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/VirtualList.VirtualList` and `ui/VirtualList.VirtualGridList` not to scroll while pressing 5-way keys after wheeling.
```
MikyungKim: How about "not to suddenly jump when pressing directional keys after wheeling"?

ybsung: Fix it as you mentioned.

packages/ui/useScroll/useScroll.js
```diff
@@ -728,6 +728,10 @@ const useScrollBase = (props) => {
 			setScrollTop(scrollTop);
 		}
 
+		if (scrollMode === 'native') {
```
MikyungKim: I believe we are safe not to check `scrollMode` here.

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -832,6 +832,10 @@ class VirtualListBasic extends Component {
 		}
 	}
 
+	updateScrollPositionTarget (x, y) {
```
MikyungKim: If you are going to define this function, why not change the code block from `setScrollPosition` and `scrollToPosition` to this function?

packages/ui/useScroll/useScroll.js
```diff
@@ -728,6 +728,10 @@ const useScrollBase = (props) => {
...
+			scrollContentHandle.current.updateScrollPositionTarget(mutableRef.current.scrollLeft, mutableRef.current.scrollTop);
+		}
+
 		if (scrollContentHandle.current.didScroll) {
```
MikyungKim: Why don't we set `scrollPositionTarget` in `didScroll` function?
It seems either `native` or `translate` mode are all calling `didScroll` after scroll.

MikyungKim: Maybe we can remove updateScrollPositionTarget function and the same code block from `setScrollPosition` and `scrollToPosition`.

MikyungKim: So, in `didScroll`,
```
this.scrollPosition = pos;
this.scrollPositionTarget = pos;
```
will do the same thing in updateScrollPositionTarget.

ybsung:
Closing in favor of https://github.com/enactjs/enact/pull/2715.

PR #2715, opened 2020-03-04
PLAT-101792: Fix not to scroll abnormally while pressing 5-way keys after wheeling

ybsung:
### Checklist

1. After wheeling in native VirtualList,
2. if pressing 5-way keys,
3. then the list suddenly scrolls.
4. So the spotted item is stick to the viewport.

### Resolution

While wheeling in native VirtualList, the final scroll position was not cached properly. So if pressing 5-way keys on the list item, then the wrong scroll position was referenced. Therefore for native VirtualList, the real scroll position instead of the cached scroll position is used.

### Additional Considerations

### Links

PLAT-101792


packages/ui/Scroller/ScrollerBasic.js
```diff
@@ -115,7 +115,30 @@ class ScrollerBasic extends Component {
...
+	}
+
+	// scrollMode 'native' only
+	didScroll (x, y) {
+		this.scrollPos.left = x;
+		this.scrollPos.top = y;
+	}
+
+	// scrollMode 'translate' only
+	setScrollPositionTarget (x, y) {
```
MikyungKim: I don't think we need this function in `Scroller`.
There is no reference case for `scrollPositionTarget` from themed `Scroller`.
Why did you add this?

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -470,6 +470,7 @@ class VirtualListBasic extends Component {
 		if (this.props.scrollMode === 'native') {
 			this.scrollToPosition(x, y, rtl);
 		} else {
+			this.setScrollPositionTarget (x, y);
 			this.setScrollPosition(x, y, rtl, x, y);
```
MikyungKim: You are passing additional parameter `x`, `y` here.

packages/ui/Scroller/ScrollerBasic.js
ybsung: Oh, my mistake. I should be removed.

packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Scroller.Scroller`, `ui/VirtualList.VirtualList` and `ui/VirtualList.VirtualGridList` not to suddenly jump when pressing directional keys after wheeling.
```
MikyungKim: Please omit Scroller here.

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -829,19 +824,21 @@ class VirtualListBasic extends Component {
...
 	// scrollMode 'translate' only
-	setScrollPosition (x, y, rtl = this.props.rtl, targetX = 0, targetY = 0) {
+	setScrollPositionTarget (x, y) {
+		// The `x`, `y` as parameters in scrollToPosition() are the position when stopping scrolling.
+		// But the `x`, `y` as parameters in setScrollPosition() are the position between current position and the position stopping scrolling.
+		// To know the position when stopping scrolling properly, `x` and `y` are passed and cached in `this.scrollPositionTarget`.
+		if (this.isPrimaryDirectionVertical) {
+			this.scrollPositionTarget = y;
+		} else {
+			this.scrollPositionTarget = x;
```
jeonghee27: `setScrollPositionTarget` `setScrollPosition` both function are for 'translate'.
I wonder the reason why you divide it to two function?

ybsung: `setScrollPositionTarget` is needed to set the position that a scroller will stop.
` setScrollPosition` is needed to update the scroll position while animating and going to the position that a scroller will stop.

To handle 5-way keys in sandstone VirtualList onAcceleratedKeyDown() properly, we need to know the position that a scroller will stop to make a decision if scrolling into a view or not.

packages/ui/useScroll/useScroll.js
```diff
@@ -1076,11 +1076,14 @@ const useScrollBase = (props) => {
...
 			showThumb(bounds);
+			if (scrollContentHandle.current && scrollContentHandle.current.setScrollPositionTarget) {
+				scrollContentHandle.current.setScrollPositionTarget(targetX, targetY);
+			}
 
 			if (animate) {
 				mutableRef.current.animator.animate(scrollAnimation(mutableRef.current.animationInfo));
 			} else {
-				scroll(targetX, targetY, targetX, targetY);
+				scroll(targetX, targetY);
```
jeonghee27: Now  `scroll` function get two parameter always.
I think the third parameter make confusing to understand further behavior. 
`scroll = (left, top, ...rest) => {`
How about remove the `...rest` also?

packages/ui/VirtualList/VirtualListBasic.js
jeonghee27: I understand. You wanted to omit update the `scrollPositionTarget` in situations such as animations. Please let me know If I misunderstand,

packages/ui/useScroll/useScroll.js
ybsung: For `translate` VirtualList, the `rest` should be passed to the `setScrollPosition`.

ybsung: So the `this.scroll` usage is
```
this.scroll(targetX, targetY, targetX, targetY);
```

MikyungKim: ? I think @jeonghee27 is right.

MikyungKim: Now we are setting `scrollPositionTarget` seperately.

ybsung: Oh, my misunderstanding. I fixed it.

PR #2716, opened 2020-03-05
PLAT-101903: Add proxy support into mergeClassNames

ryanjduffy:

### Checklist

When using CSS overrides with CSS modules in unit tests, the reflection proxy plugin is subverted by merging that object into a new one.

### Resolution
Add a test-only proxy that emulates the same behavior with support for merged class names.

### DIFF
--- a/packages/core/util/util.js
+++ b/packages/core/util/util.js
@@ -188,6 +188,15 @@ const mergeClassNameMaps = (baseMap, additiveMap, allowedClassNames) => {
 				css[key] = baseMap[key] + ' ' + additiveMap[key];
 			}
 		});
+
+		if (process.env.NODE_ENV === 'test') {
+			return new Proxy({}, {
+				get (target, key) {
+					// use the merged value if it exists and the key otherwise
+					return css[key] || key;
+				}
+			});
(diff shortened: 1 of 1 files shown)


packages/core/util/util.js
```diff
@@ -188,6 +188,15 @@ const mergeClassNameMaps = (baseMap, additiveMap, allowedClassNames) => {
 				css[key] = baseMap[key] + ' ' + additiveMap[key];
 			}
 		});
+
+		if (process.env.NODE_ENV === 'test') {
```
ryanjduffy: Wondering if we should add a special enact-specific guard here as well since it is possible (though unlikely) that a third party might be using this without the same plugins and this could fail their tests since it would change behavior.

JayCanuck: We should be good. with just `process.env.NODE_ENV === 'test'`. Why it is generic enough to match other testing tools, the csss identity name spec matches our primary compatibility targetof create-react-app, both for modular and static css. Any custom enact env var would potentially break CRA from working on our tests as expects, so I'm fine with this current propsal as-is personally.

PR #2721, opened 2020-03-09
PLAT-101531: Update VirtualList and Scroller styles in default samplers

ybsung:
### Checklist

According to the theme libraries, the styles of the VirtualList and the Scroller updates.

### Additional Considerations

If we don't need to align with the theme sampler's UX/GUI, this PR could be discarded.

### Links

PLAT-101531

### DIFF
--- a/packages/sampler/stories/default/VirtualGridList.js
+++ b/packages/sampler/stories/default/VirtualGridList.js
@@ -67,7 +67,7 @@ storiesOf('UI', module)
 				horizontalScrollbar={select('horizontalScrollbar', prop.scrollbarOption, VirtualGridListConfig)}
 				itemRenderer={uiRenderItem}
 				itemSize={{
-					minWidth: ri.scale(number('minWidth', VirtualGridListConfig, 180)),
+					minWidth: ri.scale(number('minWidth', VirtualGridListConfig, 100)),
 					minHeight: ri.scale(number('minHeight', VirtualGridListConfig, 270))
 				}}
 				noScrollByWheel={boolean('noScrollByWheel', VirtualGridListConfig)}
--- a/packages/sampler/stories/default/VirtualGridList.js
+++ b/packages/sampler/stories/default/VirtualGridList.js
@@ -27,6 +27,7 @@ const
 				{...rest}
 				caption={text}
 				source={source}
+				style={{width: '100%'}}
 				subCaption={subText}
 			/>
 		);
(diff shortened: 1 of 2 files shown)


packages/sampler/stories/default/VirtualList.js
```diff
@@ -54,8 +46,8 @@ storiesOf('UI', module)
 				<VirtualList
 					dataSize={updateDataSize(number('dataSize', VirtualListConfig, defaultDataSize))}
 					horizontalScrollbar={select('horizontalScrollbar', prop.scrollbarOption, VirtualListConfig)}
-					itemRenderer={uiRenderItem(ri.scale(number('itemSize', VirtualListConfig, 72)))}
-					itemSize={ri.scale(number('itemSize', VirtualListConfig, 72))}
+					itemRenderer={uiRenderItem(ri.scale(number('itemSize', VirtualListConfig, 156)))}
+					itemSize={ri.scale(number('itemSize', VirtualListConfig, 156))}
```
jeonghee27: image

I don't think we should update it. 
Let's update sandstone only.

MikyungKim: I agree.

MikyungKim:
Could you also take a look at VirtualGridList sample? I think item needs `width=100%`.

packages/sampler/stories/default/VirtualList.js
MikyungKim: Let's not change 72 to 156 here.

packages/sampler/stories/default/VirtualList.js
```diff
@@ -8,21 +8,13 @@ import {storiesOf} from '@storybook/react';
...
 	uiRenderItem = (size) => ({index, ...rest}) => {
-		const itemStyle = {
-			borderBottom: ri.unit(3, 'rem') + ' solid #202328',
-			boxSizing: 'border-box',
-			height: size + 'px'
-		};
-
 		return (
-			<UiItem {...rest} style={itemStyle}>
+			<UiItem {...rest} style={{height: size + 'px'}}>
```
MikyungKim: let's keep the border here too.

packages/sampler/stories/default/VirtualGridList.js
```diff
@@ -67,7 +67,7 @@ storiesOf('UI', module)
 				horizontalScrollbar={select('horizontalScrollbar', prop.scrollbarOption, VirtualGridListConfig)}
 				itemRenderer={uiRenderItem}
 				itemSize={{
-					minWidth: ri.scale(number('minWidth', VirtualGridListConfig, 180)),
+					minWidth: ri.scale(number('minWidth', VirtualGridListConfig, 100)),
```
MikyungKim: Sorry.. but.. I didn't mean this. Please add `width=100%` to the style.

PR #2723, opened 2020-03-10
Add text-index for Marquee

Baekwoo-s:
### Checklist

Fix for characters like `j` being left-truncated because of being too close to the edge

### Resolution
Adjust indentation of the marquee text 

### Comments

### DIFF
--- a/packages/ui/Marquee/Marquee.module.less
+++ b/packages/ui/Marquee/Marquee.module.less
@@ -27,6 +27,7 @@
 		transition-property: left, right;
 		transition-timing-function: linear;
 		position: relative;
+		text-indent: 0.05em;  // Fix for characters like `j` being left-truncated because of being too close to the edge
 
 		&.willAnimate {
 			will-change: left, right;


packages/ui/Marquee/Marquee.module.less
```diff
@@ -27,6 +27,7 @@
 		transition-property: left, right;
 		transition-timing-function: linear;
 		position: relative;
+		text-indent: 0.05em;  // Fix for characters like `j` being left-truncated because of being too close to the edge
```
seunghoh: 0.05em looks also cut
image

Baekwoo-s: I have confirmed this too. The existing 0.05 em is not enough. I think this figure is 0.06 em is appropriate.

Baekwoo-s:
Note: Some entries in the changelog are incorrectly entered at https://github.com/enactjs/enact/pull/2719. So I erased

PR #2727, opened 2020-03-16
PLAT-102262: Migrate i18n/I18nDecorator to support hooks

ryanjduffy:
### Checklist

Refactor `i18n/I18nDecorator` to use React hooks

### Resolution
* Moves business logic in `I18n` class
* Adds `useI18n` and `useI18nContext` hooks
* Refactors HOCs to use new hooks

### Additional Considerations
Added UI tests for i18n to be able to test resource loading in a browser for a more representative test environment.


packages/i18n/I18nDecorator/tests/useI18n-specs.js
```diff
@@ -0,0 +1,110 @@
...
+		const expected = [
+			'enact-locale-ar',
+			'enact-locale-ar-SA',
+			'enact-locale-SA',
+			'enact-locale-non-italic',
+			'enact-locale-right-to-left'
+		].sort();
+		const actual = subject.prop('className').split(' ').sort();
+
+		expect(actual).toMatchObject(expected);
```
webOS101: `toEqual` works with arrays.  Any reason to use `toMatchObject`?

ryanjduffy: Not that I recall now.

packages/i18n/I18nDecorator/useI18n.js
```diff
@@ -0,0 +1,64 @@
...
+
+	const i18n = useClass(I18n, config);
+	i18n.setContext(locale, setState);
+
+	// Add/remove listeners on mount/unmount
+	React.useEffect(() => {
+		i18n.load();
+
+		return () => i18n.unload();
+	}, [i18n]);
```
ybsung, line 53: Hello @ryanjduffy 

Many components based on React.Component have `componentDidMount` and `componentWillUnmount` lifecycle methods. So what if providing the lifecycle for hook in the `useClass` like the following way. Then we might remove those redundant hook lifecycle definition in the components and we could have the same method naming like the `load` and the `unload` or something else.
```
function useClass (Ctor, ...args) {
	const ref = React.useRef(null);
	ref.current = ref.current || new Ctor(...args);

	React.useEffect(() => {
		ref.current.load && ref.current.load();

		return () => ref.current.unload && ref.current.unload();
	}, [i18n]);

	return ref.current;
}
```

ryanjduffy: Thanks for the feedback, @ybsung. I'm hesitant to recreate the React lifecycle in a hook but I agree there is some utility here. Curious what others this.

webOS101: Hmm...  I think I would keep the `useEffect` out of `useClass`.  It's easy enough for the consumer to add that functionality if needed and the way args are used it could be tricky to add the `useEffect` body.

ybsung:
I'm curious about the current behavior in the PR. Please let me know if the following issue is a bug or not.

If passing different `i18n` configs to the `I18nDecorator`, then some configs could not be passed to the `useI18n`.

The steps to reproduce
1. Update `sandstone/samples/sampler/src/ThemeEnvironment/ThemeEnvironment.js` to pass different `i18n` configs. One for `i18n: {sync: false}`. The other for `i18n: {sync: true}`.
```
const Theme = ThemeDecorator({overlay: false, i18n: {sync: false}}, PanelsBase);
const ThemeFullscreen = ThemeDecorator({overlay: false, i18n: {sync: true}}, FullscreenBase);
```

2. Add logs in `Enact/packages/i18n/I18nDecorator/useI18n.js`
```
	console.log('useI18n out of useEffect', config.sync)
	debugger;
	React.useEffect(() => {
		console.log('useI18n inside of useEffect', config.sync)
		i18n.load();

		return () => i18n.unload();
	}, [i18n]);
```

3. Launch sandstone qa-sampler and select "Button with long text"

Expected Result:
There are two logs.
`useI18n out of useEffect false`
`useI18n out of useEffect true`

Actual Result:
There are two logs.
`useI18n out of useEffect false`
`useI18n out of useEffect false`

`i18n: {sync: true}` config does not be passed to the `useI18n`. So 'useI18n inside of useEffect' was logged once.

Working branch
Enact: https://github.com/enactjs/enact/tree/feature/PLAT-102262-ybsung
sandstone: https://github.com/enyojs/sandstone/tree/feature/PLAT-102262-ybsung

ryanjduffy:
@ybsung - there is a problem but I suspect it lies outside of the hook. The right values reach `I18nDecorator` hoc callback but the wrong value is reported by the return `i18nDecorator` function. I suspect pass-by-reference issue but haven't tracked it down yet. Thanks for the report!!

ryanjduffy:
On second thought, I _think_ it's working as expected. In sampler, only 1 of the two components -- `Theme` and `ThemeFullscreen` -- is in use. You'll see both output for `console.log('I18nDecorator', config.sync);` in the HOC wrapper because the HOCs are created at runtime but the component returned from the HOC is only created when used. Currently, only the `Theme` instance is used which is why you only see `false` reported in the logs from your example.

ybsung:
@ryanjduffy 
Oh, I see. I totally misunderstood the current behavior in the PR. Sorry for the false alarm and for wasting your time.

ryanjduffy:
No worries! Glad you asked so we can confirm it is working as expected.

PR #2730, opened 2020-03-18
ENYO-6435: Use classNames rather than joining

0x64:
### Checklist

We joined classNames for scrollers and lists, so we had to take care of falsy values when we deal with class names. For better and safe management, replaced related logic with `classNames` module.

### Resolution

### Additional Considerations

### Links
ENYO-6435


ybsung:
LGTM

PR #2732, opened 2020-03-25
PLAT-103307: [VirtualList and Scroller] Make fade-out effect dynamically

MikyungKim:
### Checklist

Current VirtualList and Scroller have static fade-out area for 48px around the content area.
This was mainly for the ImageItem to be expanded in the VirtualGridList.
But we've found out that the case for Item in VirtualList, the fade-out padding is not needed. Also, this area makes the item area smaller in the dropdown list.
So we'd like to remove the padding for fade-out and make the effect dynamically.

### Resolution
Reverted PLAT-100713 and ENYO-6427 since it caused by padding for the fade-out.
Added `applyFadeoutEffect` call from theme lib.
Added `optionalOffset` to scrollTo function to avoid item is overlapped with fade-out effect area.

### Additional Considerations

### Links
PLAT-103307

### DIFF
--- a/packages/ui/useScroll/useScroll.js
+++ b/packages/ui/useScroll/useScroll.js
@@ -116,6 +116,7 @@ const useScrollBase = (props) => {
 	// So it is better to use props[propName]
 	// instead of extracting it from the `props` and renaming it
 	delete rest.addEventListeners;
+	delete rest.applyFadeoutEffect;
 	delete rest.applyOverscrollEffect;
 	delete rest.cbScrollTo;
 	delete rest.clearOverscrollEffect;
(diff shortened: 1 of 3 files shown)


packages/ui/useScroll/useScroll.js
```diff
@@ -947,6 +948,8 @@ const useScrollBase = (props) => {
 	function setScrollLeft (value) {
 		const bounds = getScrollBounds();
 
+		props.applyFadeoutEffect(mutableRef.current.scrollLeft, value, bounds.maxLeft);
```
jeonghee27: We need to check the existence of the applyFadeoutEffect() for agate.

MikyungKim:
Closing in favor of https://github.com/enactjs/enact/pull/2737

PR #2734, opened 2020-03-26
ENYO-6430: Fix console error when wheel on ui native Scroller

seunghoh:
### Checklist

Fixed Uncaught TypeErrorr when wheeling on ui native Scroller

### Resolution
Change to call getter `getContainerRef()` which referred to the wrong ref in the `onWheel`

### Links
ENYO-6430


jeonghee27:
image

Another console error still occurs.

seunghoh:
I Checked only on Scroller which has horizontal ref, vertical ref both regardless of orientation. Added check on VirtualList case which has only oneway directional ref

jeonghee27:
LGTM

MikyungKim:
When Scroller with native mode and direction vertical, I got 
```
Uncaught TypeError: Cannot read property 'getContainerRef' of null
    at HTMLDivElement.onWheel (useScroll.js:658)
```
I think we need to check `horizontalScrollbarRef.current` and `verticalScrollbarRef.current` 's exsitence.

MikyungKim:
LGTM!

PR #2738, opened 2020-03-30
PLAT-99658: Fixed to cover scrollbars also with scrim when container is muted

0x64:
### Checklist

A scrim for a muted container of a scroller or a list does not cover scrollbars properly.

### Resolution
Moved related CSS to the top node of scrollable components.

### Additional Considerations
This change is also required to support a muted container in other theme libs.

### Links
PLAT-99658

### DIFF
--- a/packages/ui/useScroll/useScroll.js
+++ b/packages/ui/useScroll/useScroll.js
@@ -98,7 +98,7 @@ const useScrollBase = (props) => {
 			noScrollByDrag,
 			noScrollByWheel,
 			overhang,
-			overscrollEffectOn = {drag: false, pageKey: false, wheel: false},
+			overscrollEffectOn,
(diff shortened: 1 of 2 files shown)


0x64:
Note: we have a `ref` issue for `touchable` scroller now. It needs to be fixed. @MikyungKim any idea?

MikyungKim:
@0x64, I think we could pass `ref` directly to the `ScrollContentWrapper` instead of using `ForwardRef`. WDYT?

0x64:
> @0x64, I think we could pass `ref` directly to the `ScrollContentWrapper` instead of using `ForwardRef`. WDYT?

👍

0x64:
The `ref` issue resolved with prop bypassing on `Touchable`.

PR #2739, opened 2020-03-30
PLAT-97858: Migrate ui/Toggleable to use hooks

ryanjduffy:
### Checklist

First step in migrating our HOCs to composable hooks to reduce component layers and improve readability through less "prop magic" from the HOC stacks.

### Resolution
* Add `useControlledState` hook to standardize controlled state mgmt
* Add `useToggle` hook, adopting `useControlledState`, to support toggleable behavior
* Add `core/handle.not` to simply complementary conditions (e.g. prop is falsy checks)
* Adopt `ui/Toggleable` to use `useToggle`
* Adds a unit test for the regression that caused #2735

### Additional Considerations
Need to determine documentation approach for hooks. For the time being, it is private API so we can defer that for a little while if we choose.


ryanjduffy:
The issue with #2679 was storing the computed value rather than the prop value in `instance` so that when:
* the prop was null-ish,
* the value was true via the `toggle` callback (not via props), and
* the component was re-rendered
the value would be changed to `false`.

I've updated `instance` to only store the previous prop value and added a unit test to verify this behavior.

PR #2741, opened 2020-04-02
PLAT-102877: Display fill and load base on center

changgilee:
### Checklist

Add new type of Slider for display fill and load base on center

### Resolution
`balanced` prop is added for this feature

### Additional Considerations

### Links
PLAT-102877

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `ui/ProgressBar` display fill and load base on center using `balanced` prop
+
 ## [3.3.0-alpha.4] - 2020-03-30
 
 ### Deprecated
--- a/packages/ui/ProgressBar/ProgressBar.js
+++ b/packages/ui/ProgressBar/ProgressBar.js
@@ -19,9 +19,25 @@ import {validateRange} from '../internal/validators';
 import componentCss from './ProgressBar.module.less';
 
 const progressToProportion = (value) => clamp(0, 1, value);
-const calcBarStyle = (prop, anchor, value = anchor, startProp, endProp) => {
+const calcBarStyle = (prop, balanced, anchor, value = anchor, startProp, endProp) => {
 	let start = Math.min(anchor, value);
 	let end = Math.max(anchor, value) - start;
+
+	if (balanced) {
+		if (prop === 'progress') {
+			if (value <= 0.5) {
+				end = 0.5 - (end - start);
+				start = 0.5 - end;
+			} else {
+				end = (end - start) - 0.5;
+				start = 0.5;
+			}
+		} else if (prop === 'backgroundProgress') {
+			const gap = (end - start) / 2;
+			start = 0.5 - gap;
+			end = gap * 2;
+		}
+	}
 
 	if (__DEV__) {
 		validateRange(start, 0, 1, 'ProgressBar', prop, 'min', 'max');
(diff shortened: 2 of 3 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `ui/ProgressBar` add `separator` for support balance slider
```
webOS101, line 9: Should read like the sentence starts with 'Added'.  So, '(addd) `ui/ProgressBar` prop `separator`'  (I think we don't have to specify what it's for exactly, or, if we do, we should mention the prop it works with.

packages/ui/ProgressBar/ProgressBar.js
```diff
@@ -121,7 +121,15 @@ const ProgressBar = kind({
 		 * @default 0
 		 * @public
 		 */
-		progressAnchor: PropTypes.number
+		progressAnchor: PropTypes.number,
+
+		/**
+		 * Display separator at the center
```
webOS101, line 127: Can you please update this to use the position of the `progressAnchor` instead of just center?  Particularly for `ui`, we can't assume it will only be center.

packages/ui/ProgressBar/ProgressBar.module.less
```diff
@@ -10,11 +10,18 @@
 	direction: ltr;
 
 	.bar,
+	.separator,
 	.fill,
 	.load {
 		position: absolute;
 	}
 
+	.separator {
```
webOS101, line 19: For reference, the variable for the progress anchor is `var(--ui-progressbar-proportion-start)`

packages/ui/ProgressBar/ProgressBar.module.less
```diff
@@ -35,6 +42,11 @@
 			bottom: ~"calc(var(--ui-progressbar-proportion-start-background) * 100%)";
 			height: ~"calc(var(--ui-progressbar-proportion-end-background) * 100%)";
 		}
+
+		.separator {
+			width: 48px;
```
webOS101, line 47: My initial thought is that there shouldn't be any style applied in `ui`.  If we really felt there needed to be, these should become LESS variables and placed into the appropriate less file.

packages/ui/CHANGELOG.md
changgilee: I changed the comment

packages/ui/ProgressBar/ProgressBar.js
changgilee: The position of the separator has been modified to change according to the progressAnchor value

packages/ui/ProgressBar/ProgressBar.module.less
changgilee: The values are move from `ui` to `sandstone`

packages/ui/ProgressBar/ProgressBar.module.less
changgilee: It seems that `--ui-progressbar-proportion-start` does not synchronized `progressAnchor` value. So, I add `--ui-progressbar-progress-anchor` for handle it.

Djspaceg:
Honestly, I don't think we need to alter `enact/ui/ProgressBar` at all to support this feature in Sandstone. All of this can be done using the existing CSS variables and even an `::after` pseudo-element on `.bar` from within the Sandstone code.

If we were going to implement this feature here, we'd need it to be much more complete, and use a different name. "separator" isn't specific enough for the semantics of the feature it's offering. There would only be one "tick mark" that we could add, which doesn't allow for multiple ticks, like a graduated progress bar or a slider with "stops". If we're _really_ interested in adding support for such a feature, I'm in favor of that, but we should do it right and do it completely. Support for an arbitrary amount of separators (tick marks) with the ability to show or hide them.

changgilee:
@Djspaceg Is there a good prop name to recommend instead of `separator`?

Djspaceg:
@changgilee, I'd recommend not editing enact/ui at all and do all of this in Sandstone. We can add a `showAnchor` prop in sandstone, and use the ::after pseudo-element to display the visuals. That prop can add a `.showAnchor` class which configure the selectors in the ProgressBar.module.less file to display the anchor at `--ui-progressbar-proportion-start`. We don't need any new variables in enact/ui or anything.

PR #2744, opened 2020-04-07
PLAT-102572: Refactored ScrollbarBase

MikyungKim:
### Checklist

There was remained render prop in ScrollbarBase.
ScrollThumb is a somewhat ambiguous name for the component.

### Resolution
Refactored render prop pattern to custom hook `useScrollbar`
Renamed `ScrollThumb` to `ScrollbarTrack`
Renamed thumb related name to scrollbarTrack.

### Additional Considerations

### Links
PLAT-102572


packages/ui/useScroll/Scrollbar.js
```diff
@@ -33,87 +34,120 @@ const setCSSVariable = (element, variable, value) => {
...
-			hideThumbJob.current.start();
+		restProps: rest,
+		scrollbarProps: {
+			className: classNames(
+				className,
+				{[corner]: css.corner},
+				css.scrollbar,
+				vertical ? css.vertical : css.horizontal
+			),
+			ref: uiScrollbarContainerRef
```
ybsung, line 106: I don't think that we need the `ui` prefix for this `ref` as we did with `scrollContainerRef`. For `theme` refs, we could add a `theme` word like `useThemeScroll`.

packages/ui/useScroll/Scrollbar.js
```diff
@@ -33,87 +34,120 @@ const setCSSVariable = (element, variable, value) => {
...
+
+	useImperativeHandle(ref, () => {
+		const handles = pick([
+			'getContainerRef',
+			'showScrollbarTrack',
+			'startHidingScrollbarTrack',
+			'update'
+		], imperativeHandles);
+		return handles;
+	});
```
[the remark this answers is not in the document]
MikyungKim: 👍

packages/ui/useScroll/Scrollbar.js
MikyungKim: 👍

packages/ui/useScroll/Scrollbar.js
```diff
@@ -33,87 +33,110 @@ const setCSSVariable = (element, variable, value) => {
 };
 
 /**
- * An unstyled base component for a scroll bar.
+ * A custom hook that passes scrollbar behavior information as its render prop.
  *
- * @function ScrollbarBase
+ * @class
```
jeonghee27, line 38: why it is class?

jeonghee27: for documentation, let's keep it now.

MikyungKim: I added this to refer this as ui/useScroll.useScrollbar since there is no format for hooks. I think we could revisit all this type later.

PR #2747, opened 2020-04-13
PLAT-102254: Migrate ui/Cancelable to use hooks

ybsung:
### Checklist

Migrate ui/Cancelable to use hooks.

### Resolution

- Add `Cancel` for cancelable behavior.
- Add `useCancel` hook to manage the life cycle of the cancelable behavior.
- Adopt ui/Cancelable to use `useCancel`.
- Wrap `Cancelable` functional component with React.forwardRef()
- Replace `shallow` with `mount` in the unit test.
- Add `useCancel-specs.js`.

### Additional Considerations

### Links

PLAT-102254

### Comments


ybsung:
I reverted to pass `ref` with `React.forwardRef`.

ryanjduffy:
Responding to the questions in the description:

> 1> If there is the props, but no context member variable in the Cancel class, the hasPropsAndContext return false in https://github.com/enactjs/enact/blob/master/packages/core/handle/handle.js#L201 while executing the handleCancel (https://github.com/enactjs/enact/pull/2747/files#diff-7a9525ffab5b0d95304bb442d0aef0a8R27). So this.context = {} was defined in the Cancel constructor. Is there any efficient way without the context? Or it would be just fine.

Assigning `this.context` is the correct approach right now.

> 2> Could you let me know any history why the shallow and the prop were replaced with the mount and the invoke in https://github.com/enactjs/enact/pull/2739/files#diff-eed78280e9a44563cf719509de3fa7e9R119? I also replaced the shallow with the mount in my PR.

I can't recall off hand. In general, prefer `shallow` over `mount` but sometimes `mount` is required.

> 3> Cancelable's onCancel configuration could be a function or a string. But use Cancel accepts it as a configuration with onCancelWithStopPropagation property name to distinguish between the props's onCancel and the config's onCancel.

I'm not sure yet how this should be handled. I'm reviewing the PR in more detail now and will update this after that.

ybsung:
Hello @ryanjduffy 

> 3> Cancelable's onCancel configuration could be a function or a string. But use Cancel accepts it as a configuration with onCancelWithStopPropagation property name to distinguish between the props's onCancel and the config's onCancel.

Today I added one more commit to simplify `Cancel` class by moving forward('onCancel') out of it (https://github.com/enactjs/enact/pull/2747/commits/4de8e13ad3b25d0f4da0d688752da2edfae02dfd).

By doing it,
1. forward('onCancel') could be called not in `useCancel` but in `Cancelable` as we did in `Toggleable`.
2. Logically the `Cancel` class is much simpler.
3. The `useCancel` accepted the `onCancel` for props.onCancel and the `onCancelWithStopPropagation` for the `Cancelable` HOC's config before. I thought it could be confusable for users to distinguish them. But for now, the `onCancel` for the `Cancelable` HOC's config is only used in the `useCancel`.

PR #2749, opened 2020-04-15
PLAT-102264: Convert AnnounceDecorator to use hooks

ryanjduffy:
### Checklist

Adopt hooks for `ui/AnnounceDecorator`

### Additional Considerations
Chose to use `useCallback` rather than `useClass` in this case because it seemed like unnecessary complexity without any handlers and only one method. Thanks to @sugardave for reminding me of that earlier in the day. :)

### DIFF
--- a/packages/ui/AnnounceDecorator/useAnnounce.js
+++ b/packages/ui/AnnounceDecorator/useAnnounce.js
@@ -2,6 +2,34 @@ import React from 'react';
 
 import Announce from './Announce';
 
+/**
+ * Object returned by `useAnnounce`
+ *
+ * @typedef {Object} useAnnounceInterface
+ * @property {Function} announce Called to alert the user of behavior for accessibility
+ * @property {Node}     children An additional element which must be added to the component tree to
+ *                               support notifying the user
+ * @private
+ */
+
+/**
+ * Provides a method to alert the user of behavior for accessibility.
+ *
+ * ```
+ * 	function Component () {
+ *		const {announce, children} = useAnnounce();
+ *
+ *		return (
+ *			<Base onClick={() => announce('Unusually important notification for accessibility')}>
+ *				{children}
+ *			</Base>
+ *		);
+ *	}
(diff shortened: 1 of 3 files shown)


packages/ui/AnnounceDecorator/useAnnounce.js
```diff
@@ -0,0 +1,50 @@
...
+ *	}
+ * ```
+ *
+ * @returns {useAnnounceInterface}
+ * @private
+ */
+function useAnnouce () {
+	const ref = React.useRef(null);
+	const announce = React.useCallback(message => {
+		if (ref && ref.current) {
```
ybsung, line 36: Could we remove the `ref &` because the useRef returns {current: null} and the Boolean(`ref`) is always true ?

ryanjduffy: Yep. You are correct!

ryanjduffy: Seeing if I can "suggest" to myself to update it ... :)

```suggestion
		if (ref.current) {
```

packages/ui/AnnounceDecorator/useAnnounce.js
```diff
@@ -0,0 +1,50 @@
...
+ * ```
+ * 	function Component () {
+ *		const {announce, children} = useAnnounce();
+ *
+ *		return (
+ *			<Base onClick={() => announce('Unusually important notification for accessibility')}>
+ *				{children}
+ *			</Base>
+ *		);
+ *	}
```
webOS101, line 27: I feel like this sample glosses over how to combine incoming children with `children` from `useAnnounce()`.  Perhaps the `children` prop could be renamed to not conflict with incoming children?  The whole DX around `Announce` isn't that great.  Could we improve it?  Does the child even need to be a child?  Could it be a sibling of the `Base` component in this case, since we can pass the announce separately?

PR #2750, opened 2020-04-16
PLAT-102257: Migrate ui/Skinnable to use hooks

ybsung:
### Checklist

Migrate ui/Skinnable to use hooks

### Resolution

- Add `useSkinContext` hook like the way that we did in `useI18nContext`.

### Additional Considerations

- The `kind` will be a functional component in the https://github.com/enactjs/enact/pull/2745. So I don't think that we need to migrate from the `kind` to a functional component any more for the `Skinnable` kind.

- Switch the `SkinContext` and the `defaultConfig` to stick the `defaultConfig` to the `Skinnable` because the the `defaultConfig` is only used in the `Skinnable` and the `SkinContext` is used both in the `Skinnable` and `useSkinContext`.

### Links

PLAT-102257

### Comments

** CAUTION **

- The PR - https://github.com/enactjs/enact/pull/2745 should be merged first before this PR is merged.
- The CHANGELOG.md will be udpated.

### DIFF
--- a/packages/ui/Skinnable/Skinnable.js
+++ b/packages/ui/Skinnable/Skinnable.js
@@ -20,6 +20,48 @@ import classnames from 'classnames';
 import {objectify, preferDefined} from './util';
 
 /**
+ * Allows a component to respond to skin changes via the Context API
+ *
+ * Example:
+ * ```
+ * <App skin="dark">
+ * 	<Section>
+ * 		<Button>Gray Button</Button>
+ * 	<Section>
+ * 	<Popup skin="light">
+ * 		<Button>White Button</Button>
+ * 	</Popup>
+ * </App>
+ * ```
+ *
+ * @class SkinContext
+ * @memberof ui/Skinnable
+ * @hoc
+ * @public
+ */
+const SkinContext = React.createContext(null);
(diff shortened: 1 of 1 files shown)


ryanjduffy:
Thanks for the PR, @ybsung! In order to achieve the goal of collapsing our HOC stacks, I'd like to take this work a step further and create a hook that implements the logic within `Skinnable`.

I had started a version of this yesterday as well. It's not complete but I hope it gives a little more context on what I'm thinking.

https://github.com/enactjs/enact/compare/develop...feature/PLAT-102257-ryan?expand=1

ybsung:
> I had started a version of this yesterday as well. It's not complete but I hope it gives a little more context on what I'm thinking.

@ryanjduffy 
Thank you for filling the gap that I missed :)

ybsung:
Hello @ryanjduffy 

> Thanks for the PR, @ybsung! In order to achieve the goal of collapsing our HOC stacks, I'd like to take this work a step further and create a hook that implements the logic within `Skinnable`.

According to your comment, I'd like to close my PR. It would be much better to accept your branch. Thank you for taking this work.

PR #2751, opened 2020-04-16
PLAT-105257: Enact repo package.json cleanup

JayCanuck:
### Checklist

* If people `npm link` a library or `npm link-all`, then a bootstrap is required to fix the repo interlinking. This proves costly in time when done repeatedly/regularly.

### Resolution
* Adds 2 new `npm run <script>` entries:
    * `npm run bootstrap-link` - Includes the functionality of bootstrap with npm global linking built-in.  Should be used whenever linking to global NPM userspace is desired.
    * `npm run interlink` - Reconnects monorepo libraries together without the need to re-install/re-bootstrap.

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,9 +6,11 @@
   "scripts": {
     "lerna": "lerna",
     "bootstrap": "lerna bootstrap --concurrency 1",
+    "bootstrap-link": "npm run link-all && npm run interlink",
+    "interlink": "lerna link --loglevel error",
+    "link-all": "lerna --concurrency 1 exec -- npm --loglevel error --no-package-lock link",
+    "unlink-all": "lerna --concurrency 1 exec -- npm --loglevel error --no-package-lock unlink",
     "publish": "lerna publish --skip-npm --skip-git",
-    "link-all": "lerna --concurrency 1 --ignore enact-sampler exec -- npm --loglevel error link",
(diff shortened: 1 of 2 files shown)


ryanjduffy:
@JayCanuck How should we document these new commands so new developers can discover them?

package.json
```diff
@@ -6,9 +6,11 @@
   "scripts": {
     "lerna": "lerna",
     "bootstrap": "lerna bootstrap --concurrency 1",
+    "bootstrap-link": "npm run link-all && npm run interlink",
```
ryanjduffy, line 9: This sounds like "bootstrap + link" but seems to been just linking. The description sounds like it should be bootstrap + link as well. Can you clarify?

JayCanuck: "npm link"  installs as well as symlinks into global area. So this `bootstrap-link` has lerna `npm link` each package (installing deps and linking globally) then runs the `interlink` script which has lerna symlinking the packages all together appropriately.

JayCanuck:
These commands are solely for Enact developers, so we could include in the root level repo readme.md if desired. Any general public or even internal devs working on apps, under most circumstances, would use NPM releases, which do not need/use these.

ryanjduffy:
:+1: Thanks for clarifying. Let's add these to the README.md. That's a good call.

JayCanuck:
> 👍 Thanks for clarifying. Let's add these to the README.md. That's a good call.

Added :+1:

PR #2753, opened 2020-04-16
PLAT-102259: Convert Slottable to use hooks

ryanjduffy:
### Checklist

Convert `ui/Slottable` to use hooks. Not much else to say about it. :)

### DIFF
--- a/packages/ui/Slottable/Slottable.js
+++ b/packages/ui/Slottable/Slottable.js
@@ -45,23 +45,28 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 	const slots = config.slots;
 
 	// eslint-disable-next-line no-shadow
-	return function Slottable ({children, ...rest}) {
+	return function Slottable (props) {
+		// extract the slots into a new object but populating the default value to be null to allow
+		// the current "harmful" behavior below.
+		const slotProps = {children: props.children};
+		slots.forEach(k => (slotProps[k] = null));
+
 		// Slottable allows there to be other values in the destination slot and merges them.
 		// However, consumers can't avoid key warnings when merging the two lists so we should
 		// "consider this harmful" and not continue to support this with the hook and instead
 		// encourage using the slot as the default with the prop as a fallback as implemented by the
 		// hook.
-		const distributed = useSlots({slots, props: {children}});
+		const distributed = useSlots(slotProps);
 		slots.forEach(slot => {
 			const dist = distributed[slot];
-			const prop = rest[slot];
+			const prop = props[slot];
 			if (prop && dist) {
 				distributed[slot] = [].concat(prop, dist);
 			}
 		});
 
 		return (
-			<Wrapped {...rest} {...distributed} />
+			<Wrapped {...props} {...distributed} />
 		);
 	};
 });
--- a/packages/ui/Slottable/Slottable.js
+++ b/packages/ui/Slottable/Slottable.js
@@ -46,13 +46,12 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 
 	// eslint-disable-next-line no-shadow
 	return function Slottable ({children, ...rest}) {
-		const distributed = useSlots({slots, children});
-
 		// Slottable allows there to be other values in the destination slot and merges them.
 		// However, consumers can't avoid key warnings when merging the two lists so we should
 		// "consider this harmful" and not continue to support this with the hook and instead
-		// encourage "either/or". We should continue to support it here so this block ensures
-		// backwards compatibility.
+		// encourage using the slot as the default with the prop as a fallback as implemented by the
+		// hook.
+		const distributed = useSlots({slots, props: {children}});
 		slots.forEach(slot => {
 			const dist = distributed[slot];
 			const prop = rest[slot];
--- a/packages/ui/Slottable/Slottable.js
+++ b/packages/ui/Slottable/Slottable.js
@@ -62,6 +62,8 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 			const prop = props[slot];
 			if (prop && dist) {
 				distributed[slot] = [].concat(prop, dist);
+			} else if (prop) {
+				distributed[slot] = prop;
 			}
 		});
 
--- a/packages/ui/Slottable/Slottable.js
+++ b/packages/ui/Slottable/Slottable.js
@@ -60,7 +60,7 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 		slots.forEach(slot => {
 			const dist = distributed[slot];
 			const prop = props[slot];
-			if (prop && dist) {
+			if (prop != null && dist != null) {
 				distributed[slot] = [].concat(prop, dist);
 			} else if (prop) {
 				distributed[slot] = prop;
--- a/packages/ui/Slottable/Slottable.js
+++ b/packages/ui/Slottable/Slottable.js
@@ -54,4 +54,7 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 });
 
 export default Slottable;
-export {Slottable};
+export {
+	Slottable,
+	useSlots
+};
--- a/packages/ui/Slottable/Slottable.js
+++ b/packages/ui/Slottable/Slottable.js
@@ -8,84 +8,24 @@
  */
 
 import hoc from '@enact/core/hoc';
-import kind from '@enact/core/kind';
 import React from 'react';
-import warning from 'warning';
 
-// ** WARNING ** This is an intentional but likely dangerous hack necessary to clone a child while
-// omitting the `slot` property. It relies on the black box structure of a React element which could
-// change breaking this code. Without it, the slot property will cascade to a DOM node causing a
-// React warning.
-const cloneElement = (child) => {
-	const newProps = Object.assign({}, child.props);
-	delete newProps.slot;
+import useSlots from './useSlots';
 
-	return React.createElement(child.type, newProps);
-};
-
-const distributeChild = (child, index, slots, props) => {
-	let c, slot;
-	const hasSlot = (name) => slots.indexOf(name) !== -1;
-
-	if (!React.isValidElement(child)) {
-		return false;
-	} else if (child.props.slot) {
-		const hasUserSlot = hasSlot(slot = child.props.slot);
-		warning(hasUserSlot, 'The slot "%s" specified on %s does not exist', child.props.slot,
-			typeof child.type === 'string' ? child.type : (child.type.name || child.type.displayName || 'component')
-		);
-
-		if (hasUserSlot) {
-			c = cloneElement(child);
-		}
-	} else if (hasSlot(slot = child.type.defaultSlot)) {
-		c = child;
-	} else if (hasSlot(slot = child.type)) {
-		const propNames = Object.keys(child.props);
-		if (propNames.length === 1 && propNames[0] === 'children') {
-			c = child.props.children;
-		} else {
-			c = child;
-		}
-	}
-
-	if (c) {
-		let prop = props[slot];
-		if (prop) {
-			if (Array.isArray(prop)) {
-				prop.push(c);
-			} else {
-				prop = [prop, c];
-			}
-		} else {
-			prop = c;
-		}
-		props[slot] = prop;
-
-		return true;
-	}
-
-	return false;
-};
-
-const distribute = function (slots, props) {
-	if (slots) {
-		const children = [];
-		const adjusted = {...props};
-		React.Children.forEach(props.children, (child, index) => {
-			if (!distributeChild(child, index, slots, adjusted)) {
-				children.push(child);
-			}
-		});
-
-		adjusted.children = children.length > 0 ? children : null;
-		return adjusted;
-	}
-
-	return props;
-};
-
+/**
+ * Default config for `Slottable`.
+ *
+ * @memberof ui/Slottable.Slottable
+ * @hocconfig
+ * @public
+ */
 const defaultConfig = {
+	/**
+	 * Array of slot names which will be extracted from `children` and distributed to props.
+	 *
+	 * @type {String[]}
+	 * @memberof ui/Slottable.Slottable.defaultConfig
+	 */
 	slots: null
 };
 
@@ -103,15 +43,14 @@ const defaultConfig = {
  */
 const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 	const slots = config.slots;
-	return kind({
-		name: 'Slottable',
-		render: (props) => {
-			const adjusted = distribute(slots, props);
-			return (
-				<Wrapped {...adjusted} />
-			);
-		}
-	});
+
+	// eslint-disable-next-line no-shadow
+	return function Slottable (props) {
+		const distributed = useSlots({slots, children: props.children});
+		return (
+			<Wrapped {...props} {...distributed} />
+		);
+	};
 });
 
 export default Slottable;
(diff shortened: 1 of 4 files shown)


packages/ui/Slottable/tests/useSlots-specs.js
```diff
@@ -0,0 +1,212 @@
...
+			expect(actualErrorsLength).toBe(expectedErrorLength);
+
+			const actualError = console.error.mock.calls[0][0];
+			const expectedError = 'Warning: The slot "c" specified on div does not exist';
+
+			expect(actualError).toBe(expectedError);
+		}
+	);
+
+	test(
```
webOS101, line 178: Might be nice to see tests for multiple children with the same slot.

packages/ui/Slottable/Slottable.js
```diff
@@ -103,16 +43,32 @@ const defaultConfig = {
...
+		// Slottable allows there to be other values in the destination slot and merges them.
+		// However, consumers can't avoid key warnings when merging the two lists so we should
+		// "consider this harmful" and not continue to support this with the hook and instead
+		// encourage "either/or". We should continue to support it here so this block ensures
+		// backwards compatibility.
+		slots.forEach(slot => {
+			const dist = distributed[slot];
+			const prop = rest[slot];
+			if (prop && dist) {
+				distributed[slot] = Array.isArray(dist) ? [prop, ...dist] : [prop, dist];
```
webOS101, line 60: Do we need to check if `prop` is an array as well?  Also, should we add any tests around this functionality?

webOS101: Maybe `distributed[slot] = [].concat(prop, dist)`?

webOS101: Maybe it's fine as-is.  Technically `dist` could have been a single slot that contains an array.  It could be a bit difficult to figure out.

ryanjduffy: The current code does seem to handle this scenario (surprisingly to me). I'll push an update!

ryanjduffy: Thanks for the tip, @webOS101! Fixed using your suggestion.

packages/ui/Slottable/tests/useSlots-specs.js
ryanjduffy: I added a test for this as well.

ryanjduffy:
Regarding incorporating into `kind`, generally we've tried to use `Slottable` on the outer boundary of the decorated component since it's purpose is primarily a DX improvement to source props from children and we want to operate on the "true" API everywhere else.

You're right that `useSlots` doesn't have the typical signs of a hook (state/effects). I think there is some simplicity in treating it as such anyway so that all "behaviors" are hooks but I'm not convinced it's the best approach. This _might_ be the only case where a HOC could be converted to a utility so perhaps killing the hook makes sense to avoid the exception. On the other hand, treating it as a hook now means that we could add state/effects later if we needed to without changing the API.

Thoughts?

packages/ui/Slottable/Slottable.js
```diff
@@ -103,16 +43,32 @@ const defaultConfig = {
...
+		// Slottable allows there to be other values in the destination slot and merges them.
+		// However, consumers can't avoid key warnings when merging the two lists so we should
+		// "consider this harmful" and not continue to support this with the hook and instead
+		// encourage "either/or". We should continue to support it here so this block ensures
+		// backwards compatibility.
+		slots.forEach(slot => {
+			const dist = distributed[slot];
+			const prop = rest[slot];
+			if (prop && dist) {
+				distributed[slot] = [].concat(prop, dist);
```
Djspaceg, line 60: This is weird. I don't think I like it. I didn't even realize the old version had this behavior, and I'm probably the most frequent user of Slots. Mostly don't like it because A) I don't know that it's documented anywhere B) you can't control the order, so if it doesn't accidentally do what you want perfectly, you're SOL.

I feel like it's totally acceptable for Slottable (useSlots) to simply decide one of the two is "more valid" than the other, the slot or the prop and always prefer that one, provided it's consistent, predictable, overridable, and documented. I thin the following patterns could support those requirements with intuitive DX:

```jsx
<Component
   title="My Title"
   bodyText="potatoes are tasty"
>
   <bodyText>Only with salt on them</bodyText>
   <buttons><button>A Button!</button></buttons>
</Component>
```
The props argument can accept "fallback" values for if child-slots are not provided by the end-developer, in this example we have only props with names mapping 1:1 with existing slots.
```javascript
{children, title, bodyText} = props;
slots = ['title', 'bodyText', 'buttons'];
distributed = useSlots({slots, {children, title, bodyText}); // didn't include "buttons" prop, don't care about `buttons` prop in this case

// where distributed ==
{
   // title is fed back as a prop, so it's indistinguishable from whether it came in as a prop or a child-slot
   title: 'My Title',

   // bodyText child-slot is preferred since it seems more intentional, given that if the component author had wanted this to just be a prop, it wouldn't have been one, yet can be supported if a child-slot isn't provided, for convenience
   bodyText: 'Only with salt on them',

   // buttons comes in as a child-slot and we return it into the distributed collection. The author expects no comparison or override, since it wasn't intentionally included in the useSlots props argument.
   buttons: <button>A Button!</button>
}
```

ryanjduffy: Agreed! I'll make an update and we can review it again.

ryanjduffy:
@webOS101 I pushed a material change to the API so I'd like to cover it with you before sending back to @Djspaceg for integration.

packages/ui/Slottable/useSlots.js
```diff
@@ -77,36 +78,60 @@ function distribute (slots, children) {
...
 
 /**
  * Distributes `children` into the configured `slots`.
  *
+ * `useSlots` iterates over all of the `children` in `props` and distributes any children based on
+ * the followig rules:
+ *
+ * * If the child has a `slot` property matching a valid slot, or
+ * * If the component for the child has the `defaultSlot` static member matching a valid slot, or
+ * * If the child component is a string matching a valid slot.
```
webOS101, line 108: This wording is a little unclear.    'child component's type is  a string', maybe?

ryanjduffy: Fixed!

ryanjduffy:
Examples of alternatives:

```js
// 1. First API but without the fallback to prop value
const {slotBefore = props.slotBefore, slotAfter = props.slotAfter, children} = useSlots({slots: ['slotBefore', 'slotAfter'], children: props.children});
```

```js
// 2. Current API with fallback to prop value
const {slotBefore, slotAfter, children} = useSlots({slots: ['slotBefore', 'slotAfter'], props});
```

The only difference between 2 and 3 is using two args vs an object arg. We had defined a default position that hooks would take a single object arg so that favors 2 but 3 is more terse. I think I prefer 2 because the object keys are a bit more clear of an API than fn args.

```js
// 3. Separate arg API with current behavior to fallback to prop value
const {slotBefore, slotAfter, children} = useSlots(['slotBefore', 'slotAfter'], props);
```

In 4, I like that it only passes the necessary prop and each prop is explicitly used triggering our lint validation. However, it's more verbose and you can't destructure both `props` and the return from useSlots without renaming things which is annoying.

```js
// 4. Explicit prop API with current behavior to fallback to prop value
const {slotBefore, slotAfter, children} = useSlots({
	slotBefore: props.slotBefore,
	slotAfter: props.slotAfter,
	children: props.children,
});
```

Using `pick()` simplifies things a bit and avoids sending all the props into `useSlots` but it requires `ramda` and is effectively the same as the current API in 2.

```js
// 5. Explicit prop API using ramda pick() with current behavior to fallback to prop value
const {slotBefore, slotAfter, children} = useSlots(pick(['slotBefore', 'slotAfter', 'children'], props));
```

I'm leaning toward the current implementation but sharing for feedback from the team first.

webOS101:
I prefer the slightly less verbose option 3 but don't feel super-duper strongly about it.

Djspaceg:
I think I prefer #3. Here's why.
```
// 2. Current API with fallback to prop value
const {slotBefore, slotAfter, children} = useSlots({slots: ['slotBefore', 'slotAfter'], props});
```
In the #2 version, I think it only looks simple because we happen to have named the incoming `props` argument to the wrapping function "props" and that coincides with the key that the hook is looking for. If we didn't, and did say: "rest" instead, it would look more awkward:
```
const {slotBefore, slotAfter, children} = useSlots({slots: ['slotBefore', 'slotAfter'], props: rest});
```
And we could optionally pass in different values, targeting them one at a time, which makes it look much more complicated:
```
const {slotBefore, slotAfter, children} = useSlots({slots: ['slotBefore', 'slotAfter'], props: {children: someOtherChildren, slotBefore, slotAfter: customSlotAfter}});
```
I think the _extremely mildly irksome_ thing is needing to supply that "props" key (it would be a total non-issue in most cases). The #2 approach just _seems_ more bulky. It does offer the ability to extend the API more easily over time though, which I appreciate.

What I like about #3 is it does still lend itself to that "hooks use 1 arg" mentality, essentially. The second argument, an object of props is technically optional, which I think makes good sense. It's like how the props vs props+context works. The single arg covers the primary usecase, and the bonus argument gets you _extra_ functionality. (Side benefit is we don't have to specify the word "slots" or "props", and also don't have to deal with objects within objects; it's flatter.

Ultimately I would be content with 2, 3, or even the ramda thing (once I understood it better). However, I do prefer #3.

ryanjduffy:
The second arg in 3 is not optional. Without the props, `useSlots` can't do anything as it needs the props obj to get at `children` and the prop fallbacks.

ryanjduffy:
I agree that the use of `props` seems easier in this simple scenario and may not pan out as such in practice.

Djspaceg:
> The second arg in 3 is not optional. Without the props, `useSlots` can't do anything as it needs the props obj to get at `children` and the prop fallbacks.

You're right. I forgot about the passing of `children`.

ryanjduffy:
I'm now leaning toward 4 after the discussion here. Consider the following slightly more complex scenario. We're likely to want to deconstruct in the function (or immediately after at the top of the block) and there is some readability value in using `slots` elsewhere if you're interacting with the "computed" value or spreading it onto the wrapped component if it's just a pass through.

```
function ComponentDecorator ({slotBefore, slotAfter, children, myProp, anotherProp, ...rest}) {
    const slots = useSlots({slotBefore, slotAfter, children});
    const someComputedValue = fn(myProp, anotherProp);

    return <Wrapped {...rest} {...slots} myProp={someComputedValue} />;
```

Djspaceg:
I like this option too. I was a bit concerned about it initially since the slot names aren't explicit, but they sort of are, actually. It could lead to some interesting scenarios and code (that we, as a framework, shouldn't do internally):
```javascript
const everythingIsSlots = useSlots(props);
```
Not that we'd do that, but it technically wouldn't be forbidden.

I'm changing my answer. I like #4 the most now. :)

webOS101:
Wow, a real world example looks so much better.  It's just what I had been hoping for but it didn't seem like it worked out.  :+1:

ryanjduffy:
Pushed an update that uses the 4th option. I haven't checked over docs and run unit tests in sandstone yet. I'll do that tomorrow and then route back for integration.

packages/ui/Slottable/useSlots.js
```diff
@@ -82,7 +83,7 @@ function distribute (slots, {children, ...fallback}) {
 	// will append to existing props and we want the distributed value to override the fallback
 	// value.
 	return {
-		...fallback,
+		...slots,
```
webOS101, line 86: What about deleting `slots` props whose value is `undefined`?

Djspaceg: Yeah, that's probably necessary.

packages/ui/Slottable/tests/useSlots-specs.js
```diff
@@ -9,14 +9,15 @@ import useSlots from '../useSlots';
 describe('useSlots', () => {
 
 	test('should distribute children with a \'slot\' property', () => {
-		function Component (props) {
-			const {a, b, c} = useSlots({slots: ['a', 'b', 'c'], props});
+		// eslint-disable-next-line enact/prop-types
```
webOS101, line 12: Can you add a test that `children` is actually modified on output?

packages/ui/Slottable/useSlots.js
```diff
@@ -72,11 +72,7 @@ function distribute ({children, ...slots}) {
 			}
 		});
 
-		if (remaining.length > 0) {
-			props.children = remaining;
-		} else {
-			delete props.children;
-		}
+		props.children = remaining.length === 0 ? null : remaining;
```
webOS101, line 75: I'm not sure this was strictly necessary but it's shorter than the previous so... win?

packages/ui/Slottable/Slottable.js
```diff
@@ -62,6 +62,8 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 			const prop = props[slot];
 			if (prop && dist) {
 				distributed[slot] = [].concat(prop, dist);
+			} else if (prop) {
+				distributed[slot] = prop;
```
webOS101, line 66: Why is this necessary?  Don't we spread `props` below to achieve this same effect?

ryanjduffy: that's fair!

packages/ui/Slottable/useSlots.js
ryanjduffy: I've added some inline docs for this. In short, retaining `children` ensure that if the author does `<Wrapped {...props} {...distributed} />` that `props.children` is overwritten by `distributed.children` (since we already distributed the children to other props).

packages/ui/Slottable/Slottable.js
```diff
@@ -60,7 +60,7 @@ const Slottable = hoc(defaultConfig, (config, Wrapped) => {
 		slots.forEach(slot => {
 			const dist = distributed[slot];
 			const prop = props[slot];
-			if (prop && dist) {
+			if (prop != null && dist != null) {
 				distributed[slot] = [].concat(prop, dist);
 			} else if (prop) {
```
webOS101, line 65: I think this one should be `!= null`, too.

packages/ui/Slottable/useSlots.js
webOS101: That makes sense!

packages/ui/Slottable/Slottable.js
```diff
@@ -103,16 +43,38 @@ const defaultConfig = {
...
+		const slotProps = {children: props.children};
+		slots.forEach(k => (slotProps[k] = null));
+
+		// Slottable allows there to be other values in the destination slot and merges them.
+		// However, consumers can't avoid key warnings when merging the two lists so we should
+		// "consider this harmful" and not continue to support this with the hook and instead
+		// encourage using the slot as the default with the prop as a fallback as implemented by the
+		// hook.
+		const distributed = useSlots(slotProps);
+		slots.forEach(slot => {
```
webOS101, line 60: Do we need to ensure that `children` never appears in `slots`?

packages/ui/Slottable/Slottable.js
ryanjduffy: Yeah, i agree.

packages/ui/Slottable/Slottable.js
ryanjduffy: We don't currently but that's a reasonable condition to add.

packages/ui/Slottable/tests/useSlots-specs.js
ryanjduffy: Done!

packages/ui/Slottable/Slottable.js
ryanjduffy: Actually, I'd like to skip this. We don't test for this currently and it isn't possible with the hook implementation so it's out of scope and really edge case-y.

webOS101: Makes sense!  Was just asking as food for thought.

PR #2754, opened 2020-04-19
PLAT-102261: Migrate ui/A11yDecorator to use hooks

ybsung:

### Checklist

Migrate ui/A11yDecorator to use hooks

### Resolution

Add useA11y hook.

### Additional Considerations

N/A

### Links

PLAT-102261

### Comments

The `useA11y` returns `a11y` object with the `aria-label` property so that the object could be spreaded into the JSX directly.

### DIFF
--- a/packages/ui/A11yDecorator/A11yDecorator.js
+++ b/packages/ui/A11yDecorator/A11yDecorator.js
@@ -6,9 +6,10 @@
  */
 
 import hoc from '@enact/core/hoc';
-import kind from '@enact/core/kind';
 import PropTypes from 'prop-types';
 import React from 'react';
+
+import useA11y from './useA11y';
 
 /**
  * Default config for {@link ui/A11yDecorator.A11yDecorator}.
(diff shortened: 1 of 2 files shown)


packages/ui/A11yDecorator/useA11y.js
```diff
@@ -0,0 +1,62 @@
...
+
+/**
+ * Manages a accessibility label.
+ * The accessibility label is decorated with `accessibilityPreHint`, `accessibilityHint`, and `content`.
+ *
+ * @param {useA11yConfig} config Configuration options
+ * @returns {useA11yInterface}
+ * @private
+ */
+const useA11y = ({accessibilityHint, accessibilityPreHint, 'aria-label': ariaLabel, content}) => {
```
ryanjduffy, line 50: Suggest we shorten these to be `hint`, `preHint`, `label`, and `content`. The extra `accessibility` is redundant given the `useA11y` context.

packages/ui/A11yDecorator/A11yDecorator.js
```diff
@@ -60,66 +61,54 @@ const defaultConfig = {
 const A11yDecorator = hoc(defaultConfig, (config, Wrapped) => {
 	const {prop} = config;
 
-	return kind({
-		name: 'A11yDecorator',
+	// eslint-disable-next-line no-shadow
+	const A11yDecorator = React.forwardRef(({accessibilityHint, accessibilityPreHint, 'aria-label': ariaLabel, [prop]: content, ...rest}, ref) => {
```
ryanjduffy, line 65: Let's use `function A11yDecorator () { ... }` to assign the name here rather than explicitly after the fact.

Also, can you expand on why we've included `forwardRef` here?

seunghoh:
> Couple comments/questions. More importantly, I'm curious about the utility of this feature for our consumers. Do we see apps using `A11yDecorator`? The hook is rudimentary and I'm curious if it will provide much value. Are there a set of a11y features (e.g. `Announce`) that we should combine into one hook to give a more complete capability?

@ryanjduffy Good point...
I don't see any app is using `A11yDocorator` (but can't 100% sure. just checked TV and Auto apps in the webOS) It looks just migrated from Enyo to Enact as a legacy supporting perspective. I don't see much value either since Enact app can edit `aria-label` directly wheres Enyo app couldn't. I can't say let's remove A11yDecorator now but, I prefer not to migrate hook on this

PR #2755, opened 2020-04-20
ENYO-6462: Fix spinner to align center

xbinary:
### Checklist

If the containing block of  `Spinner`has `padding-left` , the `Spinner` moves to the right by the size of the `padding-left`.
Because `left` is set to `padding-left` value. (for now, `left` is set to default value. not specified)
In moonstone, there is no issue because the `section` of `Panel` has no `padding-left`.

### Resolution
`left` is set to 0. (if rtl, right is set to 0)

### Additional Considerations

### Links
ENYO-6462

### DIFF
--- a/packages/ui/Spinner/Spinner.module.less
+++ b/packages/ui/Spinner/Spinner.module.less
@@ -8,12 +8,9 @@
 
 	// centered
 	&.centered {
-		left: 0;
-		margin: 0;
-		margin-left: 50%;
+		left: 50%;
 		position: absolute;
 		top: 50%;
-		-webkit-transform: translateX(-50%) translateY(-50%);
 		transform: translateX(-50%) translateY(-50%);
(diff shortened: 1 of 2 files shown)


xbinary:
@Djspaceg - I apply your feedback.
In addition, `margin` is also deleted. This seems to have no effect anymore.

packages/ui/Spinner/Spinner.module.less
```diff
@@ -23,9 +21,8 @@
 	:global(.enact-locale-right-to-left) & {
 		// centered
 		&.centered {
-			margin: 0;
-			margin-right: 50%;
-			-webkit-transform: translateX(50%) translateY(-50%);
+			left: initial;
```
Djspaceg, line 24: I suppose we do need to reset that assignment. Good find!

PR #2756, opened 2020-04-20
PLAT-102260: Migrate IdProvider to use hooks

ybsung:
### Checklist

Migrate IdProvider to use hooks

### Resolution

- Add `useIdProvider` to return the object including `generateId` function.

### Additional Considerations

### Links

PLAT-102260


packages/ui/internal/IdProvider/useId.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {useIdProviderConfig} config Configuration options
+ * @returns {useIdProviderInterface}
+ * @private
+ */
+const useIdProvider = ({prefix}) => {
+	const provider = useClass(Provider, prefix);
+
+	React.useEffect(() => {
+		return () => provider.unload();
+	}, [provider]);
```
webOS101, line 36: Technically not required since `provider` doesn't change, but then lint will complain.  I'm not sure if there's any shortcut in the comparison logic if the array is empty.

ryanjduffy: Agreed. Seemed easiest to follow the rule since a strict equality check on the object isn't an expensive operation.

PR #2757, opened 2020-04-20
PLAT-105314: Fix ViewManager to only fire onTransition once per transition

ryanjduffy:
### Checklist

ViewManager fired two `onTransition` events for every transition

### Resolution
* Prevent firing `onTransition` for views that weren't transitioning
* Add some useful data to the event payload.

### DIFF
--- a/packages/ui/ViewManager/TransitionGroup.js
+++ b/packages/ui/ViewManager/TransitionGroup.js
@@ -293,10 +293,12 @@ class TransitionGroup extends React.Component {
 	}
 
 	completeTransition (key) {
-		delete this.currentlyTransitioningKeys[key];
+		if (key in this.currentlyTransitioningKeys) {
+			delete this.currentlyTransitioningKeys[key];
 
-		if (Object.keys(this.currentlyTransitioningKeys).length === 0) {
-			forwardOnTransition(null, this.props);
+			if (Object.keys(this.currentlyTransitioningKeys).length === 0) {
+				forwardOnTransition(null, this.props);
+			}
 		}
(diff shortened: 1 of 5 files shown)


packages/ui/ViewManager/TransitionGroup.js
```diff
@@ -293,10 +293,12 @@ class TransitionGroup extends React.Component {
 	}
 
 	completeTransition (key) {
-		delete this.currentlyTransitioningKeys[key];
+		if (key in this.currentlyTransitioningKeys) {
+			delete this.currentlyTransitioningKeys[key];
 
-		if (Object.keys(this.currentlyTransitioningKeys).length === 0) {
-			forwardOnTransition(null, this.props);
+			if (Object.keys(this.currentlyTransitioningKeys).length === 0) {
```
webOS101, line 299: Theoretically, this can still fire a spurious event if a panel is removed that was transitioning and a new transition will begin.  But, it's probably OK?

ryanjduffy:
I don't think this is quite right yet. I think there's another scenario that will cause problems.

webOS101:
I do think this looks cleaner than my previous approach and covers more scenarios.  I'm not sure what happens when animations are disabled.

ryanjduffy:
I suspect there's an issue with multiple visible views transitioning to a single view (e.g. with 
 moonstone's breadcrumb decorator) but I wasn't able to create a unit test to verify that. I think there's more refactoring to do to fix this better but I'd like to consider moving this ahead as is to at least address the common case.

PR #2758, opened 2020-04-21
PLAT-102253: Migrate spotlight/Spottable to use hooks

ybsung:
### Checklist

Migrate spotlight/Spottable to use hooks

### Resolution

Add `Spot` for cancelable behavior.
Add `useSpot` hook to manage the life cycle of the Spottable behavior.
Adopt `spotlight/Spottable` to use `useSpot`.
Add `useSpot-specs.js`.

### Additional Considerations

1. I updated the default value to `true` for the `emulateMouse` in the JS Doc.
2. I renamed `focusedWhenDisabled` as `isFocusedWhenDisabled` like `isFocused`.
3. <Question> Do we need to set `selectCancelled` to `false` when `Spot.unload()` is called?
4. `onMouseUp` was forwarded in the `componentDidUpdate` before. To do it with the `useSpot`, the `onSelectionCancel` prop with `onMouseUp` function is passed to the `useSpot` now.

Before
```
// Spottable.js
componentDidUpdate (prevProps) {
	...
	if (this.isFocused && this.props.disabled && lastSelectTarget === this && !selectCancelled) {
		selectCancelled = true;
		forward('onMouseUp', null, this.props);
	}
```
After
```
// Spottable.js
const spot = useSpot({..., onSelectionCancel: rest.onMouseUp, ...});

// Spot.js
didUpdate = () => {
	...
	if (this.isFocused && this.props.disabled && lastSelectTarget === this && !selectCancelled) {
		selectCancelled = true;
		forward('onSelectionCancel', null, this.props);
	}
```
5. There is the concern about my comment ( https://github.com/enactjs/enact/pull/2758#discussion_r412094605 ) for the following Warning.
```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
      
      Check the render method of `Spottable`.
          in mockConstructor (at Spottable.js:148)
          in Spottable (created by WrapperComponent)
          in WrapperComponent
```
### Links

PLAT-102253


packages/spotlight/Spottable/useSpot.js
```diff
@@ -0,0 +1,97 @@
...
+	const mutableRef = React.useRef({
+		prevSpotlightDisabled: spotlightDisabled,
+		spotlightDisabled
+	});
+
+	spot.setContext(mutableRef);
+
+	React.useEffect(() => {
+		// eslint-disable-next-line react/no-find-dom-node
+		spot.load(ReactDOM.findDOMNode(ref.current));
```
ybsung, line 70: Hello @ryanjduffy 

I had one problem to migrate Spottable to use Hooks.
In detail, I don't know how to change the `ReactDOM.findDOMNode(this)` properly on the `useSpot`. With the PR, there is the following warning. Do we have to keep using the `Spottable` as the React Component?

```
Warning: Function components cannot be given refs. Attempts to access this ref will fail. Did you mean to use React.forwardRef()?
      
      Check the render method of `Spottable`.
          in mockConstructor (at Spottable.js:148)
          in Spottable (created by WrapperComponent)
          in WrapperComponent
```

ybsung: In my opinion, there is no problem that the `Spottable` works. And there are no errors to run lint in enact and sandstone. So could we go ahead and fix it later? Or should we have to fix it in this PR?

packages/spotlight/Spottable/Spottable.js
```diff
@@ -65,7 +38,7 @@ const defaultConfig = {
 	 * to Spotlight 5-way events.
 	 *
 	 * @type {Boolean}
-	 * @default false
```
webOS101, line 68: :doh:

packages/spotlight/Spottable/useSpot.js
ybsung: To maintain `ref` properly, I defined `Spottable` based on React.Component and passed the `componentRef` from `ReactDOM.findDOMNode(this)` to `SpottableBase` and `useSpot`.

```
	class Spottable extends React.Component {
		componentDidMount () {
		}

		get componentRef () {
			return this.node;
		}

		render () {
			return <SpottableBase {...this.props} componentRef={this.componentRef} />;
		}
	}
```

0x64 (approved):
:lgtm:

packages/spotlight/Spottable/Spottable.js
```diff
@@ -96,353 +103,178 @@ const defaultConfig = {
...
+			onSpotlightDisappear,
+			onSpotlightDown,
+			onSpotlightLeft,
+			onSpotlightRight,
+			onSpotlightUp,
+			selectionKeys,
+			spotlightDisabled,
+			...rest
+		} = props;
+		const spot = useSpot({
```
jeonghee27, line 120:
```suggestion
		const hook = useSpot({
```

PR #2759, opened 2020-04-22
PLAT-102257: Migrate Skinnable to hooks

ryanjduffy:
### Checklist

Migrates Skinnable to use hooks

### Additional Considerations
This is the first hook (I believe) to expose a context provider. I've gone with `provideSkins` as a member of the returned object from the hook. I like that this avoids exporting the context and doesn't require the user to assemble the provider but it's still a little awkward looking

```jsx
const skins = useSkins(config);
return skins.provideSkins(
    <Component {...props} />
);
```

This model works okay with stacking multiple providers too
```
skins.provideSkins(
    other.provideOther(
        <Comp />
    )
)
```

I've also adjusted the config parameter names a bit for consistency. We had `skins` and `allowedVariants` but both referred to the allowed values. I chose to use `skins` and `variants` rather than `allowedSkins` and `allowedVariants`. I'm not :100: on this, however, since the config also accepts the current values in `skin` and `skinVariants` so we have `skins`/`skin` and `variants`/`skinVariants`.

### DIFF
--- a/packages/ui/Skinnable/useSkins.js
+++ b/packages/ui/Skinnable/useSkins.js
@@ -56,7 +56,7 @@ const SkinContext = React.createContext(null);
  */
 
 /**
- * Deteremines the effective skin and skin variants for a component and provides a method to provide
+ * Determines the effective skin and skin variants for a component and provides a method to provide
  * those values to other contained components in this subtree.
  *
  * ```
--- a/packages/ui/Skinnable/useSkins.js
+++ b/packages/ui/Skinnable/useSkins.js
@@ -19,7 +19,6 @@ import {determineSkin, determineVariants, getClassName} from './util';
  *
  * @class SkinContext
  * @memberof ui/Skinnable
- * @hoc
  * @private
  */
 const SkinContext = React.createContext(null);
@@ -29,12 +28,12 @@ const SkinContext = React.createContext(null);
  *
  * @typedef {Object} useSkinsConfig
  * @memberof ui/Skinnable
- * @property {String}   defaultSkin     Default skin name when none has been specified or inherited.
- * @property {String[]} defaultVariants Default variants when none have been specified or inherited.
- * @property {String}   [skin]          Specific skin to apply to this instance which will take precedence over the default or the inherited value.
- * @property {String[]} skins           List of allowed skins
- * @property {String[]} [skinVariants]  Specific variants to apply to this instance which will take precedence over the default or the inherited value.
- * @property {String[]} variants        List of allowed variants
+ * @property {String}   defaultSkin       Default skin name when none has been specified or inherited.
+ * @property {String[]} [defaultVariants] Default variants when none have been specified or inherited.
+ * @property {String}   [skin]            Specific skin to apply to this instance which will take precedence over the default or the inherited value.
+ * @property {String[]} skins             List of allowed skins
+ * @property {String[]} [skinVariants]    Specific variants to apply to this instance which will take precedence over the default or the inherited value.
+ * @property {String[]} [variants]        List of allowed variants
  * @private
  */
 
--- a/packages/ui/Skinnable/useSkins.js
+++ b/packages/ui/Skinnable/useSkins.js
@@ -1,0 +1,86 @@
+import React from 'react';
+
+import {determineSkin, determineVariants, getClassName} from './util';
+
+/**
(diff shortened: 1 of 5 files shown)


packages/ui/Skinnable/useSkins.js
```diff
@@ -0,0 +1,112 @@
...
+ * 	<Section>
+ * 	<Popup skin="light">
+ * 		<Button>White Button</Button>
+ * 	</Popup>
+ * </App>
+ * ```
+ *
+ * @class SkinContext
+ * @memberof ui/Skinnable
+ * @hoc
```
webOS101, line 22: Is this really a hoc?  Is it a class?  I think we need a different thing for this.

packages/ui/Skinnable/useSkins.js
```diff
@@ -0,0 +1,112 @@
...
+ */
+const SkinContext = React.createContext(null);
+
+/**
+ * Configuration for `useSkins`
+ *
+ * @typedef {Object} useSkinsConfig
+ * @memberof ui/Skinnable
+ * @property {String}   defaultSkin     Default skin name when none has been specified or inherited.
+ * @property {String[]} defaultVariants Default variants when none have been specified or inherited.
```
webOS101, line 33: is `defaultVariants` required?

packages/ui/Skinnable/useSkins.js
```diff
@@ -0,0 +1,112 @@
...
+ * @property {String}   [skin]          Specific skin to apply to this instance which will take precedence over the default or the inherited value.
+ * @property {String[]} skins           List of allowed skins
+ * @property {String[]} [skinVariants]  Specific variants to apply to this instance which will take precedence over the default or the inherited value.
+ * @property {String[]} variants        List of allowed variants
```
webOS101, line 37: Similarly, is `variants` actually required?

packages/ui/Skinnable/useSkins.js
```diff
@@ -0,0 +1,112 @@
...
+ *     defaultSkin: 'neutral',
+ *     defaultVariants: [],
+ *     skins: ['neutral', 'bold'],
+ *     variants: ['highContrast'],
+ *     skin: props.skin,
+ *     skinVariants: props.skinVariants
+ *   });
+ *
+ *   return skins.provideSkins(
+ *     <Component className={classnames(props.className, skins.className)} />
```
webOS101, line 74: Would this be clearer showing `provideSkins` and `className` destructured from the call?  Along the same vein, can we provide a utility function from `useSkins` so developers don't have to call `classnames`?  Maybe return a utility `mergeSkinClasses()`?

packages/ui/Skinnable/util.js
```diff
@@ -26,7 +27,58 @@ const objectify = (arg) => {
 
 const preferDefined = (a, b) => ((a != null) ? a : b);
 
+function determineSkin (defaultSkin, authorSkin, parentSkin) {
+	return authorSkin || defaultSkin || parentSkin;
```
webOS101, line 31: `defaultSkin` here is the local default, not the app default.  I'm not sure if there's a way to clarify that or if it needs to, but I was confused at first.

packages/ui/Skinnable/useSkins.js
ryanjduffy: It's private so I just removed the pragma.

packages/ui/Skinnable/useSkins.js
ryanjduffy: Nope

packages/ui/Skinnable/useSkins.js
ryanjduffy: Nope

packages/ui/Skinnable/util.js
ryanjduffy: Adding some docs for this method and the one below.

packages/ui/Skinnable/useSkins.js
ryanjduffy: I can see the utility of providing that method but I think it goes beyond the scope of this feature. Most likely, consumers will need other CSS class concatenation and would rather not introduce overlapping capabilities in the hooks.

Sound reasonable?

webOS101: Seems reasonable!

PR #2762, opened 2020-04-28
PLAT-104500: Add support for hooks within core/kind

ryanjduffy:
Update to #2745 with the following changes:
* Correct default export of `ui/Layout` to use the decorated version rather than the base version
* Update `useHandlers` to return the result of the handler as we did before. This resolves the failure when using `handlers` for the `itemRenderer` for a `VirtualList`.

### Checklist

Adds support for using hooks within `core/kind`

### Resolution
* Adds (private) `core/useClass` from `ui/Toggleable`
* Adds (private) `core/useHandlers` to support bound handlers compatible with `core/handle`
* Refactors `core/kind` back to a functional component
* Adds unit tests for the above

### Additional Considerations
This *does not* address the eslint warning from `react/rules-of-hooks` when using a hook within the render method of kind.


webOS101:
```
// This can be replaced with the kind config contextType when it's supported
PickerButtonBase.contextType = MarqueeControllerContext;
```
How old is that?!

ryanjduffy:
Pretty old, right?

Forgot I made that change in sandstone. i'll push a PR for that.

ryanjduffy:
Should we make this opt-in via a config for 3.x and then either opt-out or the only supported config for 4.0?

webOS101:
🤔 We _could_ have an alternate export.  I'm open to making it optional for 3.x.

ryanjduffy:
Or, if we think the risk is low, we could make it opt-out now as an escape hatch if devs run into errors. I think making it optional is the right choice even though it is more work for us if we want to adopt broadly.

Djspaceg:
Personally, I'd prefer the new functionality become the default behavior and provide a backward compatibility opt-out. It stabilizes the API sooner and establishes the future more concretely.

ryanjduffy:
I'd prefer that but I think there's a non-trivial amount of risk as illustrated by the fact that we had regressions from this change.

ryanjduffy:
I chose to add a new key -- `functional` -- to `core/kind` to allow authors to opt-in to hook support. My plan is to remove this key in 4.0 and only support functional components.

Djspaceg:
This manual approach seemed better than a check for whether the `handlers` block/key existed?

ryanjduffy:
The presence of handlers isn't the issue in this case. React treats functional components differently (e.g. they don't support the `contextType` static member and they don't support `ref`s) so we will break existing apps that are currently valid by making functional components the default.

PR #2767, opened 2020-05-11
PLAT-102263: Migrate SpotlightContainerDecorator to use hooks

ybsung:
### Checklist

Migrate spotlight/SpotlightContainerDecorator to use hooks

### Resolution

Add `SpotlightContainer` for spotlight container behavior.
Add `useSpotlightContainer` hook to manage the life cycle of the SpotlightContainer behavior.
Adopt `spotlight/SpotlightContainerDecorator` to use `useSpotlightContainer`.
Add `useSpot-specs.js`.

### Additional Considerations

- Moved `releaseContainer` and `stateFromProps` in useSpotlightContainer
- Added the missed JS Doc for `navigableFilter` in spotlight/SpotlightContainerDecorator.SpotlightContainerDecorator.defaultConfig
- Define `ForwardSpotlightContainerDecorator` to escape the following error. The `React.forwardRef` couldn't be used with `propTypes` and `defaultProps`.
```
This will give me the following warning in the console: Warning: forwardRef render functions do not support propTypes or defaultProps. Did you accidentally pass a React component?
```

### Links

PLAT-102263


packages/spotlight/SpotlightContainerDecorator/SpotlightContainer.js
```diff
@@ -0,0 +1,141 @@
...
+				this.releaseContainer(this.id);
+			}
+
+			this.id = Spotlight.add({
+				...this.config,
+				id,
+				restrict,
+				navigableFilter: this.navigableFilter
+			});
+			this.preserveId = this.canPreserveId && this.id === id;
```
webOS101, line 56: Does this imply `preserveId` only works if you specify an id?  I think this behaves differently from the previous implementation insofar as a generated id would be passed to `stateFromProps`.  In this case, we should test if `id` is `null` and allow that to preserve the id.

packages/spotlight/SpotlightContainerDecorator/useSpotlightContainer.js
```diff
@@ -0,0 +1,100 @@
...
+
+/**
+ * Configuration for `useSpotlightContainer`
+ *
+ * @typedef {Object} useSpotlightContainerConfig
+ * @memberof spotlight/SpotlightContainerDecorator
+ * @property {Object}   [containerConfig]                   Spotlight container configuration.
+ * @property {Function} [navigableFilter]                   Called to determine if a spottable
+ *                                                          component is accessible.
+ * @property {Boolean}  [preserveId]                        Preserves the spotlight id when the
```
webOS101, line 14: is it important to call out which properties can be updated and which are only set once at creation?

ryanjduffy:
Pushed doc updates and updated the hook to allow omitting the config object (along with a unit test for that).

I wonder if we should revisit our arg decision to separate design-time and run-time configurations into separate args. I didn't address that here but we can revisit later.

packages/spotlight/SpotlightContainerDecorator/tests/useSpotlightContainer-specs.js
```diff
@@ -48,6 +48,25 @@ describe('useSpotlightContainer', () => {
 	// remove all containers after each test
 	afterEach(Spotlight.clear);
 
+	test('should support omitting the config object', () => {
```
webOS101, line 51: :nice:

PR #2770, opened 2020-05-14
PLAT-107424: Reduce the number of DOM elements in VirtualList

ybsung:
### Checklist

Removed the DOM element having the `role` attribute.
Then moved the `role` attribute` to the upper DOM element.

By doing it, the `itemsRenderer` returns not the items, but only the placeholders from theme libraries. So it was renamed to the `placeholderRenderer`.

### Additional Considerations

We need to consider the compatibility for all theme libraries. So we could not remove the DOM elements aggressively ui library side. But we could do it theme libraries.

### Links

PLAT-107424

### DIFF
--- a/packages/ui/VirtualList/VirtualListBasic.js
+++ b/packages/ui/VirtualList/VirtualListBasic.js
@@ -1,5 +1,4 @@
 import classNames from 'classnames';
-import EnactPropTypes from '@enact/core/internal/prop-types';
 import {forward} from '@enact/core/handle';
 import {platform} from '@enact/core/platform';
 import PropTypes from 'prop-types';
--- a/packages/ui/VirtualList/VirtualListBasic.js
+++ b/packages/ui/VirtualList/VirtualListBasic.js
@@ -1226,9 +1226,8 @@ class VirtualListBasic extends Component {
 		return (
 			<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} data-webos-voice-disabled={voiceDisabled} ref={this.props.scrollContentRef} style={style}>
 				<div {...rest} className={contentClasses} ref={this.contentRef} role={role}>
-					{cc.length ? cc : null}
+					{[...cc, placeholderRenderer && placeholderRenderer(primary)]}
 				</div>
-				{placeholderRenderer && placeholderRenderer(primary)}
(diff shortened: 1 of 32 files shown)


packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1215,9 +1215,10 @@ class VirtualListBasic extends Component {
 
 		return (
 			<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} data-webos-voice-disabled={voiceDisabled} ref={this.props.scrollContentRef} style={style}>
-				<div {...rest} className={contentClasses} ref={this.contentRef}>
-					{itemsRenderer({cc, primary})}
+				<div {...rest} className={contentClasses} ref={this.contentRef} role={role}>
+					{cc.length ? cc : null}
 				</div>
+				{placeholderRenderer && placeholderRenderer(primary)}
```
MikyungKim, line 1221: Is it okay to move placeholder out from the above `div`?

packages/ui/useScroll/useScroll.js
```diff
@@ -1532,7 +1530,7 @@ const useScrollBase = (props) => {
 	};
 
 	return {
-		scrollContentWrapper: noScrollByDrag ? 'div' : TouchableDiv,
+		scrollContentComp: noScrollByDrag ? 'div' : TouchableDiv,
```
MikyungKim, line 1533: Can we rename `scrollContentComp` to `scrollContentComponent`?

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1216,9 +1224,10 @@ class VirtualListBasic extends Component {
 
 		return (
 			<div className={containerClasses} data-webos-voice-focused={voiceFocused} data-webos-voice-group-label={voiceGroupLabel} data-webos-voice-disabled={voiceDisabled} ref={this.props.scrollContentRef} style={style}>
-				<div {...rest} className={contentClasses} ref={this.contentRef}>
-					{itemsRenderer({cc, primary})}
+				<div {...rest} className={contentClasses} ref={this.contentRef} role={role}>
+					{cc.length ? cc : null}
```
MikyungKim, line 1229: I think we could omit the condition here.

PR #2772, opened 2020-05-18
PLAT-102268: Migrate ui/Touchable to use hook

0x64:
### Checklist

Migrates ui/Touchable to use hook.

### Resolution
To convert existing Touchable HoC into hook-based HoC;
- Added `Touchable/Touch` to have most gesture handling logic in a class instance
- Added `Touchable/useTouch` hook to serve existing `Touchable` features
- Added unit tests for `useTouch` hook with all the same cases as `Touchable`

Major changes (in a viewpoint of interface)
- `onMove` custom event is added to `Touchable` as a prop.

Minor changes
- Unit tests code for `Touchable` is slightly changed with `activeProp` name which does not affect actual tests.

### Additional Considerations

#### Interfaces

`Touchable` requires to catch the following events to handle gestures;
onBlur, onClick, onMouseDown, onMouseEnter, onMouseLeave, onMouseMove, onMouseUp, onTouchEnd, onTouchMove, onTouchStart
So, all these event handlers should be passed to the target component that is matching with the wrapped component in existing `Touchable` usage. To provide a simple way to use `useTouch` hook, handlers to be passed are grouped in an object named `handlers`.

`Touchable` provides a config `activeProp` as a prop name passed to the wrapped component, but hook does not handle any wrapped component. So the actual meaning of `activeProp` in hook version is just to serve the boolean `active` value or not. To make it clearer I named it `getActive` but not looks good. Hope better suggestion.

During this work, I found that `onMove` custom event is triggered and it missed in props list. I added it as a prop of `Touchable`.

### Links
PLAT-102268


MikyungKim (approved):
LGTM

PR #2773, opened 2020-05-18
PLAT-107660: Add forwardCustom handler

ryanjduffy:
### Checklist

We have often created one-off handlers to emit custom events.

### Resolution
Added `forwardCustom` to standardize emitting custom events to always include a `type` member along with the value returned from `adapter`.

### DIFF
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -725,7 +725,6 @@ const adaptEvent = handle.adaptEvent = curry(function (adapter, handler) {
  *                                     (suitable for passing to handle or used directly within
  *                                     `handlers` in [kind]{@link core/kind}) that will forward the
  *                                     custom event.
- * @curried
  * @memberof core/handle
  * @public
  */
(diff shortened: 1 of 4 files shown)


packages/core/handle/handle.js
```diff
@@ -695,6 +695,58 @@ const adaptEvent = handle.adaptEvent = curry(function (adapter, handler) {
...
+ *
+ * @method   forwardCustom
+ * @param    {String}        name      Name of method on the `props`
+ * @param    {EventAdapter}  [adapter] Function to adapt the event payload
+ *
+ * @returns  {HandlerFunction}         Returns an [event handler]{@link core/handle.EventHandler}
+ *                                     (suitable for passing to handle or used directly within
+ *                                     `handlers` in [kind]{@link core/kind}) that will forward the
+ *                                     custom event.
+ * @curried
```
webOS101, line 728: Is this really curried?

webOS101: I removed it.

packages/core/handle/handle.js
```diff
@@ -695,6 +695,58 @@ const adaptEvent = handle.adaptEvent = curry(function (adapter, handler) {
 	}, 'adaptEvent');
 });
 
+/**
+ * Forwards the event to a function at `name` on `props`.
```
webOS101, line 699: How do you feel about changing this to something like:  Creates a handler that will forward ...

PR #2782, opened 2020-06-08
PLAT-109453: Add useChainRefs

ryanjduffy:
### Checklist

With the adoption of hooks, we need to be able to invoke ref functions from multiple sources for the same reference.

### Resolution
Adds (private) `core/useChainRefs` to support updating multiple refs

### Additional Considerations

No change log since this is private at the moment


webOS101:
My first thought on looking at this is...  This seems like it's addressing the wrong problem.  The problem seems like the code that's trying to use the ref.  How likely is this scenario that we need to add it in core?

PR #2783, opened 2020-06-08
PLAT-100212: Export objectify in Skinnable

ybsung:
### Checklist

### Issue Resolved / Feature Added /Resolution

If passing the `skinVariants` prop to the sandstone `AccessibilityDecorator`,
the `AccessibilityDecorator` should merge its own `skinVariants` object into the `AccessibilityDecorator`'s `skinVariants` prop. But the `AccessibilityDecorator`'s `skinVariants` prop could be a string, an array, or an object. There is the `objectify` to support merging an object into those type objects. So I hope I could use it by exporting it from `Skinnable` instead of creating a new one.

So I used the `objectify` in https://github.com/enyojs/sandstone/pull/418/files#diff-1a21eb0b9c3bc7df485048524c544490R112 and added some unit tests related with that case in https://github.com/enyojs/sandstone/pull/418/files#diff-2f55a7ca463aa02732246d17fcf0aa0fR137.

### Links

PLAT-100212

### Comments

### DIFF
--- a/packages/ui/Skinnable/Skinnable.js
+++ b/packages/ui/Skinnable/Skinnable.js
@@ -209,6 +209,7 @@ const Skinnable = hoc(defaultConfig, (config, Wrapped) => {
 
 export default Skinnable;
 export {
+	objectify,
 	Skinnable,
 	useSkins
 };


ryanjduffy:
@ybsung - Can you add an explanation to the PR description clarifying why we need to export this utility?

ybsung:
> @ybsung - Can you add an explanation to the PR description clarifying why we need to export this utility?

Hello @ryanjduffy 

To fix the PLAT-100212 issue, the sandstone Accessibility should merge it's own skinVariants into the Accessibility prop. But the Accessibility prop could be a string, an array, or an object. There is the `objectify` to support merging an object into those type objects. So I hope I could use is instead of creating a new one.

So I used the `objectify` in https://github.com/enyojs/sandstone/pull/418/files#diff-1a21eb0b9c3bc7df485048524c544490R112 and added some unit tests related with that case in https://github.com/enyojs/sandstone/pull/418/files#diff-2f55a7ca463aa02732246d17fcf0aa0fR137.

If I have to find another approach, please let me know. Thanks for looking into my PR.

webOS101:
Would rather we reach into the submodule than export this.

PR #2784, opened 2020-06-09
ENYO-6470: Fix items are not rendered properly in case of various item sizes

jeonghee27:

### Checklist

The items disappear in VL with different item sizes.

### Resolution
* Cause 
The item is wrapped in another div (itemContaineRef) and the position value is already assigned by composeStyle(). applyItemPositionToDOMElement did not consider this case, and assign "translate3d" style to internal item elements. 

* Resolution
The node to which "translate3d" style is applied should be a direct descendant of VL.
Store the VL's direct descendant in the itemContainerRefs and apply "translate3d" style to it.

### Additional Considerations

### Links
ENYO-6470

### DIFF
--- a/packages/ui/VirtualList/VirtualListBasic.js
+++ b/packages/ui/VirtualList/VirtualListBasic.js
@@ -303,6 +303,7 @@ class VirtualListBasic extends Component {
 		super(props);
 
 		this.contentRef = React.createRef();
+		this.itemContainerRefs = [];
 
 		if (props.clientSize) {
 			this.calculateMetrics(props);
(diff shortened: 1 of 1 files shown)


MikyungKim:
I think `applyItemPositionToDOMElement` was there for a reason in https://github.com/enactjs/enact/pull/2486. I'm not willing to remove it right away. Could you please investigate more why the current sample is broken?

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -1068,6 +1069,11 @@ class VirtualListBasic extends Component {
 					itemRefs.current[key] = (parseInt(itemNode.dataset.index) === index) ?
 						itemNode :
 						ref.querySelector(`[data-index="${index}"]`);
+
+					if (itemRefs.current[key]) {
```
MikyungKim, line 1073: Do we need this code?
I think we can just store `ref` in `this.itemContainerRefs[key]` like
this.itemContainerRefs[key] = ref;
We already know the container's ref.

PR #2789, opened 2020-06-12
PLAT-109159: Add Button prop iconFlip

ryanjduffy:
### Checklist

* Users of `ui/Button` cannot easily flip the icon without overriding `iconComponent`.

### Resolution
* Add `ui/Button.iconFlip` to forward `flip` to `Icon`.

### Additional Considerations

### DIFF
--- a/packages/ui/Button/Button.js
+++ b/packages/ui/Button/Button.js
@@ -94,7 +94,7 @@ const ButtonBase = kind({
 		/**
 		 * Flips the icon.
 		 *
-		 * @see {@link ui/Icon.IconDecorator#flip}
+		 * @see {@link ui/Icon.Icon#flip}
 		 * @type {String}
 		 * @public
 		 */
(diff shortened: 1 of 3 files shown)


ryanjduffy:
Forgot we're not using `i18n` in `ui` so this needs to be moved to the theme layer.

packages/ui/Button/Button.js
```diff
@@ -99,13 +99,22 @@ const ButtonBase = kind({
 		 * will not be rendered.
 		 *
 		 * If this is a component rather than an HTML element string, this component will also
-		 * receive the `size` property and should be configured to handle it.
+		 * receive the `size` and `flip` properties and should be configured to handle it.
```
webOS101, line 102: This should be `iconFlip`, I think.  Or, rather, reworded to receive the `iconFlip` prop as `flip`.

ryanjduffy: :+1: I'm on it

PR #2791, opened 2020-06-15
PLAT-110783: Fix spotlight to trigger onLeaveContainerFail when leaveFor is configured

ryanjduffy:
### Checklist

When `leaveFor` for a container, failing to navigating in a direction due to that configuration should emit `onLeaveContainerFail `.

### Resolution
Return `null` when the earlier bail-out returning the source element occurs.

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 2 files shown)


packages/spotlight/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `spotlight` to trigger `onContainerLeaveFail` when `leaveFor` prevents navigation
```
germboy, line 9: Should be `onLeaveContainerFail`

ryanjduffy: :oops: fixed! :)

PR #2796, opened 2020-06-23
PLAT-111412: VirtualList: Fix items are not shown when changing RTL locales

MikyungKim:
### Checklist

The items are gone when changing RTL locales.
The issue is only for scrollMode="native".
The internal scrollPosition was reset by 0 but the actual native scroller didn't scroll to 0.

### Resolution
Make the native scroller scroll to 0 position.

### Additional Considerations

### Links
PLAT-111412

### DIFF
--- a/packages/ui/VirtualList/VirtualListBasic.js
+++ b/packages/ui/VirtualList/VirtualListBasic.js
@@ -594,7 +594,7 @@ class VirtualListBasic extends Component {
 
 	calculateMetrics (props) {
 		const
-			{clientSize, direction, itemSize, overhang, spacing} = props,
+			{clientSize, direction, itemSize, overhang, scrollMode, spacing} = props,
 			node = this.props.scrollContentRef.current;
 
 		if (!clientSize && !node) {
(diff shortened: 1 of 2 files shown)


0x64:
LGTM

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -651,8 +651,12 @@ class VirtualListBasic extends Component {
 
 		// reset
 		this.scrollPosition = 0;
-		if (this.props.scrollMode === 'translate' && this.contentRef.current) {
+		if (scrollMode === 'translate' && this.contentRef.current) {
 			this.contentRef.current.style.transform = null;
+		} else if (scrollMode === 'native' && node) {
+			node.style.scrollBehavior = null;
+			this.updateScrollPosition(this.getXY(this.scrollPosition, 0));
+			node.style.scrollBehavior = 'smooth';
```
jeonghee27, line 659: How to save and restore the previous value without specifying the specific value?'

const scrollBehavior = node.style.scrollBehavior;  
..
node.style.scrollBehavior = scrollBehavior;

MikyungKim: I think we should clean up this pattern later.

0x64:
Reverted in #2799 due to unexpected unit test failures.

PR #2798, opened 2020-06-24
PLAT-110767: Add spotlight container option straightOnlyLeave

germboy:
### Checklist

This adds a container config option `straightOnlyLeave` that prevents navigation to oblique containers when leaving the current container (as opposed to `straightOnly`, which prevents navigation to elements in an oblique position to the currently focused element).

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -336,11 +336,11 @@ const Spotlight = (function () {
 					return false;
 				}
 
-				if (getContainerConfig(currentContainerId).straightOnlyLeave) {	
+				if (getContainerConfig(currentContainerId).straightOnlyLeave) {
 					const currentContainerRect = getContainerRect(currentContainerId);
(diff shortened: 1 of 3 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -327,10 +328,27 @@ const Spotlight = (function () {
 			const currentContainerId = last(currentContainerIds);
 			const nextContainerIds = getContainersForNode(next);
 
-			// prevent focus if 5-way is being held and the next element isn't wrapped by
-			// the current element's immediate container
-			if (_5WayKeyHold && nextContainerIds.indexOf(currentContainerId) < 0 && !isContainer5WayHoldable(currentContainerId)) {
-				return false;
+			if (nextContainerIds.indexOf(currentContainerId) < 0) {
+				const currentContainerRect = getContainerRect(currentContainerId);
+				const nextContainerRect = getContainerRect(last(nextContainerIds));
```
ryanjduffy, line 333: Let's move these into the next block to avoid DOM operations until required.

PR #2804, opened 2020-06-26
PLAT-111891: Add core/util method mapAndFilterChildren

webOS101:

### Checklist

When iterating over React children, it's easy to forget that they could be `null`.

### Resolution
Add a convenience method to simplify access.

### Additional Considerations
The filtering out of `null` results could be optional to maintain them in cases where we want to preserve the number of children or the `null` for some reason.

### Links
PLAT-111891


webOS101:
`UNSAFE_filterMapChidlren`?

PR #2807, opened 2020-06-30
PLAT-111149: Replace Cell and Row to div for performance

jeonghee27:

### Checklist

- Current ImageItem has a performance issue in VirtualList.

### Resolution
Replace "Cell" and "Row" to "div" to become a performance

### Additional Considerations

### Links
PLAT-111149

### DIFF
--- a/packages/sampler/stories/default/VirtualGridList.js
+++ b/packages/sampler/stories/default/VirtualGridList.js
@@ -26,8 +26,8 @@ const
 		return (
 			<UiImageItem
 				{...rest}
-				style={{width: '100%'}}
 				src={source}
+				style={{width: '100%'}}
 			>
 				{text}
 			</UiImageItem>
--- a/packages/sampler/stories/default/VirtualGridList.js
+++ b/packages/sampler/stories/default/VirtualGridList.js
@@ -26,6 +26,7 @@ const
 		return (
 			<UiImageItem
 				{...rest}
+				style={{width: '100%'}}
 				src={source}
 			>
 				{text}
--- a/packages/sampler/stories/default/VirtualGridList.js
+++ b/packages/sampler/stories/default/VirtualGridList.js
@@ -27,7 +27,6 @@ const
 			<UiImageItem
 				{...rest}
 				src={source}
-				style={{width: '100%'}}
 			>
 				{text}
 			</UiImageItem>
--- a/packages/ui/ImageItem/ImageItem.js
+++ b/packages/ui/ImageItem/ImageItem.js
@@ -12,7 +12,6 @@ import React from 'react';
 
 import ComponentOverride from '../ComponentOverride';
 import Image from '../Image';
-import {Cell, Column, Row} from '../Layout';
 
 import componentCss from './ImageItem.module.less';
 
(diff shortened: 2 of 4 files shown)


packages/sampler/stories/default/VirtualGridList.js
```diff
@@ -26,8 +26,8 @@ const
 		return (
 			<UiImageItem
 				{...rest}
-				src={source}
 				style={{width: '100%'}}
+				src={source}
```
ybsung, line 30: Do we need to change it?

packages/ui/ImageItem/ImageItem.js
```diff
@@ -126,33 +125,29 @@ const ImageItemBase = kind({
...
 		delete rest.selected;
-
 		const isHorizontal = orientation === 'horizontal';
-		const Component = isHorizontal ? Row : Column;
 
 		return (
-			<Component {...rest}>
-				<Cell
+			<div {...rest}>
+				<ImageOverride
```
ybsung, line 134: It seems that we don't need the `ImageOverride` anymore.

Please remove the `ImageOverride` and use `<ImageComponent` here.
```
function ImageOverride ({imageComponent, ...rest}) {
	return ComponentOverride({
		component: imageComponent,
		...rest
	});
}
```

jeonghee27: When I tried to remove this and use 
```{imageComponent} +  className={css.image} placeholder={placeholder} src={src}  ```
it was not possible. because imageComponent is not Component and it's React Element.
So I don't know another way to pass the `className` or other props to `imageComponent` instead of `ImageOverride` .

File: packages/ui/ImageItem/ImageItem.module.less [hunk 3]
```diff
@@ -4,26 +4,39 @@
...
-	
+
 	&.horizontal {
 		.image {
+			display: inline-block;
+			vertical-align: middle;
 			width: 25%;
 			height: 100%;
 		}
+		.image+div {
```
ybsung, line 25: `.image + div`

packages/ui/ImageItem/tests/ImageItem-specs.js
```diff
@@ -35,12 +35,12 @@ describe('ImageItem', () => {
...
 	});
 
-	test('should use a `Row` when `orientation="horizontal"`', () => {
+	test('should use a `div` when `orientation="horizontal"`', () => {
 		const subject = shallow(
 			<ImageItemBase orientation="horizontal" />
 		);
 
-		const actual = subject.find('Row.imageItem');
+		const actual = subject.find('div.imageItem');
```
ybsung, line 43: I think that the we might use something else except `'div.imageItem'` because whatever value for `orientation` has `horizontal` or `vertical`, there is the `div.imageitem`.

packages/ui/ImageItem/ImageItem.js
ybsung: Could you check the following code? It seems to work well to me.

```
	render: ({children, css, imageComponent, orientation, placeholder, src, ...rest}) => {
		...

		return (
			<div {...rest}>
				<ComponentOverride
					className={css.image}
					component={imageComponent}
					placeholder={placeholder}
					shrink={isHorizontal}
					src={src}
				/>
```

packages/ui/ImageItem/ImageItem.module.less
```diff
@@ -4,26 +4,39 @@
...
 	.image {
+		position: relative;
 		display: block; // leaving this to prevent whitespace between inline-block elements
 		margin: 0;
 	}
-	
+
 	.caption,
 	&.selected {
+		width: 100%;
```
ybsung, line 14: Could you check if the `width` is needed here?

packages/ui/ImageItem/ImageItem.module.less
ybsung: My preference is that..

```
	&.horizontal {
        .image,
        .image + .children {
            display: inline-block;
			vertical-align: middle;
        }

		.image {
			width: 25%;
			height: 100%;
		}
		.image + .children {
			width: 75%;
		}
	}
```

packages/ui/ImageItem/ImageItem.module.less
```diff
@@ -4,26 +4,39 @@
...
+
 	&.vertical {
+		height: 100%;
 		.image {
 			width: 100%;
-			height: 100%;
+			height: calc(100% - 50px);
+		}
+		.image+div {
+			height: 50px;
```
ybsung, line 39: We need to find the way to set a proper height.

packages/ui/ImageItem/ImageItem.module.less
```diff
@@ -4,26 +4,39 @@
...
+		}
 	}
-	
+
 	&.vertical {
+		height: 100%;
 		.image {
 			width: 100%;
-			height: 100%;
+			height: calc(100% - 50px);
```
ybsung, line 36: Please confirm that the height is the right value that you expected.

packages/ui/ImageItem/ImageItem.module.less
```diff
@@ -4,26 +4,39 @@
...
+			vertical-align: middle;
+			width: 75%;
+		}
 	}
-	
+
 	&.vertical {
+		height: 100%;
 		.image {
 			width: 100%;
```
ybsung, line 35: I couldn't understand why we need the width 100% here.
For `block`, it's the 100% by default. For `inline`, the width won't work.

PR #2808, opened 2020-07-02
PLAT-112276: Provides `stopPropagation` methods from `forwardWithPrevent` handler

0x64:
### Checklist

`forwardWithPrevent` handler provides most required event info including `preventDefault` function, but not `stopPropagation` and `persist` functions. So if a parent component uses `onKeyDown` prop for a child component that forwards event via `forwardWithPrevent`, it is impossible to call `stopPropagation` or `persist` function in the parent component's scope.

### Resolution
`stopPropagation` and `persist` functions are added to the forwarded event info from `forwardWithPrevent`.

### Additional Considerations
Not like `stopPropagation`, `persist` is not so essential since we provide cloned event info rather than the original info. But, in a viewpoint of _API compatibility_, it looks better to provide `persist` also.

### Links
PLAT-112276

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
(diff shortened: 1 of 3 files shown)


0x64:
Closed in favor of #2809

PR #2812, opened 2020-07-07
PLAT-113213: Fix global class name leaks

webOS101:

### Checklist

Some components are either improperly adding class names that do not exist in their less files or they are missing placeholder classnames in cases where there is no additional functionality needed.

### Resolution
Added a develop-time utility to log global class name leakage.

### Additional Considerations
Currently pre-configured with some sandstone global class names.  Could make it configurable.  Could also probably use some sort of global toggle on/off.

### Links
PLAT-113213


packages/core/kind/debugClassnames.js
```diff
@@ -0,0 +1,49 @@
+/* !
+  classnames Copyright (c) 2018 Jed Watson.
+  Licensed under the MIT License (MIT), see
+  http://jedwatson.github.io/classnames
+*/
+
+const hasProp = {}.hasOwnProperty;
```
JayCanuck, line 7: Will want to change this to `Object.prototype.hasOwnProperty` to avoid upcoming ESLint `no-prototype-builtins` rule issue

packages/core/kind/debugClassnames.js
```diff
@@ -0,0 +1,53 @@
...
+
+// When passed as strings (not in objects!), these strings are acceptable globals
+const allowedGlobalClasses = ['neutral', 'highContrast', 'spottable', 'enact-fit'];
```
ryanjduffy, line 10: As you note in your description, since these are sandstone-specific, they will be problematic to be declared in `core`.

Djspaceg: Correct. These are skinning specific and spotlight specific (enact-fit is fine). This would need to at least be design-time configurable so they can be set up in the ThemeDecorator.

Djspaceg:
I think you're on to a solution there @ryanjduffy, with the class name comparison. This could be reconfigured to look for class names that match a certain obfuscation pattern, and any that don't match that are global, whether on purpose or accident. (globally set intentionally with `:global()` or ones without matching assignments in the JS(X).

webOS101:
Perhaps we could intercept the css import and proxy it to warn on property access instead?

webOS101:
I removed the custom `classnames` module and just left the leaking classname fixes.  This should be good to merge now.

PR #2813, opened 2020-07-08
PLAT-111012: Update framework and sample apps for new linting

JayCanuck:
Requires https://github.com/enactjs/eslint-config-enact/pull/32

### Checklist

* Current linting uses `eslint-config-enact` for in-editor linting, which doesn't work globally in ESLint >=6.x.
* Recent rule changes in https://github.com/enactjs/eslint-config-enact/pull/32 fail in current codebase.

### Resolution
* Switches local in-editor eslint config setup to use `eslint-config-enact-proxy` for ESLint 6.x & 7.x support (adding that package as a standalone dev dependency). This package auto-detects global `eslint-config-enact` or `@enact/cli` and uses that for any config/rules/plugins/parsers needed.
* Updates code to meet current linting standards, including support for unit/screenshot/ui test environments without need for additional eslint config files. See https://github.com/enactjs/eslint-config-enact/pull/32 for full linting rules.


packages/core/handle/handle.js
```diff
@@ -131,9 +131,11 @@ const makeHandler = (handlers) => {
 };
 
 // Loose check to determine if obj is component-ish if it has both props and context members
-const hasPropsAndContext = (obj) => {
-	return obj && obj.hasOwnProperty && obj.hasOwnProperty('props') && obj.hasOwnProperty('context');
-};
+const hasPropsAndContext = (obj) => (
+	obj &&
+		Object.prototype.hasOwnProperty.call(obj, 'props') &&
+		Object.prototype.hasOwnProperty.call(obj, 'context')
```
webOS101, line 137: These two lines shouldn't need to be indented further, should they?

JayCanuck: I believe that's a product of linting auuto-fix, so I suppose they do. And it makes sense to indent next-line continuationations of expressions, right?

JayCanuck:
The ignore patterns are embedded in the eslint config so saeparate `eslintignore` files are no longer needed, though I need to update the root gitignore for screenshot/ui test result directories, I believe

packages/core/handle/handle.js
webOS101: Hmm, no, we didn't before, AFAIK.  It doesn't look like it updated similar lines in sandstone.

webOS101: I guess we don't have anything else exactly like this.  I guess it's fine!

PR #2814, opened 2020-07-08
PLAT-110252: Sampler - WizardPanel: 'onTransition' event emit in Actions when story first loads

sugardave:
### Checklist

When the `stay` transition was added to `ui/ViewManager.Arranger.BasicArranger` it caused `onTransition` events to be sent even when no transition had actually occurred (i.e. on mount) for consumers using `BasicArranger`.

### Resolution
The fix is to modify `ui/ViewManager.TransitionGroup` to not send transition events when a view "appears" so that only `leave` and `enter` will send them.

### Links
PLAT-110252

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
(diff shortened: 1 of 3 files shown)


ryanjduffy:
This surfaces some arch issues in that `View` controls that there isn't animation on "appear" but `TransitionGroup` controls emitting the event. Something to consider refactoring later!

PR #2815, opened 2020-07-10
PLAT-113458: Replace straightOnlyLeave with partition config option

ryanjduffy:
### Checklist

When navigating from a container within a `straightOnlyLeave` container, the straight-only would be ignored.

### Resolution

Update the logic to:
* check upstream containers to find the first with `straightOnlyLeave` configured rather than only checkout the nearest container
* treat restricted containers (those with `enterTo` configured) as the "straight only" comparisons falling back to the target itself if it doesn't exist in any restricted containers
* limit the container search to those unique to the target to avoid comparing bounds of shared upstream containers

### Additional Considerations

* We should refactor this logic somewhere else later. It's too much (imo) to be embedded within `spotNext`
* This lacks any automated testing atm which is definitely suboptimal but out of scope for now

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,6 @@ No significant changes.
 ### Added
 
 - `core/util` function `mapAndFilterChildren` to safely iterate over React `children`
-- `spotlight` container config prop `straightOnlyLeave` to prevent navigation to oblique containers when leaving the current container
 
 ### Fixed
 
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -352,22 +352,28 @@ const Spotlight = (function () {
 						if (config && config.enterTo) {
 							return getContainerRect(id);
 						}
-					}, null) || getRect(next);
+					}, null);
 
-					const currentContainerRect = getContainerRect(straighOnlyLeaveContainer);
+					// If we find a container around the target that is within the first shared
+					// container and is a restricted container, verify that it matches straightOnly
+					// rules. If we don't find one, it means that the target is not within a
+					// restricted container and normal 5-way rules should apply so we bail out.
+					if (nextRect) {
+						const currentContainerRect = getContainerRect(straighOnlyLeaveContainer);
 
-					// prevent focus for obliquely leaving straightOnlyLeave containers
-					if (
-						(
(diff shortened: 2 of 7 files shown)


ryanjduffy:
@germboy - Would like your feedback on this on Monday. This "fix" is really a different implementation of the requirement that led to `straightOnlyLeave`. Let's catch up as soon as our schedules align :)

germboy:
This change appears to look for any parent container that is configured with `straightOnlyLeave`, then use that container/dom as the reference point for moving to the next focusable element. However, it doesn't appear to actually restrict focus movement in cases where `straightOnlyLeave` is used. I think this logic will either need to be moved back to `spotNext` or there will be more to consider in the `target.js` / `navigate.js` bits of spotlight to address this.
I can provide a sample to better demonstrate my concern based on this approach, if needed.

packages/spotlight/src/spotlight.js
```diff
@@ -335,25 +334,6 @@ const Spotlight = (function () {
...
-							(nextContainerRect.bottom < currentContainerRect.top || nextContainerRect.top > currentContainerRect.bottom)
-						) ||
-						(
-							(direction === 'up' || direction === 'down') &&
-							(nextContainerRect.right < currentContainerRect.left || nextContainerRect.left > currentContainerRect.right)
-						)
-					) {
-						return false;
-					}
-				}
```
germboy, line 356: We might want to re-consolidate the multiple `if` statements above, like it was prior to the old `straightOnlyLeave` implementation
```
if (_5WayKeyHold && nextContainerIds.indexOf(currentContainerId) < 0 && !isContainer5WayHoldable(currentContainerId)) {
```

CHANGELOG.md
```diff
@@ -11,7 +11,6 @@ No significant changes.
 ### Added
 
 - `core/util` function `mapAndFilterChildren` to safely iterate over React `children`
-- `spotlight` container config prop `straightOnlyLeave` to prevent navigation to oblique containers when leaving the current container
```
webOS101, line 14: :iseewhatyoudidthere:

packages/spotlight/src/spotlight.js
germboy: I pushed this change :)

PR #2823, opened 2020-07-17
PLAT-113895: Fix Spottable hovered state

germboy:
### Checklist

`Spottable` does not update its `isHovered` state correctly in all cases

### Resolution
A change in https://github.com/enactjs/enact/pull/1667 added guards to ensure the pointer moved before changing the hovered state during mouse enter/leave events. This apparently was due to maintain consistency with webOS, which originally did not fire mouse enter/leave events when a background spottable element moved/transitioned past a stationary pointer. That does not appear to be the case in webOS now, so we can remove these guards.

### Links
https://github.com/enactjs/enact/pull/1667

### Comments
I've run UI tests and they all pass with this change.

### DIFF
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -14,7 +14,6 @@ import React from 'react';
 import ReactDOM from 'react-dom';
 
 import {getContainersForNode} from '../src/container';
-import {hasPointerMoved} from '../src/pointer';
 import {getDirection, Spotlight} from '../src/spotlight';
 
 /**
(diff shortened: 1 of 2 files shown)


ryanjduffy:
I think the proposed fix is safe. We'll want to do some thorough testing but it seems the conditions for which it was added have long been sunset.

packages/spotlight/Spottable/Spottable.js
```diff
@@ -394,16 +393,12 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 
 		handleEnter = (ev) => {
 			forward('onMouseEnter', ev, this.props);
-			if (hasPointerMoved(ev.clientX, ev.clientY)) {
-				this.isHovered = true;
-			}
+			this.isHovered = true;
```
webOS101, line 396: Do we need to update the similar code in `SpotlightContainerDecorator`?

PR #2824, opened 2020-07-21
PLAT-114153: Fix Scroller and VirtualList to re-render when its size changed

MikyungKim:
### Checklist

When a Scroller or VirtualList changed its size dynamically by other elements, for example, in Panel from header height changing, it is not re-rendered so focus can be truncated.

### Resolution
Added `ResizeObserver` to make Scroller or VirtualList re-render when its size has been changed.

### Additional Considerations
I left `resize` handler as is because it has an additional feature like scrollTo 0 position logic.

### Links
PLAT-114153

### DIFF
--- a/packages/ui/useScroll/useScroll.js
+++ b/packages/ui/useScroll/useScroll.js
@@ -297,6 +297,7 @@ const useScrollBase = (props) => {
 			const {animator, resizeRegistry, scrolling, scrollStopJob} = mutableRef.current; // eslint-disable-line react-hooks/exhaustive-deps
 
 			resizeRegistry.parent = null;
+			mutableRef.current.resizeObserver.unobserve(scrollContainerRef.current);
 
 			// Before call cancelAnimationFrame, you must send scrollStop Event.
 			if (scrolling) {
--- a/packages/ui/useScroll/useScroll.js
+++ b/packages/ui/useScroll/useScroll.js
@@ -309,7 +309,6 @@ const useScrollBase = (props) => {
 				}
 			};
 		}
-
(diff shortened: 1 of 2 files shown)


packages/ui/useScroll/useScroll.js
```diff
@@ -289,6 +290,7 @@ const useScrollBase = (props) => {
 
 	useEffect(() => {
 		mutableRef.current.resizeRegistry.parent = context;
+		mutableRef.current.resizeObserver.observe(scrollContainerRef.current);
```
ybsung, line 293: Don't we need unobserve() when unmounting?

MikyungKim: good point!

packages/ui/useScroll/useScroll.js
```diff
@@ -287,6 +288,29 @@ const useScrollBase = (props) => {
...
+			return;
+		}
+		if (typeof ResizeObserver === 'function') {
+			let resizeObserver = new ResizeObserver(() => {
+				if (scrollContentHandle.current && scrollContentHandle.current.syncClientSize) {
+					scrollContentHandle.current.syncClientSize();
+				}
+			});
+
+			resizeObserver.observe(containerRef);
```
jeonghee27, line 303: No need to do resizeObserver.unobserve()?

MikyungKim: I did disconnect().

jeonghee27: Ah, I saw resizeObserver.disconnect()

PR #2827, opened 2020-07-23
PLAT-114518: Fix Dropdown to not lose focus when dataSize has changed

MikyungKim:

### Checklist

Sandstone Dropdown loses its focus when `dataSize`  of VL has changed to smaller.
The reason why Dropdown loses its focus is after calling `scrollTo` 0 index from Dropdown, VirtualList itself calls `scrollTo` to adjust its `scrollPosition` to update it with shrunk `maxPos` due to shrunk `dataSize`.
Before calling `scrollTo`, VirtualList checks current `scrollPosition` is bigger than `masPos`.
At this point, `scrollPosition` supposed to be '0' and VL shouldn't call `scrollTo` because Dropdown called `scrollTo` with `animate: false` option. However, by the nature of "native" scroller, `onScroll` event happens asynchronously and `scrollPosition` had the wrong value and unnecessary `scrollTo` happened from VirtualList.

### Resolution
I introduced `scrollToPositionTarget` in VirtualListBasic. When scrollTo called with `animate: false` in scrollMode native, we set this variable and see this value from componentDidUpdate phase to determine to call `scrollTo` or not.

### Additional Considerations

### Links
PLAT-114518

### DIFF
--- a/packages/ui/VirtualList/VirtualListBasic.js
+++ b/packages/ui/VirtualList/VirtualListBasic.js
@@ -2,6 +2,7 @@ import classNames from 'classnames';
 import EnactPropTypes from '@enact/core/internal/prop-types';
 import {forward} from '@enact/core/handle';
 import {platform} from '@enact/core/platform';
+import {clamp} from '@enact/core/util';
 import PropTypes from 'prop-types';
 import equals from 'ramda/src/equals';
 import React, {Component} from 'react';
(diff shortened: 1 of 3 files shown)


packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -577,6 +585,7 @@ class VirtualListBasic extends Component {
 		}
 
 		position.primaryPosition -= offset;
+		position.primaryPosition = clamp(0, maxPos, position.primaryPosition);
```
jeonghee27, line 588: Could you merge this two lines?

		position.primaryPosition = clamp(0, maxPos, position.primaryPosition - offset);

MikyungKim: 👌

PR #2829, opened 2020-07-27
PLAT-114531: Add rtl prop to ViewManager

ryanjduffy:
### Checklist

Arrangers lack the data necessary to adjust their animation for locales with right-to-left reading order

### Resolution
Add `rtl` prop to `ViewManager` and the arranger config.


packages/ui/ViewManager/tests/ViewManager-specs.js
```diff
@@ -434,4 +434,38 @@ describe('ViewManager', () => {
 
 		expect(spy).toHaveBeenLastCalledWith({index: 0, previousIndex: 1});
 	});
+
+	test('should pass `rtl` prop to arranger when `true`', () => {
```
webOS101, line 438: Thank you for new unit tests!

PR #2834, opened 2020-07-31
PLAT-109024: Added "tallglyph" locale support

Djspaceg:
### Checklist

Often, languages are given incorrect rules due to a naive categorization which states that all non-Latin languages are tall-height glyph languages. This isn't true and we need a more granular way to group these languages that need some extra height for tall glyphs vs languages that fit within the bounds of "normal" glyph metrics.

### Resolution
This PR adds a mechanism in LESS to apply rules to those languages specifically. This also updates the base-approach of how to access the `:global(.enact-locale-*) &` selector by adding a mixin that can target any feature of the localization classes.

The new mixins are as follows:
* `.enact-locale-line-height()` - A convenience function, since this is one of the most common cases
* `.enact-locale-tallglyph()` - A generalized function that has both shorthand and full rule support
* `.enact-locale-rtl()` - A convenience function, since this is one of the most common cases
* `.enact-locale()` - A generalized metod to add rulesets to a given locale case (e.g. `right-to-left`, `ar`, `en-US`, etc)

This also includes a new `variables.less` file which contains a necessary variable for the `tallglyph` set of mixins. This variable is a list of language codes that should be included when generating rules. This variable can be overridden by a theme.

### Additional Considerations
An additional PR can be created for the conversion of all existing references of the global class name to the new mixin format. This could additionally be bundled with this PR, but was left out to keep the focus of the review on the "work" parts of the change, rather than the conversion.

### DIFF
--- a/packages/ui/styles/mixins.less
+++ b/packages/ui/styles/mixins.less
@@ -1,3 +1,6 @@
+@import "./variables.less";
+
+
 // Disabled elements
 .disabled(@rules; @target) when (isruleset(@rules)) and (@target = parent) {
 	[disabled] {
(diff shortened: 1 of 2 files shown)


packages/ui/styles/mixins.less
```diff
@@ -345,8 +348,89 @@
...
+// Accept a ruleset and apply it to the list of tall-glyph languages
+// Ex:
+//   .enact-locale-tallglyph({
+//     font-size: 1.6em;
+//   });
+.enact-locale-tallglyph(@rules) when (isruleset(@rules)) {
+	// :global(.enact-locale-th) &,
+	// :global(.enact-locale-si) & {
+	// 	@rules();
+	// }
```
ryanjduffy, line 410: Going to remove these

packages/ui/styles/mixins.less
```diff
@@ -345,8 +348,89 @@
...
+	each(@locale-tallglyph-languages, {
+		// Take each language in the list, apply the @rules to them.
+		// @value is used to generate an appropriate selector for the @rules.
+		// Ex: `.enact-locale-th { @rules(); }`
```
ryanjduffy, line 414: Add a ` &` in the selector of this example

packages/ui/styles/mixins.less
```diff
@@ -345,8 +348,89 @@
...
+// Assign rules specifically to RTL locales with a convenient shorthand.
+.enact-locale-rtl(@rules) when (isruleset(@rules)) {
+	.enact-locale(right-to-left, @rules);
+}
+
+// Generates an appropriate selector given the locale-target, and injects the given rules into it.
+.enact-locale(@locale-target; @rules) when (iskeyword(@locale-target)) and (isruleset(@rules)) {
+	:global(.enact-locale-@{locale-target}) & {
+		@rules();
+		// content: 'you got it';
```
ryanjduffy, line 428: Remove this debug code

packages/ui/styles/mixins.less
Djspaceg: I had actually left this here on purpose, since this (commented out) approach is more verbose and less configurable on the development side, but it produces slightly smaller, more compact output code. Eg:
Commented out version:
```css
.enact-locale-si .BodyText_bodyText__1Fj1C,
.enact-locale-th .BodyText_bodyText__1Fj1C,
.enact-locale-vi .BodyText_bodyText__1Fj1C {
    line-height: 1.7em;
}
```
vs Current:
```css
.enact-locale-si .BodyText_bodyText__1Fj1C {
    line-height: 1.7em;
}
.enact-locale-th .BodyText_bodyText__1Fj1C {
    line-height: 1.7em;
}
.enact-locale-vi .BodyText_bodyText__1Fj1C {
    line-height: 1.7em;
}
```
This more verbose version that produces more compact output code could be desirable for efficiency and I think is possible to automate the generation of, but I just couldn't figure it out at this time.

PR #2835, opened 2020-07-31
PLAT-102263: Refactoring useSpotlightContainer

ryanjduffy:
Proposing some changes to #2767 based on practices we've learned since this work was started

* change the API of `SpotlightContainer` such that static values are passed to the constructor and variable values are passed via `setProps`.
* rename the config members for `useSpotlightContainer` to drop the `spotlight` prefix (e.g. from `spotlightRestrict` to `restrict`). The longer name is useful as a prop but redundant for the hook. It requires a little rewiring but we shouldn't be trying to pass props straight through to the hook anyway.
* remove event forwarding from `useSpotlightContainer`. I'd like to minimize forwarding from hooks because it makes composing them more difficult. If you have multiple hooks that need to handle the same event, how do you line them all up? Easier to handle imperatively (or via some new, future utility) than reconstructing prop objects just so they can be forwarded. I did retain forwarding in the HOC but took advantage of the `useHandlers` hook previously added.
* remove the `state` parameter from `SpotlightContainer`. This is needed when state updates trigger re-renders since it allows us to decouple the process from the react-unaware class. In this case, the state is all local (e.g. the current id) so we don't need to externalize it.
* move attributes into hook. This is a new idea but I think it makes sense in this case. Spotlight containers must have a set of attributes on the DOM in order to function. Rather than requiring consumers wire them up correctly (error prone), it's better to build that into the hook and document that spreading `attributes` onto the root node is required (less error prone).

Open questions:
* Should `restrict` really be the only container config prop that is variable via the hook? That matches the current HOC API but I'm not sure it makes much sense.
* I removed the event handler unit tests from the hook. They're pretty wonky for the HOC and are probably better suited to UI tests since the unit test implementation is very mocked.


ybsung:
Hello @ryanjduffy 

I think everything looks good to me. Thanks for refactoring `useSpotlightContainer `. I'm merging the PR to my PR.

PR #2839, opened 2020-08-03
PLAT-113429: Add preferredEnterTo parameter to getTargetByContainer

ryanjduffy:
### Checklist

We have cases (e.g. enyojs/sandstone#667) in which we want to imperatively focus either the last focused or default element of a container without requiring that the container always use that config for 5-way navigation.

### Resolution
Add a `preferredEnterTo` parameter to `getTargetByContainer` (and downstream files) to favor a particular target.

### Additional Considerations
We should consider making this a new top-level `Spotlight` API (e.g. `focusContainer`) in a future release.

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -623,7 +623,9 @@ const getAllContainerIds = () => {
 /**
  * Returns the default focus element for a container
  *
- * @param   {String}  containerId  ID of container
+ * @param   {String}                             containerId        ID of container
+ * @param   {('last-focused'|'default-element')} [preferredEnterTo] Prefer the given enterTo
+ *                                                                  configuration
  *
  * @returns {Node|null}                 Default focus element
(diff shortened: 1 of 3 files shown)


packages/spotlight/src/container.js
```diff
@@ -638,6 +640,12 @@ function getContainerDefaultElement (containerId) {
 	}
 
 	defaultElementSelector = coerceArray(defaultElementSelector);
+
+	// If a preferred enterTo has been provided, we will favor it by making it first in search array
+	if (preferredEnterTo && typeof preferredEnterTo === 'string' && preferredEnterTo !== 'default-element') {
+		defaultElementSelector.unshift(preferredEnterTo);
```
webOS101, line 646: If only `default-element` and `last-focus` are allowed, should we just force `last-focus` here for simplicity?  If other values are allowed, should we update the docs above?

PR #2840, opened 2020-08-04
PLAT-115535: Wrap the created portal with div having `aria-owns`

ybsung:
### Checklist

When moving focus from the component in the panel to the component in the floatLayer and moving back,
the screen reader doesn't read out the panel header because the panel has `aria-owns="floatLayer"` prop.

But when the panel is in the floatLayer like `PopupTabLayout`, the `aria-owns` won't work. Then the screen reader reads the panel title first, then a focused component when moving focus from out of the panel to in the panel.

### Resolution

When creating a component in the floatLayer, the component has the id and is wrapped with `div` including `aria-owns` prop of the id.

### Additional Considerations

### Links

PLAT-115535

### Comments

### DIFF
--- a/packages/ui/FloatingLayer/FloatingLayer.js
+++ b/packages/ui/FloatingLayer/FloatingLayer.js
@@ -122,6 +122,7 @@ class FloatingLayerBase extends React.Component {
 	constructor (props) {
 		super(props);
 		this.node = null;
+		this.id = this.generateId();
 		this.state = {
 			nodeRendered: false
 		};
@@ -178,6 +179,10 @@ class FloatingLayerBase extends React.Component {
 			this.controller.unregister();
 		}
 	}
+
+	generateId = () => {
+		return Math.random().toString(36).substr(2, 8);
+	};
 
 	handleNotify = oneOf(
 		[forEventProp('action', 'close'), call('handleClose')],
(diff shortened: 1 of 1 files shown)


MikyungKim (approved):
lgtm

packages/ui/FloatingLayer/FloatingLayer.js
```diff
@@ -252,13 +257,15 @@ class FloatingLayerBase extends React.Component {
...
-				<div {...rest}>
+			const portal = ReactDOM.createPortal(
+				<div {...rest} id={id}>
 					{scrimType !== 'none' ? <Scrim type={scrimType} onClick={this.handleClick} /> : null}
 					{React.cloneElement(children, {onClick: this.stopPropagation})}
 				</div>,
 				this.node
 			);
+
+			return <div aria-owns={id}>{portal}</div>;
```
ryanjduffy, line 268: Adding DOM here could be problematic. Since the portal is rendering in another subtree, this `<div />` ends up being dropped wherever `FloatingLayer` was used and those component might not be expecting it. 

I like the instinct to build this into `ui` but I'm not sure this is the right way. It might not be feasible to do since some outer component needs the `id` in order to set `aria-owns` correctly.

An alternative might be to add some `Context` to this to help auto-wire IDs up since they are required (instead of something generic like CSS classes) for ARIA to behave correctly. I'm not sure what that would be yet but something to consider.

PR #2847, opened 2020-08-17
PLAT-116120: Quickfix for Marquee within Marquee

teckliew:
### Checklist

Parent `Marquee` cancels child `Marquee` animation when the parent `Marquee` executes `measureWidth`.

### Resolution
Detect if there are marquees as its child to bypass its own measurements and prevent from marqueeing, leaving the children to handle marquee animation.
Added a warning against nested `Marquee`'s

### Additional Considerations
This opens a :canofworms: with many marquee inception edge cases.
Doesn't work for multiple layers of marquees, or if we want the parent to marquee its children with a marquee child.

### Links


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `ui/MarqueeDecorator` warning against nested `Marquee`'s
+
```
germboy, line 10: This changelog should go in `packages/ui`. I can take care of that during integration.

PR #2853, opened 2020-08-19
PLAT-117042: Fix VirtualList to call `onScrollStop` when scrollbar's visibility changed while scrolling

MikyungKim:

### Checklist

When the visibility of scrollbars changed, current `scrollStopJob` stopped and we create `scrollStopJob` to reference the recent function. But if `scrollStopJob` is already started and did not call its function, some internal status may not be correct. In PLAT-117042, `scrolling` did not reset.

### Resolution
Checked if the internal status `scrolling` is true then call `run` before `stop` it.

### Additional Considerations
In Scrollbar component, sometimes an error that showing `scrollbarTrackRef.current` is null. So I've added a guard for it.

### Links
PLAT-117042


packages/ui/useScroll/Scrollbar.js
```diff
@@ -50,7 +50,9 @@ const useScrollbar = (props) => {
 	hideScrollbarTrackJob.current = hideScrollbarTrackJob.current || new Job(hideScrollbarTrack, scrollbarTrackHidingDelay);
 
 	function hideScrollbarTrack () {
-		removeClass(scrollbarTrackRef.current, css.scrollbarTrackShown);
+		if (scrollbarTrackRef.current !== null) {
+			removeClass(scrollbarTrackRef.current, css.scrollbarTrackShown);
```
ybsung, line 54: How about moving the condition checking `null` into the `removeClass`? If the `removeClass` is called several times, we need to wrap it with condition several times with your code.

packages/ui/useScroll/useScroll.js
```diff
@@ -392,6 +392,9 @@ const useScrollBase = (props) => {
 		}
 
 		return () => {
+			if (ref.scrollStopJob.alive()) {
+				ref.scrollStopJob.run();
+			}
```
ybsung, line 397: There is another scrollStopJob.stop() call in the following code. If this code works first, then the `ref.scrollStopJob.alive()` would be false and it works incorrectly. Am I right?

```
	useEffect(() => {
		mutableRef.current.resizeRegistry.parent = context;

		// componentWillUnmount
		return () => {
			const {... scrollStopJob} = mutableRef.current;

			...
			scrollStopJob.stop();
```

ybsung: I got it! Their lifecycles are different.

ybsung (approved):
LGTM

packages/core/util/Job.js
```diff
@@ -25,6 +25,8 @@ class Job {
 		this.timeout = timeout;
 	}
 
+	alive = () => Boolean(this.id);
```
MikyungKim, line 28: @ryanjduffy I'm going to add this function in Job. WDYT?

packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Scroller` and `ui/VirtualList` to call remaining scrollStopJob function in cleanup phase
```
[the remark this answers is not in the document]
MikyungKim: Actually the bug itself is from theme(sandstone) side not ui side.
Can I mention the theme-side bug here then?
e.g. Fixed `ui/VirtualList` to not lose focus when scrollbar visibility has changed

packages/core/util/Job.js
MikyungKim: I see your point. I'll try the first suggestion of yours. Thanks!

0x64:
Looks sound!

PR #2856, opened 2020-08-25
PLAT-101327: Fix Routable to respect current path at render

ryanjduffy:
### Checklist

Double clicking on a `Link` (or `Linkable`-wrapped component) that used a relative path (e.g. `path="./page2"`) would result in concatenating the path twice (e.g. `/app/page2/page2`).

### Resolution
The `Routable` code always used the current path to evaluate the relative path so executing it again after the current path had changed would append the path again.

The proposed fix caches the current path in `Linkable` on first render so that it is always used to resolve relative paths.

### Additional Considerations
I included a bit of hook migration since doing so made it simpler to perform the caching.

### Links


0x64 (approved):
:lgtm:

PR #2858, opened 2020-08-28
PLAT-116511 Fix a view re-entering

webOS101:

### Checklist

If a `View` was transitioning out and then was transitioned back before it was completely removed, the `enteringProp` was not reset.  This could cause the children to not update when the view arrived.  In `Panels`, this caused the child `Panel` to not get its `hideChildren` prop set and then cleared.  Because the view did not re-render after transitioning back, auto focus logic could not fire to focus the panel contents.

### Resolution
If `componentWillEnter` is called and `entering` isn't set, set it to allow for component to re-enter.

### Additional Considerations
Conceivably, this could cause regressions in cases not considered.  In theory, it shouldn't cause problems if a view wasn't already being removed from the active children and then added back.

### Links
PLAT-115879

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Scroller` and `ui/VirtualList` to call `onScrollStop` when scrollbar's visibility changed while scrolling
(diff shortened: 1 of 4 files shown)


ryanjduffy:
Based on pairing with @webOS101, pushed another commit to ensure the `changeDirection` member is updated when a view is leaving.

ryanjduffy:
Merged this to `develop` instead of the release branch so I fast forwarded the release branch to include it.

PR #2859, opened 2020-09-02
PLAT-118682: Refactored RTL mixins for better interoperability added extract mixin

Djspaceg:
### Checklist

All themes deserve to use the `.extract` mixin, and the `*-start-end` mixins should work with the developer instead of against.

### Resolution
* Migrated style-utils from Sandstone into ui
* Updated the `position-start-end`, `margin-start-end`, and `padding-start-end` mixins to support a single variable containing a list of values that can be parsed as shorthand to apply RTL-safe rules using the associated values.

This is a good addition and causes no known negative side effects. It encourages better, more readable, more maintainable coding practices.

### Examples

Using `padding` as an example below, but `margin` and `position` work exactly the same way.

Here's the old usage, and the result:
```css
.padding-start-end(@start, @end);
.padding-start-end(12px, 24px);

.selector {
    padding-left: 12px;
    padding-right: 24px;
}
.enact-locale-right-to-left .selector {
    padding-right: 12px;
    padding-left: 24px;
}
```
This has the drawback of requiring two lines and multiple variables to pass around, especially if a vertical margin is desirable. It's also backward from standard shorthand.

The old way still works, but now the same mixin also supports the following arguments:
```css
.padding-start-end(@shorthand-paddings, @target);
.padding-start-end(12px);
.padding-start-end(12px 24px);
.padding-start-end(12px 24px 36px);
.padding-start-end(12px 24px 36px 48px);
.padding-start-end(12px 24px 36px 48px, self);

@my-component-padding: 12px;  // or
@my-component-padding: 12px 24px;  // or
@my-component-padding: 12px 24px 36px;  // or
@my-component-padding: 12px 24px 36px 48px;
.padding-start-end(@my-component-padding);
.padding-start-end(@my-component-padding, self);

// Resulting in:
.selector {
    padding: 12px 24px 36px 48px;
}
.enact-locale-right-to-left .selector {
    padding: 12px 48px 36px 24px;
}
// or in the case of `self`
.enact-locale-right-to-left.selector {  // note the lack of a space between the locale and selector.
    padding: 12px 48px 36px 24px;
}
```

### Important Note

* The old approach supports exactly and only 2 arguments.
* The new approach supports 1 or 2 arguments
   1. The first being a single value or a _(space separated)_ list of values.
   2. The second being an optional target to attach this to. "self" is the only supported value currently, but is valuable if you're trying to use the mixin on the root node which also has the `.enact-locale-right-to-left` class applied.

The new method effectively allows you to only have to specify one variable and one rule assignment to get automatic RTL support, just like a "smart" or RTL-safe `margin`/`padding`/`position`.

### DIFF
--- a/packages/ui/styles/mixins.less
+++ b/packages/ui/styles/mixins.less
@@ -1,5 +1,5 @@
+@import "./internal/utils.less";
 @import "./variables.less";
-
 
 // Disabled elements
 .disabled(@rules; @target) when (isruleset(@rules)) and (@target = parent) {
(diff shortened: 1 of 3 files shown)


packages/ui/styles/mixins.less
```diff
@@ -22,56 +22,78 @@
...
+		padding:
+			// Note: swapping left and right values
+			.extract(@trbl; top)[]
+			.extract(@trbl; left)[]
+			.extract(@trbl; bottom)[]
+			.extract(@trbl; right)[]
+		;
+	}; @target);
+}
+.padding-start-end(@start; @end; @target: "") when (default()) {
```
MikyungKim, line 89: Do we need an empty line here?

MikyungKim (changes requested):
I've checked sandstone sampler and the Scroller sampler's padding seems broken. Would you please check?

packages/ui/styles/mixins.less
Djspaceg: I often "group" mixins together that have the same but different argument signatures by omitting the empty line between them. I can add it if you'd prefer.

MikyungKim: Nope, I agree with your opinion. Thanks for clarification.

MikyungKim (approved):
:awesome:

Djspaceg:
Any idea when this will be merged?

ybsung:
> Any idea when this will be merged?

I think that we might merge this PR after creating a new tag at the middle of this month. :)

PR #2836, opened 2020-08-01
PLAT-113686: Fix Touchable to invoke event callbacks updated during gesture

RenukaAtale:
### Checklist

* [ ] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [ ] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [ ] At least one test case is included for this feature or bug fix
* [ ] Documentation was added or is not needed

* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)

PR #2860, opened 2020-09-28
PLAT-120781: Fix console error when flick

jeonghee27:

### Checklist

There is console error when flick Agate Picker. 
 - TypeError: Cannot set property 'onFlick' of undefined

### Resolution
Change unintended assigned.
- It looks like regression from https://github.com/enactjs/enact/pull/2836/
- Flick.js has different structure with Drag.js and Hold.js. Flick store own config in `this` not `this.flickConfig`

### Additional Considerations

### Links
PLAT-120781

### DIFF
--- a/packages/ui/Touchable/Flick.js
+++ b/packages/ui/Touchable/Flick.js
@@ -31,7 +31,7 @@ class Flick {
 		if (!this.tracking) return;
 
 		// This will update the `flickConfig` with the new value
-		this.flickConfig.onFlick = onFlick;
+		this.onFlick = onFlick;
 	};
 
 	move = ({x, y}) => {
--- a/packages/ui/Touchable/Flick.js
+++ b/packages/ui/Touchable/Flick.js
@@ -25,12 +25,12 @@ class Flick {
 		this.move(coords);
 	};
 
-	// This method will get the `onFlick` props in the existing `flickConfig` values.
+	// This method will get the `onFlick` props.
 	updateProps = ({onFlick}) => {
-		// Check `tracking` gesture is not in progress. Check if gesture exists before updating the references to the `flickConfig`
+		// Check `tracking` gesture is not in progress. Check if gesture exists before updating the references to the `onFlick`
 		if (!this.tracking) return;
 
-		// This will update the `flickConfig` with the new value
+		// This will update the `onFlick` with the new value
 		this.onFlick = onFlick;
 	};
 


packages/ui/Touchable/Flick.js
```diff
@@ -31,7 +31,7 @@ class Flick {
 		if (!this.tracking) return;
 
 		// This will update the `flickConfig` with the new value
```
ybsung, line 33: Do we need to update `flickConfig` comment?

ybsung:
> Flick.js has different structure with Drag.js and Hold.js. Flick store own config in this not this.flickConfig

I don't know why `Flick` has to have the different structure with others.

In addition, there is the `flickConfigPropType ` variable in the `Flick`. So the related values like `maxDuration`, `maxMoves`, and `maxVelocity` might be better in the `flickConfig` variable at least.
```
const flickConfigPropType = PropTypes.shape({
	maxDuration: PropTypes.number,
	maxMoves: PropTypes.number,
	maxVelocity: PropTypes.number
});
```

So my suggestion is...
how about updating not `updateProps` as you did but `begin` function like the following way? Then we might need to update variable names related with the properties in the this.flickConfig.

```suggestion
	begin = ({maxDuration, maxMoves, minVelocity}, {onFlick}, coords) => {
		this.flickConfig = {
			minVelocity,
			maxMoves,
			onFlick
		};

		if (maxDuration !== null) {
			this.cancelJob.startAfter(maxDuration);
		}
		this.tracking = !!onFlick;
		this.moves.length = 0;

		this.move(coords);
	};
```

packages/ui/Touchable/Flick.js
```diff
@@ -31,7 +31,7 @@ class Flick {
...
-		this.flickConfig.onFlick = onFlick;
+		this.onFlick = onFlick;
```
ybsung, line 34: Could you consider the unit test case to cover this line?

packages/ui/Touchable/Flick.js
ybsung: My comment might be ignored if keeping using `this.flickConfig`.

packages/ui/Touchable/Flick.js
jeonghee27: I tried to invoke onFlick by simulating flick with touchstart and touchend like onTap.
But it doesn't work. I am not sure flick is simulateable.

Perhaps there was a reasone why Flick.js and Drag.js have low coverage.
I think it is correct to revisit this part again and review it in the direction of increasing the test coverage.

jeonghee27:
> > Flick.js has different structure with Drag.js and Hold.js. Flick store own config in this not this.flickConfig
> 
> I don't know why `Flick` has to have the different structure with others.
> 
> In addition, there is the `flickConfigPropType ` variable in the `Flick`. So the related values like `maxDuration`, `maxMoves`, and `maxVelocity` might be better in the `flickConfig` variable at least.
> 
> ```
> const flickConfigPropType = PropTypes.shape({
> 	maxDuration: PropTypes.number,
> 	maxMoves: PropTypes.number,
> 	maxVelocity: PropTypes.number
> });
> ```
> 
> So my suggestion is...
> how about updating not `updateProps` as you did but `begin` function like the following way? Then we might need to update variable names related with the properties in the this.flickConfig.
> 
> ```
> 	begin = ({maxDuration, maxMoves, minVelocity}, {onFlick}, coords) => {
> 		this.flickConfig = {
> 			minVelocity,
> 			maxMoves,
> 			onFlick
> 		};
> 
> 		if (maxDuration !== null) {
> 			this.cancelJob.startAfter(maxDuration);
> 		}
> 		this.tracking = !!onFlick;
> 		this.moves.length = 0;
> 
> 		this.move(coords);
> 	};
> ```

I don't think it is the flick must have the same structure with Drag and Hold. We need an more understanding of Touchable.
Let's discuss this again after the M version.

ybsung (approved):
LGTM

0x64:
Related PR is #2836

PR #2861, opened 2020-10-06
PLAT-108976: Fix Marquee to restart animation when text changed while focus retained

MikyungKim:

### Checklist

The Marquee does not restart its animation when text changed while focus retained.

### Resolution
The state `animating` from `MarqueeDecorator` had not been reset when the text changed and it leads Marquee not to restart its animation even the new text is long enough.
So I called `restartAnimation` function followed by `cancelAnimation`.
I'm not sure what impacts would it be here since I'm not familiar with the logic.

### Additional Considerations

### Links
PLAT-108976


ybsung (approved):
I agree with the approach in this PR.

`tryStartingAnimation` was called when changing the text. But `this.state.animating` was `TimerState.CLEAR`, so `startAnimation` was not called to start a marquee again.

The way to set `this.state.animating` to `TimerState.SYNCSTART_PENDING` and to call `startAnimation` properly would be the one calling the `resetAnimation`.

In my opinion, the current approach would be the best for now.

LGTM

PR #2864, opened 2020-10-28
PLAT-123775: Spotlight is on a component when it is active in touch mode

ybsung:

### Checklist

If touch events generates, then spottable visual effect should be shown when it is not focus but only active.

### Resolution

When pointerMode is `touch`, then Spotlight is on a component when it is active

### Additional Considerations

### Links

PLAT-123775


jeonghee27 (approved):
Looks good to me. Could you just fix lint error?

jeonghee27:
One thing I am concerning,
when there is no focus effect, it should not work when user press the OK key. Should we add that part?

PR #2865, opened 2020-10-28
PLAT-123726: Heading: italic style font cut off

Sangwook-Lee:
### Checklist

Although Korean is a language to which italic font styles can be applied, non-italic classes are being added.

### Resolution
It is deciding whether to add the `non-italic` class by checking the `scriptName`. 
We can check `Kore` to apply the italic font style to Korean.

### Additional Considerations

### Links
PLAT-123726

### DIFF
--- a/packages/i18n/I18nDecorator/getI18nClasses.js
+++ b/packages/i18n/I18nDecorator/getI18nClasses.js
@@ -45,7 +45,7 @@ function getClassesForLocale (li, options) {
 		})
 	], function (classes) {
 		const scriptName = li.getScript();
-		if (scriptName !== 'Latn' && scriptName !== 'Cyrl' && scriptName !== 'Grek') {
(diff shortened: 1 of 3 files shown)


MikyungKim (approved):
:lgtm:

packages/i18n/I18nDecorator/getI18nClasses.js
```diff
@@ -45,7 +45,7 @@ function getClassesForLocale (li, options) {
 		})
 	], function (classes) {
 		const scriptName = li.getScript();
-		if (scriptName !== 'Latn' && scriptName !== 'Cyrl' && scriptName !== 'Grek') {
+		if (scriptName !== 'Latn' && scriptName !== 'Cyrl' && scriptName !== 'Grek' && scriptName !== 'Kore') {
```
ybsung, line 48: I'm not sure that this condition is fine to all products because the font in Auto doesn't have italic font in it. There are 2 options.
1) Merge first, then update the auto font or create an additional PR for auto
2) Add more condition to check if it is on TV from enact/webOS

PR #2866, opened 2020-11-09
PLAT-125200: Call stopPropagation when enter key is pressed in touch mode 

ybsung:
### Checklist

When touching a component, focus effect has been gone, but focus is still on the component. So if pressing an `enter` key, the, the component receives the key and could execute something.

### Resolution

If pressing an `enter` key after touching, the key is ignored.

### Additional Considerations

### Links

PLAT-125200

### Comments

### DIFF
--- a/packages/spotlight/SpotlightRootDecorator/SpotlightRootDecorator.js
+++ b/packages/spotlight/SpotlightRootDecorator/SpotlightRootDecorator.js
@@ -6,8 +6,8 @@
  * @exports SpotlightRootDecorator
  */
 
-import {is} from '@enact/core/keymap';
 import hoc from '@enact/core/hoc';
+import {is} from '@enact/core/keymap';
 import React from 'react';
 
 import Spotlight from '../src/spotlight';
--- a/packages/spotlight/SpotlightRootDecorator/SpotlightRootDecorator.js
+++ b/packages/spotlight/SpotlightRootDecorator/SpotlightRootDecorator.js
@@ -80,6 +80,7 @@ const SpotlightRootDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			}
 
 			this.rootContainer = document.querySelector('#root > div');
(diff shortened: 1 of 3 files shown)


packages/spotlight/SpotlightRootDecorator/SpotlightRootDecorator.js
```diff
@@ -104,9 +104,15 @@ const SpotlightRootDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			}
 		};
 
-		handleKeyDown = () => {
-			this.rootContainer.classList.add('mouse-mode');
-			this.rootContainer.classList.remove('touch-mode');
+		handleKeyDown = (ev) => {
+			const {keyCode} = ev;
+
+			if (keyCode === 13 && this.rootContainer.classList.contains('touch-mode')) {
```
0x64, line 110:
```suggestion
			if (is('enter', keyCode) && this.rootContainer.classList.contains('touch-mode')) {
```
We need to handle 13 and 16777221 for an _enter_ key.

0x64:
Little out of this scope, I revisit already merged #2864 also.

For now, we're only interested in 'enter/OK' keys and we ignored other keys like 'cancel' key, color keys, and media keys. It looks fine as our current approach is quite defensive for various combinations of input types. If we need to take care of other keys also, then we may need to keep a state of 'non-touch' case and 'touch' case. (I believe it would be not 'mouse-mode' and 'touch-mode' since our approach in #2864 does not distinguish key input and mouse input actually.)

packages/spotlight/SpotlightRootDecorator/SpotlightRootDecorator.js
ybsung: I applied your proposal to my PR. Thanks.

0x64 (approved):
:lgtm:

PR #2868, opened 2020-11-12
PLAT-125337: Wrap the enact root DOM element with touch related CSS class

ybsung:
### Checklist

In High Contrast mode, the font color doesn't change in a Button when tapping.
Then there are many screenshot test fails.

### Resolution

`enact-a11y-high-contrast` and `non-touch-mode` CSS classes are in one DOM element. But the CSS selector related with focus effect was defined as if in different DOM elements. So I wrapped the Enact root DOM element with a div element including `non-touch-mode` or `touch-mode` CSS class.

### Additional Considerations

### Links

PLAT-125337

### Comments

I think that CHANGELOG is not needed in this PR. But if having a different opinion, please let me know.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,12 +2,6 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
-## [unreleased]
-
-### Fixed
-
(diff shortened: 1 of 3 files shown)


jeonghee27 (approved):
LGTM

PR #2871, opened 2020-11-24
PLAT-125200: Fix SpotlightRootDecorator to ignore `enter` key in touch mode

jeonghee27:

### Checklist

Image Item is checked when pressing an enter key in touch mode

### Resolution
Prevent default to prevent `onClick` event when pressing an enter key in touch mode

### Additional Considerations
This patch preventDefault onKeyDown, it can block further `onClick` event firing.
But it does not call propergation, because key event is necessary when closing an input by enter.

Since the keyDown event is propagated as it is, the side effects of it have not yet been found, but it can be careful. 
App should use the onClick event as possible.

### Links
PLAT-125200

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `spotlight/SpotlightRootDecorator` to call preventDefault when `enter` key is pressed in touch mode
+
 ## [3.4.9-experimental-1] - 2020-11-19
 
 ### Changed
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `spotlight/SpotlightRootDecorator` to call preventDefault when `enter` key is pressed in touch mode
```
MikyungKim, line 9:
```suggestion
- `spotlight/SpotlightRootDecorator` to ignore `enter` key in touch mode
```

MikyungKim (approved):
LGTM

PR #2881, opened 2020-12-16
PLAT-125327: Fix condition of .position-start-end mixin for backward compatibility

ybsung:

### Checklist

Newly added `.position-start-end` mixin broke the layout of `sandstone/Button` with `color` prop.

### Resolution

`sandstone/Button` uses `.position-start-end(-@sand-button-colordot-width, initial)` to set `left` and `right` styles to the color bar. But after the https://github.com/enactjs/enact/pull/2859/ PR, `top` and `bottom` styles were also added to the color bar and the position was broken. Because the following new mixin was called unexpectedly.

```
.position-start-end (@trbl; @target: "") when (length(@trbl) > 1) or ((iskeyword(@target)) or (@target = "")) {
	.position(@trbl);

	.enact-locale-rtl({
		.position(
			// Note: swapping left and right values
			.extract(@trbl; top)[]
			.extract(@trbl; left)[]
			.extract(@trbl; bottom)[]
			.extract(@trbl; right)[]
		);
	}; @target);
}
```

The new mixin's condition seems wrong. If passing the `initial` as the second parameter in the mixin to set the `end` value, then the second parameter will be treated as a `@target` since it is a `keyword` and the first parameter will be assigned to top, left, right, and bottom. That's the reason why unexpected top and bottom were added.

So I fixed the condition of `.position-start-end` so that the right mixin could be called.

### Additional Considerations

### Links

PLAT-125327

### DIFF
--- a/packages/ui/styles/mixins.less
+++ b/packages/ui/styles/mixins.less
@@ -24,7 +24,7 @@
 // Applies RTL-compatible start and end position to a selector
 // Simple "when" here just assumes that if the first argument is a list with more than 1 entry,
 // then units are correct. If it's just 1, then we varify the value is a number (measurement).
-.position-start-end (@trbl; @target: "") when (length(@trbl) > 1) or ((iskeyword(@target)) or (@target = "")) {
+.position-start-end (@trbl; @target: "") when (length(@trbl) > 1) or (@target = parent) or (@target = self) or (@target = "") {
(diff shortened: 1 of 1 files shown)


0x64 (approved):
:lgtm:

MikyungKim (approved):
:lgtm:

PR #2891, opened 2021-01-20
PLAT-130865: Supports a custom default event dispatching target

0x64:

### Checklist

Due to React's changes of event delegation from version 17, event callback order is changed in some components which add event handlers to `document`.

### Resolution
An API is added to set up React DOM tree root for internal components which cannot determine which node is the right event target.

### Additional Considerations
APIs to add/remove event handlers (`on` and `off`) already provide a parameter to specify the target node to attach or detach event handlers. So if an app uses Enact libraries without Enact CLI, it can simply use our existing APIs with the parameter without setting up the target node with the newly added API. The new API is for internal use.

### Links
PLAT-130865

### DIFF
--- a/packages/core/dispatcher/dispatcher.js
+++ b/packages/core/dispatcher/dispatcher.js
@@ -11,6 +11,21 @@ import curry from 'ramda/src/curry';
 
 import {getListeners, addListener} from './listeners';
 
+let defaultTarget = typeof document === 'object' && document;
+
+/*
+ * Sets a selector for the default target. If no selector is set, `document` is the default target.
+ *
+ * @function
+ *
+ * @returns {undefined}
+ * @memberof core/dispatcher
+ * @private
+ */
+const setRootId = (rootId) => {
+	defaultTarget = typeof document === 'object' && document.querySelector('#' + rootId) || false;
+};
+
 /*
  * Checks if the default target of `document` exists before returning it, otherwise returns `false`.
  *
@@ -20,7 +35,7 @@ import {getListeners, addListener} from './listeners';
  * @memberof core/dispatcher
  * @private
  */
-const getDefaultTarget = () => typeof document !== 'undefined' && document;
+const getDefaultTarget = () => defaultTarget;
 
 /*
  * Wraps event callbacks with a try-catch block to prevent unrelated code from blocking.
@@ -137,5 +152,6 @@ const once = function (name, fn, target) {
 export {
 	off,
 	on,
-	once
+	once,
+	setRootId
 };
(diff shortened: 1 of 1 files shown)


0x64:
Note:
- The new API is private since it is for internal use only.
- Unit test fails with related React 17 testing modules yet. Need to check after related modules are properly updated.
- No change log included here. We may need to add a changelog for React 17 migration not for each module packages but for the release notes.

packages/core/dispatcher/dispatcher.js
```diff
@@ -11,6 +11,21 @@ import curry from 'ramda/src/curry';
...
+ * Sets a selector for the default target. If no selector is set, `document` is the default target.
+ *
+ * @function
+ *
+ * @returns {undefined}
+ * @memberof core/dispatcher
+ * @private
+ */
+const setRootId = (rootId) => {
+	defaultTarget = typeof document === 'object' && document.querySelector('#' + rootId) || false;
```
MikyungKim, line 26: Shouldn't `false` be `defaultTarget`?

0x64: Good point! Thanks!

packages/core/dispatcher/dispatcher.js
```diff
@@ -11,6 +11,21 @@ import curry from 'ramda/src/curry';
...
+/*
+ * Sets a selector for the default target. If no selector is set, `document` is the default target.
+ *
+ * @function
+ *
+ * @returns {undefined}
+ * @memberof core/dispatcher
+ * @private
+ */
+const setRootId = (rootId) => {
```
MikyungKim, line 25:
```suggestion
const setDefaultTargetById = (id) => {
```

packages/core/dispatcher/dispatcher.js
```diff
@@ -11,6 +11,21 @@ import curry from 'ramda/src/curry';
 
 import {getListeners, addListener} from './listeners';
 
+let defaultTarget = typeof document === 'object' && document;
+
+/*
+ * Sets a selector for the default target. If no selector is set, `document` is the default target.
+ *
+ * @function
+ *
```
MikyungKim, line 20: `@param`

0x64: ✅

MikyungKim (approved):
LGTM

packages/core/dispatcher/dispatcher.js
```diff
@@ -11,6 +11,22 @@ import curry from 'ramda/src/curry';
 
 import {getListeners, addListener} from './listeners';
 
+let defaultTarget = typeof document === 'object' && document;
+
+/*
+ * Sets a selector for the default target. If no selector is set, `document` is the default target.
+ *
+ * @function
+ * @param	{String}	id	Node id of the default target
```
jeonghee27, line 20: Tab is right?

PR #2895, opened 2021-02-02
Revert "PLAT-108976: Fix Marquee to restart animation when text changed while focus retained"

0x64:
Reverts enactjs/enact#2861


MikyungKim:
Due to sandstone/dropdown regression, we want to revert this and revisit it in the future.

MikyungKim (approved):
LGTM

PR #2905, opened 2021-03-11
PLAT-137759: Fix FloatingLayerDecorator to render the floating node properly

MikyungKim:

### Checklist

From the theme libs, the Popup kind is not opened in some samples.

### Resolution
While we migrated `FloatingLayerDecorator` to use hooks, handler that uses modules from `@enact/core/handle` had some problem with binding `this`. So I replaced the related logic to not use the related modules. 

### Additional Considerations
N/A

### Links
PLAT-137759


0x64 (approved):
:lgtm:

PR #2916, opened 2021-05-03
PLAT-139596: Fix console error on Spottable

MikyungKim:
### Checklist

In some samples, we've seen the below console error
```Uncaught Error: Invalid hook call. Hooks can only be called inside of the body of a function component. This could happen for one of the following reasons:...```

### Resolution
Removed invalid hook calling in a class and replaced it with a function call that has `forceUpdate()`.

### Additional Considerations

### Links
PLAT-139596


jeonghee27 (approved):
LGTM

PR #2917, opened 2021-05-10
PLAT-126935: Refined overscroll effect

0x64:

### Checklist

For better UX in theme libraries, the default overscroll effect factor needs to be refined. From this PR, the default effect factor is 0.5 which is half of the maximum effect (1.0) and it will help theme libraries to make effects bigger than the default easily for theme-specific cases.

### Resolution
Change the default effect factor from 1 to 0.5.

### Additional Considerations

### Links
PLAT-126935


packages/ui/Scrollable/Scrollable.js
```diff
@@ -35,6 +35,7 @@ const
 		isPageDown: is('pageDown'),
 		isPageUp: is('pageUp'),
 		nop: () => {},
+		overscrollDefaultRatio: 0.5,
```
MikyungKim, line 38: Why did you also change here?
I think the effect of moonstone is different from sandstone. I think we don't need to change `Scrollable`.

MikyungKim (approved):
LGTM

PR #2929, opened 2021-06-17
PLAT-143666: Fix MarqueeDecorator to stop marquee properly after hiding pointer when `marqueeOn` is `hover`

jeonghee27:
### Checklist

Header title is not stopped marquee when focused on slotafter button.

### Resolution
Pointer hide key event was occurred on body, because there is not focus.
So the key event handler should be added on document instead of root node.

### Additional Considerations

### Links
PLAT-143666


0x64 (approved):
:lgtm:

packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -344,7 +344,7 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			if (this.props.marqueeOn === 'render') {
 				this.startAnimation(this.props.marqueeOnRenderDelay);
 			}
-			on('keydown', this.handlePointerHide);
+			on('keydown', this.handlePointerHide, document);
```
MikyungKim, line 347: I think we should add `document` to `off` function at line 405. Please check.

MikyungKim (approved):
:shipit:

PR #2932, opened 2021-07-07
WRN-1635: Make ui/ViewManager Sampler

YunhoLee0:
### Checklist

There is no ViewManager sampler in ui/ViewManger to verify the behavior with Arranger.
ViewManager sample needs to be added.

### Resolution
Add ViewManager sample that works with Arranger.

### Links
WRN-1635

### DIFF
--- a/packages/sampler/stories/default/ViewManager.js
+++ b/packages/sampler/stories/default/ViewManager.js
@@ -46,7 +46,7 @@ class ViewManagerLayout extends Component {
 
 	render () {
 		const {selected} = this.state;
-	
+
 		return (
 			<Layout>
 				<Cell component={Scroller} size="7%">
--- a/packages/sampler/stories/default/ViewManager.js
+++ b/packages/sampler/stories/default/ViewManager.js
@@ -4,6 +4,7 @@ import {mergeComponentMetadata} from '@enact/storybook-utils';
 import Group from '@enact/ui/Group';
 import Item from '@enact/ui/Item';
 import Layout, {Cell} from '@enact/ui/Layout';
+import ri from '@enact/ui/resolution';
 import Scroller from '@enact/ui/Scroller';
 import {arrange, SlideArranger, SlideBottomArranger, SlideLeftArranger, SlideRightArranger, SlideTopArranger, ViewManager, ViewManagerBase} from '@enact/ui/ViewManager';
 import {useCallback, useState} from 'react';
@@ -21,13 +22,17 @@ const prop = {
 		SlideArranger: null,
 		'CustomArranger (FadeAndSlideArranger)': null
 	},
-	direction: ['bottom', 'left', 'right', 'top']
+	direction: ['bottom', 'left', 'right', 'top'],
+	end: ['index', 'index+1', 'index+2'],
+	start: ['index', 'index-1', 'index-2']
 };
 
-const views = new Array(10).fill().map((i, index) => {
+const itemSize = 10;
+
+const views = new Array(itemSize).fill().map((i, index) => {
 	return {
-		title: `Item ${index + 1}`,
-		content: `This is View ${index + 1}`
+		title: `Item ${index}`,
+		content: `This is View ${index}`
 	};
 });
 
@@ -36,6 +41,27 @@ const ViewManagerLayout = (props) => {
 	const handleChangeView = useCallback((ev) => {
 		setSelected(ev.selected);
 	}, [setSelected]);
+
+	const selectedEnd = select('end', prop.end, ViewManagerConfig, prop.end[0]);
+	const selectedStart = select('start', prop.start, ViewManagerConfig, prop.start[0]);
+	const endRange = [selected, selected+1, selected+2];
+	const startRange = [selected, selected-1, selected-2];
+	let end, start;
+
+	for (let i = 0; i < 3; i++) {
+		if (selectedEnd === prop.end[i]) {
+			end = endRange[i];
+		}
+		if (selectedStart === prop.start[i]) {
+			start = startRange[i];
+		}
+	}
+	if (start < 0) {
+		start = 0;
+	}
+	if (end >= itemSize) {
+		end = itemSize - 1;
+	}
 
 	return (
 		<Layout>
@@ -50,9 +76,11 @@ const ViewManagerLayout = (props) => {
 				</Group>
 			</Cell>
 			<Cell
-				className={css.viewport}
 				component={ViewManager}
+				end={end}
 				index={selected}
+				start={start}
+				style={{height: ri.scale(42 * (end - start + 1)), overflow: 'hidden'}}
 				{...props}
 			>
 				{views.map((view, i) => (
@@ -107,7 +135,6 @@ export const _ViewManager = () => {
 		<ViewManagerLayout
 			arranger={arranger}
 			duration={number('duration', ViewManagerConfig)}
-			end={number('end', ViewManagerConfig)}
 			enteringDelay={number('enteringDelay', ViewManagerConfig)}
 			noAnimation={boolean('noAnimation', ViewManagerConfig)}
 			onAppear={action('onAppear')}
@@ -118,7 +145,6 @@ export const _ViewManager = () => {
 			onWillTransition={action('onWillTransition')}
 			reverseTransition={boolean('reverseTransition', ViewManagerConfig)}
 			rtl={boolean('rtl', ViewManagerConfig)}
-			start={number('start', ViewManagerConfig)}
 		/>
 	);
 };
--- a/packages/sampler/stories/default/ViewManager.js
+++ b/packages/sampler/stories/default/ViewManager.js
@@ -5,12 +5,12 @@ import Group from '@enact/ui/Group';
 import Item from '@enact/ui/Item';
 import Layout, {Cell} from '@enact/ui/Layout';
 import Scroller from '@enact/ui/Scroller';
-import {arrange, SlideArranger, SlideBottomArranger, SlideLeftArranger, SlideRightArranger, SlideTopArranger, ViewManager} from '@enact/ui/ViewManager';
+import {arrange, SlideArranger, SlideBottomArranger, SlideLeftArranger, SlideRightArranger, SlideTopArranger, ViewManager, ViewManagerBase} from '@enact/ui/ViewManager';
 import {useCallback, useState} from 'react';
 
 import css from './ViewManager.module.less';
 
-const ViewManagerConfig = mergeComponentMetadata('ViewManager', ViewManager);
+const ViewManagerConfig = mergeComponentMetadata('ViewManager', ViewManagerBase, ViewManager);
 
 ViewManagerConfig.defaultProps = {
 	duration: 300,
@@ -36,7 +36,7 @@ const views = new Array(10).fill().map((i, index) => {
 	};
 });
 
-function ViewManagerLayout (props) {
+const ViewManagerLayout = (props) => {
 	const [selected, setSelected] = useState(0);
 	const handleChangeView = useCallback((state) => {
 		setSelected(state.selected);
@@ -56,19 +56,20 @@ function ViewManagerLayout (props) {
 				</Group>
 			</Cell>
 			<Cell
+				className={css.viewport}
 				component={ViewManager}
 				index={selected}
 				{...props}
 			>
 				{views.map((view, i) => (
-					<div className={css.view} key={i}>
+					<div className={css.box} key={i}>
 						{view.content}
 					</div>
 				))}
 			</Cell>
 		</Layout>
 	);
-}
+};
 
 export default {
 	title: 'UI/ViewManager',
--- a/packages/sampler/stories/default/ViewManager.js
+++ b/packages/sampler/stories/default/ViewManager.js
@@ -44,8 +44,8 @@ const ViewManagerLayout = (props) => {
 
 	const selectedEnd = select('end', prop.end, ViewManagerConfig, prop.end[0]);
 	const selectedStart = select('start', prop.start, ViewManagerConfig, prop.start[0]);
-	const endRange = [selected, selected+1, selected+2];
-	const startRange = [selected, selected-1, selected-2];
+	const endRange = [selected, selected + 1, selected + 2];
+	const startRange = [selected, selected - 1, selected - 2];
 	let end, start;
 
 	for (let i = 0; i < 3; i++) {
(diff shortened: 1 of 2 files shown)


packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+const views = new Array(10).fill().map((i, index) => {
+	return {
+		title: `Item ${index + 1}`,
+		content: `This is View ${index + 1}`
+	};
+});
+
+class ViewManagerLayout extends Component {
```
MikyungKim, line 39: Can we use `useState`?

YunhoLee0: I've changed the class to function with useState.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,134 @@
...
+import Group from '@enact/ui/Group';
+import Item from '@enact/ui/Item';
+import Layout, {Cell} from '@enact/ui/Layout';
+import Scroller from '@enact/ui/Scroller';
+import {arrange, SlideArranger, SlideBottomArranger, SlideLeftArranger, SlideRightArranger, SlideTopArranger, ViewManager} from '@enact/ui/ViewManager';
+import {useCallback, useState} from 'react';
+
+import css from './ViewManager.module.less';
+
+const ViewManagerConfig = mergeComponentMetadata('ViewManager', ViewManager);
```
MikyungKim, line 13: Let's add `VIewManagerBase` for default props.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,134 @@
...
+
+ViewManagerConfig.defaultProps = {
```
MikyungKim, line 15: Maybe this is not needed...

packages/sampler/stories/default/ViewManager.module.less
```diff
@@ -0,0 +1,18 @@
...
+
+.navItem {
+	border-bottom: solid black;
+}
+
+.view {
+	height: 100px;
+	width: 600px;
+	background-color: @bg-color;
+}
```
MikyungKim, line 18: Please add one empty line.

packages/sampler/stories/default/ViewManager.module.less
```diff
@@ -0,0 +1,18 @@
...
+
+.navItem {
+	border-bottom: solid black;
+}
+
+.view {
+	height: 100px;
+	width: 600px;
+	background-color: @bg-color;
```
MikyungKim, line 17: do we need this?

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,134 @@
...
+};
+
+const views = new Array(10).fill().map((i, index) => {
+	return {
+		title: `Item ${index + 1}`,
+		content: `This is View ${index + 1}`
+	};
+});
+
+function ViewManagerLayout (props) {
```
MikyungKim, line 39: It's okay but let's make this an arrow function.

MikyungKim (commented):
Let's make a viewport to hide the away view.
And how about adjusting the height of the content to fit the navigation area?

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I've added it.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I've added `ViewManagerBase`, but defaultProps prop is not attached. So I think this is needed.

packages/sampler/stories/default/ViewManager.module.less
YunhoLee0: OK

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I've changed this to an arrow function.

YunhoLee0:
I've added a style to hide the away view.
And currently, the height of the content fits the navigation area.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,135 @@
...
+const views = new Array(10).fill().map((i, index) => {
+	return {
+		title: `Item ${index + 1}`,
+		content: `This is View ${index + 1}`
+	};
+});
+
+const ViewManagerLayout = (props) => {
+	const [selected, setSelected] = useState(0);
+	const handleChangeView = useCallback((state) => {
```
MikyungKim, line 41:
```suggestion
	const handleChangeView = useCallback((ev) => {
```

YunhoLee0: OK

MikyungKim (approved):
:lgtm:

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,135 @@
...
+		arranger = FadeAndSlideArranger;
+	} else {
+		arranger = arrangers[arrangerType];
+	}
+
+	return (
+		<ViewManagerLayout
+			arranger={arranger}
+			duration={number('duration', ViewManagerConfig)}
+			enteringDelay={number('enteringDelay', ViewManagerConfig)}
```
jeonghee27, line 116: is this prop working? or is there bug on this prop?

packages/sampler/stories/default/ViewManager.js
jeonghee27: As far as I checked, defaultProps is attached well.
But there is no `enteringDelay` default value on the ViewManger component side.
If there is bug about `enteringDelay` let's make new issue for fix this and define default value.
or let's define default value on the ViewManagerBase component.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,135 @@
...
+				], animationOptions);
+			}
+		};
+		arranger = FadeAndSlideArranger;
+	} else {
+		arranger = arrangers[arrangerType];
+	}
+
+	return (
+		<ViewManagerLayout
```
jeonghee27, line 113: Why there is no `start` and `end` prop?

In fact the reason I need Viewmanager sample is fot this two props. it releated with WRN-853.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,135 @@
...
+	arranger: ['SlideBottomArranger', 'SlideLeftArranger', 'SlideRightArranger', 'SlideTopArranger', 'SlideArranger', 'CustomArranger (FadeAndSlideArranger)'],
+	direction: ['bottom', 'left', 'right', 'top']
+};
+
+const arrangers = {
+	'SlideBottomArranger': SlideBottomArranger,
+	'SlideLeftArranger': SlideLeftArranger,
+	'SlideRightArranger': SlideRightArranger,
+	'SlideTopArranger': SlideTopArranger
+};
```
jeonghee27, line 30:
```suggestion
const arrangers = {
	SlideBottomArranger,
	SlideLeftArranger,
	SlideRightArranger,
	SlideTopArranger
};
```
But, it still not enough.
If i were you I will use only one variable.
it duplicated with prop.arranger

```

const prop = {
	arranger: {
		SlideBottomArranger,
		SlideLeftArranger,
		SlideRightArranger,
		SlideTopArranger,
		SlideArranger: null,
		CustomArranger: null
	},
	direction: ['bottom', 'left', 'right', 'top']
};


	const arrangerType = select('arranger', Object.keys(prop.arranger), ViewManagerConfig, 'SlideBottomArranger');
	
		arranger = prop.arranger[arrangerType];
```

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,135 @@
...
+	let arranger;
+
+	if (arrangerType === 'SlideArranger') {
+		const amount = number('amount', ViewManagerConfig, {range: true, min: 0, max: 100}, 100);
+		const direction = select('direction', prop.direction, ViewManagerConfig, prop.direction[0]);
+		arranger = SlideArranger({amount, direction});
+	} else if (arrangerType === 'CustomArranger (FadeAndSlideArranger)') {
+		// The following arranger is from sandstone/internal/Panels/Arrangers.
+		const quadInOut = 'cubic-bezier(0.455, 0.030, 0.515, 0.955)';
+		const animationOptions = {easing: quadInOut};
```
jeonghee27, line 90: Did you define a variable that used only once in order to give it a meaning of quadInOut ?

packages/sampler/stories/default/ViewManager.module.less
```diff
@@ -0,0 +1,24 @@
+@import "../../src/styles/colors.less";
+
+.group {
+	> * {
+		padding: 9px;
+		margin-right: 12px;
+	}
+}
```
jeonghee27, line 8: I don't think it is necessary styles. Is there any reason to need some pading or margin?
Also I don't think the selecting way is good. It's not intuitive which one will be chosen. It would have been better to pass a class to the Item.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -0,0 +1,135 @@
...
+	component: 'ViewManager'
+};
+
+export const _ViewManager = () => {
+	const arrangerType = select('arranger', prop.arranger, ViewManagerConfig, prop.arranger[0]);
+	let arranger;
+
+	if (arrangerType === 'SlideArranger') {
+		const amount = number('amount', ViewManagerConfig, {range: true, min: 0, max: 100}, 100);
+		const direction = select('direction', prop.direction, ViewManagerConfig, prop.direction[0]);
```
jeonghee27, line 85: These two prop is not ViewManager prop. and it make confusion.
Let's make another knob tab 

```suggestion
		const amount = number('amount', SliderArrangerConfig, {range: true, min: 0, max: 100}, 100);
		const direction = select('direction', prop.direction, SliderArrangerConfig, prop.direction[0]);
```

packages/sampler/stories/default/ViewManager.js
YunhoLee0: By rechecking it, defaultProps is attached well. I guess there were some mistakes to check. So above code that defines defaultProps is not needed.
Furthermore, `enteringDelay` default value is missing in the component code.
I'd create a new issue to fix it after this is merged.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: That was because those props can't be set by fixed default value.
I've added those props with an empty initial state.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: That's what I should know. Thank you.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: You're right. It's fixed.

packages/sampler/stories/default/ViewManager.module.less
YunhoLee0: It's removed.
It was for giving more visibility similar to the navigation bar on the left side and contents on the right side.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: OK

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I'm looking at how this prop working. It seems the prop doesn't work.

jeonghee27: Please make new issue about that.

packages/sampler/stories/default/ViewManager.js
jeonghee27: This is not what I meant, and I cannot any  behavior about `start` and `end` in this current sample.
Since `start` and `end` are values that depend on the current index, So they cannot be received as single knobs.
If we need knobs, I think we can only delta value based on the index.

Please check the behavior using the code below.
```
				start={selected - 1}
				end={selected + 1}
				index={selected}
```

```
.viewport {
	// height: 42px;
.box {
	// height: 100%;
```
I thought the reason I expected this issue to take long was because of these props.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I've created the issue. (WRN-3728)

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I've changed how `start` and `end` work in this sample.
It works with some selected values among the range that depend on the current index.
Also, the height of the viewport is dependent on how many views show.

packages/sampler/stories/default/ViewManager.js
```diff
@@ -37,6 +42,27 @@ const ViewManagerLayout = (props) => {
...
+		if (selectedStart === prop.start[i]) {
+			start = startRange[i];
+		}
+	}
+	if (start < 0) {
+		start = 0;
+	}
+	if (end >= itemSize) {
+		end = itemSize - 1;
+	}
```
jeonghee27, line 64: Please use Math.min Math.max

packages/sampler/stories/default/ViewManager.js
```diff
@@ -37,6 +42,27 @@ const ViewManagerLayout = (props) => {
 		setSelected(ev.selected);
 	}, [setSelected]);
 
+	const selectedEnd = select('end', prop.end, ViewManagerConfig, prop.end[0]);
+	const selectedStart = select('start', prop.start, ViewManagerConfig, prop.start[0]);
```
jeonghee27, line 46:
```suggestion
	const selectedStart = select('start', Object.keys(prop.start), ViewManagerConfig, prop.start[0]);
```

Can't we use object for end and start?
end : {'index':0, 'index+1':1, 'index+2':2},

packages/sampler/stories/default/ViewManager.js
YunhoLee0: I've changed to use them.

packages/sampler/stories/default/ViewManager.js
YunhoLee0: Thanks. I've changed to use object for them.

jeonghee27 (approved):
LGTM

PR #2935, opened 2021-07-27
WRN-1264 : Add support for iterating over HOC configs

sjro:

### Checklist

Need to compare if HOC default config is changed like API change 

### Resolution
Add a displayName and defaultConfigs 

### Additional Considerations

### Links
WRN-1264

### DIFF
--- a/packages/core/hoc/hoc.js
+++ b/packages/core/hoc/hoc.js
@@ -70,9 +70,7 @@ const hoc = (defaultConfig, hawk) => {
 		defaults = null;
 	}
 
-	let Component;
-
-	Component = (config, maybeWrapped) => {
+	const Component = (config, maybeWrapped) => {
 		if (isRenderable(config)) {
 			return factory(defaults, config);
 		} else {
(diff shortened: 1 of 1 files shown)


packages/core/hoc/hoc.js
```diff
@@ -70,7 +70,9 @@ const hoc = (defaultConfig, hawk) => {
 		defaults = null;
 	}
 
-	return (config, maybeWrapped) => {
+	let Component;
```
0x64, line 73: It looks better to define `Component` as `const` variable with assignment below.

packages/core/hoc/hoc.js
```diff
@@ -70,7 +70,9 @@ const hoc = (defaultConfig, hawk) => {
...
+
+	Component = (config, maybeWrapped) => {
```
0x64, line 75:
```suggestion
	const Component = (config, maybeWrapped) => {
```

0x64 (approved):
:lgtm:

seunghoh (approved):

PR #2937, opened 2021-08-04
WRN-3773: Fix VirtualList not to provide invalid item refs to theme libraries

0x64:

### Checklist

Theme libraries cannot access item's rendered DOM node when item's content is updated without updating a list.

### Resolution
`enact/ui` provides rendered item nodes to theme libraries to help access those DOM nodes.
If item's DOM node is updated by content update without updating a list (by prop changing or any other reason), a list does NOT refresh nodes' info so theme libraries tries to access invalid DOM nodes which are not on the DOM tree anymore.
In this PR, `enact/ui` provides rendered item *container* nodes instead of item nodes. Since item container nodes are not updated by item renderers, theme libraries are able to valid DOM nodes through item container nodes.

### Additional Considerations

### Links
WRN-3773


0x64:
Close as we decided not to apply this change.

PR #2942, opened 2021-09-07
WRN-6365: Fixed Marquee to not separate layer when short text

jeonghee27:

### Checklist

Marquee text with short content has willAnimate class

### Resolution
`willAnimate` prop follows `state.promoted`.

Logic about `promoted` It came from #1959, but the text length was not considered  from the beginning. It make  `state.promoted=true`  according to marqueeOn condition.

`state.promoted` prop becomes `true` in almost case by `promoteJob`. This is because the default value of `contentFits` was `false`. it was set on `invalidateMetrics()`;
on the other side, `InvalidateMetrics()` set `state.promoted` to `false` for reset, but in `promotedJob()` it is set to true because of the contentFit=false, so this has wrong logic.
`contentFits` is calculated on `calculateMetrics` . We should refer the value after calculating. 

In this PR, I set `null` for default `contentFits` value

### Additional Considerations

### Links
WRN-6365

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,12 +2,6 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
-## [unreleased]
-
-### Changed
-
-- `ui/MarqueeDecorator` to not separate layer when short text
-
 ## [4.0.5] - 2021-08-02
 
 No significant changes.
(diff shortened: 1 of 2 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Changed
+
+- `ui/MarqueeDecorator` to not separate layer when short text
```
MikyungKim, line 9: I think we don't need a change log for this.

MikyungKim (approved):
LGTM 👍

MikyungKim:
But, we may need to test Marquee TCs for this change manually.

PR #2957, opened 2021-10-20
WRN-5708, WRN-8317: Added support for multiple ilib resources paths and detecting changed resources

MikyungKim:

### Checklist

WRN-5708:
For webOS that supports multiple resources, we need to invalidate the cache of strings. We've added a timestamp in ilibmanifest.json file so that we could validate the resources.
WRN-8317:
Added support for loading given multiple resources paths, e.g. resources, resources_0,...

### Resolution
WRN-5708:
Checking timestamp of `ilibmanifest.json` file and remove the cache of the strings.
WRN-8317:
Retrieve the result of additional resources path and combine the result with the default resources path

### Additional Considerations
WRN-5708 is dependent with WRN-8317 so the PR is combined.

### Links
WRN-5708, WRN-8317


packages/i18n/src/Loader.js
```diff
@@ -221,10 +242,23 @@ EnyoLoader.prototype.loadFiles = function (paths, sync, params, callback, rootPa
...
+						// Overwrite the _root/path result
+						Object.assign(ret[ret.length - 1], json);
+					}
+				};
 
 				if (this.isAvailable(_root, path)) {
 					getSync(this._pathjoin(_root, path), handler);
+
+					if (this.addPaths && Array.isArray(this.addPaths)) {
+						for (let addedRoot of this.addPaths) {
```
gouniLee, line 256: fyi. addPath() takes "String" or "Array". so If ILIB_ADDITIONAL_RESOURCES_PATH value is a string, this.addPaths array length 1 always here.

packages/i18n/src/Loader.js
```diff
@@ -221,10 +242,23 @@ EnyoLoader.prototype.loadFiles = function (paths, sync, params, callback, rootPa
 					}
 				};
 
-				this.loadManifestsSync(_root);
+				const handleAdditionalResourcesPath = (json, err) => {
+					if (found && !err && typeof json === 'object') {
```
gouniLee, line 246: Question) doesn't need to consider the returning `text` type case here?
somethig like: `if (typeof(text) === "string") { text = JSON.parse(text);}`

packages/i18n/src/Loader.js
MikyungKim: I know. I wrote this for general usage.

packages/i18n/src/Loader.js
MikyungKim: We've already parsed the result from `getImpl` function.

packages/i18n/src/Loader.js
```diff
@@ -50,6 +50,8 @@ const iLibResources = typeof ILIB_RESOURCES_PATH === 'undefined' ? '/locale' : I
 const cachePrefix = 'ENACT-ILIB-';
 const cacheKey = cachePrefix + 'CACHE-ID';
 const cacheID = typeof ILIB_CACHE_ID === 'undefined' ? '$ILIB' : ILIB_CACHE_ID;
+const timeStampKey = 'l10n_timestamp';
```
jeonghee27, line 53: 10n?

MikyungKim: It means localization.

jeonghee27 (approved):
LGTM

PR #2981, opened 2021-12-08
WRN-3728: Ui/ViewManager: 'enteringDelay' prop seems doesn't work and missing default value

sjro:
### Checklist

Resolve 'enteringDelay' control not working

### Resolution
Add a control to set 'enteringProp' and change the content of the view to show that the transition of the view is finished after 'enteringDelay'

### Additional Considerations
In storybook version 6.4.x, there was a problem that the control of the storybook was not visible, so it was fixed to 6.3.14

### Links
WRN-3728

### Comments


packages/sampler/stories/default/ViewManager.js
```diff
@@ -70,13 +79,15 @@ const ViewManagerLayout = (props) => {
...
 				index={selected}
 				start={start}
+				enteringProp={enteringProp}
+				enteringDelay={enteringDelay}
 				style={{height: ri.scale(42 * (end - start + 1)), overflow: 'hidden'}}
 				{...rest}
 			>
 				{views.map((view, i) => (
-					<div className={css.box} key={i}>
+					<ChildrenDiv className={css.box} enteringProp={enteringProp} key={i} >
```
0x64, line 88: Let's update this code to use another way (e.g. passing 'args') rather than using `enteringProp` directly since it could lead to a misunderstanding that an app developer should pass `enteringProp` to children components.

sjro: fixed it.

0x64 (approved):
:lgtm:

package-lock.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "enact",
-  "version": "4.1.0",
+  "version": "4.1.1",
```
jeonghee27, line 3: please exclude the package.jsonpac, pakage-lock, and npm-shrinkwrap.json

packages/sampler/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact Sampler, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `ui/ViewManager` sample
```
jeonghee27, line 9: We usually don't leave Changelog when we fix stories.
Let's remove this also

packages/sampler/stories/default/ViewManager.js
```diff
@@ -74,9 +87,9 @@ const ViewManagerLayout = (props) => {
 				{...rest}
 			>
 				{views.map((view, i) => (
-					<div className={css.box} key={i}>
+					<ChildrenDiv args={args} className={css.box} key={i} >
```
jeonghee27, line 90:
```suggestion
					<ChildrenDiv args={args} className={css.box} key={i}>
```

package-lock.json
sjro: fixed it

packages/sampler/CHANGELOG.md
sjro: remove it

PR #2987, opened 2021-12-16
WRN-6761: VirtualGridListNative keeps scrolling after rotated while scrolling in Android

alexandrumorariu:
### Checklist

Changed the structure of the handleResizeWinodw from both Scrollable Native and useScroll so that the problem is less visible.
This problem was reproduced on all the components that used Scrollable Native and useScroll across all the libraries. This solution does not cover the Safari browser for IOS because of a strange interaction between overflow hidden and it (the items remain invisible until you scroll a bit).

### Resolution

### Additional Considerations
The solution is not a perfect fix as in some rare cases the list is not 100% at the top(the worst I've seen it was a little bit past it but the first element was still visible) but it should stop the scrolling effect from after the screen was rotated.

### Links
WRN-6761

### DIFF
--- a/packages/ui/Scrollable/ScrollableNative.js
+++ b/packages/ui/Scrollable/ScrollableNative.js
@@ -489,9 +489,20 @@ class ScrollableBaseNative extends Component {
 			if (handleResizeWindow) {
 				handleResizeWindow();
 			}
+			if ( platform.platformName === 'androidChrome' || platform.platformName === 'ios' ) {
+				this.childRefCurrent.containerRef.current.style.overflow = 'hidden';
+
+			}
 			this.childRefCurrent.containerRef.current.style.scrollBehavior = null;
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,8 @@ The following is a curated list of changes in the Enact project, newest changes
 
 ### Fixed
 
+- `ui/Scrollable/ScrollableNative` changed the behaviour of the window resize method
+- `ui/useScroll/useScroll` changed the behaviour of the window resize method
```
daniel-stoian-lgp, line 18:
```suggestion
- `ui/Scrollable/ScrollableNative` stopped scroller momentum when the window resizes
- `ui/useScroll/useScroll` stopped scroller momentum when the window resizes
```

adrian-cocoara-lgp (approved):
LGTM

PR #3006, opened 2022-02-15
WRN-17310: Add sampler for Drawing in enact/packages/sampler

adrian-cocoara-lgp:
Enact-DCO-1.0-Signed-off-by: Adrian Cocoara adrian.cocoara@lgepartner.com

### Checklist

Added sampler for ui/Drawing

### Resolution

### Additional Considerations

### Links
WRN-17310

### DIFF
--- a/packages/sampler/stories/default/Drawing.js
+++ b/packages/sampler/stories/default/Drawing.js
@@ -1,0 +1,31 @@
+import {mergeComponentMetadata} from '@enact/storybook-utils';
+import {boolean, range, select} from '@enact/storybook-utils/addons/controls';
+
+import Drawing from '../../../ui/Drawing';
+
+const DrawingConfig = mergeComponentMetadata('Drawing', Drawing);
+
+export default {
+	title: 'UI/Drawing',
+	component: 'Drawing'
+}
+
+export const _Drawing = (args) => (
+	<Drawing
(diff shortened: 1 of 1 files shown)


packages/sampler/stories/default/Drawing.js
```diff
@@ -0,0 +1,31 @@
+import {mergeComponentMetadata} from '@enact/storybook-utils';
+import {boolean, range, select} from '@enact/storybook-utils/addons/controls';
+
+import Drawing from '../../../ui/Drawing';
```
stanca-pop-lgp, line 4: Could you import this from the enact package, not locally?

packages/sampler/stories/default/Drawing.js
```diff
@@ -0,0 +1,31 @@
...
+
+const DrawingConfig = mergeComponentMetadata('Drawing', Drawing);
+
+export default {
+	title: 'UI/Drawing',
```
stanca-pop-lgp, line 9: Please sort.

adrian-cocoara-lgp: this is the order used in every sample. we should leave this as it is now

stanca-pop-lgp (approved):
LGTM

PR #3018, opened 2022-02-25
WRN-12923: Fixed children of `Button` when `icon` is true

sjro:
### Checklist

If `icon` prop is `true`, it should not be passed as children.

### Resolution
If `icon` is true, don't pass this as children.

### Additional Considerations

### Links
WRN-12923

### Comments
Enact-DCO-1.0-Signed-off-by: SJ RO ([sj.ro@lge.com](mailto:sj.ro@lge.com))


0x64 (approved):

packages/ui/Button/Button.js
```diff
@@ -216,11 +216,17 @@ const ButtonBase = kind({
...
-				</ComponentOverride>
-			);
+			if (icon === true) {
+				return (
+					<ComponentOverride {...props} />
+				);
+			} else {
+				return (
+					<ComponentOverride {...props}>
+						{icon}
```
MikyungKim, line 226: How about
```suggestion
						{icon === true ? null : icon}
```
instead of `if/else`?

packages/ui/Button/Button.js
```diff
@@ -216,11 +216,17 @@ const ButtonBase = kind({
...
+					</ComponentOverride>
+				);
+			}
```
MikyungKim, line 229: how about ..
```suggestion
			if (typeof icon === 'string') {
				props.children = icon;
			}

			return (
				<ComponentOverride {...props} />
			);
```

MikyungKim (commented):
Please add a change log for this.

MikyungKim (commented):
Please update your branch to the recent develop.

MikyungKim (approved):
:lgtm:

PR #3035, opened 2022-03-25
WRN-18494: Remove unnecessary wrapper node on FloatingLayer

jeonghee27:
### Checklist

When two floatinglayers are displayed, there is a bug that a popup which open later shown on down of other popup .
Because `FloatingLayer `reuses nodes , If you reopen a previously opened Popup, it may open below other popups.

### Resolution
I removed unnecessary wrapper node that is reused on FloatingLayer

### Additional Considerations
I wanted to remove the `state.nodeRendered` also, but I failed. There was much bug when I remove the variable
In the case of a popup that is start with open, re-rendering is required after the ref of this.floatingLayer is bound.

### Links
WRN-18494

### Comments

### DIFF
--- a/packages/ui/FloatingLayer/FloatingLayer.js
+++ b/packages/ui/FloatingLayer/FloatingLayer.js
@@ -104,7 +104,7 @@ class FloatingLayerBase extends Component {
 	static contextType = FloatingLayerContext;
 
 	static defaultProps = {
-		floatLayerClassName: 'enact-fit enact-clip enact-untouchable',
+		floatLayerClassName: 'enact-fit enact-clip enact-untouchable enact-z-index',
 		floatLayerId: 'floatLayer',
 		noAutoDismiss: false,
 		open: false,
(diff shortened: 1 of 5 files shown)


0x64 (approved):
:lgtm:

packages/ui/FloatingLayer/FloatingLayer.js
```diff
@@ -211,44 +213,37 @@ class FloatingLayerBase extends Component {
...
 
-		this.node = document.createElement('div');
-		this.node.className = floatLayerClassName;
-		this.node.style.zIndex = 100;
-
-		this.floatingLayer.appendChild(this.node);
-		on('scroll', this.handleScroll, this.node);
+		on('scroll', this.handleScroll, this.floatingLayer);
 
 		// render children when this.node is inserted in the DOM tree.
```
juwonjeong, line 226: Would it be better to remove this comment as this.node is no longer used?

juwonjeong (approved):
LGTM

PR #3039, opened 2022-04-01
WRN-19696 : [React18] Fix unit test fail

sjro:
### Checklist

following error fix at enact test 
console.error
    Warning: An update to xxxx inside a test was not wrapped in act(...).
    When testing, code that causes React state updates should be wrapped into act(...):
    act(() => {
      /* fire events that update state */
    });

### Resolution
React is trying to warn us that something happened to our component, when we weren't expecting anything to happen at all. We're supposed to wrap every interaction we make with our component in an act function call, so that React know that we expect our component to perform some updates.
So I wrap every our own function to wrap it in act() and call it.
And use jest.useFakeTimer() to activate the faketimer for test. 

### Additional Considerations

### Links
WRN-19696

### Comments


packages/ui/ViewManager/tests/View-specs.js
```diff
@@ -132,7 +141,8 @@ describe('View', () => {
 				done();
 			});
 
-			ref.current.componentWillStay(spy);
+			act(() => ref.current.componentWillStay(spy));
+			jest.runAllTimers();
```
jeonghee27, line 145: Please check whether this is  necessary or not.

jeonghee27:
Could you please resolve conflicts?

packages/ui/ViewManager/tests/View-specs.js
jeonghee27: maybe..

```suggestion
			done();
```

packages/ui/ViewManager/tests/ViewManager-specs.js
```diff
@@ -1,12 +1,20 @@
...
 describe('ViewManager', () => {
 	// Suite-wide setup
+	beforeEach(() => {
+		jest.useFakeTimers();
+	});
+
+	afterEach(() => {
+		jest.useRealTimers();
+	});
+
```
jeonghee27, line 17: How about move this "useFakeTimers" to TC.

I think it is required on one TC

sjro: Since setTimeout is used in the Mock Arranger function, it is required for all test cases that use the Mock Arranger.

packages/ui/ViewManager/tests/View-specs.js
sjro: Since setTimeout is used in the Mock Arranger function, we need to advance the timer's time with jest.runAllTimers() .

jeonghee27 (approved):
LGTM

jeonghee27 (approved):
LGTM

PR #3047, opened 2022-04-15
WRO-692: Changed Spotlight to not focus on an invisible element

jeonghee27:
### Checklist

The current spotlight found target including invisible elements.
In that case the element cannot get focus. So consequently focus is trapped.

### Resolution
- Exclude invisible element when searching target (display none / visibility hidden)
- Improve performance of "querySelector" function
- Use isNavigable rather than navigableFilter to exclude disabled  container.

### Additional Considerations

### Links
WRO-692

### Comments

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -4,7 +4,7 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
-- `spotlight` to not focus an element that is invisible
+- `spotlight` to not focus an invisible element
 
 ## [4.5.0-alpha.1] - 2022-04-15
 
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -4,6 +4,8 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
+### Changed
+
 - `spotlight` to not focus an invisible element
 
 ## [4.5.0-alpha.1] - 2022-04-15
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ### Changed
 
-- `spotlight` to not focus an invisible element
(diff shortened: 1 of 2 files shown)


packages/spotlight/src/container.js
```diff
@@ -88,17 +88,14 @@ let GlobalConfig = {
  * @private
  */
 const querySelector = (node, includeSelector, excludeSelector) => {
-	const include = Array.prototype.slice.call(node.querySelectorAll(includeSelector));
+	let include = new Set(node.querySelectorAll(includeSelector));
```
0x64, line 91: > const include = Array.prototype.slice.call(node.querySelectorAll(includeSelector));
> const exclude = node.querySelectorAll(excludeSelector);
> return include.filter(el => !Array.prototype.includes.call(exclude, el));

Not sure but what about an approach like the above?

0x64, line 91:
```suggestion
	const include = new Set(node.querySelectorAll(includeSelector));
```

packages/spotlight/src/container.js
```diff
@@ -328,7 +364,7 @@ const getSpottableDescendants = (containerId) => {
 
 	candidates.push(...getOwnedNodes(node, selector));
 
-	return candidates.filter(n => navigableFilter(n, containerId));
+	return candidates.filter(n => isNavigable(n, containerId));
```
0x64, line 367: We need a discussion about function names, especially between the terms 'Spottable' and 'Navigable'.

0x64 (approved):
:lgtm:

packages/spotlight/src/container.js
0x64: We're going to change `navigableFilter` rather than `getSpottableDescendants`.

packages/spotlight/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+- `spotlight` to not focus an invisible element
```
MikyungKim, line 7:
```suggestion
### Changed

- `spotlight` to not focus on an invisible element
```

MikyungKim (approved):
:lgtm:

PR #3054, opened 2022-05-02
WRO-2712: Use fakeTimers for unit test

sjro:
### Checklist

In some cases, when the code uses timers (setTimeout, setInterval, clearTimeout, clearInterval), the tests may become unpredictable, slow and flaky.
In fact, an unpredictable error occurs when performing unit tests on travis of this repository.
To solve this problem, most testing frameworks provide an option to replace the real timer in the test with a fake timer.

### Resolution
Use fake timers in all unit test cases that require timers.
(https://testing-library.com/docs/using-fake-timers)

### Additional Considerations

### Links
WRO-2712

### Comments

### DIFF
--- a/packages/ui/Touchable/tests/Touchable-specs.js
+++ b/packages/ui/Touchable/tests/Touchable-specs.js
@@ -4,6 +4,13 @@ import {configure, getConfig, resetDefaultConfig} from '../config';
 import Touchable from '../Touchable';
 
 describe('Touchable', () => {
+	beforeEach(() => {
+		jest.useFakeTimers();
+	});
+	afterEach(() => {
+		jest.useRealTimers();
+	});
+
 	let data;
 
 	const DivComponent = ({children = 'Toggle', id, onClick, onMouseDown, onMouseLeave, onMouseUp, onTouchStart, onTouchEnd, ...props}) => {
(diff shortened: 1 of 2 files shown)


juwonjeong (approved):
LGTM

packages/ui/Touchable/tests/Touchable-specs.js
```diff
@@ -55,10 +62,10 @@ describe('Touchable', () => {
 			fireEvent.mouseDown(component, ev);
 			rerender(<Component holdConfig={holdConfig} onHold={handler} onHoldStart={() => {}} />);
 
-			setTimeout(() => {
-				expect(handler).toHaveBeenCalled();
-				done();
-			}, 20);
+			jest.runOnlyPendingTimers();
```
jeonghee27, line 65: I'm curious about the difference between using `advancetimersbyTime` and using `runonlypendingTimers`.

sjro: As described in the link (https://jestjs.io/docs/timer-mocks), when `advancetimersbyTime(msToRun)` is called, all timers are advanced by msToRun milliseconds. **All pending** "macro-tasks" that have been queued via setTimeout() or setInterval(), and would be executed during this time frame, will be executed. 
In the Hold class that is the subject of this unit test, setInterval() is called, so there may be multiple timers and there may be cases where all timers cannot advance at once, so I used `runOnlyPendingTimers()` that only runs pending timers.

PR #3066, opened 2022-06-24
WRO-7190: Fix timing issue when rerende VL by resizing

jeonghee27:
### Checklist

Item list is disappear after the `VirtualList size is changed".

### Resolution
This problem is because the virtualList's `firstIndex` state is not updated properly.

Complex to explain, this is a timing issue.
After resizing VL, VL tried to sets a new `firstIndex`.
The problem is that immediately after this, another rerender is occured by the App.

In between, the `didScroll` event handler is called.
In this `didScroll`, It `setState` the previous `firstIndex`, and as a result, the `firstIndex` state is not updated properly.

Even if rerender occurs by the app, the state value should be updated by VL normally.
Due to the asynchronous handling of `useEffect` this was not guaranteed.
**In order to prevent of invoking `didScroll` in the middle, it should be handled synchronously by using `useLayoutEffect` instead of `useEffect`.**

### Additional Considerations

### Links
WRO-7190

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Scroller` and `ui/VirtualList` to rerender property when `clientSize` is changed
+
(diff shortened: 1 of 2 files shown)


changgilee (approved):
LGTM

PR #2982, opened 2021-12-08
WRN-848: Updated to use `forwardCustom` and add `type` when forwarding custom events

MikyungKim:
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [x] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [x] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
We've added `forwardCustom` in https://github.com/enactjs/enact/pull/2773.
And some of the custom events have no `type` when firing.

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
Fixed to use `forwardCustom` and add `type` when forwarding custom events.

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)
WRN-848

### Comments
Enact-DCO-1.0-Signed-off-by: Mikyung Kim (mikyung27.kim@lge.com)

PR #3067, opened 2022-06-24
WRO-6848: Provide `stopPropagation` and `preventDefault` for custom events

0x64:
### Checklist

`focusCustom` method of `handle` module does not include `stopPropagation` in the event payload that it dispatches, so there is no way to stop propagation of the event in Enact apps when they use `focusCustom`.

### Resolution
This PR changes `forwardCustom` to include `stopPropagation` and `preventDefault` of the original event in a forwarded event payload.

### Additional Considerations
The original issue is only about `stopPropagation` but `preventDefault` is also commonly called by apps. So two methods are supported.
DOM `Event` interface has [two more methods](https://developer.mozilla.org/en-US/docs/Web/API/Event#methods) -`composedPath` and `stopImmediatePropagation`- but they are not a member of React synthetic event interface and actually not so meaningful in React apps. So those methods are not included.

### Links
WRO-6848
enactjs/enact#2982
enactjs/sandstone#1120
enactjs/sandstone#1136

### Comments

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -5,6 +5,11 @@ The following is a curated list of changes in the Enact core module, newest chan
 ## [4.5.0-rc.1] - 2022-06-23
 
 No significant changes.
+## [unreleased]
+
+### Changed
+
+- `core/handle.forwardCustom` handler to include `preventDefault` and `stopPropagation` methods in the forwarded event payload
 
 ## [4.5.0-beta.1] - 2022-05-31
 
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,14 +2,15 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
-## [4.5.0-rc.1] - 2022-06-23
-
-No significant changes.
 ## [unreleased]
 
 ### Changed
 
 - `core/handle.forwardCustom` handler to include `preventDefault` and `stopPropagation` methods in the forwarded event payload
+
+## [4.5.0-rc.1] - 2022-06-23
+
+No significant changes.
 
 ## [4.5.0-beta.1] - 2022-05-31
 
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -436,7 +436,7 @@ const preventDefault = handle.preventDefault = callOnEvent('preventDefault');
  * @param    {Object}    ev     Event payload
  * @param    {Object}    props  Props object
  *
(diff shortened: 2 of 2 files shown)


packages/core/handle/handle.js
```diff
@@ -732,22 +732,105 @@ const adaptEvent = handle.adaptEvent = curry(function (adapter, handler) {
...
+ * ```
+ *
+ * @method   forwardCustomWithPrevent
+ * @param    {String}        name      Name of method on the `props`
+ * @param    {EventAdapter}  [adapter] Function to adapt the event payload
+ *
+ * @returns  {HandlerFunction}         Returns an [event handler]{@link core/handle.EventHandler}
+ *                                     (suitable for passing to handle or used directly within
+ *                                     `handlers` in [kind]{@link core/kind}) that will forward the
+ *                                     custom event and will return `false` if default action is prevented
```
jeonghee27, line 794: Since the return type is not bool, I am not sure if this part is correct.
It is also private api, and i can see what it means. so there is no better idea to me.
Please check it when merging.

jeonghee27 (approved):
LGTM

packages/core/CHANGELOG.md
```diff
@@ -5,6 +5,11 @@ The following is a curated list of changes in the Enact core module, newest chan
 ## [4.5.0-rc.1] - 2022-06-23
 
 No significant changes.
+## [unreleased]
```
MikyungKim, line 8: Please make this right.

packages/core/handle/handle.js
```diff
@@ -732,22 +732,105 @@ const adaptEvent = handle.adaptEvent = curry(function (adapter, handler) {
...
+ * @memberof core/handle
+ * @private
+ */
+const forwardCustomWithPrevent = handle.forwardCustomWithPrevent = named((name, adapter) => {
+	let prevented = false;
+
+	const adapterWithPrevent = (ev, ...args) => {
+		let customEventPayload = adapter ? adapter(ev, ...args) : null;
+		let preventDefaultFromAdapter = null;
```
MikyungKim, line 803: Please rename this.

```suggestion
		let existingPreventDefault = null;
```

packages/core/handle/handle.js
MikyungKim: I agree. But I think this part makes sense since the subject is "an event handler".
Let's fix line 760 part.

0x64: Done!

MikyungKim (approved):
LGTM

PR #3070, opened 2022-07-01
WRO-4003: check parent's container restrict condition when change activate container

jeonghee27:
### Checklist

Focus can leave the restrict container after when returning from another app.
(Element outside the popup receives focus.)

### Resolution
Cause: 
If the "lastContainer" is `popup container`, focus cannot escape because `popup container` is restrict container. ('self-only')
When button inside popup get focus, spotlight set `lastContainer` to `popup content container`.  `popup content container` is not restrict container.

When pointer leave the popup, spotlight set `lastContainer` to `popup container`.  In that case, there is no problem.
In the reproducible case, the pointer leave event did not occur and another app was opened and closed.
Spotlight try to set `container1`(out of popup) as the new activeContainer because user clicked on the center.
If `lastContainer` is a restrict container, Spotlight does not change activeConatienr to `container1`.
But since `lastContainer` is `popup content container`, activeConatiener is changed to `container1` and the problem is reproduced.

The problem was caused by checking only the `lastContainer` when checking `mayActivateContainer`.

Resolution:
Check parent's container restrict condition when change activate container

### Additional Considerations

### Links
WRO-4003

### Comments

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -955,7 +955,7 @@ function mayActivateContainer (containerId) {
 	// activated is not inside the restrict container, the next container should not be activated.
 	const currentContainerNode = getContainerNode(currentContainerId);
 	const restrictContainer = getContainersForNode(currentContainerNode).reduceRight((result, outerContainerId) => {
-		return result || (isRestrictedContainer(outerContainerId) ? outerContainerId : null) ;
+		return result || (isRestrictedContainer(outerContainerId) ? outerContainerId : null);
 	}, null);
 
 	return !restrictContainer || containsContainer(restrictContainer, containerId);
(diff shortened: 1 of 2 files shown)


juwonjeong (approved):
LGTM

packages/spotlight/src/container.js
```diff
@@ -951,13 +951,14 @@ function containsContainer (outerContainerId, innerContainerId) {
...
 	const currentContainerId = getLastContainer();
 
-	// If the current container is restricted to 'self-only' and if the next container to be
-	// activated is not inside the currently activated container, the next container should not be
-	// activated.
-	return (
-		!isRestrictedContainer(currentContainerId) ||
-		containsContainer(currentContainerId, containerId)
-	);
+	// If the current container or Its outer containers are restricted to 'self-only' and if the next container to be
```
0x64, line 954:
```suggestion
	// If the current container or its outer containers are restricted to 'self-only' and if the next container to be
```

0x64: The line became longer, so what about refining line breaks?

jeonghee27: I fixed. Thanks

0x64 (approved):

PR #3072, opened 2022-07-12
WRO-4765: Fixed `i18n` resources loader to override strings where the original strings file does not exist

MikyungKim:
### Checklist

When we added the support for overriding strings with additional resources path, we assumed the original strings are always given to the same locale folder. But some apps need to override strings when the original strings file does not exist(e.g. when the keys and the values pair are all the same, the original app doesn't have ko resources but override strings are in ko folder ).

### Resolution
Changed to read the resources file from the additional resource's path even if the original file does not exist.
Added the read json object to the cache array.

### Additional Considerations

### Links
WRO-4765

### Comments


gouniLee (approved):
Confirmed. it works as expected.
note) the cli version has to match `5.0.0-rc.1`

seunghoh (approved):

PR #3080, opened 2022-08-04
WRO-9768: Fix `Marquee` to stop at the proper position when scaled

jeonghee27:
### Checklist

After one cycle of Marquee, the text stops over the start line and comes back incase of scaled Item.

### Resolution
Apply the scaled value to the spacing calculation so that it stops at the correct point.

### Additional Considerations

When an item of size 100 is scaled by 120%,
The marquee component is well calculated the translateX as "120+spacing size" (=distance).
However, in browsers, it translate by 120px * 120%, eventually causing the marquee to pass more than expected.

We can consider the scale value when cacluate the distance value.
In this case, you need to apply the scale value in the `measureWidths` function of `MarqueeDecorator`.
(The code in `MarqueeBase` still needs modifications.)
However, in order to make the modifications as safe as possible,  I have determined that this  `MarqueeDecorator` modification is not essential at this time.

### Links
WRO-9768

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/Marquee` to stop at the starting point after one cycle when scale is applied
+- `ui/Marquee` to stop at the starting point after one cycle when scaled
 
 ## [4.5.1] - 2022-08-03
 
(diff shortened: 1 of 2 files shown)


0x64 (approved):

packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Marquee.MarqueeBase` to stop at the starting point after one cycle when scale is applied
```
```suggestion
- `ui/Marquee` to stop at the starting point after one cycle when a scale is applied
```

MikyungKim (approved):
LGTM

PR #3096, opened 2022-10-17
WRO-13025: Set default target for Event listeners properly when snapshot build

jeonghee27:
### Checklist

ContextualPopup is not dismissed when click ouside aria in Enact browser with Agate

### Resolution
Cause :
`onClick` event listeners cannot get event, because the `defaultTarget` is `false` when snapshot build.

Resoution:
setDefaultTarget again when defaultTarget is falsy.

### Additional Considerations
There could be other approaches that we fix Agate. 
But this is a more general and safe way to cover all timing and themes.

### Links
WRO-13025

### Comments

### DIFF
--- a/packages/core/dispatcher/dispatcher.js
+++ b/packages/core/dispatcher/dispatcher.js
@@ -36,7 +36,11 @@ const setDefaultTargetById = (id) => {
  * @memberof core/dispatcher
  * @private
  */
-const getDefaultTarget = () => defaultTarget;
+const getDefaultTarget = () => {
+	console.log("getDefaultTarget defaultTarget= ");
+	console.log(defaultTarget);
+	return defaultTarget || (typeof document === 'object' && document);
+};
 
 /*
  * Wraps event callbacks with a try-catch block to prevent unrelated code from blocking.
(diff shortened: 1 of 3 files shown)


packages/core/dispatcher/dispatcher.js
```diff
@@ -36,7 +38,12 @@ const setDefaultTargetById = (id) => {
  * @memberof core/dispatcher
  * @private
  */
-const getDefaultTarget = () => defaultTarget;
+const getDefaultTarget = () => {
+	if (!defaultTarget) {
+		setDefaultTargetById(rootId);
+	}
+	return defaultTarget;
```
MikyungKim, line 45:
```suggestion

	return defaultTarget;
```

packages/core/dispatcher/dispatcher.js
```diff
@@ -36,7 +38,12 @@ const setDefaultTargetById = (id) => {
  * @memberof core/dispatcher
  * @private
  */
-const getDefaultTarget = () => defaultTarget;
+const getDefaultTarget = () => {
+	if (!defaultTarget) {
+		setDefaultTargetById(rootId);
```
MikyungKim, line 43: Should we add some documentation for this behavior to the above or not?

packages/core/dispatcher/dispatcher.js
jeonghee27: I fixed

packages/core/dispatcher/dispatcher.js
jeonghee27: I fixed

MikyungKim (approved):
LGTM

PR #3098, opened 2022-10-19
WRO-12239: Update unit test for packages/core files

0x64:
### Checklist

Unit tests could be added to packages/core files for quality management.

### Resolution
Added unit tests to cover more cases.

### Additional Considerations

### Links
WRO-12239

### Comments

### DIFF
--- a/packages/core/platform/tests/platform-specs.js
+++ b/packages/core/platform/tests/platform-specs.js
@@ -1,8 +1,8 @@
-import {parseUserAgent} from '../platform';
+import {parseUserAgent, platform} from '../platform';
 
 describe('platform', () => {
 
-	describe('webOS', () => {
+	describe('parseUserAgent for webOS', () => {
 		// From http://webostv.developer.lge.com/discover/specifications/web-engine/
 		const webOSTVNext = 'Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/87.0.4280.88 Safari/537.36 WebAppManager';
 		const webOSTV6 = 'Mozilla/5.0 (Web0S; Linux/SmartTV) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/79.0.3440.106 Safari/537.36 WebAppManager';
@@ -101,6 +101,48 @@ describe('platform', () => {
 			const expected = {webos: -1, legacy: 2};
 
 			expect(parseUserAgent(webOS2)).toMatchObject(expected);
+		});
+	});
+
+	describe('parseUserAgent for User-Agent Reduction', () => {
+		const testVersion = '113';
+		const uaGenerator = (unifiedPlatform, deviceCompatibility = '', majorVersion = testVersion) => (
+			`Mozilla/5.0 (${unifiedPlatform}) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/${majorVersion}.0.0.0 ${deviceCompatibility} Safari/537.36`
+		);
+		const testCases = [
+			uaGenerator('Macintosh; Intel Mac OS X 10_15_7'),
+			uaGenerator('Windows NT 10.0; Win64; x64'),
+			uaGenerator('X11; Linux x86_64'),
+			uaGenerator('X11; CrOS x86_64 14541.0.0'),
+			uaGenerator('Fuchsia')
+			// uaGenerator('Linux; Android 10; K', 'Mobile') // Currently not supported
+		];
+
+		test(`should return object including chrome ${testVersion}`, () => {
+			for (let i = 0; i < testCases.length; i++) {
+				expect(parseUserAgent(testCases[i])?.chrome?.toString()).toBe(testVersion);
+			}
+		});
+	});
+
+	describe('platform', () => {
+		test('should have truthy value for `node` if window does not exist', () => {
+			function returnsUndefined () {}
+			const windowSpy = jest.spyOn(window, 'window', 'get').mockImplementation(returnsUndefined);
+
+			expect(platform['node']).toBe(true);
+
+			windowSpy.mockRestore();
+		});
+
+		test('should have truthy value for `unknown` in the testing environment', () => {
+			// The first access invokes detecting based on user agent value
+			expect(platform['unknown']).toBe(true);
+			// The second access makes the module to return already detected platform information
+			expect(platform['unknown']).toBe(true);
+		});
+
+		test('should have platform information', () => {
 		});
 	});
 });
--- a/packages/core/platform/tests/platform-specs.js
+++ b/packages/core/platform/tests/platform-specs.js
@@ -126,23 +126,19 @@ describe('platform', () => {
 	});
 
 	describe('platform', () => {
-		test('should have truthy value for `node` if window does not exist', () => {
-			function returnsUndefined () {}
-			const windowSpy = jest.spyOn(window, 'window', 'get').mockImplementation(returnsUndefined);
+		test('should return `true` for `node` if window does not exist', () => {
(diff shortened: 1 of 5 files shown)


packages/core/platform/tests/platform-specs.js
```diff
@@ -103,4 +103,46 @@ describe('platform', () => {
...
+		});
+
+		test('should have truthy value for `unknown` in the testing environment', () => {
+			// The first access invokes detecting based on user agent value
+			expect(platform['unknown']).toBe(true);
+			// The second access makes the module to return already detected platform information
+			expect(platform['unknown']).toBe(true);
+		});
+
+		test('should have platform information', () => {
```
MikyungKim, line 145: Are you intended to leave this empty?

0x64: Oops, no. I will remove this.

packages/core/platform/tests/platform-specs.js
```diff
@@ -103,4 +103,46 @@ describe('platform', () => {
...
+			for (let i = 0; i < testCases.length; i++) {
+				expect(parseUserAgent(testCases[i][0])?.[testCases[i][1]]?.toString()).toBe(testVersion);
+			}
+		});
+	});
+
+	describe('platform', () => {
+		test('should have truthy value for `node` if window does not exist', () => {
+			function returnsUndefined () {}
+			const windowSpy = jest.spyOn(window, 'window', 'get').mockImplementation(returnsUndefined);
```
MikyungKim, line 131: how about
```suggestion
			const windowSpy = jest.spyOn(window, 'window', 'get').mockImplementation(() => {});
```

packages/core/platform/tests/platform-specs.js
```diff
@@ -103,4 +103,46 @@ describe('platform', () => {
...
+
+			expect(platform['node']).toBe(true);
+
+			windowSpy.mockRestore();
+		});
+
+		test('should have truthy value for `unknown` in the testing environment', () => {
```
MikyungKim, line 138: how about 
```suggestion
		test('should return `true` for `unknown` in the testing environment', () => {
```

packages/core/platform/tests/platform-specs.js
```diff
@@ -103,4 +103,46 @@ describe('platform', () => {
...
+
+		test(`should return object including chrome ${testVersion}`, () => {
+			for (let i = 0; i < testCases.length; i++) {
+				expect(parseUserAgent(testCases[i][0])?.[testCases[i][1]]?.toString()).toBe(testVersion);
+			}
+		});
+	});
+
+	describe('platform', () => {
+		test('should have truthy value for `node` if window does not exist', () => {
```
MikyungKim, line 129: same here.

packages/core/util/tests/util-specs.js
```diff
@@ -68,6 +122,46 @@ describe('util', () => {
...
+		const additiveMap = {
+			'class-shared': 'real-class-shared-additive',
+			'class-shared-another': 'real-class-shared-another-additive',
+			'class-additive-only': 'real-class-additive-only'
+		};
+
+		test('should return a base map if an additive map is not given', () => {
+			const expected = mergeClassNameMaps(baseMap);
+
+			expect(expected['class-base-only']).toEqual('real-class-base-only');
```
MikyungKim, line 140: Can we just use toEqual? like expect(expected).toEqual(baseMap).

packages/core/util/tests/util-specs.js
```diff
@@ -68,6 +122,46 @@ describe('util', () => {
...
+			'class-shared-another': 'real-class-shared-another-base'
+		};
+		const additiveMap = {
+			'class-shared': 'real-class-shared-additive',
+			'class-shared-another': 'real-class-shared-another-additive',
+			'class-additive-only': 'real-class-additive-only'
+		};
+
+		test('should return a base map if an additive map is not given', () => {
+			const expected = mergeClassNameMaps(baseMap);
```
MikyungKim, line 138: Shouldn't this be 'actual'?

packages/core/util/tests/util-specs.js
```diff
@@ -68,6 +122,46 @@ describe('util', () => {
...
+
+			expect(expected['class-base-only']).toEqual('real-class-base-only');
+			expect(expected['class-shared']).toEqual('real-class-shared-base');
+			expect(expected['class-shared-another']).toEqual('real-class-shared-another-base');
+			expect(expected['class-additive-only']).not.toEqual('real-class-additive-only');
+		});
+
+		test('should return a merged map containing shared class names', () => {
+			const expected = mergeClassNameMaps(baseMap, additiveMap);
```
MikyungKim, line 147: Also, shouldn't it be 'actual'?

MikyungKim: Can we make an expected object to compare? Please refer to `usePublicClassNames-specs`.

packages/core/util/tests/util-specs.js
0x64: It is possible only for when an additive map is not given since other cases need to be checked with merged result. And, I want to make `expect`s format as same as each other. WDYT?

packages/core/util/tests/util-specs.js
0x64: You're right. 😓

packages/core/util/tests/util-specs.js
0x64: Ah, not I got your point. Good idea.

0x64: I created such a long test code since the below code blocks to change the result.
https://github.com/enactjs/enact/blob/develop/packages/core/util/util.js#L195
Rather than the merged `css` object, it returns a Proxy instance, so the returned value is actually an empty object and but we can access its properties via getters.

0x64: I added a way to avoid proxy instance in unit test code, so finally it compares via `toEqual` between objects.

packages/core/util/tests/util-specs.js
```diff
@@ -68,6 +122,69 @@ describe('util', () => {
...
+				if (keys[i] !== proxy[keys[i]]) {
+					obj[keys[i]] = proxy[keys[i]];
+				}
+			}
+
+			return obj;
+		};
+
+		test('should return a base map if an additive map is not given', () => {
+			const expected = {
```
MikyungKim, line 152: Can't this be just baseMap?

packages/core/util/tests/util-specs.js
```diff
@@ -68,6 +122,69 @@ describe('util', () => {
...
+			'class-shared-another': 'real-class-shared-another-base'
+		};
+		const additiveMap = {
+			'class-shared': 'real-class-shared-additive',
+			'class-shared-another': 'real-class-shared-another-additive',
+			'class-additive-only': 'real-class-additive-only'
+		};
+
+		// Do NOT use destructuring operator as 'pxory' does not have properties but have only getters for properties.
+		const avoidProxy = (proxy) => {
```
MikyungKim, line 138: How about 
```suggestion
		// Helper function to get an object from the proxy object that has only getters for properties to make testing easier
		const getResultFromProxy = (proxy) => {
```

packages/core/util/tests/util-specs.js
0x64: This is intended. The interface requirement is to have the same class info, not the original reference. So, I don't make it to check the reference just to simplify the test.

packages/core/util/tests/util-specs.js
0x64: Look clearer. I will apply it. Thanks for good suggestion.

MikyungKim (approved):
LGTM

PR #3127, opened 2023-03-06
Add isStandardFocusableMode private globalConfig 

juwonjeong:
### Checklist

Spotlight needs to navigate standard focusable element when isStandardFocusableMode is set true.
If focusRingElement exists, when the spotlight focuses a certain element, focusRingElement needs to move.

### Resolution
Implement spotlight logic to determine if the element is standard focusable element.
Add `isStandardFocusMode` config in GlobalConfig(https://github.com/enactjs/enact/blob/master/packages/spotlight/src/container.js#L43).
`isStandardFocusMode` config can be set with spotlight.initialize function.

### Additional Considerations
isStandardFocusMode config is the private config.

### Links
WRP-10235

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -599,7 +599,7 @@ const isNavigable = (node, containerId, verify) => {
 	}
 
 	const config = getContainerConfig(containerId);
-	if (verify && config && config.selector && !isContainer(node) && !(matchSelector(config.selector, node) || (GlobalConfig.isStandardFocusableMode && isStandardFocusable(node)))) {
+	if (verify && config && config.selector && !isContainer(node) && !matchSelector(config.selector, node) && !(GlobalConfig.isStandardFocusableMode && isStandardFocusable(node))) {
 		return false;
 	}
 
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -90,9 +90,8 @@ let GlobalConfig = {
  * @private
  */
 const querySelector = (node, includeSelector, excludeSelector) => {
-	let include = new Set(node.querySelectorAll(includeSelector));
 	const focusables = GlobalConfig.isStandardFocusableMode ? Array.prototype.filter.call(node.getElementsByTagName('*'), isStandardFocusable) : [];
-	include = new Set([...include, ...focusables]);
+	const include = new Set([...node.querySelectorAll(includeSelector), ...focusables]);
 
 	const exclude = node.querySelectorAll(excludeSelector);
 
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -12,7 +12,7 @@ import {coerceArray} from '@enact/core/util';
 import intersection from 'ramda/src/intersection';
 import last from 'ramda/src/last';
 
-import {contains, matchSelector, getContainerRect, getRect, intersects} from './utils';
+import {contains, matchSelector, getContainerRect, getRect, intersects, isStandardFocusable} from './utils';
 
 const containerAttribute = 'data-spotlight-id';
 const containerConfigs   = new Map();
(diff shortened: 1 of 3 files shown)


packages/spotlight/src/container.js
```diff
@@ -89,7 +90,10 @@ let GlobalConfig = {
  * @private
  */
 const querySelector = (node, includeSelector, excludeSelector) => {
-	const include = new Set(node.querySelectorAll(includeSelector));
+	let include = new Set(node.querySelectorAll(includeSelector));
```
0x64, line 93: Let's move `querySelectorAll` call into two lines below as we a new `Set` here is not used actually.

packages/spotlight/src/utils.js
```diff
@@ -172,13 +172,47 @@ function getContainerRect (containerId) {
 	return getRect(containerNode);
 }
 
+function isStandardFocusable (element) {
+	if ((element.tabIndex < 0) || isAtagWithoutHref(element) || isActuallyDisabled(element) || isExpresslyInert(element) || isElementHidden(element)) {
```
0x64, line 176: For later visits, could you please add some comments for clauses in this function?

packages/spotlight/src/utils.js
```diff
@@ -172,13 +172,47 @@ function getContainerRect (containerId) {
...
+		return false;
+	} else if ((!element.parentElement) || (element.tabIndex >= 0)) {
+		return true;
+	}
+}
```
0x64, line 181: Please add `return false;` explicitly.

packages/spotlight/src/utils.js
```diff
@@ -172,13 +172,47 @@ function getContainerRect (containerId) {
 	return getRect(containerNode);
 }
 
+function isStandardFocusable (element) {
+	if ((element.tabIndex < 0) || isAtagWithoutHref(element) || isActuallyDisabled(element) || isExpresslyInert(element) || isElementHidden(element)) {
+		return false;
+	} else if ((!element.parentElement) || (element.tabIndex >= 0)) {
+		return true;
+	}
```
0x64, line 180: Please add `else` case to return falsy value for rest conditions.

packages/spotlight/src/utils.js
```diff
@@ -172,13 +172,47 @@ function getContainerRect (containerId) {
...
+}
+
+function isAtagWithoutHref (element) {
+	return (element.tagName === 'A' && element.getAttribute('href') === null && element.getAttribute('tabIndex') === null);
+}
+
+function isActuallyDisabled (element) {
+	if (['BUTTON', 'INPUT', 'SELECT', 'TEXTAREA', 'OPTGROUP', 'OPTION', 'FIELDSET'].includes(element.tagName)) {
+		return (element.disabled);
```
0x64, line 189: Not an issue but just curious that why do you use parenthesis here? :>

packages/spotlight/src/utils.js
```diff
@@ -172,13 +172,47 @@ function getContainerRect (containerId) {
...
+		return false;
+	}
+}
+
+function isExpresslyInert (element) {
+	return ((element.inert) && (!element.ownerDocument.documentElement.inert));
+}
+
+function isElementHidden (element) {
+	let elemRect = element.getBoundingClientRect();
```
0x64, line 200:
```suggestion
	const elemRect = element.getBoundingClientRect();
```

packages/spotlight/src/utils.js
juwonjeong: I think it is unnecessary. The related code has been updated.

packages/spotlight/src/utils.js
juwonjeong: I added else case. Thank you.

packages/spotlight/src/utils.js
juwonjeong: I added some comments

packages/spotlight/src/container.js
juwonjeong: It's fixed.

0x64 (approved):

PR #3128, opened 2023-03-07
WRP-6257: Cell: Add `grow` prop to expand the size to its container

hoeunryuLG:
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [x] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [x] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
A `grow`able cell will expand to its maximum size according to the remaining space of the container. This is used when you want to grow the size of this cell so that it fills the container.

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
add `grow` prop to `Cell` component.

A `grow`able Cell has a declaration of `flex: 1 1 100%` in CSS by default. However, the `flex-basis` property is overridden by the `var(--cell-size)` value passed from JS. And by combining `max-width: none` and `flex-grow: 1` declaration, its size can be increased by the size of its parent container.

Also, by the declaration of `flex: 1 1 100%`, the `flex-shrink` property has a value of `1`, which means that its size can be reduced as much as necessary, but because of `min-width: var(--cell-size)` and `min-height: var(--cell-size)` declaration, it does not decrease below the specified cell size.

When combined with `shrink`, `shrink` prop takes precedence over `grow` prop and `grow` prop is simply ignored. When used in conjunction with `size`, the size will be the minimum size, growing as necessary, to fit the container.

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)
Screenshot tests are added to Sandstone (enactjs/sandstone#1417)

Unit Test Cases
- `grow` CSS class is applied when no props (`shrink`, `size`) are  given
- `grow` CSS class and `flex-basis` CSS style are applied when `grow` and `size` props are given
- `grow` CSS class is not applied when used with `shrink` prop

Sampler
- The thrid Cell is modified to change the `grow` and `size` props.The part where `grow` is automatically activated is still checkable when `shrink` prop is disabled in the second Cell.
- Clarify the names of each Cell and control to make them easier to distinguish.

### Links
[//]: # (Related issues, references)
WRP-6257
enactjs/sandstone#1417

### Comments
[...]

PR #3138, opened 2023-03-21
WRP-6257: Cell: Add grow prop to expand the size to its container

hoeunryuLG:
### Checklist

A `grow`able cell will expand to its maximum size according to the remaining space of the container. This is used when you want to grow the size of this cell so that it fills the container.

This PR is reopened (#3128) to fix a screenshot test failing on Item component in sandstone.

### Resolution
add `grow` prop to `Cell` component.

A `grow`able Cell has a declaration of `flex: 1 1 100%` in CSS by default. However, the `flex-basis` property is overridden by the `var(--cell-size)` value passed from JS. And by combining `max-width: none` and `flex-grow: 1` declaration, its size can be increased by the size of its parent container.

Also, by the declaration of `flex: 1 1 100%`, the `flex-shrink` property has a value of `1`, which means that its size can be reduced as much as necessary, but because of `min-width: var(--cell-size)` and `min-height: var(--cell-size)` declaration, it does not decrease below the specified cell size.

`size` CSS class is introduced to distinguish between implict `grow` (`!shrink && !size`) and explicit `grow` used with `size` prop. Explicit `grow` is always used with `size` prop and the `size` CSS class is used for two purposes:
- To invalidate `max-width/height: var(--cell-size)` declaration when `size` prop is used to `max-width/height: none`. It is needed to increase its size by the size of its parent container.
- To declare `min-width/height: var(--cell-size)`. It is needed for its size not to decrease below the specified size.

When combined with `shrink`, `shrink` prop takes precedence over `grow` prop and `grow` prop is simply ignored. When used in conjunction with `size`, the size will be the minimum size, growing as necessary, to fit the container.

### Additional Considerations
Screenshot tests are added to Sandstone ()
Make sure that this change doesn't affect the screenshot test results of other components.

Unit Test Cases
- `grow` CSS class is applied when no props (`shrink`, `size`) are  given
- `grow` and `size` CSS class and `flex-basis` CSS style are applied when `grow` and `size` props are given
- `grow` CSS class is not applied when used with `shrink` prop

Sampler
- The thrid Cell is modified to change the `grow` and `size` props.The part where `grow` is automatically activated is still checkable when `shrink` prop is disabled in the second Cell.
- Clarify the names of each Cell and control to make them easier to distinguish.

### Links
WRP-6257
reopened #3128
screenshot test enactjs/sandstone#1427

### Comments
Enact-DCO-1.0-Signed-off-by: Hoeun Ryu <hoeun.ryu@lge.com>

### DIFF
--- a/packages/sampler/stories/default/Layout.js
+++ b/packages/sampler/stories/default/Layout.js
@@ -21,19 +21,22 @@ export const _Layout = (args) => (
 			orientation={args['orientation']}
 		>
 			<Cell
-				size={args['cell size'] + 'px'}
+				size={args['first cell size'] + 'px'}
 				shrink
 			>
-				<Item>First</Item>
+				<Item>First&#xa0;(shrink)</Item>
 			</Cell>
-			<Cell shrink={args['shrinkable cell']}>
-				<Item>Second</Item>
+			<Cell shrink={args['shrinkable second cell']}>
+				<Item>Second&#xa0;(shrinkable)</Item>
 			</Cell>
-			<Cell>
-				<Item>Third</Item>
+			<Cell
+				grow={args['growable third cell']}
+				size={args['third cell size'] + 'px'}
+			>
+				<Item>Third&#xa0;(growable)</Item>
 			</Cell>
 			<Cell shrink>
-				<Item>Last</Item>
+				<Item>Last&#xa0;(shrink)</Item>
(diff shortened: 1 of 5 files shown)


MikyungKim (approved):
LGTM

seunghoh (approved):
LGTM

PR #3162, opened 2023-07-14
WRP-21128: Fixed components used in enact/ui/Layout/Cell which cannot have their css customized in consumer apps.

alexandrumorariu:
Enact-DCO-1.0-Signed-off-by: Alexandru Morariu alexandru.morariu@lgepartner.com

### Checklist

While working on Agate-Apps we noticed that components used in enact/ui/Layout/Cell cannot have their css customized. In order to customize it we passed the component as a child to Cell instead of using the component prop. Now we can pass the css directly.

### Resolution
I added the componentCss prop to Cell component in order to increase customisation.

### Additional Considerations

### Links
WRP-21128


packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `ui/Layout.Cell` prop `componentCss` in order to pass classes to the `Cell` component
```
daniel-stoian-lgp, line 9:
```suggestion
- `ui/Layout.Cell` prop `componentCss` in order to customize the component used in `Cell`
```

daniel-stoian-lgp (commented):
see my comments

daniel-stoian-lgp (approved):

packages/ui/Layout/Cell.js
```diff
@@ -59,6 +59,17 @@ const CellBase = kind({
 		 */
 		component:  EnactPropTypes.renderable,
 
+		/**
+		 * Customizes the component used in the `Cell` by mapping the supplied collection of CSS class names to the corresponding internal elements and states of this component.
+		 *
+		 * When using {@link ui/Layout.Cell}, the `css` prop is forwarded to this component
```
0x64, line 65: "this component" makes a big misunderstanding. It can be read like "if a developer set `componentCss` prop to this component, then `css` prop will be forwarded to this component."

0x64 (approved):

PR #3167, opened 2023-08-04
WRP-539: Add docs about `bindAs`

sjro:
### Checklist

There is no description about 'bindAs' of handle.

### Resolution
Add the docs for 'bindAs'

### Additional Considerations

### Links
WRP-539

### Comments
Enact-DCO-1.0-Signed-off-by: SJ RO ([sj.ro@lge.com](mailto:sj.ro@lge.com))

### DIFF
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -28,7 +28,7 @@
  * `props` and `context`. This allows you to write consistent event handlers for components created
  * either with `kind()` or ES6 classes without worrying about from where the props are sourced.
  *
- * Handlers can either be bound directly using the native `bind()` method or using the `bindAs()`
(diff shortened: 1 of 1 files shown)


seunghoh (commented):
Thanks for the PR. I have some comments
1. Would be better to add an explanation just below(or above) Example 1, and Example 2 so that what's your purpose.
2. You shall mention the advantage of using `bindAs`. It makes the function name to be displayed in the development tool call stacks. We had difficulties to debug due to lots of `anonymous` call stacks. Please refer to the original PR - https://github.com/enactjs/enact/pull/1652

By the way, I am not sure if we need to export and make `bindAs` public. I just created one more PR to decide which way we choose. https://github.com/enactjs/enact/pull/3170

seunghoh:
Discussed and decided WNF

PR #3170, opened 2023-08-24
WRP-539: Update handle docs to remove bindAs example

seunghoh:
### Checklist

`bindAs` utility is for internal use only and is not exported but docs example uses.
It could cause confusion for readers.

### Resolution
Removed `bindAs` usage from the example and added some inline comments for internal developers about the handle.bindAs purpose.

### Additional Considerations
I still don't know if we need to export `bindAs` function to the public. This function is definitely useful.
If we decide to make it public, we shall add docs, example.

### Links
WRP-539
https://github.com/enactjs/enact/pull/1652

### Comments

Enact-DCO-1.0-Signed-off-by: Seungho Park <seunghoh.park@lge.com>

### DIFF
--- a/packages/core/handle/handle.js
+++ b/packages/core/handle/handle.js
@@ -28,8 +28,7 @@
  * `props` and `context`. This allows you to write consistent event handlers for components created
  * either with `kind()` or ES6 classes without worrying about from where the props are sourced.
  *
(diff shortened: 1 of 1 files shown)


seunghoh:
Discussed and decided WNF

PR #3183, opened 2023-11-14
WRQ-2421: Fix MarqueeDecorator to re-render marquee when its size changed

juwonjeong:
### Checklist

When Marquee size is changed dynamically(for example, changing the screen size or rotating the screen) it is not re-rendered, so the text will not be able to start marquee animation and may be truncated.

### Resolution
Added `ResizeObserver` to make Marquee re-render when its size is changed. 

### Additional Considerations

### Links
WRQ-2421


seunghoh (approved):
LGTM
I've added unit test for ResizeObserver

PR #3193, opened 2023-12-11
WRQ-171: Added internal scrolling option to disallow negative offset

0x64:
### Checklist

When a theme library call `scrollTo` with `optionalOffset` for internal uses, sometimes the finally applied offset could be negative value but it is not desired behavior.

### Resolution
Simply added an option to avoid negative offset.

### Additional Considerations
No changelog as only internal code is changed without impact for developers.

### Links
WRQ-171

### Comments


juwonjeong (approved):
LGTM

PR #3208, opened 2024-02-06
WRQ-188: Update core/platform and webos/platform

0x64:
### Checklist

Some platforms detected by `core/platform` and `webos/platform` are outdated.
* There are platforms over ten years from their end-of-life.
* Some platforms are not rigorously tested several years.
* Thanks to interop activities of browser vendors, app developers need just a major compatible browser name and version like Chrome, 120.

Some APIs look unclear in a viewpoint of modules' roles and responsibilities.
* core/platform provides 'touch' and 'touchscreen', and it could lead misunderstanding or a human error.
* core/platform's role is to support app devs to cover compatibility issues, so detecting 'webos' is a different kind of feature. It looks better to provide 'webos' info in webos/platform since apps using webOS features should import @enact/webos anyway.

### Resolution
core/platform
* Outdated platforms will be deprecated in Enact 5.0.0
* 'type' is added that is one of 'desktop', 'mobile', 'webos', 'node', and 'unknown'
* 'browserName' is added that is of 'chrome', 'safari', or 'firefox' (Note that the value of this new property is different from the existing 'platformName' that will be deprecated in Enact 5.0.0)
* 'browserVersion' is added as a number (Note that its value is a float for 'safari' so that minor numbers can be compared for compatibility issues)
* 'touch' is replaced with 'touchEvent' as its actual meaning
* 'touchscreen' is replaced with 'touchScreen' to match camel case (Note that we're trying to detect touch screen devices, but it can't be perfect.)

core/webos
* 'webos' is added to check if the platform is webOS
* 'version' is added to check webOS version (from 3.x to 23)
* 'chrome' is added to check chrome version (just for detected webOS vrersions; recommend to use core/platform in general purpose)

### Additional Considerations

### Links
WRQ-188

### Comments


packages/core/platform/platform.js
```diff
@@ -7,38 +8,41 @@
...
-// Adapted from https://patrickhlauke.github.io/touch/touchscreen-detection/
-// I've omitted the touch event fallback since that is covered by hasTouch and we're less concerned
-// with legacy browsers used in touchscreen environments.
-const hasTouchScreen = () => {
-	return (
-		// if Pointer Events are supported, just check maxTouchPoints
-		(window.PointerEvent && ('maxTouchPoints' in window.navigator) && window.navigator.maxTouchPoints > 0) ||
-		// check for any-pointer:coarse which mostly means touchscreen
-		(window.matchMedia && window.matchMedia('(any-pointer:coarse)').matches)
+		featureTouchEvent() ||
```
juwonjeong, line 42: Do we need to update `hasTouch` logic which will be deprecated with featureTouchEvent()?

.travis.yml
```diff
@@ -1,8 +1,8 @@
 dist: focal
 language: node_js
 node_js:
+    - node
```
juwonjeong, line 4: As I checked, it is different from sandstone .travis.yml. Will the sandstone repo also be updated?

packages/core/platform/platform.js
```diff
@@ -152,29 +150,102 @@ const parseUserAgent = (userAgent) => {
...
+ * @property {Boolean} node - `true` only if `window` is `undefined`. Deprecated: will be removed in 5.0.0. Use `type` instead.
+ * @property {String} [platformName] - The name of the platform, if detected. Deprecated: will be removed in 5.0.0. Use `browserName` instead for browser names.
+ * @property {Number} safari - The version of the detected browser, if safari browser is detected
+ * @property {Boolean} touch - `true` if the platform has native single-finger events. Deprecated: will be removed in 5.0.0. Use `touchEvent` instead.
+ * @property {Boolean} touchEvent - `true` if the browser has native touch events
+ * @property {Boolean} touchscreen - `true` if the platform has a touch screen. Deprecated: will be removed in 5.0.0. Use `touchScreen` instead.
+ * @property {Boolean} touchScreen - `true` if the platform has a touch screen
+ * @property {String} type - The type of the detected platform. One of 'desktop', 'mobile', 'webos', 'node', or 'unknown'
+ * @property {Boolean} unknown - `true` for any unknown system. Deprecated: will be removed in 5.0.0. Use `type` instead.
```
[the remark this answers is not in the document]
0x64: I also considered that ordering, but we normally explain props in alphabetical order. 🤔

packages/core/platform/platform.js
0x64: `hasTouch` will be removed on the next major version update, but `featureTouchEvent` will remain.

.travis.yml
0x64: That's my intention. Right. (If no issue occurs due to this change.)

0x64: BTW, I want to change the order in this repo. I will update this.

juwonjeong (commented):
LGTM

packages/core/platform/platform.js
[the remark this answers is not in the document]
juwonjeong: @seunghoh, I think it would make more sense to put deprecated properties at the bottom to let developers know which properties to use. But since we normally place the props in alphabetical order, I'll leave it as is.
CC. @0x64

seunghoh:
Let's add the Changelog

packages/core/platform/platform.js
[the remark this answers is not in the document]
seunghoh: @0x64 Sorry, I changed my mind :( 
Let's move deprecated props to the bottom for enactjs.com API  page

packages/webos/platform/platform.js
```diff
@@ -71,11 +122,14 @@ function detect () {
...
+ * @property {Boolean} webos `true` for webOS
  * @property {Boolean} [tv] `true` for LG webOS SmartTV
- * @property {Boolean} [watch] `true` for LG webOS SmartWatch
+ * @property {Boolean} [watch] `true` for LG webOS SmartWatch. Deprecated: will be removed in 5.0.0.
  * @property {Boolean} [open] `true` for Open webOS
- * @property {Boolean} [legacy] `true` for legacy webOS (Palm and HP hardware)
+ * @property {Boolean} [legacy] `true` for legacy webOS (Palm and HP hardware). Deprecated: will be removed in 5.0.0.
  * @property {Boolean} [unknown] `true` for any unknown system
+ * @property {Number}  [version] The version of the platform if detected
+ * @property {Number}  [chrome] The version of Chrome if detected
```
seunghoh, line 132: @0x64  here as well, Let's move deprecated to the bottom

packages/core/platform/platform.js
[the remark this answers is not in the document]
0x64: @seunghoh @juwonjeong As far as I know, the order of comments is not the same as the order of documents. So, I cannot control it.

packages/webos/platform/platform.js
0x64: It doesn't work. I explained at the first comment.

seunghoh (approved):
LGTM

PR #3210, opened 2024-02-20
WRQ-6220: Fixed `spotlight` navigation from the focused element clipped by an overflow container

0x64:
### Checklist

Spotlight cannot choose the next focus target if the currently focused node is partially visible as clipped by an overflow container. Note that a clipped node within an overflow container can be a navigable target.

### Resolution
Updated a navigation logic to count only a visible part if the currently focused node is partially visible and its closest container is an overflow container.

### Additional Considerations

### Links
WRQ-6220

### Comments

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `spotlight` navigation from the focused element clipped by an overflow container
+
 ## [4.8.0] - 2024-02-08
 
 No significant changes.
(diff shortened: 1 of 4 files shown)


packages/spotlight/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact spotlight module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `spotlight` navigation from the focused element clipped by an overflow container
```
juwonjeong, line 9: How about
```suggestion
- `spotlight` to find a navigable element when the focused element clipped by an overflow container
```
Or, I think it would be okay to leave it as is.

juwonjeong (approved):
LGTM

PR #3215, opened 2024-02-26
WRQ-6229: Add config `scrollTargetOnDescendantsFocus` to container

juwonjeong:

### Checklist

The container needs additional config to define whether the container will be the scroll target when its child items get focused. 

### Resolution
Add private container config scrollTargetOnDescendantsFocus.
Add private method `getScrollTargetOnDescendantsFocus` to find the scroll target container when the element is focused.

### Additional Considerations

### Links
WRQ-6299

### DIFF
--- a/packages/spotlight/src/container.js
+++ b/packages/spotlight/src/container.js
@@ -70,6 +70,7 @@ let GlobalConfig = {
 	partition: false, // use the container bounds for partitioning when leaving
 	rememberSource: false,
 	restrict: 'self-first', // 'self-first', 'self-only', 'none'
+	scrollTargetOnDescendantsFocus: false,    // @private - use the container for the scroll target when its descendants get focused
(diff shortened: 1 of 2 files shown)


packages/spotlight/src/container.js
```diff
@@ -1101,6 +1102,22 @@ function notifyEnterContainer (direction, previous, previousContainerIds, curren
 	});
 }
 
+/**
+ * Returns a closest container that wrap the element and has positionTargetOnDescendantsFocus configured
```
0x64, line 1106: It looks that this function finds the furthest container, not the closest container. And, in my opinion, the current behavior makes sense. Let's discuss about it. @juwonjeong

juwonjeong: I fixed to find the closest container. Thank you

0x64 (approved):

PR #3217, opened 2024-03-04
WRQ-11247: Added fontScale prop for large text mode scale values

mmyelyn:
### Checklist

There is a requirement for large text mode for a11y.
I have modified the app (including sampler) to review the screen when large text is applied.

### Resolution
Modified to update the base font size when the fontScale prop changes.

### Additional Considerations

### Links
WRQ-11247

### Comments
Enact-DCO-1.0-Signed-off-by: Hyelyn Kim ([myelyn.kim@lge.com](mailto:myelyn.kim@lge.com))

### DIFF
--- a/packages/ui/resolution/ResolutionDecorator.js
+++ b/packages/ui/resolution/ResolutionDecorator.js
@@ -9,7 +9,7 @@ import ReactDOM from 'react-dom';
 import PropTypes from 'prop-types';
 import hoc from '@enact/core/hoc';
 
-import {init, config as riConfig, defineScreenTypes, getResolutionClasses} from './resolution';
+import {init, config as riConfig, defineScreenTypes, getResolutionClasses, updateBaseFontSize, calculateFontSize} from './resolution';
 
 /**
  * Default config for `ResolutionDecorator`.
@@ -118,6 +118,12 @@ const ResolutionDecorator = hoc(defaultConfig, (config, Wrapped) => {
 
 		componentWillUnmount () {
 			if (config.dynamic) window.removeEventListener('resize', this.handleResize);
+		}
+
+		componentDidUpdate (prevProps) {
+			if(prevProps.fontScale !== this.props.fontScale) {
+				updateBaseFontSize(calculateFontSize(null, this.props.fontScale));
+			}
 		}
 
 		handleResize = () => {
--- a/packages/ui/resolution/resolution.js
+++ b/packages/ui/resolution/resolution.js
@@ -194,7 +194,7 @@ function getScreenType (rez) {
  * @returns {String}          The calculated pixel size (with unit suffix. Ex: "24px").
  * @public
  */
-function calculateFontSize (type, fontScale) {
+function calculateFontSize (type, fontScale = 1) {
 	const scrObj = getScreenTypeObject(type);
 	const shouldScaleFontSize = (config.intermediateScreenHandling === 'scale') && (config.matchSmallerScreenType ? workspaceBounds.width > scrObj.width && workspaceBounds.height > scrObj.height :
 		workspaceBounds.width < scrObj.width && workspaceBounds.height < scrObj.height);
--- a/packages/ui/resolution/resolution.js
+++ b/packages/ui/resolution/resolution.js
@@ -194,18 +194,18 @@ function getScreenType (rez) {
  * @returns {String}          The calculated pixel size (with unit suffix. Ex: "24px").
  * @public
  */
-function calculateFontSize (type) {
+function calculateFontSize (type, fontScale) {
 	const scrObj = getScreenTypeObject(type);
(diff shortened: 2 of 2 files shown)


packages/ui/resolution/ResolutionDecorator.js
```diff
@@ -97,7 +97,8 @@ const ResolutionDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		static displayName = 'ResolutionDecorator';
 
 		static propTypes = /** @lends ui/resolution.ResolutionDecorator.prototype */ {
-			className: PropTypes.string
+			className: PropTypes.string,
+			fontScale: PropTypes.number
```
juwonjeong, line 101: Please add detailed docs and also set `fontScale` default value to be 1

packages/ui/resolution/ResolutionDecorator.js
```diff
@@ -116,6 +117,12 @@ const ResolutionDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			this.rootNode = ReactDOM.findDOMNode(this);
 		}
 
+		componentDidUpdate (prevProps) {
+			if (prevProps.fontScale !== this.props.fontScale) {
+				updateBaseFontSize(calculateFontSize(null, this.props.fontScale));
```
juwonjeong, line 122: Is it ok to pass null in calculateFontSize?

mmyelyn: Yes it is.
If the type value is falsy, the current screenType is used by the getScreenTypeObject function.
In fact, when calling this function in init, type is not passed, so it is passed as undefined.

packages/ui/resolution/ResolutionDecorator.js
mmyelyn: Added.

juwonjeong:
@mmyelyn, font scale does not work with http://SERVERIP:8080/?path=/story/about-a-tour-of-sampler--a-tour-of-sampler&globals=font+scale:1.6. Could you please check it?

packages/ui/resolution/ResolutionDecorator.js
```diff
@@ -28,6 +28,18 @@ const defaultConfig = {
 	 */
 	dynamic: true,
 
+	/**
```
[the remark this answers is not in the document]
mmyelyn: Sorry. I confused defaultProps and defaultConfig. Updated to defaultProps from defaultConfig

mmyelyn:
> @mmyelyn, font scale does not work with http://SERVERIP:8080/?path=/story/about-a-tour-of-sampler--a-tour-of-sampler&globals=font+scale:1.6. Could you please check it?

This is because fontScale was not passed when updating the base font in init. In the process of fixing, i modified fontScale to be managed as an internal resolution variable. So, the fontScale parameter that was passed to the calculate function has been deleted.

juwonjeong:
I think it is better to apply 4b93d18861378ed94dfeb9a6febd4c9be9fb8f3f commit to not change the original init function, but I need the merger's opinion.

packages/ui/resolution/ResolutionDecorator.js
```diff
@@ -114,6 +129,15 @@ const ResolutionDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			if (config.dynamic) window.addEventListener('resize', this.handleResize);
 			// eslint-disable-next-line react/no-find-dom-node
 			this.rootNode = ReactDOM.findDOMNode(this);
+
+			updateFontScale(this.props.fontScale);
```
juwonjeong, line 133:
```suggestion
```
I think we should move `updateFontScale` into the constructor before init function call

packages/ui/resolution/ResolutionDecorator.js
```diff
@@ -97,7 +97,22 @@ const ResolutionDecorator = hoc(defaultConfig, (config, Wrapped) => {
 		static displayName = 'ResolutionDecorator';
 
 		static propTypes = /** @lends ui/resolution.ResolutionDecorator.prototype */ {
-			className: PropTypes.string
+			className: PropTypes.string,
+
+			/**
+			 * Font Scale for large text mode.
+	 		 * Use this value to set the scale of the font.
+	 		 * This is the value that will be multiplied by pxPerRem, which is determined by resolution.
```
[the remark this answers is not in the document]
mmyelyn: I used it because the currently used term is large text mode. It may be confused with large text in ThemeDecorator, but since there is currently no fixed name, it would be a good idea to change it after the name is fixed. This is a draft and will not be merged.

juwonjeong (approved):
LGTM

packages/ui/resolution/ResolutionDecorator.js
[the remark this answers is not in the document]
juwonjeong: I agree. I think we need to set a proper name for this mode when merging it. I will approve it since this is a draft PR.

PR #3239, opened 2024-05-21
WRQ-20844: Fixed React 18.3 warnings in enact/ui/Layout

daniel-stoian-lgp:
### Checklist

In React 18.3, There are warnings for key spreading. The warning is reproducible when using agate/LabeledIconButton

### Resolution
Passed key attribute directly to JSX without using spread in enact/ui/Layout/Cell

### Additional Considerations

### Links
WRQ-19328

### Comments


packages/ui/Layout/Cell.js
```diff
@@ -96,6 +96,8 @@ const CellBase = kind({
 		 */
 		grow: PropTypes.bool,
 
+		key: PropTypes.string,
```
daniel-stoian-lgp, line 99: here I need help to properly describe this prop

adrian-cocoara-lgp: My suggestion is:
`
/**
	* A unique key used by React to identify the component.
        * @type {String}
        * @public
	 */
`
but let's see SWP team's opinion.

adrian-cocoara-lgp (approved):

packages/ui/Layout/Cell.js
```diff
@@ -173,13 +180,13 @@ const CellBase = kind({
 		}
 	},
 
-	render: ({component: Component, componentCss, componentRef, ...rest}) => {
+	render: ({component: Component, componentCss, componentRef, key, ...rest}) => {
```
juwonjeong, line 183:
```suggestion
	render: ({component: Component, componentCss, componentRef, ...rest}) => {
```

packages/ui/Layout/Cell.js
```diff
@@ -173,13 +180,13 @@ const CellBase = kind({
...
 		delete rest.align;
 		delete rest.grow;
 		delete rest.shrink;
 		delete rest.size;
 
-		return <Component css={componentCss} ref={componentRef} {...rest} />;
+		return <Component css={componentCss} ref={componentRef} key={key} {...rest} />;
```
juwonjeong, line 189: I think instead of adding docs of key, we can just destructure rest props.
```suggestion
		const {key, ...restProps} = {...rest};

		return <Component css={componentCss} key={key} ref={componentRef} {...restProps} />;
```

packages/ui/Layout/Cell.js
juwonjeong: I think we can remove this docs.

juwonjeong (approved):
LGTM

PR #3247, opened 2024-06-24
WRQ-24303: Fix `marqueeDecorator` and `marqueeController` to start animation properly when content changed

juwonjeong:

### Checklist

When <Marquee> is wrapped with MarqueeController to be synchronized and marqueeOn props is 'render' and the content changes, MarqueeDecorator cannot start the marquee animation. 

### Resolution
`MarqueeDecorator.componentDidUpdate()` called when content changed. 
As children is changed, `this.invalidateMetrics()`, `this.cancelAnimation()` are called and `forceRestartMarquee` variable set to true in `componentDidUpdate` function,

`this.tryStartingAnimation()` function is called because `forceRestartMarquee` is true.
However, In `tryStartingAnimation()`, it returns without calling `startAnimation()` function becuase `this.timerState` is 1(=TimerState.START_PENDING). 
`this.timerState` is set to 1 in `MarqueeDecorator.start()` function which is called before `componentDidUpdate`. 

So I added `retryStartingAnimation` flag in `MarqueeDecorator.cancelAnimation()` to check whether we need to restartAnimation after cancelJob completes.

### Additional Considerations
In additional, I think handleComplete in useMarqueeController needs to be fixed too.
When <Marquee> is wrapped with a MarqueeController, I found that the MarqueeDecorator.start function is constantly being called, even though content is short that doesn't need to be animate.

### Links
WRQ-24303

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by MarqueeController and text changed
+- `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by `ui/Marquee.MarqueeController` and text changed
 
 ## [4.9.0-beta.1] - 2024-06-17
 
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by MarqueeController and text changed
+
 ## [4.9.0-beta.1] - 2024-06-17
 
 No significant changes.
--- a/packages/ui/Marquee/useMarqueeController.js
+++ b/packages/ui/Marquee/useMarqueeController.js
@@ -193,7 +193,7 @@ const useMarqueeController = (props) => {
 	 */
 	const handleComplete = useCallback((component) => {
 		const complete = markReady(component);
-		if (complete) {
+		if (complete && !component.contentFits) {
 			markAll(STATE.ready);
 			dispatch('start');
 		}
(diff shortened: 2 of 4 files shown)


juwonjeong:
Since the patch is implemented in cancelJob which is called with setTimeout, I cannot add unit test of it.

packages/ui/Marquee/useMarqueeController.js
```diff
@@ -193,7 +196,7 @@ const useMarqueeController = (props) => {
 	 */
 	const handleComplete = useCallback((component) => {
 		const complete = markReady(component);
-		if (complete) {
+		if (complete && !component.contentFits) {
```
MikyungKim, line 199: I think we can have this change in develop branch, but we need to discuss when we make a change to deployed one.

packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by MarqueeController and text changed
```
```suggestion
- `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by `ui/Marquee.MarqueeController` and text changed
```

packages/ui/Marquee/useMarqueeController.js
```diff
@@ -178,9 +181,9 @@ const useMarqueeController = (props) => {
 	 *
```
MikyungKim, line 181: Could you fix the comments at line 180?

packages/ui/CHANGELOG.md
juwonjeong: It's fixed

packages/ui/Marquee/useMarqueeController.js
juwonjeong: Okay I agree.

packages/ui/Marquee/useMarqueeController.js
juwonjeong: I fixed the comment.

MikyungKim (approved):
LGTM

0x64:
Let's add 'restart' function to MarqueeDecorator.js that simply calls 'restartAnimation' for consistent API sets. Looks great except this.

juwonjeong:
It's fixed :)

0x64:
Checked codecov results and it looks okay.

PR #3166, opened 2023-08-02
WRP-21845: update lerna v7

hong6316:
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [ ] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [ ] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
`lerna` v7 has been released.
This module needs to be updated to the latest node modules to fix security vulnerabilities and prevent malware.

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
Update `lerna` v7

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)
- Fix the 'bootstrap' and 'interlink' scripts to work with `lerna` v7
`lerna` v7 removed the `lerna bootstrap`, `lerna add`, and `lerna link` commands by default.
(Reference: https://lerna.js.org/docs/legacy-package-management)
So this repo needs to be migrated to work with `lerna` v7.

- Fix the `unlink-all` script that has not worked.

### Links
[//]: # (Related issues, references)
WRP-21845

### Comments
Enact-DCO-1.0-Signed-off-by: Taeyoung Hong (taeyoung.hong@lge.com)

PR #3248, opened 2024-06-27
WRQ-26798: Remove Lerna and refine run scripts

0x64:
### Checklist

After #3166, 'npm run bootstrap' does not link internal packages automatically. 

### Resolution
As we use npm higher than 7 now, we can use npm workspaces.
* Updated our monorepo to use npm workspaces
* Removed Lerna as we do not use any complex function of Lerna
* Made 'npm install' to make an internal link also
* Made 'npm run clean' to remove 'node_modules' of the root path also
* Removed 'npm run publish' as we don't use it anymore
* But, kept the existing run scripts rather than removing them for any scripts
* Updated a related document

### Additional Considerations

### Links
WRQ-26798

### Comments


hong6316 (dismissed):
👍 
But please fix the travis fail

PR #3252, opened 2024-07-09
WRQ-28000: Fix `VoiceControlDecorator` and `PlaceholderControllerDecorator` not to remount its children when it rerenders

juwonjeong:

### Checklist

When `VoiceControlDecorator` wrapped video player component in the app, video player component remounted continuously.

### Resolution
In the previous logic, every time `VoiceControlDecorator` render, the children of the `VoiceControlDecorator` are unmounted and then remounted. 
It means that `Video` and `Media` components are re-mounted continuously. 
To fix this issue, we modify `VoiceControlDecorator` and `PlaceholderControllerDecorator` to not use HOCs in the render method.

### Additional Considerations

### Links
WRQ-28000

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by `ui/Marquee.MarqueeController` and text changed
+- `ui/Placeholder.PlaceholderControllerDecorator` to not remount its children when it rendered
 
 ## [4.9.0-beta.1] - 2024-06-17
 
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -7,7 +7,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by `ui/Marquee.MarqueeController` and text changed
-- `ui/Placeholder.PlaceholderControllerDecorator` to not remount its children when it rendered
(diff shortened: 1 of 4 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -7,6 +7,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 ### Fixed
 
 - `ui/Marquee.MarqueeDecorator` to start animation properly when synchronized by `ui/Marquee.MarqueeController` and text changed
+- `ui/Placeholder.PlaceholderControllerDecorator` to not remount its children when it rendered
```
MikyungKim, line 10:
```suggestion
- `ui/Placeholder.PlaceholderControllerDecorator` to not remount its children when it rerenders
```

packages/webos/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact webos module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `webos/speech.VoiceControlDecorator` to not remount its children when it rendered
```
MikyungKim, line 9:
```suggestion
- `webos/speech.VoiceControlDecorator` to not remount its children when it rerenders
```

MikyungKim:
To be clearer for the remounting the children, if we define component in render function, React will see a new component type on every render and destory the entire subtree's DOM nodes and state (https://legacy.reactjs.org/docs/reconciliation.html#elements-of-different-types).

packages/ui/CHANGELOG.md
juwonjeong: It's fixed

packages/webos/CHANGELOG.md
juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

0x64 (approved):
Thank you for catch my 🐛!

PR #3254, opened 2024-07-15
WRQ-28801: Rollback `handleComplete` function in `useMarqueeController` to not decide whether to start by checking contentFits

juwonjeong:

### Checklist

In https://github.com/enactjs/enact/pull/3247, I fixed the issue to start animation properly by `marqueeController` when content changed. 
As mentioned in the Additional Consideration section, I added the additional patch in that PR which fix `handleComplete` in `useMarqueeController` becuase it seems that MarqueeDecorator.start function is constantly being called in sync mode even though content that doesn't need to be animate. 
However we found that this patch cause some regression in non-sync mode. So we rollback this additional patch in `useMarqueeController`.

### Resolution
Rollback the patch in useMarqueeController.handleComplete.

### Additional Considerations

### Links
WRQ-28801

### DIFF
--- a/packages/ui/Marquee/useMarqueeController.js
+++ b/packages/ui/Marquee/useMarqueeController.js
@@ -196,7 +196,7 @@ const useMarqueeController = (props) => {
 	 */
 	const handleComplete = useCallback((component) => {
 		const complete = markReady(component);
-		if (complete && !component.contentFits) {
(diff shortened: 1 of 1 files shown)


MikyungKim (approved):
LGTM

0x64 (approved):

PR #3260, opened 2024-07-23
WRQ-27279:  Enable VirtualList and VirtualGridList to read the order of item when each item set role=`listitem`

juwonjeong:

### Checklist

Sandstone virtualList and virtualGridList cannot read out the order of item even when item set its role to `listitem`.
VirtualList and virtualGridList have item wrapper divs, which prevents the list and listitems from being direct parent-child DOM element.

### Resolution
Set the `presentation` role to wrapper divs so that there are not exposed in the accessibility tree.

### Additional Considerations

### Links
WRQ-27279


seunghoh (approved):
LGTM

PR #3264, opened 2024-08-06
WRQ-28860: Modified so that related warnings appear only when accessing the webos property of core/platform

mmyelyn:
### Checklist

Previously, when accessing the core/platform module, you could see the deprecated warning below.

> DEPRECATED: webos. Will be removed in 5.0.0. Refer `@enact/webos`'s `platform` for webOS specific information..

It would be better to appear this to users accessing webos properties.
Users accessing other properties may not be interested in that information.

### Resolution
The warning has been modified to only be appeared to users accessing platform.webos

### Additional Considerations

### Links
WRQ-28860

### Comments

### DIFF
--- a/packages/core/platform/platform.js
+++ b/packages/core/platform/platform.js
@@ -150,13 +150,7 @@ const parseUserAgentLegacy = (userAgent) => {
 		}
 	}
 
-	if ('webos' === plat.platformName) {
-		deprecate({
-			name: plat.platformName,
-			message: 'Refer `@enact/webos`\'s `platform` for webOS specific information.',
-			until: '5.0.0'
-		});
-	} else if (!['chrome', 'safari', 'firefox'].includes(plat.platformName)) {
+	if (!['chrome', 'safari', 'firefox'].includes(plat.platformName)) {
 		deprecate({
 			name: plat.platformName,
 			until: '5.0.0'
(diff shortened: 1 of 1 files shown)


juwonjeong:
There are still Deprecated message when accessing the core/platform module
> DEPRECATED: webos. Will be removed in 5.0.0.
It seems like this message needs to be a bit more detailed.

packages/core/platform/platform.js
```diff
@@ -150,13 +150,7 @@ const parseUserAgentLegacy = (userAgent) => {
...
 	}
 
-	if ('webos' === plat.platformName) {
-		deprecate({
-			name: plat.platformName,
-			message: 'Refer `@enact/webos`\'s `platform` for webOS specific information.',
-			until: '5.0.0'
-		});
-	} else if (!['chrome', 'safari', 'firefox'].includes(plat.platformName)) {
+	if (!['chrome', 'safari', 'firefox'].includes(plat.platformName)) {
```
juwonjeong, line 153: Could you please remove this if statement too?

mmyelyn: It seems like this log is not needed now. I removed

juwonjeong (approved):
LGTM

PR #3238, opened 2024-05-17
WRQ-19328: Fix React 18.3 warnings in enact

juwonjeong:
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [ ] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [ ] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
In React 19, defaultProps in function components will be removed.
In React 18.3, There are warnings for key spreading

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
Remove defaultProps for function components and replaced them with ES6 default parameters.
Pass key directly to JSX without using spread

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)

### Links
[//]: # (Related issues, references)
WRQ-19328

### Comments
Enact-DCO-1.0-Signed-off-by: Juwon Jeong (juwon.jeong@lge.com)

PR #3265, opened 2024-08-08
WRQ-31515: Fixed `Scroller`, `VirtualList`, and `VirtualGridList` to set prop value to default when `undefined` is passed for the prop value

juwonjeong:
### Checklist

In https://github.com/enactjs/enact/pull/3238, we removed defaultProps and implemented this feature manually.
After this implementation, if `undefined` is passed for a prop value, the prop value remains `undefined`. 
However, defaultProps(https://legacy.reactjs.org/docs/typechecking-with-proptypes.html#default-prop-values) sets prop value to default value when `undefined` value is passed for the prop. 

### Resolution
Fix `Scroller`, `VirtualList` and `VirtualGridList` to set prop value to default value when `undefined` is passed for the prop value

### Additional Considerations
I checked React's defaultProps for falsy values(null, undefined, false, NaN, 0, -0, "") and found that it sets the prop value to its default value only when undefined is passed.

And it seems that React's defaultProps only checked `undefined`.
- https://react.dev/reference/react/Component#static-defaultprops
- https://github.com/facebook/react/pull/28733/files/41b4a691b2ccf01a224a4e7c6e4f8e979a125876#diff-dbd72a473871eeddbe00ceed54a59bff33e6324ffdea15cb4ee3abbdad988cfbR348

### Links
WRQ-31515

### DIFF
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `core/util` function `setDefaultProps` to set props to default values when props are missing or `undefined`
+
 ## [4.9.0] - 2024-07-17
 
 No significant changes.
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -6,7 +6,7 @@ The following is a curated list of changes in the Enact core module, newest chan
 
 ### Added
 
-- `core/util` function `setDefaultProps` to set props to default values when props are missing or `undefined`
+- `core/util` function `setDefaultProps` to set props that are missing or `undefined` to default values
 
 ## [4.9.0] - 2024-07-17
 
--- a/packages/core/util/util.js
+++ b/packages/core/util/util.js
@@ -267,7 +267,7 @@ const mapAndFilterChildren = (children, callback, filter) => {
 };
 
 /**
- * Set props to default values when props are missing or `undefined`
+ * Sets props that are missing or `undefined` to default values
  *
  * @function
  * @param {Obejct}        props           Props object
--- a/packages/core/util/util.js
+++ b/packages/core/util/util.js
@@ -267,6 +267,30 @@ const mapAndFilterChildren = (children, callback, filter) => {
 };
 
 /**
+ * Set props to default values when props are missing or `undefined`
+ *
+ * @function
+ * @param {Obejct}        props           Props object
+ * @param {Obejct}        defaultProps    Default value object
+ *
+ * @returns {Object}                      Props with default values
+ * @memberof core/util
+ * @public
+ */
+const setDefaultProps = (props, defaultProps = {}) => {
+	const result = Object.assign({}, props);
+
+	for (const prop in defaultProps) {
+		// eslint-disable-next-line no-undefined
+		if (props[prop] === undefined) {
+			result[prop] = defaultProps[prop];
+		}
+	}
+
+	return result;
+};
+
+/**
  * Performs shallow comparison for given objects.
  *
  * @function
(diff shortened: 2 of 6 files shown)


packages/ui/Scroller/Scroller.js
```diff
@@ -53,7 +53,13 @@ const scrollerDefaultProps = {
 const Scroller = (props) => {
 	// Hooks
 
-	const scrollerProps = Object.assign({}, scrollerDefaultProps, props);
+	const scrollerProps = Object.assign({}, props);
+	for (const prop in scrollerDefaultProps) {
+		// eslint-disable-next-line no-undefined
+		if (scrollerProps[prop] === undefined) {
```
seunghoh, line 59: Don't we need general check for the prop like `if (!scrollerProps[prop])` ?  
Is it OK to check only `=== undefined` ?

juwonjeong: I checked React's defaultProps for falsy values(null, undefined, false, NaN, 0, -0, "") and found that it sets the prop value to its default value only when `undefined` is passed. So I implemeted it to check only `undefined` case. I will add this to Additional Considerations sections.

packages/ui/Scroller/Scroller.js
```diff
@@ -53,7 +53,13 @@ const scrollerDefaultProps = {
 const Scroller = (props) => {
 	// Hooks
 
-	const scrollerProps = Object.assign({}, scrollerDefaultProps, props);
+	const scrollerProps = Object.assign({}, props);
+	for (const prop in scrollerDefaultProps) {
```
MikyungKim, line 57: Let's make a util function for this and use it through ui, sandstone, and agate.
Input would be props and defaultProps objects and it will return a object that has the default prop set if a prop set to `undefined`.

juwonjeong: I added util function

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
+## [unreleased]
+
+### Added
+
+- `core/util` function `setDefaultProps` to set props to default values when props are missing or `undefined`
```
MikyungKim, line 9:
```suggestion
- `core/util` function `setDefaultProps` to set props that are missing or `undefined` to default values
```

packages/core/util/util.js
```diff
@@ -266,6 +266,30 @@ const mapAndFilterChildren = (children, callback, filter) => {
 	}
 };
 
+/**
+ * Set props to default values when props are missing or `undefined`
```
MikyungKim, line 270:
```suggestion
 * Sets props that are missing or `undefined` to default values
```

packages/ui/CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/Scroller`, `ui/VirtualList.VirtualList` and `ui/VirtualList.VirtualGridList` to set prop value to default when `undefined` is passed for the prop value
```
```suggestion
- `ui/Scroller` and `ui/VirtualList` to have default prop when `undefined` prop is passed
```

packages/core/CHANGELOG.md
juwonjeong: It's fixed.

packages/ui/CHANGELOG.md
juwonjeong: It's fixed.

packages/core/util/util.js
juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

0x64 (approved):

PR #3274, opened 2024-09-04
WRR-1010: Fixed to recognize the last input type properly

0x64:
### Checklist

The last input type is not recognized properly when `requestLastInputType` is called on some platforms since the function do nothing on unsupported platforms.

### Resolution
The issue occurs after #3218 that adds `requestLastInputType`. At the moment we added this API, it is unclear that who is in charge of handling unsupported platforms among Enact, theme libraries, and apps.
The `requestLastInputType` function's role is simply to call LS2Request so the usage of this API depends on apps' logic. We'd like to remove platform checking and make this function to call LS2Request always. It means that Enact does not involve checking if the platform is supported. So, theme libraries or apps should check the availability.

### Additional Considerations
On unsupported platforms, LS2Request API covers warnings so `requestLastInputType` does not need to cover the case.

### Links
WRR-1010
enactjs/sandstone#1682
Related: WRQ-6498 / #3218

### Comments

### DIFF
--- a/packages/webos/lastInputType/lastInputType.js
+++ b/packages/webos/lastInputType/lastInputType.js
@@ -6,20 +6,15 @@
  */
 
 import LS2Request from '../LS2Request';
-import platform from '../platform';
 
 const requestLastInputType = ({onSuccess, onFailure}) => {
-	if (platform.tv) {
-		return new LS2Request().send({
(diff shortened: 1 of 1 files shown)


MikyungKim (approved):
LGTM

juwonjeong (approved):
LGTM

PR #3275, opened 2024-09-06
WRR-2172: Fix `VirtualList` to update its scroll bounds when the total size of items is changed

juwonjeong:

### Checklist

There was an issue where the scroller could not move to the end of the scroller when the content of a VirtualList was changed.
This issue occurs when the content of virtualList is updated between fixed size content and variable size content and its minSize and dataSize is the same. 

For example, the size of the item contents is as follows:
contents A = [120, 120, 120]
contents B = [120, 240, 240]
In this case, the condition for recalculating the scroll bounds value is not met in the existing code. Because the minSizes and datasizes of the two content lists are the same.

If the scroll bounds value is not updated, scrollTo will not work because it will assume that maxLeft position of the scroller has already been reached.

### Resolution
In the existing code, we checked whether the minimum size of the item has changed, whether the data size has changed, etc. In this commit, we added a condition to check whether the total size of the items has changed.

### Additional Considerations

### Links
WRR-2172

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -12,7 +12,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 - `ui/Marquee.MarqueeController` to start animation properly when `marqueeOnFocus` is set to `true` and text changed
 - `ui/Scroller` and `ui/VirtualList` to have default prop when `undefined` prop is passed
-- `ui/VirtualList.VirtualListBasic` to update its scroll bounds when the total width of items is changed
+- `ui/VirtualList` to update its scroll bounds when the total size of items is changed
 
 ## [4.9.0] - 2024-07-17
 
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -12,6 +12,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 - `ui/Marquee.MarqueeController` to start animation properly when `marqueeOnFocus` is set to `true` and text changed
 - `ui/Scroller` and `ui/VirtualList` to have default prop when `undefined` prop is passed
+- `ui/VirtualList.VirtualListBasic` to update its scroll bounds when the total width of items is changed
 
 ## [4.9.0] - 2024-07-17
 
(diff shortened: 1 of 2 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 - `ui/Marquee.MarqueeController` to start animation properly when `marqueeOnFocus` is set to `true` and text changed
 - `ui/Scroller` and `ui/VirtualList` to have default prop when `undefined` prop is passed
+- `ui/VirtualList.VirtualListBasic` to update its scroll bounds when the total width of items is changed
```
MikyungKim, line 15:
```suggestion
- `ui/VirtualList` to update its scroll bounds when the total width of items is changed
```

MikyungKim: Could you also change `ui/VirtualList.VirtualListBasic` to `ui/VirtualList` at line 9?
It's enough to say `ui/VirtualList`.

MikyungKim: 🤔 Is `total width of items` right term?

seunghoh: How about 
"to update its scroll bounds when the total width of all items changes"

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -406,7 +406,8 @@ class VirtualListBasic extends Component {
 			prevProps.direction !== this.props.direction ||
 			prevProps.overhang !== this.props.overhang ||
 			prevProps.spacing !== this.props.spacing ||
-			!equals(prevProps.itemSize, this.props.itemSize)
+			!equals(prevProps.itemSize, this.props.itemSize) ||
+			prevProps.itemSizes?.reduce((acc, cur) => acc + cur, 0) !== this.props.itemSizes?.reduce((acc, cur) => acc + cur, 0)
```
MikyungKim, line 410: 👍 works like charm!

packages/ui/CHANGELOG.md
MikyungKim: `total size of items` would be better :)

juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

0x64 (approved):

PR #3293, opened 2024-11-13
WRR-9866: Fix VirtualListBasic to update its item position when the size of each item is changed

juwonjeong:

### Checklist

There was an issue that VirtualList did not re-calculate the positions of items when items in the VirtualList was changed. 

For example, the size(=width) of the item contents is as follows:
contents A = [120, 240, 480]
contents B = [480, 240, 120]

In this case, VirtualList needs to re-calculate the positions of items to render them properly. 

### Resolution
I fixed the code to compare arrays of item sizes instead of the total size of the items.

### Additional Considerations

### Links
WRR-9866

### DIFF
--- a/packages/ui/VirtualList/VirtualListBasic.js
+++ b/packages/ui/VirtualList/VirtualListBasic.js
@@ -2,7 +2,7 @@ import classNames from 'classnames';
 import EnactPropTypes from '@enact/core/internal/prop-types';
 import {forward} from '@enact/core/handle';
 import {platform} from '@enact/core/platform';
-import {clamp} from '@enact/core/util';
+import {clamp, shallowEqual} from '@enact/core/util';
 import PropTypes from 'prop-types';
 import equals from 'ramda/src/equals';
 import {createRef, Component} from 'react';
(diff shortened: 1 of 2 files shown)


0x64 (approved):

packages/ui/VirtualList/VirtualListBasic.js
```diff
@@ -407,7 +407,7 @@ class VirtualListBasic extends Component {
 			prevProps.overhang !== this.props.overhang ||
 			prevProps.spacing !== this.props.spacing ||
 			!equals(prevProps.itemSize, this.props.itemSize) ||
-			prevProps.itemSizes?.reduce((acc, cur) => acc + cur, 0) !== this.props.itemSizes?.reduce((acc, cur) => acc + cur, 0)
+			!shallowEqual(prevProps.itemSizes, this.props.itemSizes)
```
seunghoh, line 410: This change looks better than before as we are comparing shallowEqual which we can early exit if the array objects are identical. 
I just want to check one thing, if the itemSize array doesn't change, when entering the componentDidUpdate function(eg. force update), check if the itemSize array objects are the same so that it is really possible to exit early from shallowcompare.

seunghoh (approved):
LGTM

PR #3295, opened 2024-11-19
WRR-287: Update the minimum supported versions of browsers

0x64:
### Checklist

Along with theme library's change, the minimum version of browsers need to be updated and our documents also need to be updated.
In detail, we start to use relative RGB that is supported from safari 16.4, chrome 119, and firefox 128.

### Resolution
* Updated the minimum supported versions of browsers: safari 16.4, chrome 119, and firefox 128
* Updated docs for https://enactjs.com/docs/developer-guide/creating-components/

### Additional Considerations

### Links
WRR-287

### Comments


MikyungKim:
Based on MDN, safari 16.4 supports relative RGB partially.

>  Partial support
Implementation based on older spec version. As a result, calculations with channel values do not work correctly, requiring values to be specified as percentages with units (%).

Did you know that?

0x64:
> Based on MDN, safari 16.4 supports relative RGB partially.
> 
> > Partial support
> > Implementation based on older spec version. As a result, calculations with channel values do not work correctly, requiring values to be specified as percentages with units (%).
> 
> Did you know that?

MDN explains like below.
> Implementation based on older spec version. As a result, channel value calculations do not work correctly, requiring values to be specified as percentages with units (e.g. 30%, which would be equivalent to a 76.5 <number> value). See [bug 267647](https://webkit.org/b/267647).

CanIUse (https://caniuse.com/?search=relative%20RGB) explains like below.
> Implementation based on older spec version. As a result, channel value calculations do not work correctly, requiring values to be specified as percentages with units (e.g. 30%, which would be equivalent to a 76.5 <number> value). See [bug 267647](https://webkit.org/b/267647).

From the webkit bug link https://webkit.org/b/267647, we can find several facts.
- > The mixing of number and percentage is from an earlier, and incorrect, version of the spec where channel keywords could resolve to either number or percentage as desired. Now the spec is corrected to say that they all resolve to number.
- So, the issue comes from an implementation of the old spec.
- This sample does not work on Safari 16.4: https://developer.chrome.com/blog/css-relative-color-syntax#saturate_a_color.
- And it works from Safari 18.0.
- I tested, it works on Safari 18.1. So, at least CanIUse shows a wrong info.

Actually, relative RGB spec is changed before becoming the official standard, so Safari's implementation from 16.4 before 18.0 is based on the old spec. That is why it has an issue.

So, for us, it does not matter since we simply use R,G,B value as derived without any additional calculation. If we need to calculate later, the supported version of Safari needs to be updated at least 18.0 or higher after testing.

MikyungKim (approved):
LGTM

juwonjeong (approved):
LGTM

PR #3296, opened 2024-11-25
WRR-11077: Fixed Scroller to prevent the native scrolling behavior caused by keydown events when the popup is opened

juwonjeong:

### Checklist

In this case, the Scroller has a button that opens FixedPopupPanel. The issue is that when press the down key at the moment the popup is opened, the scroller behind the FixedPopupPanel scrolls down.

### Resolution
To fix this issue, we have to prevent the native scrolling behavior caused by keydown events. 
So I fixed keydown handler of scrollerContainer to call `preventDefault` when a keydown event is caused by a directional key and the spotlight is paused.
To not to have `spotlight` dependency, the preventing function resides in sandstone and enact/Scroller just calls the funcion passed by prop like other cases.

### Additional Considerations

### Links
WRR-11077

### DIFF
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@enact/core": "^4.9.4",
+    "@enact/spotlight": "^4.9.4",
     "classnames": "^2.5.1",
     "direction": "^1.0.4",
     "invariant": "^2.2.4",
--- a/packages/ui/package.json
+++ b/packages/ui/package.json
@@ -33,7 +33,6 @@
   },
   "dependencies": {
     "@enact/core": "^4.9.4",
-    "@enact/spotlight": "^4.9.4",
     "classnames": "^2.5.1",
     "direction": "^1.0.4",
     "invariant": "^2.2.4",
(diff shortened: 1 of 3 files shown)


packages/ui/package.json
```diff
@@ -33,6 +33,7 @@
   },
   "dependencies": {
     "@enact/core": "^4.9.4",
+    "@enact/spotlight": "^4.9.4",
```
MikyungKim, line 36: Please revert this change.

juwonjeong: It's removed

packages/ui/useScroll/useScroll.js
```diff
@@ -132,6 +132,7 @@ const useScrollBase = (props) => {
 	delete rest.onScrollStart;
 	delete rest.onScrollStop;
 	delete rest.onWheel;
+	delete rest.preventScroll;
```
MikyungKim, line 135:
```suggestion
	delete rest.preventScroll; // scrollMode 'native'
```

juwonjeong: It's fixed.

MikyungKim (approved):
LGTM

0x64 (approved):

PR #3302, opened 2024-12-12
WRR-9415: Remove styles for legacy IE 5 to 8 from vendor-opacity mixin

0x64:
### Checklist

As we stopped supporting Internet Explorer browsers from 5.0.0-alpha.1, we don't need to keep IE related code. `.vendor-opacity()` LESS mixin has code for IE so that I want to remove it. But the mixin function itself is possibly used in apps and I will keep the mixin function even it simply provides the standard opacity CSS attribute.

### Resolution
Removed IE related code from `.vendor-opacity()` mixin.

### Additional Considerations

### Links
WRR-9415

### Comments

### DIFF
--- a/packages/ui/styles/mixins.less
+++ b/packages/ui/styles/mixins.less
@@ -153,12 +153,6 @@
 	&:fullscreen { @rule(); }
 }
 
-.vendor-opacity(@opacity) {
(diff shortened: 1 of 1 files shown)


MikyungKim (approved):
LGTM

PR #3321, opened 2025-04-07
WRR-22638: Added reset method to spotlight to reseting an instance of Spotlight Accelerator

mmyelyn:
### Checklist

There is an issue where the instance value of `SpotlightAccelerator` is not updated when passing spotlight as `inputField`. This causes the callback to not be called and the spotlight to not move.

### Resolution
Added it as a private function so that reset can be called from onKeyDown of `InputFieldSpotlightDecorator`.

### Additional Considerations

### Links
WRR-22638

### Comments

### DIFF
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -967,7 +967,16 @@ const Spotlight = (function () {
 		 * @param {Object} position `x` and `y` coordinates for the pointer
 		 * @private
 		 */
-		focusNextFromPoint: spotNextFromPoint
+		focusNextFromPoint: spotNextFromPoint,
+
+		/**
+		 * Resets spotlight accelerate
+		 *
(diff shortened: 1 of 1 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -967,7 +967,16 @@ const Spotlight = (function () {
...
 		 */
-		focusNextFromPoint: spotNextFromPoint
+		focusNextFromPoint: spotNextFromPoint,
+
+		/**
+		 * Resets spotlight accelerate
+		 *
+		 * @private
+		 */
+		resetSpotlightAccelerate: function () {
```
juwonjeong, line 977:
```suggestion
		resetSpotlightAccelerator: function () {
```

juwonjeong (approved):
LGTM

PR #3324, opened 2025-04-21
WRR-24070: Changed function to reset 5WayKeyHold value

mmyelyn:
### Checklist

There is an issue where Spotlight does not move if there is an InputField and it is not wrapped with the same SpotlightContainerDecorator. this is because the 5WayKeyHold value is not updated when passing spotlight as inputField. This causes the callback to not be called and the spotlight to not move.

### Resolution
Fixed to reset 5WayKeyHold by extending the `resetSpotlightAccelerator` function to `resetSpotlightAcceleratorAndHoldKey`

### Additional Considerations

### Links
WRR-24070

### Comments


juwonjeong (approved):
LGTM

PR #3327, opened 2025-04-30
WRR-22888: Remove eslint-related configs in package.json

hong6316:
### Checklist

We have already implemented eslint 9, and going forward, configs other than eslint.config.js (eslintrc files, eslintignore files, package.json, etc.) will no longer work.

### Resolution
Remove eslint-related configs in package.json

### Additional Considerations

### Links
WRR-22888

### Comments
Enact-DCO-1.0-Signed-off-by: Taeyoung Hong (taeyoung.hong@lge.com)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,16 +1,15 @@
 {
   "name": "enact",
-  "version": "5.0.0-alpha.5",
+  "version": "5.0.0-beta.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "enact",
-      "version": "5.0.0-alpha.5",
+      "version": "5.0.0-beta.1",
       "license": "Apache-2.0",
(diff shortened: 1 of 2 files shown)


package-lock.json
```diff
@@ -1,16 +1,15 @@
 {
   "name": "enact",
-  "version": "5.0.0-alpha.5",
```
daniel-stoian-lgp, line 3: i'm concerned about deleting package-lock and then running a npm install for this case.
When we add/remove dependencies or update a major version of a dependency, we should not delete package-lock. Deleting will cause installs of latest minor versions of other dependencies and we did not trace any changelog/breaking changes for those.
Please reinstate original package-lock, perform an npm i and then push the file.

daniel-stoian-lgp (approved):

PR #3375, opened 2026-01-12
NXT-7706: Applied item margin in VirtualList when scroll by pressing key

vJIYEv:
### Checklist

In VirtualList when scroll by focus an item with key, sometimes focused item was cut and not fully shown.

### Resolution
Item margin was not considered when calculate the scroll position.
If an item has margin-left (or margin-top), scroll length is not enough so cut area width (or height) is same as the item margin-left (or margin-top).
The item will not cut if scroll more as much as the item margin-left (or margin top).
And because item margin-left is shown (margin-top) in VirtualList, I thought margin-right (or margin-bottom) shoud be shown  too.
To show the item margin-right/bottom on the right side of VirtualList, just scroll as much as the item margin-right/bottom like the above.
So I changed the scroll position to consider both margin-left/top and margin-right/bottom. 

### Additional Considerations

### Links
NXT-7706

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact project, newest changes 
 
 ### Fixed
 
+- `ui/VirtulList` not to cut off focused item when scroll by pressing key
 - `ui/VirtualList.VirtualGridList` to adjust itemSize properly when resizing window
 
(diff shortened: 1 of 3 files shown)


daniel-stoian-lgp (approved):
LGTM

ion-andrusciac-lgp (approved):

PR #3378, opened 2026-01-15
NXT-560: Updated minimum supported safari version

daniel-stoian-lgp:
### Checklist

We need to update supported browser versions, based on usage and webOS needs

### Resolution

- webOS 2025 TV products have Chrome 120, so at least we need to support Chrome 120.

For general environment including desktop and mobile that browsers are frequently updated on, we don't have a clear year-based rule.
More reasonable rule would be usage-based.
I refer https://caniuse.com/usage-table and https://www.whatismybrowser.com/guides/the-latest-version/
Blink (Chrome/Edge): version 103 is the oldest version of at least 0.1% usage
WebKit (Safari): 16.6 for desktop(released 2023) and 15.8 for mobile (released 2021) are oldest version that has a usage over 0.1%.
Gecko (Firefox): 115 version is the oldest version that has a usage over 0.1% and this version is released 2023. Since Firefox is NOT shipped as a pre-installed manner, users of Firefox usually update their browser.
Trident (Internet Explorer): we should not support that.
 
In #3295, minimum versions were raised safari: 16.4, chrome: 119, firefox: 128 
Currently, we don't check node version since it is related to isomorphic build only that is possibly covered by polyfills handled by CLI.
 

So, our supported oldest versions of rendering engines are:

Blink (Chrome): 119
WebKit (Safari): 16.6
Gecko (Firefox): 128

Only Safari requires change

### Additional Considerations

### Links
NXT-560

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact project, newest changes 
 
 ## [unreleased]
 
+### Changed
+
+- `core/platform` to support `safari` 16.6, `chrome` 119, and `firefox` 128 or later
+
 ### Fixed
 
 - `ui/VirtulList` not to cut off focused item when scroll by pressing key
--- a/packages/core/CHANGELOG.md
+++ b/packages/core/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact core module, newest changes on the top.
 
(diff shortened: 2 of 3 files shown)


alexandrumorariu (approved):
LGTM

ion-andrusciac-lgp (approved):

PR #3379, opened 2026-01-22
NXT-8994: Adapted `Enact` code to respect `React Compiler`

ion-andrusciac-lgp:
### Checklist

Adapted `Enact` code to respect `React Compiler`

### Resolution
`eslint-plugin-react-hooks 7.x` has `React Compiler` embedded. The code needs to be adapted to respect `Eslint` rules

- Adapted how the `refs` were changed during render, and also cannot read `.current` during render. This is not allowed outside `handlers` or `effects` [REFERENCE](https://react.dev/reference/react/useRef#:~:text=You%20can%20read%20or%20write%20refs%20from%20event%20handlers%20or%20effects%20instead.)
- Modifying a variable defined outside a component or hook is not allowed
- Modifying component props or hook arguments is not allowed
- State update inside `useEffect` is not recommended 

### Additional Considerations
Some of the errors were silenced as they were considered false positives

### Links
NXT-8994

### Comments


packages/core/useClass/useClass.js
```diff
@@ -29,10 +29,9 @@ import {useRef} from 'react';
  * @private
  */
 function useClass (Ctor, ...args) {
-	const ref = useRef(null);
-	ref.current = ref.current || new Ctor(...args);
+	const [state] = useState(new Ctor(...args));
```
daniel-stoian-lgp:
```suggestion
	const [state] = useState(() => new Ctor(...args));
```

This will make it run only once, not on every rerender. What do you think?

packages/ui/Skinnable/useSkins.js
```diff
@@ -84,8 +84,14 @@ function useSkins (config) {
 
 	const {parentSkin, parentVariants} = use(SkinContext) || {};
 
+	const [currentParentVariants, setCurrentParentVariants] = useState(parentVariants);
+
+	if (parentVariants !== currentParentVariants) {
+		setCurrentParentVariants(parentVariants);
+	}
```
daniel-stoian-lgp: this causes an additional rerender on every parentVariants change. When parentVariants changes, the first render uses the old currentParentVariants, then setState triggers a second render with the new value

How about just dropping useMemo() ? Let me know your opinion.

 "const effectiveVariants = determineVariants(defaultVariants, variants, skinVariants, parentVariants);"

packages/ui/Toggleable/Toggleable.js
```diff
@@ -179,9 +179,12 @@ const ToggleableHOC = hoc(defaultConfig, (config, Wrapped) => {
 		// FIXME: Current behavior is to use `false` when switching from a truthy value to
 		// either null or undefined. The ternary below enforces that but we don't want to
 		// continue this exception in the future and should sunset it with this HOC.
-		const {current: instance} = useRef({selected: null});
+		const [instance, setInstance] = useState({selected: propSelected});
 		const selected = (instance.selected && propSelected == null) ? false : hook.selected;
-		instance.selected = propSelected;
+
+		if (propSelected && instance.selected !== propSelected) {
+			setInstance({selected: propSelected});
```
daniel-stoian-lgp: i see an issue of potential inifnite loops because of setState on render.

How about using the derivedStateFromProps pattern?


```suggestion
		const [derivedState, setDerivedState] = useState(() => ({
			prevPropSelected: null,
			selected: hook.selected
		}));

		if (propSelected !== derivedState.prevPropSelected) {
			const newSelected = (derivedState.prevPropSelected && propSelected == null) ? false : hook.selected;
			setDerivedState({
				prevPropSelected: propSelected,
				selected: newSelected
			});
		}

		const selected = derivedState.selected;
```

packages/ui/useScroll/Scrollbar.js
```diff
@@ -90,14 +88,18 @@ const useScrollbar = (props) => {
...
+			scrollbarHandleRef.current = {
+				getContainerRef,
+				showScrollbarTrack,
+				startHidingScrollbarTrack,
+				update
+			};
+		}
+
+		hideScrollbarTrackJob.current = hideScrollbarTrackJob.current || new Job(hideScrollbarTrack, scrollbarTrackHidingDelay);
+	});
```
daniel-stoian-lgp: don't we need some dependencies for this hook?

packages/ui/Toggleable/Toggleable.js
```diff
@@ -179,9 +179,12 @@ const ToggleableHOC = hoc(defaultConfig, (config, Wrapped) => {
...
+		}
```
daniel-stoian-lgp:
```suggestion
		const [prevPropSelected, setPrevPropSelected] = useState(null);

		if (propSelected !== prevPropSelected) {
			setPrevPropSelected(propSelected);
		}

		const selected = (prevPropSelected && propSelected == null) ? false : hook.selected;
```

packages/ui/Toggleable/Toggleable.js
```diff
@@ -179,9 +179,12 @@ const ToggleableHOC = hoc(defaultConfig, (config, Wrapped) => {
 		// FIXME: Current behavior is to use `false` when switching from a truthy value to
 		// either null or undefined. The ternary below enforces that but we don't want to
 		// continue this exception in the future and should sunset it with this HOC.
-		const {current: instance} = useRef({selected: null});
+		const [instance, setInstance] = useState({selected: propSelected});
```
daniel-stoian-lgp: This initializes with the current propSelected, but we need to track the previous value. The original ref version started with null.
Minor issue: You're wrapping the value in an object {selected: propSelected} when you could just store the value directly.

packages/ui/Toggleable/Toggleable.js
```diff
@@ -179,9 +179,12 @@ const ToggleableHOC = hoc(defaultConfig, (config, Wrapped) => {
...
 		const selected = (instance.selected && propSelected == null) ? false : hook.selected;
-		instance.selected = propSelected;
+
+		if (propSelected && instance.selected !== propSelected) {
```
daniel-stoian-lgp: This only updates when propSelected is truthy

daniel-stoian-lgp (approved):
LGTM

dan-ichim-lgp (approved):
LGTM, just double check useControlledState change.

PR #3380, opened 2026-01-26
NXT-575: Fix kind tests broken by React 19 update

alexandrumorariu:
### Checklist

I fixed the tests for the kind component. I also added a snip of code in order to enable defaultProps for our functional components. 

### Resolution
I added a check in render to make sure that we have defaultProps on the first render.

### Additional Considerations
The defaultProps  fix is a temporary solution until we find a better way of implementing this.

### Links
NXT-575


daniel-stoian-lgp:
In the PR description, please explain technically the issues that were happening and also explain tehnically the resolution, as well as additional considerantions
Also, please add changelog

packages/core/kind/kind.js
```diff
@@ -163,6 +163,15 @@ const kind = (config) => {
 				...boundHandlers
 			};
 
+			if (defaultProps) {
+				Object.keys(defaultProps).forEach(key => {
+					// eslint-disable-next-line no-undefined
+					if (merged[key] === null || merged[key] === undefined) {
```
daniel-stoian-lgp: React's defaultProps only replaced undefined, not null values:


```suggestion
					if (merged[key] === undefined) {
```

daniel-stoian-lgp (approved):
LGTM

ion-andrusciac-lgp (approved):

PR #3374, opened 2026-01-05
NXT-577: Fixed VirtualGridList to be responsive while resizing window

vJIYEv:
### Checklist

* [x] I have read and understand the [contribution guide](http://enactjs.com/docs/developer-guide/contributing/)
* [x] A [CHANGELOG entry](http://enactjs.com/docs/developer-guide/contributing/changelogs/) is included
* [ ] At least one test case is included for this feature or bug fix
* [x] Documentation was added or is not needed
* [ ] This is an API breaking change

### Issue Resolved / Feature Added
[//]: # (Describe the issue resolved or feature added by this pull request)
VirtualGridList items are not resized properly before reloading the app

### Resolution
[//]: # (Does the code work as intended?)
[//]: # (What is the impact of this change and *why* was it made?)
It is because item sizes are fixed at the time when the app is loaded first, and not changed even if the resolution is updated.
I added calculation for resizing when the resolution is updated.

### Additional Considerations
[//]: # (How should the change be tested?)
[//]: # (Are there any outstanding questions?)
[//]: # (Were any side-effects caused by the change?)
I found VirtualList has the same issue.
But the same solution did not work for VirtualList.
I will fix VirtualList later.

### Links
[//]: # (Related issues, references)
NXT-577

### Comments
Enact-DCO-1.0-Signed-off-by: Jiye Kim (jiye.kim@lge.com)

PR #3382, opened 2026-01-27
NXT-8190: Fixed VirtualList to be responsive while resizing window

vJIYEv:
### Checklist

VirtualList items are not resized properly before reloading the app.

### Resolution
1. This issue is similar to https://github.com/enactjs/enact/pull/3374
but different because itemSize is Number type, immutable.
I want to change itemSize value in useScroll like mutable value, so applied `useState` to itemSize.
2. If VirtualList items have different sizes, I should change each child of itemSizes.
So I scaled each child of itemSizes but this doesn't require useState because itemSizes is Object type, mutable.
3. If "item size" is changed from VirtualList sample, "itemSize" should be changed too.
So I made "originalItemSize" state and if originalItemSize is different with new item size, change the itemSize state.

### Additional Considerations
https://github.com/enactjs/limestone/pull/297 resolves the issue on the (sample) app side.

### Links
NXT-8190

### Comments
Enact-DCO-1.0-Signed-off-by: Jiye Kim ([jiye.kim@lge.com](mailto:jiye.kim@lge.com))


ion-andrusciac-lgp (approved):

daniel-stoian-lgp (approved):
LGTM

PR #3387, opened 2026-02-06
NXT-8190: Fixed VirtualList when itemSize is null

vJIYEv:
### Checklist

After https://github.com/enactjs/enact/pull/3382 is merged,
VirtualList is rerendered infinitely if itemSize is null.

### Resolution
Set itemSize as 0 when it is null.

### Additional Considerations

### Links
NXT-8190

### Comments

### DIFF
--- a/packages/ui/useScroll/useScroll.js
+++ b/packages/ui/useScroll/useScroll.js
@@ -174,8 +174,8 @@ const useScrollBase = (props) => {
 	const [isHorizontalScrollbarVisible, setIsHorizontalScrollbarVisible] = useState(horizontalScrollbar === 'visible');
 	const [isVerticalScrollbarVisible, setIsVerticalScrollbarVisible] = useState(verticalScrollbar === 'visible');
 	const [riRatio, setRiRatio] = useState(ri.scale(1));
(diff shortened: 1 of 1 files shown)


packages/ui/useScroll/useScroll.js
ion-andrusciac-lgp, line 175:
```suggestion
	const propsItemSize = props.itemSize || 0;
	
```

Maybe it will be better to declare a local const and use it where it's needed instead of using `props.itemSize || 0` everywhere?

ion-andrusciac-lgp, line 122:
```suggestion
			itemSize: itemPropsSize = 0,
			itemSizes,
```
Another approach may be to rename `itemSize` and give it a default value, and after that, use `itemPropsSize` instead of `props.itemSize || 0`.

vJIYEv: I updated the code following this.

vJIYEv: This one made the same error because itemSize value is null.

ion-andrusciac-lgp (approved):

daniel-stoian-lgp (approved):

PR #3389, opened 2026-02-11
NXT-9810: Added support for scrollend event

daniel-stoian-lgp:
### Checklist

scrollend event was launched https://developer.mozilla.org/en-US/docs/Web/API/Document/scrollend_event 

### Resolution
Added scrollend event support, with fallback for older browsers

### Scenario: User scrolls with mouse wheel

*Before*
User scrolls
onScroll fires (many times per second)
scrollStopJob.start() - Restarts 200ms timer on EVERY scroll event
onScroll fires again... timer resets
onScroll fires again... timer resets
User stops scrolling
No more onScroll events...
Wait 200ms... 
Timer expires → scrollStopOnScroll() is called
Result - 200ms delay

*After*
User scrolls
The implementation now uses hybrid logic - different behavior for different scroll types:

1. For Mouse/Touch Scrolling
User scrolls with mouse wheel
onScroll fires (many times)
scrollStopJob.start() - Starts 200ms fallback timer
onScroll fires again... fallback timer resets
User stops scrolling
Browser fires 'scrollend' event 
onScrollEnd() called
  → isScrollAnimationTargetAccumulated? NO (mouse/touch)
  → scrollStopOnScroll() called immediately
  → scrollStopJob.stop() - Cancel fallback timer
Result: Instant scroll end detection (~0ms delay)

2. For Keyboard Scrolling (Accumulating)
User holds arrow key down
onScroll fires (many times)
scrollStopJob.start() - Starts 200ms fallback timer
onScroll fires again... fallback timer resets
scrollEndGraceTimer cleared (allows accumulation to continue)
User releases arrow key
Browser's smooth scroll continues briefly...
Browser fires 'scrollend' event
onScrollEnd() called
  → isScrollAnimationTargetAccumulated? YES (keyboard)
  → Start 100ms grace timer
  → scrollStopJob.stop() - Cancel fallback timer
More 'scrollend' events might fire (smooth scroll continuing)...
  → Each one restarts the 100ms grace timer
100ms passes with no new scrollend events
Grace timer expires → scrollStopOnScroll() called 
Result: 100ms debounced scroll end (prevents callback spam)

### Additional Considerations
No changelog because the component is private 
JSDOM does not support scrollend event , so the unit tests cannot cover scrollend

Another issue that was solved: Vertical wheel scrolling triggered onScrollStop twice - once from the 200ms fallback timer, and again from the scrollend event arriving late.
Solution:
- Added early return in onScrollEnd if scrolling is already false (prevents duplicate call when fallback timer fires first)
- Set isScrollAnimationTargetAccumulated = true in onScroll for native mode (ensures scrollend uses grace period)
- Don't restart fallback timer if already in grace period (prevents late onScroll from interfering)

### Links
NXT-9810

### Comments


ion-andrusciac-lgp (approved):

alexandrumorariu (approved):
LGTM

PR #3394, opened 2026-02-25
NXT-3891: Improved performance of useScroll and useMarqueeController 

daniel-stoian-lgp:
### Checklist

### Resolution

Scroll event is attached in useScroll, to be passive: `utilEvent('scroll').addEventListener(scrollContentRef, onScroll, {passive: true});` - By marking a touch or wheel listener as passive, the developer is promising the handler won't call preventDefault to disable scrolling. This frees the browser up to respond to scrolling immediately without waiting for JavaScript, thus ensuring a reliably smooth scrolling experience for the user.

 ` scrollContentHandle.current.didScroll(mutableRef.current.scrollLeft, mutableRef.current.scrollTop);`  triggers a React state update inside VirtualList (to recompute which items to render). 
In React 19.1, state updates called from a {passive: true} scroll event listener are now scheduled through the concurrent scheduler — they go into a low-priority lane and are deferred to a later frame. https://github.com/facebook/react/releases/tag/v19.1.0 

Solution: wrap didScroll in flushSync(); This keeps `passive: true` for smooth scrolling, but forces React to treat the didScroll state update as synchronous:

In React 19.1, a pre-existing bug in ui/Marquee/useMarqueeController became critical:
handleFocus and handleBlur had props in their useCallback dependency array. Since props is a new object reference on every render, both callbacks were recreated every render. This busted the value useMemo, producing a new MarqueeControllerContext object every render. Solution is to remove props from handleFocus/handleBlur deps, use a stable propsRef instead. Eliminates the context churn and the unconditional re-renders.

### Additional Considerations

### Links
NXT-3891

### Comments


ion-andrusciac-lgp (approved):

packages/ui/useScroll/useScroll.js
```diff
@@ -871,7 +872,9 @@ const useScrollBase = (props) => {
 		}
 
 		if (scrollContentHandle.current.didScroll) {
-			scrollContentHandle.current.didScroll(mutableRef.current.scrollLeft, mutableRef.current.scrollTop);
+			flushSync(() => {
```
daniel-stoian-lgp: this change seems to cause the scroll issue on Dropdown component

PR #3398, opened 2026-03-19
NXT-10863: Marquee - Investigate performance improvements

dan-ichim-lgp:
### Checklist

Reduced per-instance lifecycle overhead in ui/Marquee for high-churn rendering scenarios (e.g. VirtualList scrolling), without changing public API or behavior.

### Resolution

Four targeted improvements to MarqueeDecorator hot paths:

Render-mode start gating — Added an internal hasStartedRender flag so a render-mode marquee only starts once per mount. Resets when a metrics-invalidating update (content, locale, RTL, spacing) forces a real restart.
Conditional cancelAnimation() — In componentDidUpdate, cancelAnimation() is now called only when the marquee is actually active (sync, animating, promoted, or timer pending). Idle marquees no longer trigger cancel/start churn on every prop update.
Conditional validateTextDirection() — Only runs when direction-relevant inputs change (rtl, forceDirection, locale, or children).
Metrics invalidation on re-enable — When marqueeDisabled transitions true → false, cached metrics are cleared so a potential container resize during the disabled period doesn't cause a stale animation distance.
CSS — will-change: transform replaces will-change: opacity (more accurate compositing hint); 

### Additional Considerations

### Links

NXT-10863


packages/ui/Marquee/MarqueeDecorator.js
```diff
@@ -405,12 +417,26 @@ const MarqueeDecorator = hoc(defaultConfig, (config, Wrapped) => {
 				prevProps.marqueeSpeed !== marqueeSpeed ||
 				prevProps.forceDirection !== forceDirection
 			) {
-				this.cancelAnimation();
+				if (
```
daniel-stoian-lgp, line 417: before your change, it would be good to have :
				// When re-enabling after marqueeDisabled, invalidate stale metrics
				if (prevProps.marqueeDisabled && !marqueeDisabled) {
					this.invalidateMetrics();
				}


When marqueeDisabled transitions true → false, Content may have changed multiple times during scroll while measurements were suppressed

daniel-stoian-lgp: Also, check useMarqueeController() changes from https://github.com/enactjs/enact/pull/3394/changes 
Maybe we can add them to this PR

dan-ichim-lgp: > before your change, it would be good to have : // When re-enabling after marqueeDisabled, invalidate stale metrics if (prevProps.marqueeDisabled && !marqueeDisabled) { this.invalidateMetrics(); }
> 
> When marqueeDisabled transitions true → false, Content may have changed multiple times during scroll while measurements were suppressed

good catch

daniel-stoian-lgp:
LGTM, but 
1. add changelog
2. refine the PR description to contain all relevant information. Remove redundant information
3. Run jenkins checks and add the link on the user story

packages/ui/Marquee/Marquee.module.less
```diff
@@ -34,14 +34,10 @@
 		}
 
 		&.willAnimate {
-			will-change: opacity;
+			will-change: transform;
 			overflow: visible;
 		}
 
-		&.animate {
```
daniel-stoian-lgp, line 41: some ui tests are failing because of the deletion. what was the purpose of it? animate class is set in MarqueeBase

dan-ichim-lgp: Revert

alexandrumorariu (approved):
LGTM

PR #3399, opened 2026-03-20
NXT-10084: Fixed VirtualList to find item with itemRef

vJIYEv:
### Checklist

When getting item margin in VirtualList, wrong item is targeted sometimes.

### Resolution
I changed the code to get item with itemRef instead of getting item by class name.
And updating scrollBound for every component update could be the reason of the issue.
I changed the code to update the scrollBound for once when itemMargin is calculated.

### Additional Considerations
For vertical orientation, optional offset is already considered and calculated so we don't need margin offset.
So margin offset is applied only when there isn't optional offset.

### Links
NXT-10084

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,7 +6,8 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/VirtualList.VirtualList`, to adjust itemSize properly when resizing window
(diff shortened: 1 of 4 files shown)


bongsok (commented):
LGTM

bongsok (approved):
LGTM.
Check with gallery app.

hong6316 (approved):
👍

PR #3400, opened 2026-03-26
NXT-10868: Added support for linear scaling type

ion-andrusciac-lgp:
### Checklist

Added support for linear scaling type. Previously, the system primarily relied on discrete screenTypes (breakpoints) to determine pxPerRem. While effective for standard resolutions, it didn't always provide a smooth transition for intermediate window sizes or non-standard aspect ratios.

### Resolution
- When `linearScaling.active: true`, the font size scales linearly as the window resizes, avoiding sudden jumps between breakpoints.
- By using `type: 'currentScreen'`, the system guarantees that the UI looks exactly as defined when the resolution matches a known `screenType`
- Constraint logic `(Math.min/max)` prevents the scale factor from becoming too small or too large, protecting the UI on extreme display setups.

### Additional Considerations
- The choice of the geometric mean ($\sqrt{scaleX \times scaleY}$) ensures that the scaling factor accounts for both width and height changes. This prevents the UI from appearing overly large on ultra-wide monitors where only the width increases significantly.
- Using `baseScreen mode`, creates a global linear curve where defined `pxPerRem` values for non-base screens are treated as suggestions/breakpoints for standard mode, but are overridden by the fluid math in linear mode.

### Links
NXT-10868

### Comments

### DIFF
--- a/packages/ui/resolution/resolution.js
+++ b/packages/ui/resolution/resolution.js
@@ -41,12 +41,12 @@ const linearScalingType = {
 	/**
 	 * Scales based on the `baseScreen` (the screen type marked as `base: true`).
 	 */
-	baseScreenReport: 'baseScreen',
+	baseScreen: 'baseScreen',
 
 	/**
 	 * Scales based on the current detected `screenType`.
 	 */
-	currentScreenReport: 'currentScreen'
+	currentScreen: 'currentScreen'
 };
 
 /**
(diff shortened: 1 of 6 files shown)


packages/ui/resolution/resolution.js
```diff
@@ -40,9 +52,44 @@ const unitToPixelFactors = {
  */
 const configDefaults = {
 	fontSizeHandling: 'scale',
-	orientationHandling: 'normal'
+	orientationHandling: 'normal',
+	linearScaling: {
+		active: true,
```
daniel-stoian-lgp: should it be active by default?

daniel-stoian-lgp (approved):
LGTM

alexandrumorariu (approved):
LGTM

PR #3402, opened 2026-03-26
NXT-10864 [Enact] Spottable - Investigate performance improvements

dan-ichim-lgp:
### Checklist

Reduced per-instance render overhead in spotlight/Spottable, with particular impact in VirtualList scrolling where many spottable items mount and re-render frequently.

### Resolution

Five independent optimizations:

Removed redundant mount render (Spottable.js) — componentDidMount called forceUpdate(), a leftover from the ReactDOM.findDOMNode() era. With WithRef + useImperativeHandle, the DOM node is available before useLayoutEffect fires on the first render, making the second render unnecessary. Saves one full render per Spottable mount.

Converted Spottable class wrapper to a function component (Spottable.js) — the inner Spottable class existed only to hold a handleForceUpdate instance method. It is now a plain function component that calls useReducer for a stable, allocation-free force-update trigger, removing a class instance and its field allocation per mount.

Removed dead delete rest.spotlightId (Spottable.js) — spotlightId is already destructured out of props in SpottableBase, so it is never present in rest. The delete was a no-op on every render.

Conditional Spotlight.getCurrent() (SpottableCore.js) — didUpdate previously called getCurrent() on every render to re-sync this.isFocused. The call is now skipped unless spotlightDisabled is true or a selection cancel may fire. this.isFocused stays accurate via handleFocus/handleBlur events in all other cases. unload() re-queries getCurrent() directly at unmount to keep onSpotlightDisappear reliable.

Reduced per-render allocations (useSpottable.js) — three sources of unnecessary allocations eliminated:

context.current is now mutated in-place instead of being replaced with a new object each render.
A shared EMPTY_ATTRIBUTES constant is used when spotlightId is not set, avoiding an empty-object allocation on every render.
The props rest object (a fresh copy already created by destructuring) is mutated in-place before being passed to hook.setPropsAndContext, avoiding a second spread-merge allocation that previously occurred on every render.

### Additional Considerations

No API or behavioral changes. Render-count expectations in shouldComponentUpdate tests updated to reflect the removed mount-time forceUpdate().

### Links

NXT-10864
Related: findDOMNode removal — commit 33c27d5 (WRQ-19325)

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
+### Changed
+
+- `spotlight/Spottable` to reduce redundant work during high-frequency re-renders
+
 ### Fixed
 
 - `spotlight` to focus visible children of elements which do not have `overflow:true`
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,10 @@ The following is a curated list of changes in the Enact project, newest changes 
 - `core/util` function `checkPropTypes` to check the prop types of a component
 - `ui/resolution` Added support for linear scaling type
 
+### Changed
+
+- `spotlight/Spottable` to reduce redundant work during high-frequency re-renders
+
 ### Fixed
 
 - `spotlight` to focus visible children of elements which do not have `overflow:true`
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -148,7 +148,6 @@ const Spottable = hoc(defaultConfig, (config, Wrapped) => {
 		}
 		rest.tabIndex = tabIndex;
 
-		delete rest.spotlightId;
 		const handlers = useHandlers(spotHandlers, rest, spot);
 
 		return (
--- a/packages/spotlight/Spottable/Spottable.js
+++ b/packages/spotlight/Spottable/Spottable.js
@@ -13,7 +13,7 @@ import {checkPropTypes} from '@enact/core/util';
 import {WithRef} from '@enact/core/internal/WithRef';
 import classNames from 'classnames';
 import PropTypes from 'prop-types';
-import {Component, useCallback, useRef} from 'react';
+import {useCallback, useReducer, useRef} from 'react';
 
 import {spottableClass, useSpottable} from './useSpottable';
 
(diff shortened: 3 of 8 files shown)


daniel-stoian-lgp:
Looks ok, but i can suggest some improvements:

1 Spottable.js:

line 153: Dead delete rest.spotlightId : spotlightId is already destructured out of props into its own variable at the top of SpottableBase, so it's never in rest to begin with

2. Spottable.js
This class component can be made functional:

> 	class Spottable extends Component {
> 		handleForceUpdate = () => {
> 			this.forceUpdate();
> 		};
> 
> 		render () {
> 			return <SpottableBase {...this.props} handleForceUpdate={this.handleForceUpdate} />;
> 		}
> 	}

functional: 

> function Spottable (props) {
> 		// The outer Spottable class exists for one reason only — to produce a stable handleForceUpdate callback.   
>                // useReducer's dispatch is guaranteed stable across renders by React, so it can replace the class entirely:
> 		const forceUpdate = useReducer(n => n + 1, 0)[1];
> 		return <SpottableBase {...props} handleForceUpdate={forceUpdate} />;
> }

3. useSpottable.js: Double object allocation on every render

> // Object 1: created by destructuring rest
> const useSpottable = ({emulateMouse, getSpotRef, selectionKeys = [...], spotlightDisabled, ...props} = {}) => {
>     // ...
>     // Object 2: created here to merge them back together
>     hook.setPropsAndContext({selectionKeys, spotlightDisabled, ...props}, context.current);

Object 2 can be modified to:

> props.selectionKeys = selectionKeys;
> props.spotlightDisabled = spotlightDisabled;
> hook.setPropsAndContext(props, context.current);

Since props (the rest object) is already a fresh copy, it can just be mutated in-place before passing


Please test and let me know your opinion about the suggestions

daniel-stoian-lgp:
please add information about the changes made after my comments, in the PR description

packages/spotlight/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ The following is a curated list of changes in the Enact spotlight module, newest
 
 ## [unreleased]
 
+### Changed
+
+- `spotlight/Spottable` to reduce redundant work during high-frequency re-renders
```
daniel-stoian-lgp, line 9: please also add a changelog entry in the main changlog file (on the root of the project)

daniel-stoian-lgp (approved):

alexandrumorariu (approved):
LGTM

PR #3411, opened 2026-04-22
NXT-10735: Create a kind-like that can support hooks

alexandrumorariu:
### Checklist

I created a new kind-like that can support hooks and added it to the kind file. Right now it is private but can still be imported from there.

### Resolution

### Additional Considerations
use() is not the same as useContext(). The React docs are explicit: use() can read a Context or a Promise. When passed a Promise, it suspends — and suspension during SSR with renderToString does not work the same way as in a streaming (renderToPipeableStream) environment.

The risk surfaces like this:
js

// If contextType is accidentally a Promise-like object, or if
// use() is passed something that triggers Suspense...
const ctx = use(contextType); // throws a Promise → breaks renderToString

With useContext(contextType) this can't happen — it only accepts a Context object and never suspends.

The original class branch also used static contextType which is immune to this entirely.
The fix

Replace use with useContext in the rewritten component:
js

// Before
import {createContext, use} from 'react';
// ...
const ctx = use(contextType);

// After
import {createContext, useContext} from 'react';
// ...

### Links
NXT-10735

### DIFF
--- a/packages/core/kind/functionalKind.js
+++ b/packages/core/kind/functionalKind.js
@@ -65,6 +65,55 @@ const NoContext = createContext(null);
  */
 
 /*
+ * Creates a new component with some helpful declarative syntactic sugar.
+ *
+ * Example:
+ * ```
+ *	import css from './Button.module.less';
+ *	// Return a functional component
+ *	const Button = functionalKind({
+ *		name: 'Button',
+ *		// expect color and onClick properties but neither required
+ *		propTypes: {
+ *			color: PropTypes.string
+ *		},
+ *		// if no color is provided, it'll be green
+ *		defaultProps: {
+ *			color: 'green'
+ *		},
+ *		// expect backgroundColor via context
+ *		contextType: React.createContext({ backgroundColor }),
+ *		// configure styles with the static className to merge with user className
+ *		styles: {
+ *			// include the CSS modules map so 'button' can be resolved to the local name
+ *			css,
+ *			className: 'button'
+ *		},
+ *		// add event handlers that are cached between calls to prevent recreating each call. Any
+ *		// handlers are added to the props passed to `render()`.  See core/handle.
+ *		handlers: {
+ *			onKeyDown: (evt, props) => { .... }
+ *		},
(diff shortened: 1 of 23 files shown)


packages/core/kind/functionalKind.js
```diff
@@ -0,0 +1,168 @@
...
+	const renderKind = (props, context) => {
+		if (renderStyles)   props = renderStyles(props, context);
+		if (renderComputed) props = renderComputed(props, context);
+
+		return useRender(props, context); // eslint-disable-line react-hooks/rules-of-hooks
+	};
+
+	const Component = function (props) {
+		// Hooks must always be called unconditionally and in the same order.
+		const ctx = useContext(contextType);
```
daniel-stoian-lgp, line 150: why did you use useContext() instead of use() ;
Please justify all your technical solutions in the PR description

packages/core/kind/functionalKind.js
```diff
@@ -0,0 +1,168 @@
...
+ * @property {Object.<string, any>} [defaultProps] Sets the default props
+ * @property {Object} [contextType] Specifies context type
+ * @property {StylesBlock} [styles] Configures styles
+ * @property {Object.<string, HandlerFunction>} [handlers] Adds event handlers
+ * @property {Object.<string, ComputedPropFunction>} [computed] Adds computed properties
+ * @property {RenderFunction} useRender The useRender function
+ */
+
+/*
+ * Creates a new functional component with declarative syntactic sugar.
```
daniel-stoian-lgp, line 117: can you add a working example or at leat a TODO here to add a working example

packages/core/kind/functionalKind.js
alexandrumorariu: Done

packages/core/kind/functionalKind.js
alexandrumorariu: Done

daniel-stoian-lgp (approved):

ion-andrusciac-lgp (approved):

PR #3412, opened 2026-04-23
NXT-7491 [2027TV] POC on 2way focus browsing

dan-ichim-lgp:
### Checklist

This change adds Tab / Shift+Tab keyboard navigation to Spotlight for Enact apps.

Before: Spotlight did not handle the Tab key. Tab used the browser’s default focus behavior, which does not understand Enact’s focus model (spottable controls, spotlight containers, portaled content). Keyboard users could not rely on Tab to move focus in a consistent, layout-based order across the UI.

After: Spotlight handles Tab and Shift+Tab and moves focus among spottable controls in visual order on screen (top to bottom, then left to right; mirrored for RTL).

This applies broadly—to any spottable UI, including lists, grids (e.g. VirtualGridList items), buttons, and form controls. The PR also adds extra behavior for open dropdown popups (portaled lists linked to triggers via aria-owns), which was the main motivation and test focus for this story.

### Resolution

What this PR introduces
1. General Tab navigation (all spottable controls)

When the user presses Tab or Shift+Tab, Spotlight:

- finds focusable (spottable) controls in the current scope;
- sorts them by where they appear on screen, not by DOM order alone;
- moves focus to the next or previous control in that order.

This works for normal page content, grid layouts, lists, and any other component built with spottable children—including VirtualGridList, for rendered/visible items in the DOM.

Note: Virtualized lists only include items currently in the DOM, so Tab moves through visible spottable items, not the full virtual dataset.

2. Additional behavior for open dropdown popups

Dropdown option lists use a special Enact pattern: a portaled popup list in a self-only spotlight container, linked to its trigger through aria-owns. For this case, Spotlight adds:

- Tab / Shift+Tab through options inside an open list;
- at the end or start of a list, handoff to the next control in layout order;
- when the next control is another open dropdown, focus enters that list at the first (Tab) or last (Shift+Tab) option;
- when a dropdown is closed, focus lands on its trigger button only.

Example (three open dropdowns):

last option in A → first option in B → first option in C → close button
(one Tab per step; Shift+Tab reverses)

Unchanged

5-way (arrow key) navigation — not modified
Public Spotlight API — no breaking changes; apps do not need code updates

How it works (simple overview)

1. User presses Tab or Shift+Tab.
2. Spotlight builds a list of spottable controls in screen order.
3. If focus is inside an open dropdown list, Tab stays within that list until the last/first option.
4. At a list boundary, Spotlight picks the next nearby control on the page.
5. If that control opens another list, focus enters that list; otherwise focus moves to the control itself (button, grid item, etc.).

Other cases: Tab is managed while Spotlight is paused (modal); Tab is skipped if the pointer moved during the key press; RTL layouts are supported.

### Links

NXT-7491

### Comments

### DIFF
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -14,6 +14,11 @@ No significant changes.
 ## [5.4.4] - 2026-05-22
 
 No significant changes.
+## [unreleased]
+
+### Fixed
+
+- `spotlight` keyboard focus flow for open dropdown popups so Tab/Shift+Tab traversal is predictable between open lists and can continue to nearby controls after popup navigation
 
 ## [5.5.0] - 2026-05-08
 
@@ -28,7 +33,6 @@ No significant changes.
 ### Fixed
 
 - `spotlight` to focus visible children of elements which do not have `overflow:true`
-- `spotlight` keyboard focus flow for open dropdown popups so Tab/Shift+Tab traversal is predictable between open lists and can continue to nearby controls after popup navigation
 
 ## [5.4.3] - 2026-02-26
 
--- a/packages/spotlight/CHANGELOG.md
+++ b/packages/spotlight/CHANGELOG.md
@@ -28,6 +28,7 @@ No significant changes.
 ### Fixed
 
 - `spotlight` to focus visible children of elements which do not have `overflow:true`
+- `spotlight` keyboard focus flow for open dropdown popups so Tab/Shift+Tab traversal is predictable between open lists and can continue to nearby controls after popup navigation
 
 ## [5.4.3] - 2026-02-26
 
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -28,6 +28,7 @@ import {spottableClass} from '../Spottable';
 import {getPausedInstance, isPaused, pause, resume} from '../Pause';
 
 import {getInputType} from './inputType';
+import {createTabTraversal} from './tabTraversal';
 import {contains} from './utils';
 
 import {
@@ -41,8 +42,6 @@ import {
 	getContainerNode,
 	getContainersForNode,
 	getLastContainer,
-	getOwnedSelfOnlyContainerIds,
-	getPopupOwnerElement,
 	getSpottableDescendants,
 	isContainer,
 	isContainer5WayHoldable,
@@ -98,17 +97,6 @@ const isLeft = is('left');
 const isRight = is('right');
 const isTab = is('tab');
 const isUp = is('up');
-
-// Vertical centers within this distance share a Tab row; horizontal position breaks ties.
-const TAB_ROW_THRESHOLD = 24;
-
-// CSS class applied by @enact/i18n I18nDecorator to the app root for RTL locales.
-const I18N_RTL_CLASS = 'enact-locale-right-to-left';
-
-// Per-keypress cache for getLinearTargetsInContainer, scoped to one handleTab call.
-let _linearTargetsCache = null;
-
-let _tabNavTestHooks = {};
 
 /**
  * Translates keyCodes into 5-way direction descriptions (e.g. `'down'`)
@@ -322,6 +310,13 @@ const Spotlight = (function () {
 		return next.reduce((focused, target) => focused || Spotlight.focus(target), false);
 	}
 
+	const {runWithLinearTargetsCache, spotLinear} = createTabTraversal({
+		focusElement,
+		getCurrent,
+		isPaused,
+		restoreFocus
+	});
+
 	// The below should be gated on non-production environment only.
 	function assignFocusPreview (elem) {
 		const directions = ['up', 'right', 'down', 'left'],
@@ -484,311 +479,6 @@ const Spotlight = (function () {
 	}
 
 	/*
-	 * Whether the document uses right-to-left layout for Tab ordering.
-	 *
-	 * Intentionally does not import @enact/i18n:
-	 * - `locale.isRtlLocale` is callback-based and depends on ilib; Tab handling needs a
-	 *   synchronous answer on every keydown without adding that heavyweight dependency to spotlight.
-	 * - `util.isRtlText` detects RTL characters in a string, not page layout direction.
-	 *
-	 * Instead, read layout direction already reflected on the DOM by the app or I18nDecorator:
-	 * the `dir` attribute when set, and the `enact-locale-right-to-left` class on the app root.
-	 */
-	function isRtlDocument () {
-		const root = document.documentElement;
-		const body = document.body;
-		const dir = (root && root.dir) || (body && body.dir);
-
-		if (typeof dir === 'string' && dir.toLowerCase() === 'rtl') {
-			return true;
-		}
-
-		if (root?.classList.contains(I18N_RTL_CLASS) || body?.classList.contains(I18N_RTL_CLASS)) {
-			return true;
-		}
-
-		return Boolean(document.querySelector('.' + I18N_RTL_CLASS));
-	}
-
-	/*
-	 * Chooses the subtree root for Tab linear traversal. When focus sits inside a portaled
-	 * `self-only` container (e.g. contextual popup list), descendants of the app root do not
-	 * include that subtree, so using the root container would mix in unrelated controls. The
-	 * innermost `self-only` ancestor matches the modal boundary the user is in.
-	 */
-	function getLinearTabSearchContainerId (focused) {
-		if (isPaused()) {
-			return getLastContainer() || rootContainerId;
-		}
-		if (!focused) {
-			return rootContainerId;
-		}
-		const ids = getContainersForNode(focused);
-		for (let i = ids.length - 1; i > 0; i--) {
-			const cfg = getContainerConfig(ids[i]);
-			if (cfg && cfg.restrict === 'self-only') {
-				return ids[i];
-			}
-		}
-
-		return rootContainerId;
-	}
-
-	/*
-	 * Whether point (aX,aY) comes before (bX,bY) in the same top-to-bottom, left-to-right order
-	 * used by `getLinearTargetsInContainer` (must stay in sync with that sort).
-	 */
-	function comesBeforeInTabOrder (aX, aY, bX, bY, isRtl) {
-		if (Math.abs(aY - bY) > TAB_ROW_THRESHOLD) {
-			return aY < bY;
-		}
-		if (aX !== bX) {
-			return isRtl ? aX > bX : aX < bX;
-		}
-
-		return false;
-	}
-
-	function resolveTargetToOpenPopupItem (target, currentPopupContainerId, isForward) {
-		const ownerNode = target?.closest?.('[aria-owns]');
-		if (!ownerNode || ownerNode === document.body) {
-			return target;
-		}
-
-		for (const popupContainerId of getOwnedSelfOnlyContainerIds(ownerNode, currentPopupContainerId)) {
-			const popupTargets = getLinearTargetsInContainer(popupContainerId);
-			const popupTargetEntry = isForward ? popupTargets[0] : popupTargets[popupTargets.length - 1];
-			const popupTarget = popupTargetEntry?.target || getTargetByContainer(popupContainerId);
-			if (popupTarget) {
-				return popupTarget;
-			}
-		}
-
-		return target;
-	}
-
-	function isTargetInSelfOnlyContainer (target) {
-		if (!target) {
-			return false;
-		}
-
-		// Use DOM ancestry only. `getContainersForNode()` may include containers connected via
-		// `aria-owns`, which would incorrectly mark popup owner controls as inside self-only popups.
-		const containerNode = target.closest?.('[data-spotlight-container][data-spotlight-id]');
-		if (!containerNode) {
-			return false;
-		}
-
-		const containerId = getContainerId(containerNode);
-		if (!containerId || containerId === rootContainerId) {
-			return false;
-		}
-
-		const config = getContainerConfig(containerId);
-		return Boolean(config && config.restrict === 'self-only');
-	}
-
-	function findLinearTabExitTargetInTargets (targets, ax, ay, isRtl, isForward) {
-		const orderedTargets = isForward ? targets : [...targets].reverse();
-
-		for (const {target, x: cx, y: cy} of orderedTargets) {
-			if (!target || isTargetInSelfOnlyContainer(target)) {
-				continue;
-			}
-			const inOrder = isForward ?
-				comesBeforeInTabOrder(ax, ay, cx, cy, isRtl) :
-				comesBeforeInTabOrder(cx, cy, ax, ay, isRtl);
-			if (inOrder) {
-				// Do not defer popup owners; resolveTargetToOpenPopupItem redirects into open popups.
-				return target;
-			}
-		}
-
-		return null;
-	}
-
-	/*
-	 * After Tab reaches the end (or start) of a `self-only` popup list, pick the next spottable
-	 * outside the popup: the immediate successor (or predecessor) in the same global linear order
-	 * as the root container, so grids work for horizontal and vertical neighbors.
-	 */
-	function findLinearTabExitTarget (focusedElement, selfOnlyContainerId, isForward) {
-		const containerNode = getContainerNode(selfOnlyContainerId);
-		if (!containerNode) {
-			return null;
-		}
-
-		const popupOwner = getPopupOwnerElement(containerNode);
-		const focusCenter = getRect(focusedElement).center;
-		const isRtl = isRtlDocument();
-		const rootTargets = getLinearTargetsInContainer(rootContainerId);
-
-		const pickExitTarget = (x, y, searchForward) => {
-			const candidate = findLinearTabExitTargetInTargets(
-				rootTargets,
-				x,
-				y,
-				isRtl,
-				searchForward
-			);
-			return candidate ?
-				resolveTargetToOpenPopupItem(candidate, selfOnlyContainerId, isForward) :
-				null;
-		};
-
-		if (isForward) {
-			const exitTarget = pickExitTarget(focusCenter.x, focusCenter.y, true);
-			if (exitTarget) {
-				return exitTarget;
-			}
-		}
-
-		if (popupOwner) {
-			const ownerRect = getRect(popupOwner);
-			// top+1 keeps row-level neighbors in the same TAB_ROW_THRESHOLD band as a tall owner button.
-			const exitTarget = pickExitTarget(ownerRect.center.x, ownerRect.top + 1, isForward);
-			if (exitTarget) {
-				return exitTarget;
-			}
-		}
-
-		if (!isForward) {
-			return pickExitTarget(focusCenter.x, focusCenter.y, false);
-		}
-
-		return null;
-	}
-
-	function getLinearTargetContainerId (target) {
-		// Prefer DOM ancestry over getContainersForNode so aria-owns owners stay in root Tab order.
-		const containerNode = target?.closest?.('[data-spotlight-container][data-spotlight-id]');
-		if (containerNode) {
-			const domContainerId = getContainerId(containerNode);
-			if (domContainerId) {
-				return domContainerId;
-			}
-		}
-
-		const containerIds = getContainersForNode(target);
-		return last(containerIds);
-	}
-
-	function getLinearTargetsInContainer (containerId) {
-		if (_linearTargetsCache?.has(containerId)) {
-			return _linearTargetsCache.get(containerId);
-		}
-
-		const isRtl = isRtlDocument();
-		const visitedContainers = new Set();
-		const visitedTargets = new Set();
-
-		const gatherSpottableLeaves = (id) => {
-			if (!id || visitedContainers.has(id)) {
-				return [];
-			}
-
-			visitedContainers.add(id);
-
-			return getSpottableDescendants(id).reduce((result, candidate) => {
-				if (isContainer(candidate)) {
-					const nestedContainerId = getContainerId(candidate);
-					result.push(...gatherSpottableLeaves(nestedContainerId));
-				} else if (!visitedTargets.has(candidate)) {
-					visitedTargets.add(candidate);
-					result.push(candidate);
-				}
-
-				return result;
-			}, []);
-		};
-
-		const targets = gatherSpottableLeaves(containerId)
-			.filter((target) => {
-				const targetContainerId = getLinearTargetContainerId(target);
-				return isNavigable(target, targetContainerId, true);
-			})
-			.map((target, order) => {
-				const {center} = getRect(target);
-				return {order, target, x: center.x, y: center.y};
-			})
-			.sort((a, b) => {
-				if (comesBeforeInTabOrder(a.x, a.y, b.x, b.y, isRtl)) return -1;
-				if (comesBeforeInTabOrder(b.x, b.y, a.x, a.y, isRtl)) return 1;
-				return a.order - b.order;
-			});
-
-		if (_linearTargetsCache) {
-			_linearTargetsCache.set(containerId, targets);
-		}
-
-		return targets;
-	}
-
-	function spotLinear (isForward) {
-		let currentFocusedElement = getCurrent();
-
-		if (!currentFocusedElement) {
-			if (!restoreFocus()) {
-				return false;
-			}
-			currentFocusedElement = getCurrent();
-			if (!currentFocusedElement) {
-				return false;
-			}
-		}
-
-		const searchContainerId = getLinearTabSearchContainerId(currentFocusedElement);
-		const linearTargets = getLinearTargetsInContainer(searchContainerId);
-		if (!linearTargets.length) {
-			return false;
-		}
-		let currentIndex = linearTargets.findIndex(({target}) => target === currentFocusedElement);
-		if (currentIndex < 0) {
-			const nearestTarget = getNearestTargetFromPosition(getRect(currentFocusedElement).center, searchContainerId);
-			currentIndex = nearestTarget ?
-				linearTargets.findIndex(({target}) => target === nearestTarget) :
-				-1;
-			// When no nearest target can be found (e.g. position is NaN), keep currentIndex at -1
-			// so that nextIndex = -1 + 1 = 0, landing on the first element in the list.
-			// Treating 0 as the virtual "current" would skip the first element entirely.
-		}
-
-		const nextIndex = currentIndex + (isForward ? 1 : -1);
-		const scopeIsSelfOnly = searchContainerId !== rootContainerId &&
-			getContainerConfig(searchContainerId)?.restrict === 'self-only';
-
-		if (nextIndex >= linearTargets.length) {
-			if (scopeIsSelfOnly && isForward) {
-				const exitTarget = findLinearTabExitTarget(
-					currentFocusedElement,
-					searchContainerId,
-					true
-				);
-				if (exitTarget) {
-					return focusElement(exitTarget, getContainersForNode(exitTarget));
-				}
-			}
-			return false;
-		}
-		if (nextIndex < 0) {
-			if (scopeIsSelfOnly && !isForward) {
-				const exitTarget = findLinearTabExitTarget(
-					currentFocusedElement,
-					searchContainerId,
-					false
-				);
-				if (exitTarget) {
-					return focusElement(exitTarget, getContainersForNode(exitTarget));
-				}
-			}
-			return false;
-		}
-
-		const nextTarget = linearTargets[nextIndex].target;
-		return focusElement(nextTarget, getContainersForNode(nextTarget));
-	}
-
-	/*
 	 * Tab/Shift+Tab handler. Runs spotLinear with a per-keypress target cache and calls
 	 * preventDefault when focus moved or when paused (to keep Tab inside a modal overlay).
 	 */
@@ -804,13 +494,7 @@ const Spotlight = (function () {
 			}
 		}
 
-		_linearTargetsCache = new Map();
-		let handled;
-		try {
-			handled = spotLinear(!evt.shiftKey);
-		} finally {
-			_linearTargetsCache = null;
-		}
+		const handled = runWithLinearTargetsCache(() => spotLinear(!evt.shiftKey));
 
 		if (handled || (blocked && isPaused())) {
 			preventDefault(evt);
@@ -1356,33 +1040,12 @@ const Spotlight = (function () {
 
 	};
 
-	_tabNavTestHooks = {
-		findLinearTabExitTarget,
-		getLinearTabSearchContainerId,
-		comesBeforeInTabOrder,
-		resolveTargetToOpenPopupItem,
-		isTargetInSelfOnlyContainer,
-		findLinearTabExitTargetInTargets,
-		getLinearTargetContainerId,
-		getLinearTargetsInContainer,
-		runWithLinearTargetsCache: (fn) => {
-			_linearTargetsCache = new Map();
-			try {
-				return fn();
-			} finally {
-				_linearTargetsCache = null;
-			}
-		},
-		spotLinear
-	};
-
 	return exports;
 
 })();
 
 export default Spotlight;
 export {
-	_tabNavTestHooks,
 	getDirection,
 	Spotlight
 };
--- a/packages/spotlight/src/spotlight.js
+++ b/packages/spotlight/src/spotlight.js
@@ -605,46 +605,39 @@ const Spotlight = (function () {
 		const isRtl = isRtlDocument();
 		const rootTargets = getLinearTargetsInContainer(rootContainerId);
 
-		if (isForward) {
-			const fallbackTarget = findLinearTabExitTargetInTargets(
+		const pickExitTarget = (x, y, searchForward) => {
+			const candidate = findLinearTabExitTargetInTargets(
 				rootTargets,
-				focusCenter.x,
-				focusCenter.y,
+				x,
+				y,
 				isRtl,
-				true
+				searchForward
 			);
-			if (fallbackTarget) {
-				return resolveTargetToOpenPopupItem(fallbackTarget, selfOnlyContainerId, true);
+			return candidate ?
+				resolveTargetToOpenPopupItem(candidate, selfOnlyContainerId, isForward) :
+				null;
+		};
+
+		if (isForward) {
+			const exitTarget = pickExitTarget(focusCenter.x, focusCenter.y, true);
+			if (exitTarget) {
+				return exitTarget;
 			}
 		}
 
 		if (popupOwner) {
 			const ownerRect = getRect(popupOwner);
 			// top+1 keeps row-level neighbors in the same TAB_ROW_THRESHOLD band as a tall owner button.
-			const ownerTarget = findLinearTabExitTargetInTargets(
-				rootTargets,
-				ownerRect.center.x,
-				ownerRect.top + 1,
-				isRtl,
-				isForward
-			);
-			if (ownerTarget) {
-				return resolveTargetToOpenPopupItem(ownerTarget, selfOnlyContainerId, isForward);
+			const exitTarget = pickExitTarget(ownerRect.center.x, ownerRect.top + 1, isForward);
+			if (exitTarget) {
+				return exitTarget;
 			}
 		}
 
 		if (!isForward) {
-			const fallbackTarget = findLinearTabExitTargetInTargets(
-				rootTargets,
-				focusCenter.x,
-				focusCenter.y,
-				isRtl,
-				false
-			);
-			if (fallbackTarget) {
-				return resolveTargetToOpenPopupItem(fallbackTarget, selfOnlyContainerId, false);
-			}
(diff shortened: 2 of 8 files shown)


packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+
+		return false;
+	}
+
+	function getPopupOwnerElement (containerNode) {
+		if (!containerNode) {
+			return null;
+		}
+
+		const owners = document.querySelectorAll('[aria-owns]');
```
daniel-stoian-lgp: `container.js` already has `getOwnedNodes(node)` (line 291) that parses aria-owns.
This function, plus `resolveTargetToOpenPopupItem` and `isOpenSelfOnlyPopupOwner` below, re-implement aria-owns parsing inline three separate times with subtle differences. Reuse `getOwnedNodes` if possible.

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+				if (ownedNode && ownedNode.contains(containerNode)) {
+					return owner;
+				}
+			}
+		}
+
+		return null;
+	}
+
+	function resolveTargetToOpenPopupItem (target, currentPopupContainerId, isForward) {
```
daniel-stoian-lgp: this function and isOpenSelfOnlyPopupOwner contain a lot of duplicate code. why was not a helper function used?

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+	 */
+	function findLinearTabExitTarget (focusedElement, selfOnlyContainerId, isForward) {
+		const containerNode = getContainerNode(selfOnlyContainerId);
+		if (!containerNode) {
+			return null;
+		}
+
+		const popupOwner = getPopupOwnerElement(containerNode);
+		const focusRect = focusedElement.getBoundingClientRect();
+		const rowThreshold = 24;
```
daniel-stoian-lgp: what's the reason for this number?

daniel-stoian-lgp: why is it declared at function level?

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+			);
+			if (fallbackTarget) {
+				return resolveTargetToOpenPopupItem(fallbackTarget, selfOnlyContainerId, false);
+			}
+		}
+		return null;
+	}
+
+	function getLinearTargetsInContainer (containerId) {
+		const rowThreshold = 24;
```
daniel-stoian-lgp: what's the reason for this number?

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+			return false;
+		}
+		let currentIndex = linearTargets.findIndex(({target}) => target === currentFocusedElement);
+		if (currentIndex < 0) {
+			const currentRect = currentFocusedElement.getBoundingClientRect();
+			const currentCenter = {
+				x: currentRect.left + (currentRect.width / 2),
+				y: currentRect.top + (currentRect.height / 2)
+			};
+			let nearestDistance = Number.POSITIVE_INFINITY;
```
daniel-stoian-lgp: why did you choose this constant?

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+		return false;
+	}
+
+	function getPopupOwnerElement (containerNode) {
+		if (!containerNode) {
+			return null;
+		}
+
+		const owners = document.querySelectorAll('[aria-owns]');
+		for (const owner of owners) {
```
daniel-stoian-lgp: why did you not use forEach() ?

packages/spotlight/src/spotlight.js
```diff
@@ -90,6 +90,7 @@ const isEnter = is('enter');
 const isLeft = is('left');
 const isRight = is('right');
 const isUp = is('up');
+const tabKeyCode = 9;
```
daniel-stoian-lgp: keymap.js already exposes `is('tab')` (see `isEnter`, `isUp`, etc. above).

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
 		}
 	}
 
+	function isRtlDocument () {
```
daniel-stoian-lgp: i18n exports an `isRtlText`/`isRtlLocale` helper used elsewhere in Enact. Either reuse @enact/i18n or document why this is intentional.

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+		}
+
+		return rootContainerId;
+	}
+
+	/*
+	 * Whether point (aY,aX) comes before (bY,bX) in the same top-to-bottom, left-to-right order
+	 * used by `getLinearTargetsInContainer` (must stay in sync with that sort).
+	 */
+	function comesBeforeInTabOrder (aY, aX, bY, bX, rowThreshold, isRtl) {
```
daniel-stoian-lgp: Param order `(aY, aX, bY, bX, ...)` seems unusual — humans usually write `(x, y)` pairs.

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+
+	function getPopupOwnerElement (containerNode) {
+		if (!containerNode) {
+			return null;
+		}
+
+		const owners = document.querySelectorAll('[aria-owns]');
+		for (const owner of owners) {
+			const owns = owner.getAttribute('aria-owns');
+			if (typeof owns !== 'string' || owns.length === 0) {
```
daniel-stoian-lgp: `getAttribute` can only return string or null. A simple `if (!owns) continue;` is enough

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+			}
+		}
+
+		return null;
+	}
+
+	function resolveTargetToOpenPopupItem (target, currentPopupContainerId, isForward) {
+		// Only redirect when the actual target (or its nearest spottable ancestor) is a popup owner.
+		// Avoid climbing to broad ancestors (e.g. app/root) which can cause loops between open popups.
+		const ownerNode = target?.closest ? target.closest('[aria-owns]') : null;
```
daniel-stoian-lgp: redundant guard as there is  optional chaining. Just write `target?.closest('[aria-owns]')`.

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+		const containerId = getContainerId(containerNode);
+		if (!containerId || containerId === rootContainerId) {
+			return false;
+		}
+
+		const config = getContainerConfig(containerId);
+		return Boolean(config && config.restrict === 'self-only');
+	}
+
+	function isOpenSelfOnlyPopupOwner (target, currentPopupContainerId) {
```
daniel-stoian-lgp: aprox 30 lines of this function are functionally identical to `resolveTargetToOpenPopupItem` above: same DOM walk, same filtering. Please collapse into one helper

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+		}
+
+		return false;
+	}
+
+	function findLinearTabExitTargetInTargets (targets, containerNode, ax, ay, rowThreshold, isRtl, isForward) {
+		const currentPopupContainerId = getContainerId(containerNode);
+		let popupOwnerFallback = null;
+
+		if (isForward) {
```
daniel-stoian-lgp: forward and reverse branches are mirror copies. They could share one body checks if it is forward or not

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+			for (const {target, x: cx, y: cy} of targets) {
+				if (!target || containerNode.contains(target) || isTargetInSelfOnlyContainer(target)) {
+					continue;
+				}
+				if (comesBeforeInTabOrder(ay, ax, cy, cx, rowThreshold, isRtl)) {
+					if (isOpenSelfOnlyPopupOwner(target, currentPopupContainerId)) {
+						popupOwnerFallback = popupOwnerFallback || target;
```
daniel-stoian-lgp: this keeps only the first match but the loop continues. Either break after capturing the fallback, or document why subsequent iterations matter. Right now it looks like dead work.

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+		return false;
+	}
+
+	function findLinearTabExitTargetInTargets (targets, containerNode, ax, ay, rowThreshold, isRtl, isForward) {
+		const currentPopupContainerId = getContainerId(containerNode);
+		let popupOwnerFallback = null;
+
+		if (isForward) {
+			for (const {target, x: cx, y: cy} of targets) {
+				if (!target || containerNode.contains(target) || isTargetInSelfOnlyContainer(target)) {
```
daniel-stoian-lgp: `containerNode.contains(target)` runs for every candidate; the targets list is already filtered by container ID on line 63. Is this filter actually needed?

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+
+		// Anchor exit traversal to the popup owner first so Tab completes the opened dropdown and
+		// then continues through surrounding nearby controls in the base layout order.
+		if (popupOwner) {
+			const ownerRect = popupOwner.getBoundingClientRect();
+			const ownerTarget = findLinearTabExitTargetInTargets(
+				rootTargets,
+				containerNode,
+				ownerRect.left + (ownerRect.width / 2),
+				ownerRect.top + 1,
```
daniel-stoian-lgp: Why not `top + height/2` like the fotward/reverse cases ? If the +1 is to bias above vs below, please mention

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+				false
+			);
+			if (fallbackTarget) {
+				return resolveTargetToOpenPopupItem(fallbackTarget, selfOnlyContainerId, false);
+			}
+		}
+		return null;
+	}
+
+	function getLinearTargetsInContainer (containerId) {
```
daniel-stoian-lgp: This is invoked from `spotLinear` on every Tab keypress, and twice from `findLinearTabExitTarget` for the same containerId. There's no memoization and the nested `gatherSpottableLeaves` rebuilds sets each call. 
Existing 5-way navigation target.js / container.js uses cached descendants
Tab navigation should too especially since this re-reads `getBoundingClientRect()` for every spottable.

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+			.map((target, order) => {
+				const rect = target.getBoundingClientRect();
+				return {
+					order,
+					target,
+					x: rect.left + (rect.width / 2),
+					y: rect.top + (rect.height / 2)
+				};
+			})
+			.sort((a, b) => {
```
daniel-stoian-lgp: this seems to duplicate comesBeforeInTabOrder() logic .Please unify

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+		}
+
+		const searchContainerId = getLinearTabSearchContainerId(currentFocusedElement);
+		const linearTargets = getLinearTargetsInContainer(searchContainerId);
+		if (!linearTargets.length) {
+			return false;
+		}
+		let currentIndex = linearTargets.findIndex(({target}) => target === currentFocusedElement);
+		if (currentIndex < 0) {
+			const currentRect = currentFocusedElement.getBoundingClientRect();
```
daniel-stoian-lgp: target.js already exposes `getNearestTargetFromPosition`; please reuse

packages/spotlight/src/spotlight.js
```diff
@@ -464,30 +465,477 @@ const Spotlight = (function () {
...
+				}
+			}
+			return false;
+		}
+
+		const nextTarget = linearTargets[nextIndex].target;
+		return focusElement(nextTarget, getContainersForNode(nextTarget));
+	}
+
 	function onKeyDown (evt) {
```
daniel-stoian-lgp: `onKeyDown` was a clean aprox 15 lines before. It now branches on Tab in TWO places (the `shouldPreventNavigation` block and the main block) with overlapping semantics. It is worth  extracting `handleTab(evt)` so the 5-way path stays readable and the Tab path stays in one place.

packages/spotlight/src/spotlight.js
```diff
@@ -989,7 +1437,33 @@ const Spotlight = (function () {
...
+		isTargetInSelfOnlyContainer,
+		findLinearTabExitTargetInTargets,
+		getLinearTargetsInContainer,
+		spotLinear,
+		onAcceleratedKeyDown,
+		onBlur,
+		onFocus,
+		handleWebOSMouseEvent,
+		handleKeyboardStateChangeEvent,
+		onKeyUp
```
daniel-stoian-lgp: MAJOR CONCERN
This entire block exposes 13 internal functions on the public Spotlight API purely to make tests pass. Spotlight is consumed across the organization. Every export here is now a load-bearing API surface that we must support forever.
The last six are window event handlers that have absolutely no business being on the public API. 
If  you really need them for tests, maybe we should move them into a sibling module  and import in  tests directly. In this way, there will be no public API leak.

daniel-stoian-lgp (changes requested):
Before reading my comments, please do a self-review of the entire PR.
It was really hard to go through the PR as the PR description speaks very thin about navigation between some dropdowns, while de DoD of the user story mentions a broader scope. The PR Description should explain in technical details what all these changes in the PR do, how the new flow works etc. 
The PR description should be oriented towards an outside user, so it can be easily understood how it works and what it does

packages/spotlight/src/tests/spotlight-specs.js
```diff
@@ -202,4 +323,73 @@ describe('Spotlight', () => {
...
+		afterEach(() => {
+			removeAllContainers();
+			document.body.innerHTML = '';
+		});
+
+		test('should resolve forward handoff without throwing when two popups are open', () => {
+			setupTabHandoffScenario({openA: true, openB: true, openC: false});
+			const next = Spotlight.findLinearTabExitTarget(document.getElementById('a5'), 'popup-a', true);
+
+			expect(next == null || next.nodeType === 1).toBe(true);
```
daniel-stoian-lgp: This passes when `next` is null, i.e. when the production code FAILED to find the handoff target.

packages/spotlight/src/tests/spotlight-specs.js
```diff
@@ -202,4 +323,73 @@ describe('Spotlight', () => {
...
+
+		test('should resolve chained forward handoff for three open popups', () => {
+			setupTabHandoffScenario({openA: true, openB: true, openC: true});
+			const fromA = Spotlight.findLinearTabExitTarget(document.getElementById('a5'), 'popup-a', true);
+			const fromB = Spotlight.findLinearTabExitTarget(document.getElementById('b5'), 'popup-b', true);
+			const fromC = Spotlight.findLinearTabExitTarget(document.getElementById('c5'), 'popup-c', true);
+
+			expect(fromA == null || fromA.nodeType === 1).toBe(true);
+			expect(fromB == null || fromB.nodeType === 1).toBe(true);
+			expect(fromC == null || fromC.nodeType === 1).toBe(true);
```
daniel-stoian-lgp: the test asserts nothing about order or predictability. The expected chain should be a5 -> b1, b5 -> c1, c5 -> closeX

packages/spotlight/src/tests/spotlight-specs.js
```diff
@@ -202,4 +323,73 @@ describe('Spotlight', () => {
...
+			const next = Spotlight.findLinearTabExitTarget(document.getElementById('a5'), 'popup-a', true);
+
+			expect(next == null || next.nodeType === 1).toBe(true);
+		});
+
+		test('should hand off from popup A to popup B first option (deterministic)', () => {
+			setupDeterministicPopupHandoff();
+			const next = Spotlight.findLinearTabExitTarget(document.getElementById('a2'), 'popup-a', true);
+
+			expect(next == null || next.id === 'b1').toBe(true);
```
daniel-stoian-lgp: Test name says "(deterministic)" but the assertion still admits null. The assertion should be:
 expect(next).not.toBeNull();
expect(next.id).toBe('b1');
As written, a regression that returns null is silently OK.

Please check all your tests

packages/spotlight/src/tests/spotlight-specs.js
```diff
@@ -202,4 +323,73 @@ describe('Spotlight', () => {
...
+		});
+
+		test('should hand off reverse from popup B to popup A last option (deterministic)', () => {
+			setupDeterministicPopupHandoff();
+			const previous = Spotlight.findLinearTabExitTarget(document.getElementById('b1'), 'popup-b', false);
+
+			expect(previous == null || previous.id === 'a2').toBe(true);
```
daniel-stoian-lgp: Same as above

packages/spotlight/src/tests/spotlight-specs.js
```diff
@@ -202,4 +323,73 @@ describe('Spotlight', () => {
...
+		});
+
+		test('should fall back to outside control when target owner has no open popup', () => {
+			setupDeterministicPopupHandoff();
+			document.getElementById('ownerB').removeAttribute('aria-owns');
+			const next = Spotlight.findLinearTabExitTarget(document.getElementById('a2'), 'popup-a', true);
+
+			expect(next == null || next.id === 'ownerB').toBe(true);
```
daniel-stoian-lgp: Test name says "should fall back to OUTSIDE control", but the test asserts `next.id === 'ownerB'`. The DOM has an element literally called `outside`. So either the test title is wrong, the assertion is wrong, or both.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const left = i === 0 ? 20 : 160;
+		const node = document.querySelector(`[data-spotlight-id="${id}"]`);
+		Object.defineProperty(node, 'getBoundingClientRect', {
+			configurable: true,
+			value: () => ({left, top: 72, width: 110, height: 90, right: left + 110, bottom: 162})
+		});
+	});
+};
+
+const setupSpotlightWithMocks = () => {
```
daniel-stoian-lgp: This helper could be a big problem. It mocks every file of spotlight: container Pause, pointer etc
The tests below run against an entirely synthetic environment.
Refactoring the real modules silently invalidates the mocks here, so tests pass while production regresses.
For Tab navigation specifically, which is fundamentally about integration between focus, container restrictions, and DOM order, this is the wrong level of test. 
I prefer real container.js with a realistic DOM and dispatch real keydown events.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		setupSpotlightWithMocks();
+		const {first} = setupDocument();
+		mockPause.isPaused.mockReturnValue(false);
+
+		const {keydown} = getHandlers();
+		const ev = {keyCode: 9, shiftKey: false, target: first, preventDefault: jest.fn(), stopPropagation: jest.fn()};
+		keydown(ev);
+
+		expect(mockPointer.setPointerMode).toHaveBeenCalledWith(false);
+		expect(ev.preventDefault.mock.calls.length >= 0).toBe(true);
```
daniel-stoian-lgp: `mock.calls.length >= 0` is alway true. This ia a tautology

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const {second} = setupDocument();
+		second.focus();
+		mockPause.isPaused.mockReturnValue(false);
+
+		const {keydown} = getHandlers();
+		const ev = {keyCode: 9, shiftKey: true, target: second, preventDefault: jest.fn(), stopPropagation: jest.fn()};
+		keydown(ev);
+
+		expect(mockPointer.setPointerMode).toHaveBeenCalledWith(false);
+		expect(ev.preventDefault.mock.calls.length >= 0).toBe(true);
```
daniel-stoian-lgp: `mock.calls.length >= 0` is alway true. This is a tautology

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const {first} = setupDocument();
+		mockPause.isPaused.mockReturnValue(false);
+
+		const {keydown, mousemove} = getHandlers();
+		mousemove({target: first, clientX: 10, clientY: 10});
+
+		const ev = {keyCode: 9, shiftKey: false, target: first, preventDefault: jest.fn(), stopPropagation: jest.fn()};
+		keydown(ev);
+
+		expect(mockPointer.notifyPointerMove).toHaveBeenCalled();
```
daniel-stoian-lgp: The test name says "should skip Tab movement when pointer moved during key press" but the assertion is just that `notifyPointerMove` was called, which is guaranteed by the  mousemove() call two lines up. 
The test does not verify that Tab movement was actually skipped. 
To assert the real wish you'd check that preventDefault was NOT called and/or that `setContainerLastFocusedElement` wasn't invoked for a new target.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const res = Spotlight.findLinearTabExitTarget(document.getElementById('a2'), 'missing-popup', true);
+		expect(res).toBeNull();
+	});
+
+	test('should resolve forward handoff target from popup owner context', () => {
+		setupSpotlightWithMocks();
+		setupPopupDocument();
+
+		const res = Spotlight.findLinearTabExitTarget(document.getElementById('a2'), 'popup-a', true);
+		expect(res == null || res.nodeType === 1).toBe(true);
```
daniel-stoian-lgp: The test name says "should resolve forward handoff target" — so the expectation is `res != null && res.id === <expected>`. 
Assert that, not the universal "either null or any node."

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		expect(mockContainer.getSpottableDescendants).toHaveBeenCalled();
+		expect(mockContainer.getContainerConfig).toHaveBeenCalled();
+	});
+
+	test('should resolve reverse handoff target from popup owner context', () => {
+		setupSpotlightWithMocks();
+		setupPopupDocument();
+
+		const res = Spotlight.findLinearTabExitTarget(document.getElementById('b1'), 'popup-b', false);
+		expect(res == null || res.nodeType === 1).toBe(true);
```
daniel-stoian-lgp: Same null-OR-any-node tautology. 
assert `res.id === 'a2'` (or whatever the right answer is).

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const originalGetSpottableDescendants = mockContainer.getSpottableDescendants;
+		mockContainer.getSpottableDescendants = jest.fn((id) => {
+			if (id === 'popup-b') return [];
+			return originalGetSpottableDescendants(id);
+		});
+		mockTarget.getTargetByContainer.mockReturnValue(document.getElementById('b2'));
+
+		const res = Spotlight.findLinearTabExitTarget(document.getElementById('a2'), 'popup-a', true);
+
+		expect(res == null || ['b2', 'outside'].includes(res.id)).toBe(true);
```
daniel-stoian-lgp: Triple-tautology: passes if null, passes if 'b2', passes if 'outside'. 
The whole point of this test is the "getTargetByContainer fallback" path, so the strict assertion is `res.id === 'b2'`. 
If you allow 'outside' too, you're admitting you don't know which branch the code is taking.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+	});
+
+	test('should advance focus with spotLinear when next exists', () => {
+		setupSpotlightWithMocks();
+		const {first, second} = setupDocument();
+		first.focus();
+
+		const moved = Spotlight.spotLinear(true);
+
+		expect(typeof moved).toBe('boolean');
```
daniel-stoian-lgp: 1. `typeof moved === 'boolean'` — the function's return type annotation already says it returns a boolean. Always true.
2. `getSpottableDescendants` was called — implementation  detail, not behaviour.
3. `activeElement === second || activeElement === first`   passes whether focus moved or didn't. The test name says   "should advance focus" , so assert `activeElement === second`.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		document.body.appendChild(orphan);
+		Object.defineProperty(orphan, 'getBoundingClientRect', {
+			configurable: true,
+			value: () => ({left: Number.NaN, top: Number.NaN, width: 10, height: 10, right: Number.NaN, bottom: Number.NaN})
+		});
+		orphan.focus();
+
+		const moved = Spotlight.spotLinear(true);
+
+		expect(typeof moved).toBe('boolean');
```
daniel-stoian-lgp: Test title says "should fallback to index zero when nearest distance cannot be computed" , so the expectation  is "after this call, focus is on linearTargets[0]." 
That's the `ownerA` button. Please assert `document.activeElement.id === 'ownerA'`
`typeof moved === 'boolean'` is always true and tests literally nothing. This is a coverage hit, not a test.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		document.body.appendChild(orphan);
+		Object.defineProperty(orphan, 'getBoundingClientRect', {
+			configurable: true,
+			value: () => ({left: 158, top: 18, width: 10, height: 10, right: 168, bottom: 28})
+		});
+		orphan.focus();
+
+		const moved = Spotlight.spotLinear(true);
+
+		expect(typeof moved).toBe('boolean');
```
daniel-stoian-lgp: Same problem. The orphan is positioned at x=158 (right next to ownerB at x=160), so the nearest-target-index logic should pick ownerB and Tab forward should land on `outside`. Assert that. `typeof moved` is meaningless.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+
+	test('should resolve reverse self-only boundary via spotLinear handoff', () => {
+		setupSpotlightWithMocks();
+		setupPopupDocument();
+		document.getElementById('b1').focus();
+
+		const moved = Spotlight.spotLinear(false);
+
+		expect(typeof moved).toBe('boolean');
+		expect(document.activeElement == null || document.activeElement.nodeType === 1).toBe(true);
```
daniel-stoian-lgp: `typeof moved === 'boolean'` always passes.
`activeElement == null || activeElement.nodeType === 1` is ALWAYS true
 `document.activeElement` is either null or an Element (nodeType 1) by spec. 
This test is pure noise.
		
Real assertion: focus should be on the last spottable inside popup-a, `expect(document.activeElement.id).toBe('a2')`.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const current = document.getElementById('a2');
+		current.focus();
+		mockPause.isPaused.mockReturnValue(false);
+		const {keydown} = getHandlers();
+		const ev = {keyCode: 9, shiftKey: false, target: current, preventDefault: jest.fn(), stopPropagation: jest.fn()};
+
+		keydown(ev);
+
+		expect(ev.preventDefault).toHaveBeenCalled();
+		expect(['b1', 'ownerB', 'outside'].includes(document.activeElement.id)).toBe(true);
```
daniel-stoian-lgp: The set `['b1', 'ownerB', 'outside']` covers 3 possible handoff targets — but only one of them is the expected correct answer. The DOD says "overlay takes precedence", which means the next dropdown's first option (`b1`) is the right answer when popup-b is open.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		const current = document.getElementById('b1');
+		current.focus();
+		mockPause.isPaused.mockReturnValue(false);
+		const {keydown} = getHandlers();
+		const ev = {keyCode: 9, shiftKey: true, target: current, preventDefault: jest.fn(), stopPropagation: jest.fn()};
+
+		keydown(ev);
+
+		expect(ev.preventDefault).toHaveBeenCalled();
+		expect(document.activeElement.id === 'a2' || document.activeElement.id === 'ownerA').toBe(true);
```
daniel-stoian-lgp: `a2` (last item of popup-a) and `ownerA` (the dropdown trigger) are very different outcomes — landing on `ownerA` would CLOSE the dropdown UX flow. 
The DOD's "overlay takes precedence" rule says it must be `a2`. Assert one or the other, not "either."

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+		});
+		orphan.focus();
+		mockPause.isPaused.mockReturnValue(false);
+		const {keydown} = getHandlers();
+		const ev = {keyCode: 9, shiftKey: false, target: orphan, preventDefault: jest.fn(), stopPropagation: jest.fn()};
+
+		keydown(ev);
+
+		expect(mockPointer.setPointerMode).toHaveBeenCalledWith(false);
+		expect(ev.preventDefault.mock.calls.length >= 0).toBe(true);
```
daniel-stoian-lgp: `mock.calls.length >= 0` is always true. 

The orphan is at (22,22), nearest to `first` at (20,20). Tab forward should land on `second`. 

`expect(document.activeElement.id).toBe('second')`.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,1060 @@
...
+
+		const forward = Spotlight.findLinearTabExitTargetInTargets(targets, containerNode, 60, 90, 24, false, true);
+		const reverse = Spotlight.findLinearTabExitTargetInTargets(targets, containerNode, 180, 90, 24, false, false);
+		const linear = Spotlight.getLinearTargetsInContainer(rootContainerId);
+
+		expect(forward == null || forward.id === 'outside').toBe(true);
+		expect(reverse == null || reverse.nodeType === 1).toBe(true);
+		expect(linear.length > 0).toBe(true);
+		expect(typeof linear[0].x).toBe('number');
+		expect(typeof linear[0].y).toBe('number');
```
daniel-stoian-lgp: 5 assertions, none of them strict:
`forward == null || forward.id === 'outside'` — passes on null
`reverse == null || reverse.nodeType === 1` — passes on nullOR any element. Always true if a node was returned at all.
`linear.length > 0` — fine but doesn't check the contents.
`typeof linear[0].x === 'number'` — type-check rather than  value-check. NaN is also `typeof === 'number'`, so this wouldn't catch the broken-rect bug the file pretends to test.

Test name says "should filter targets and compute linear target list" — assert what was filtered (e.g. that `a2` was excluded because it's inside containerNode

daniel-stoian-lgp (changes requested):
check my tests comments.

packages/spotlight/src/spotlight.js
dan-ichim-lgp: There is memoization: _linearTargetsCache in handleTab caches getLinearTargetsInContainer results per containerId for one Tab press (sorted list + rects). Repeat lookups in the same keypress are cache hits. I can add a code comment to make this clearer if you want.

packages/spotlight/src/tests/spotlight-keydown-specs.js
```diff
@@ -0,0 +1,613 @@
...
+			<div data-spotlight-container="true" data-spotlight-id="${rootContainerId}">
+				<div data-spotlight-container="true"></div>
+			</div>
+		`;
+		configureContainer(rootContainerId, {selector: '.spottable'});
+
+		expect(tabNavTestHooks.getLinearTargetsInContainer(rootContainerId)).toEqual([]);
+	});
+
+	test('getLinearTargetContainerId: falls back to getContainersForNode when closest container has empty id', () => {
```
daniel-stoian-lgp, line 629: test title doesn't match its assertion

  test('getLinearTargetContainerId: falls back to getContainersForNode when closest container has empty id', () => {
      …
      expect(tabNavTestHooks.getLinearTargetContainerId(btn)).toBe('');
  });
  The wrap has data-spotlight-id="". The title says "falls back to getContainersForNode" , but the assertion is the empty string, which could be returned by either path (the primary path returns getContainerId(wrap) which is "", or the fallback returns last(getContainersForNode(btn)) which could also be ""). The test doesn't actually verify which code path is executed. 
Either:
  a) Mock getContainersForNode to return [rootContainerId, 'sentinel'] and assert 'sentinel' (proves fallback ran), or
  b) Rename the test to "returns empty when the closest container has empty id" (more honest about what's being checked).

packages/spotlight/CHANGELOG.md
```diff
@@ -15,6 +15,7 @@ The following is a curated list of changes in the Enact spotlight module, newest
 ### Fixed
 
 - `spotlight` to focus visible children of elements which do not have `overflow:true`
+- `spotlight` keyboard focus flow for open dropdown popups so Tab/Shift+Tab traversal is predictable between open lists and can continue to nearby controls after popup navigation
```
daniel-stoian-lgp: let's move the entry to a new unreleased version

packages/spotlight/src/spotlight.js
```diff
@@ -989,7 +1351,29 @@ const Spotlight = (function () {
...
+		getLinearTargetsInContainer,
+		runWithLinearTargetsCache: (fn) => {
+			_linearTargetsCache = new Map();
+			try {
+				return fn();
+			} finally {
+				_linearTargetsCache = null;
+			}
+		},
+		spotLinear
```
daniel-stoian-lgp: let's move all these helpers into another file (maybe tabTraversal.js? ), in order to avoid spotlight API leak. this will allow to use them in tests, but @enact/spotlight's exports stay clean

daniel-stoian-lgp (commented):
check my comments

daniel-stoian-lgp:
The function extraction looks good. 
However, the tests still reach into internals the same way they did with _tabNavTestHooks. They just import the new tabTraversal mutable export instead. So you kept a test-only export, and it now relies on import-order side effects (tabNavTestHooks is null until spotlight.js's immediately invoked function expression runs createTabTraversal). 

Could you do the following:

  1. Export the pure helpers as plain named exports from tabTraversal.js (**_comesBeforeInTabOrder, getLinearTabSearchContainerId, getLinearTargetContainerId, getLinearTargetsInContainer, isTargetInSelfOnlyContainer, resolveTargetToOpenPopupItem, findLinearTabExitTarget, findLinearTabExitTargetInTargets_**). 
Then have the helper tests import {…} from '../tabTraversal' directly and drop the export let tabTraversal / tabTraversal = api singleton entirely. _createTabTraversal_ should return only what production consumes (spotLinear, plus the cache wrapper if we keep it).
  2. In the tests, rename the two mislabeled describe blocks. 'Spotlight spotLinear (integration)' and 'Spotlight Tab helpers (integration)' aren't integration tests. They call internal functions directly. They should read as unit tests (example 'tabTraversal — spotLinear', 'tabTraversal — pure helpers'). The genuine integration coverage already lives in 'Spotlight Tab key dispatch (integration)' and 'Tab keydown (integration …)', which dispatch real keydown events.
  3. Drop the redundant spotLinear unit tests that duplicate the real dispatch tests. Specifically:
    - 'should advance focus to the next spottable element' (calls spotLinear(true) → second) duplicates 'should move focus forward on Tab'.
    - 'should return false at the end of the linear list' duplicates 'should not call preventDefault when Tab reaches end of a non-self-only linear list'.
For the remaining spotLinear cases that are real edge paths (r**estoreFocus bootstrap, nearest-target fallback, self-only-no-exit**), I prefer using them through dispatchTab(handlers.keydown) so we're testing the path users actually do, rather than the internal entry point.

daniel-stoian-lgp (approved):
LGTM. thank you for your effort

alexandrumorariu (approved):
LGTM

PR #3415, opened 2026-05-06
NXT-12921: Fix useScroll scroll animation

alexandrumorariu:
### Checklist

I fixed scrollEnd feature to work with scroll by key.

### Resolution
Scroll by key was not treated as an accumulating scroll and was not stopped by the first condition. I added a new key for the object called `keyScroll` that is applied on `onKeyUp`. This will make sure we handle the scroll properly.  I also added a check for `keyPress` on the exit condition that prevents `onScrollEnd` to trigger while we are scrolling.

### Additional Considerations

### Links
NXT-12921

### DIFF
--- a/packages/ui/useScroll/tests/useScroll-specs.js
+++ b/packages/ui/useScroll/tests/useScroll-specs.js
@@ -400,6 +400,7 @@ describe('useScroll', () => {
 				scrollMode: 'native',
 				rtl: true,
 				...mocks,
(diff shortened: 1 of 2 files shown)


daniel-stoian-lgp (approved):
LGTM
The performance on TV board is not perfect, but it is improved

ion-andrusciac-lgp (approved):

PR #3416, opened 2026-05-06
NXT-13542: Removed guard for spotlight disabled

daniel-stoian-lgp:
### Checklist

Spottable  tracks a flag called prevSpotlightDisabled. It's used inside SpottableCore.didUpdate to detect a spotlightDisabled: true → false transition and, when it sees one, restores Spotlight focus to the container.
For that detection to work, prev must reliably mean "the value of spotlightDisabled in the previous render". In NXT-8994 https://github.com/enactjs/enact/pull/3379 , code only refreshed prev when the prop changed:

```
  if (context.spotlightDisabled !== spotlightDisabled) {
      context.prevSpotlightDisabled = context.spotlightDisabled;
      context.spotlightDisabled = spotlightDisabled;
  }
```

  Once a Spottable component went through one true → false transition (which every media-control button does when the controls auto-show), prev was set to true and never cleared again. From that render onward, every render saw prev === true and re-ran the focus restoration.

In normal use you don't notice, because Spotlight.getCurrent() already returns one of the Spotlight buttons, so the `else if (!Spotlight.getCurrent()) ` branch is skipped. But the Storybook controls input is not a Spotlight element. That means Spotlight.getCurrent() returns null while the user is typing. Each keystroke triggered:

```
   else if (!Spotlight.getCurrent()) {
      Spotlight.focus(containerId);   // ← focuses into the VideoPlayer
  }
```

### Resolution

### Additional Considerations
There is no changelog as the component is private
Added unit test for the affected scenario

### Links
NXT-13542

### Comments

### DIFF
--- a/packages/spotlight/Spottable/tests/useSpot-specs.js
+++ b/packages/spotlight/Spottable/tests/useSpot-specs.js
@@ -294,4 +294,79 @@ describe('useSpottable', () => {
 			expect(spy).toHaveBeenCalledTimes(expected);
 		});
 	});
+
+	describe('focus restoration on spotlightDisabled transition', () => {
+		let focusSpy;
+		let isPausedSpy;
+		let pointerModeSpy;
+		let setPointerModeSpy;
+
+		beforeEach(() => {
+			focusSpy = jest.spyOn(Spotlight, 'focus').mockImplementation(() => true);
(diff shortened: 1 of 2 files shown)


ion-andrusciac-lgp (approved):

vJIYEv (approved):
LGTM

PR #3424, opened 2026-05-18
NXT-14097: Fixed wrong scroll position and screen freeze in Scroller

vJIYEv:
### Checklist

Scroller stutters when scrolled by long press.

### Resolution

1. Slow scroll speed during arrow key long press
- Replaced with a dynamic step of 10% of the remaining distance (minimum 8px), so scrolling starts fast and eases in as it approaches the target.

2. Screen clipped when reversing scroll direction during long press
- Added `cancelAnimationFrame` guards to cancel.
- Replaced the bounds check with a target-reached check.
- Added a condition to stop animation: if 1s has elapsed since animation started.

### Additional Considerations

### Links
NXT-14097

### Comments

### DIFF
--- a/packages/ui/Scroller/ScrollerBasic.js
+++ b/packages/ui/Scroller/ScrollerBasic.js
@@ -158,7 +158,7 @@ class ScrollerBasic extends Component {
 			const scrollTop = directionY > 0 ? node.scrollTop < top : node.scrollTop > top;
 
 			// Stop animating if the target is reached or 1s has elapsed since animation started
-			if (!scrollTop && !scrollLeft || elapsed > 1) {
+			if ((!scrollTop && !scrollLeft) || elapsed > 1) {
 				// Fallback: if timed out before reaching target, jump there with smooth scroll
 				if (elapsed > 1 && (node.scrollLeft !== left || node.scrollTop !== top)) {
 					node.scrollTo({top, left, behavior: 'smooth'});
--- a/packages/ui/Scroller/ScrollerBasic.js
+++ b/packages/ui/Scroller/ScrollerBasic.js
@@ -133,6 +133,7 @@ class ScrollerBasic extends Component {
 		if (platform.chrome && smoothBehavior && repeat) {
 			this.animateScroll(this.getRtlPositionX(left), top, node);
 		} else {
+			if (node.scrolling && this.scrollAnimationId) window.cancelAnimationFrame(this.scrollAnimationId);
 			node.scrollTo({left: this.getRtlPositionX(left), top, behavior});
 		}
 	}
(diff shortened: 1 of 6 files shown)


ion-andrusciac-lgp (approved):

packages/ui/useScroll/useScroll.js
```diff
@@ -945,9 +945,13 @@ const useScrollBase = (props) => {
 				scrollByPage(ev.keyCode);
 			}
 		} else {
+			const isArrowKey = mutableRef.current.lastInputType === 'arrowKey';
+			const isPageKey = mutableRef.current.lastInputType === 'pageKey';
```
daniel-stoian-lgp: why are these 2 variables initialized now before props.preventScroll?.(ev)  ?  It might be a problem if the consumer's preventScroll mutates lastInputType

packages/ui/Scroller/ScrollerBasic.js
```diff
@@ -133,39 +133,44 @@ class ScrollerBasic extends Component {
...
-				this.scrollAnimationId = window.requestAnimationFrame(animateScroll);
-			}
+			const horizontalScrollFactor = Math.max(Math.abs(left - node.scrollLeft) * 0.1, 8);
+			const verticalScrollFactor = Math.max(Math.abs(top - node.scrollTop) * 0.1, 8);
+
+			node.scrollBy({top: directionY * verticalScrollFactor, left: directionX * horizontalScrollFactor, behavior: 'instant'});
+			this.scrollAnimationId = window.requestAnimationFrame(animateScroll);
 		};
 
 		this.scrollAnimationId = window.requestAnimationFrame(animateScroll);
```
daniel-stoian-lgp, line 176: i see here a potential issue. This overwrites this.scrollAnimationId without cancelling the previous frame.
I would suggest to cancel any existing id at the top of animateScroll (exactly as lines 136-139)

What do you think?

`animateScroll (left, top, node) {
		**if (this.scrollAnimationId) {
			window.cancelAnimationFrame(this.scrollAnimationId);
			this.scrollAnimationId = null;
		}**

......`

packages/ui/Scroller/ScrollerBasic.js
```diff
@@ -133,39 +133,44 @@ class ScrollerBasic extends Component {
...
-			// Check if we reached the scroll bounds and cancel the animation
-			if (
-				top > this.scrollBounds.maxTop && node.scrollTop === this.scrollBounds.maxTop ||
-				left > this.scrollBounds.maxLeft && node.scrollLeft === this.scrollBounds.maxLeft ||
-				top < 0 && node.scrollTop === 0 ||
-				left < 0 && node.scrollLeft === 0
-			) {
-				window.cancelAnimationFrame(this.scrollAnimationId);
+			// Stop animating if the target is reached or 1s has elapsed since animation started
+			if (!scrollTop && !scrollLeft || elapsed > 1) {
```
daniel-stoian-lgp:
```suggestion
			if ((!scrollTop && !scrollLeft) || elapsed > 1) {
```

daniel-stoian-lgp (commented):
check my comments and apply the fixes also on the other PR.
Thank you!

packages/ui/Scroller/ScrollerBasic.js
daniel-stoian-lgp: after consultations with Ion, he convinced me that this change request might be hurtfull. I will resolve the comment

daniel-stoian-lgp (approved):

PR #3427, opened 2026-05-20
NXT-9871: Fixed overscroll effect VirtualList items have margin

vJIYEv:
### Checklist

The overscroll effect did not trigger when VirtualList items had CSS margin.

### Resolution
When item margins are first detected, `scrollHeight` (vertical) or `scrollWidth` (horizontal) was not updated alongside `maxTop`/`maxLeft`.
This broke the overscroll trigger condition `curPos >= maxPos`, because `scrollHeight - clientHeight !== maxTop`.
Fixed by incrementing both fields by the margin sum and calling `setContainerSize()`.

### Additional Considerations
Additionally, `itemSize` and `itemSizes` props were mutated in-place during window resize handling in `useScroll`, which could cause unexpected behavior when the same object reference was reused.
 Fixed `itemSize` by creating a new object via spread and also fixed `itemSizes` by using `map` and storing the result in a ref (`scaledItemSizesRef`) so re-renders reflect the scaled values without mutating the original array.

### Links
NXT-9871

### Comments

### DIFF
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -6,8 +6,10 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
-- `ui/VirtualList` overscroll effect at the bottom of the list
 - `ui/Scroller` jumping back when scrolled by long press
+- `ui/VirtualList` overscroll effect at the bottom of the list
+- `ui/VirtualList.VirtualList` overscroll effect at the bottom of the list
+- `ui/VirtualList.VirtualGridList` overscroll effect at the bottom of the list
 
 ## [5.5.0] - 2026-05-08
 
--- a/packages/ui/CHANGELOG.md
+++ b/packages/ui/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 The following is a curated list of changes in the Enact ui module, newest changes on the top.
 
+## [unreleased]
+
+### Fixed
+
+- `ui/VirtualList` overscroll effect at the bottom of the list
+
 ## [5.5.0] - 2026-05-08
 
 ### Added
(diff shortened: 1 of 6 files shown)


packages/ui/CHANGELOG.md
```diff
@@ -6,6 +6,7 @@ The following is a curated list of changes in the Enact ui module, newest change
 
 ### Fixed
 
+- `ui/VirtualList` overscroll effect at the bottom of the list
```
daniel-stoian-lgp: shouldn't we mention also VirtualGridList?

packages/ui/useScroll/useScroll.js
```diff
@@ -176,6 +176,7 @@ const useScrollBase = (props) => {
 	const [riRatio, setRiRatio] = useState(ri.scale(1));
 	const [originalItemSize, setOriginalItemSize] = useState(props.itemSize);
 	const [itemSize, setItemSize] = useState(props.itemSize);
+	const scaledItemSizesRef = useRef(null);
```
daniel-stoian-lgp, line 179: Now that scaled itemSizes are cached in scaledItemSizesRef, line 423 will keep returning the cached array forever, even if the parent component later passes a different props.itemSizes reference.

Could you add a source-tracking ref next to scaledItemSizesRef and clear the cache when the source reference changes?
Let me know if you see any problems with this suggestion.

  const scaledItemSizesRef = useRef(null);
  const scaledItemSizesSourceRef = useRef(props.itemSizes);

  if (scaledItemSizesSourceRef.current !== props.itemSizes) {
      scaledItemSizesRef.current = null;
      scaledItemSizesSourceRef.current = props.itemSizes;
  }

packages/ui/CHANGELOG.md
vJIYEv: I updated changelog.

packages/ui/useScroll/useScroll.js
vJIYEv: Thank you. I applied your suggestion.

daniel-stoian-lgp (approved):

alexandrumorariu (approved):
LGTM

PR #3429, opened 2026-05-27
NXT-13027: [i18n] Investigate usage of  useSyncExternalStore

dan-ichim-lgp:
### Checklist

 Refactor useI18n to subscribe to the i18n store via React's useSyncExternalStore, replacing the manual useState + setContext(locale, setState) pattern.

### Resolution
The I18n class now implements the external store contract (subscribe, getSnapshot, getServerSnapshot) and maintains an internal snapshot (className, loaded, locale, rtl) updated through _updateSnapshot. useI18n subscribes via useSyncExternalStore instead of passing setState into setContext.

Impact: subscribing to the i18n store is now safe under concurrent rendering (no tearing between components), with SSR support via getServerSnapshot. Observable behavior for consumers (className, loaded, locale, rtl, updateLocale) is unchanged.

### Additional Considerations

### Links
NXT-13027

### Comments

### DIFF
--- a/packages/i18n/CHANGELOG.md
+++ b/packages/i18n/CHANGELOG.md
@@ -9,15 +9,6 @@ No significant changes.
 ## [5.4.4] - 2026-05-22
 
 No significant changes.
-## [unreleased]
-
-### Changed
-
-- `i18n/I18nDecorator` to use `useSyncExternalStore` for subscribing to the ilib locale store, replacing the manual `useState` + `setContext` pattern for safer concurrent rendering
-
-### Fixed
-
-- `i18n/I18nDecorator` to handle failed async resource requests instead of silently ignoring them
 
 ## [5.5.0] - 2026-05-08
 
--- a/packages/i18n/CHANGELOG.md
+++ b/packages/i18n/CHANGELOG.md
@@ -9,6 +9,15 @@ No significant changes.
 ## [5.4.4] - 2026-05-22
 
 No significant changes.
+## [unreleased]
+
+### Changed
+
+- `i18n/I18nDecorator` to use `useSyncExternalStore` for subscribing to the ilib locale store, replacing the manual `useState` + `setContext` pattern for safer concurrent rendering
+
+### Fixed
+
+- `i18n/I18nDecorator` to handle failed async resource requests instead of silently ignoring them
 
 ## [5.5.0] - 2026-05-08
 
--- a/packages/i18n/I18nDecorator/I18n.js
+++ b/packages/i18n/I18nDecorator/I18n.js
@@ -26,7 +26,6 @@ class I18n {
 	}) {
 		this._locale = null;
 		this._ready = sync;
-		this._onLoadResources = () => {};
 		this.loadResourceJob = new Job(state => this._updateSnapshot(state));
 		this._listeners = new Set();
 		this._snapshot = {
--- a/packages/i18n/I18nDecorator/I18n.js
+++ b/packages/i18n/I18nDecorator/I18n.js
@@ -1,5 +1,4 @@
 import {on, off} from '@enact/core/dispatcher';
-import {Job} from '@enact/core/util';
 
 import {isRtlLocale, updateLocale} from '../locale';
 import {createResBundle, setResBundle} from '../src/resBundle';
@@ -10,6 +9,9 @@ import {onWindowFocus} from './windowFocus';
 
 /**
  * Manages i18n resource loading.
+ *
+ * Implements the external store contract expected by `useSyncExternalStore`:
+ * `subscribe`, `getSnapshot`, and `getServerSnapshot`.
  *
  * @class I18n
  * @private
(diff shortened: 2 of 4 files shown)


packages/i18n/CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
...
+## [unreleased]
+
+### Changed
+
+- `i18n/I18nDecorator` to use `useSyncExternalStore` for subscribing to the ilib locale store, replacing the manual `useState` + `setContext` pattern for safer concurrent rendering
+
+### Fixed
+
+- `i18n/I18nDecorator` to handle failed async resource requests instead of silently ignoring them
+
```
daniel-stoian-lgp: since i18n class is private and the public api of i18ndecorator is not affected, i think we do not need a changelog entry, at least for the changed section

packages/i18n/I18nDecorator/I18n.js
```diff
@@ -33,20 +40,81 @@ class I18n {
...
+		this._listeners.forEach(l => l());
+	}
+
+	/**
+	 * Updates the snapshot and notifies all subscribers.
+	 *
+	 * Notifications are suppressed before `load()` is called (i.e. during render
+	 * in async mode) so that `setLocale` in render does not trigger store updates
+	 * mid-render. In sync mode `_ready` is already `true`, so `useSyncExternalStore`
+	 * handles the concurrent-rendering tearing check automatically.
```
daniel-stoian-lgp: The jsdoc is misleading compared to what the function actually does. A more accurate description wuould be

```suggestion
Updates the snapshot and notifies subscribers when the store is ready.

`_ready` reflects whether `load()` has run. In async mode it starts `false`,
which suppresses notifications for the synchronous `setLocale` call made
during the first render (the initial snapshot is published by the first
`getSnapshot` read instead). In sync mode `_ready` is `true` from
construction, so updates notify immediately; `useSyncExternalStore` is
responsible for the tearing check under concurrent rendering.
```

packages/i18n/I18nDecorator/I18n.js
```diff
@@ -122,43 +189,49 @@ class I18n {
...
+					...this._snapshot,
+					loaded: true,
+					locale
+				});
 
-			this.loadResourceJob.promise(resources);
+				if (typeof console !== 'undefined') {
+					// eslint-disable-next-line no-console
+					console.error('[I18n] Failed to load resources for locale', locale, error);
+				}
```
daniel-stoian-lgp: The catch path does {...this._snapshot, loaded: true, locale}. That keeps the old className and rtl but writes the new locale. After a failed RTL switch the app will render LTR classes against an ar-SA locale. So we need to address this as well.
At the end also add a test for this scenario

packages/i18n/I18nDecorator/I18n.js
```diff
@@ -122,43 +189,49 @@ class I18n {
...
-			this._onLoadResources(state);
+			this._updateSnapshot({
+				className,
+				loaded: true,
+				locale,
+				rtl
+			});
 		} else {
-			const resources = Promise.all([
+			Promise.all([
```
daniel-stoian-lgp: The old code wrapped the async chain in a Job.  The new code just fires a fresh Promise.all from setLocale. If locale changes twice in quick succession, both chains run, and whichever resolves last wins, even if it's the older one. 
It is worth restoring some form of "latest-wins" guard (e.g., capture this._latestLoadToken in constructor, set it in loadResources and   clear _latestLoadToken = null in unload function).

Try to add a test for this scenario

daniel-stoian-lgp (commented):
Before addressing the details mentioned in the comments  below, I need to mention that the DoD of this task framed this as an investigation — "Check if we can use useSyncExternalStore in Enact… For example: useI18n.js." The expected output is a recommendation and (if positive) the smallest possible change that swaps in the hook. 

This PR is closer to a redesign of I18n than an investigative spike, and several changes touch behavior that wasn't part of the ticket.
Each of the extra changes from the PR need their own review, their own tests, and their own ticket so the diff tells a single story.

  Suggested path forward

  1. Reduce this PR to the minimal useSyncExternalStore adoption: just the subscribe/getSnapshot/getServerSnapshot surface and the hook swap in useI18n. Keep Job, keep _onLoadResources, keep the // TODO: catch comment, keep
  SSR shape. 
 2. File the other findings as separate tickets so they get their own review and changelog entries:
    - Failed async resource handling policy
    - SSR rtl computation in getServerSnapshot
    - onLoadResources deprecation/removal
    - etc

packages/i18n/I18nDecorator/I18n.js
```diff
@@ -24,7 +27,14 @@ class I18n {
 		this._locale = null;
 		this._ready = sync;
 		this._onLoadResources = () => {};
```
daniel-stoian-lgp: line 29 is dead code now. pelase remove

packages/i18n/I18nDecorator/useI18n.js
daniel-stoian-lgp: [onLoadResources]  property is not wired anywhere. please remove it

daniel-stoian-lgp (approved):

ion-andrusciac-lgp (approved):

PR #3430, opened 2026-05-28
NXT-13086: Migrated ui/Transition to functional component + added small improvements

daniel-stoian-lgp:
### Checklist

### Resolution
1. Transition.module.less:  Removed will-change: transform / opacity / height / max-width from five places. These stylings were added in 2017.
 What will-change is for: it's a hint that tells the browser "this property is about to animate, please prepare a GPU layer for it now so the animation is smooth."
The problem: these hints were set on every Transition all the time, even ones that were just sitting there idle. So the browser was permanently keeping a dedicated GPU layer for each Transition, in case it might animate. Lots of mounted Transitions = lots of wasted GPU memory. The browser is now capale to do it on its own automatically, exactly when it's needed. 

2. migrated Transition to functional component
static contextType = ResizeContext  --> const ctx = use(ResizeContext)
constructor (state init) --> useState(() => initialValue) 
static getDerivedStateFromProps --> An useRef comparison + conditional setState during render
componentDidMount -->  useEffect(() => {...}, []) 
componentDidUpdate  --> useEffect(() => {...}, [deps]) + useLayoutEffect . useLayoutEffect runs synchronously after DOM commit, before paint. scrollHeight reads need a committed layout.
shouldComponentUpdate :  Dropped
Instance methods   --> useCallback or inline functions
Instance fields (this.x) -->  useRef(initial) — read/write via .current  

3. TransitionBase
 Before: innerStyle always allocated a fresh { } and style unconditionally spread the input style into a new object for the clip branch. With many Transitions re-rendering, this adds garbage collectionn for no functional benefit.
After:
  - innerStyle returns a hoisted EMPTY_STYLE constant when neither the clip-horizontal nor the fade/slide-duration branch applies.
  - style only spreads style when it's actually provided (style ? {...style, overflow: 'hidden'} : {overflow: 'hidden'}).

### Additional Considerations

### Links
NXT-13086

### Comments


ion-andrusciac-lgp (approved):

packages/ui/Transition/Transition.module.less
```diff
@@ -12,13 +12,11 @@
 
 		.inner {
 			transition-property: transform;
-			will-change: transform;
```
dan-ichim-lgp, line 15: will-change was removed for transform, opacity, height, and max-width. Was this intentional? It can affect animation smoothness on TV targets. If this is a deliberate optimization, please mention it in the PR description or CHANGELOG so reviewers know it was considered.

packages/ui/Transition/tests/Transition-specs.js
dan-ichim-lgp, line 1: missing show/hide/show coverage => Consider adding a test for visible: true → false → true with type="clip" (with animation enabled). That’s the main accordion/header path and would increase confidence beyond the noAnimation tests.

dan-ichim-lgp (approved):
LGTM — functional migration preserves existing behavior, tests pass, and the React 19 patterns (use, useLayoutEffect, documented render-time state update) look correct. please confirm will-change removal was intentional.

packages/ui/Transition/tests/Transition-specs.js
daniel-stoian-lgp: added the test

packages/ui/Transition/Transition.module.less
daniel-stoian-lgp: added changelog

PR #3431, opened 2026-06-04
NXT-13084: Modified `resolution` behavior

ion-andrusciac-lgp:
### Checklist

Modified `resolution` behavior

### Resolution
- Added a new function `getShouldScale` which determines if the font size should be scaled 
- Modified `getClosestResolutionType` function, so now the algorithm will also consider and the aspect ratio ratio of the screen, not only the spatial distance between resolutions
- Modified the behavior for the scaling in the `portrait` mode. Now the content will be scaled based on the current resolution size (only if `orientationHandling` is set to `scale`), as it already happens and in `landscape` mode.
- Changed the scale condition so scaling now applies only when the workspace is smaller than the matched screen type in both dimensions (previously either), to avoid over-scaling at intermediate resolutions.

### Additional Considerations

### Links
NXT-13084

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,13 +2,13 @@
 
 The following is a curated list of changes in the Enact project, newest changes on the top.
 
-## [Unreleased]
+## [unreleased]
 
 ### Changed
 
 - `spotlight/SpotlightRootDecorator` to use a functional component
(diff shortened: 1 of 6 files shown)


daniel-stoian-lgp (approved):
LGTM

alexandrumorariu (approved):
LGTM

PR #3434, opened 2026-06-17
NXT-12133: Fixed native scroll animation exactly on target instead of clipping it

daniel-stoian-lgp:
### Checklist

* [x] I have read and understand the [contribution guide](https://enactjs.com/docs/developer-guide/contributing/)
* [x] A [CHANGELOG entry](https://enactjs.com/docs/developer-guide/contributing/changelogs/) is included

When using stickTo = start for limestone scroller or VirtualList, some focused items were clipped

### Resolution
The native-mode `animateScroll` in both `ScrollerBasic` and `VirtualListBasic` advanced the scroll position by a fixed step each frame (`scrollFactor`, ~gridSize/12  for VirtualList; max(remaining*0.1, 8) for Scroller) and stopped on the first frame that reached or passed the target, without clamping the final step.
The overshoot was usually invisible because it landed within the same item. But when a focused item is anchored to the start edge (e.g. `stickTo="start"`), the overshoot pushed the item's leading edge past the viewport's start edge, clipping it. 
Clamp each frame's step to the remaining distance so the final frame lands exactly on the target, eliminating the overshoot.

### Additional Considerations

### Links
NXT-12133

### Comments

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,11 @@ The following is a curated list of changes in the Enact project, newest changes 
 - `ui/resolution` to not resize too often for some intermediate resolutions
 - `ui/Transition` to remove static `will-change` declarations, reducing persistent GPU layer promotion and memory usage on TV targets
 
+### Fixed
+
+- `ui/Scroller` native scroll animation to land exactly on the target position instead of overshooting by up to one animation step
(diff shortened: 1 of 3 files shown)


ion-andrusciac-lgp (approved):

dan-ichim-lgp (approved):
LGTM

PR #3437, opened 2026-06-24
NXT-10275: Added prerender support for ui/FloatingLayer

daniel-stoian-lgp:
### Checklist

Portals are not supported by React's server renderer, so FloatingLayer renders null during a renderToString prerender pass and its content never appears in prerendered HTML. This adds an opt-in prerender prop that lets floating-layer content be prerendered.

When prerender is true, the content is rendered inline on the first render (the server prerender pass and the initial client render) and then relocated into the floating layer via createPortal once mounted. The inline first render matches the server markup, so hydration succeeds without a mismatch; the portal move happens in a post-mount commit.

When prerender is false (default), behavior is unchanged. Existing popups are unaffected.

### Resolution

### Additional Considerations

### Links
NXT-10275

### Comments

### DIFF
--- a/packages/ui/FloatingLayer/FloatingLayer.js
+++ b/packages/ui/FloatingLayer/FloatingLayer.js
@@ -99,6 +99,25 @@ class FloatingLayerBase extends Component {
 		open: PropTypes.bool,
 
 		/**
+		 * Enables prerendering support.
+		 *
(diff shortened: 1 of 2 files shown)


alexandrumorariu (approved):
LGTM

PR #3438, opened 2026-06-25
NXT-13085: Investigate performance issues

dan-ichim-lgp:
### Checklist

Refactor @enact/core to use React 18/19 patterns where they improve render-path performance and hook correctness
Primary goals:

core/kind — cache defaultPropKeys at component creation instead of calling Object.keys(defaultProps) on every functional render
core/usePublicClassNames — memoize merged CSS class maps with useMemo

### Resolution

core/kind (kind.js, functionalKind.js, util.js)

Hoist defaultPropKeys and handlerKeys before the functional/class branches
Functional render path applies defaults via applyDefaultProps with the cached key list
Extract bindInlineHandlers for the synchronous inline path
Align kind.inline with functionalKind: snapshot props before injecting handlers so each handler receives the same base props
functionalKind uses React use() for context (consistent with kind)
core/usePublicClassNames

Wrap merge logic in useMemo with deps [componentCss, customCss, publicClassNames]
Use shared normalizePublicClassNames from core/util
Memoization applies when those inputs are referentially stable (typical: module-level CSS maps, publicClassNames: true or a string)
core/kind/styles.js

Use normalizePublicClassNames at kind factory time (DRY with the hook)
core/util

Add public applyDefaultProps and normalizePublicClassNames
setDefaultProps delegates to applyDefaultProps
usePrevious uses useRef instead of updating state during render (same return values, no extra render on value change)
Impact: Same public behavior for normal usage; fewer allocations on kind functional render paths and fewer redundant CSS merges when hook inputs are stable. Not a breaking API change — new util exports are additive. Internal fix: kind.inline handler injection.

### Additional Considerations

### Links
NXT-13085

### Comments


ion-andrusciac-lgp (approved):

packages/core/util/util.js
```diff
@@ -373,19 +418,17 @@ const checkPropTypes = (component, props, prevProps) => {
  * @public
  */
 const usePrevious = (value) => {
-	const [previousTrackedValue, setPreviousTrackedValue] = useState(value);
-	const [previousValue, setPreviousValue] = useState(value);
+	const ref = useRef(value);
+	const previous = ref.current; // eslint-disable-line react-hooks/refs -- intentional previous-value tracking
 
-	if (value !== previousTrackedValue) {
-		setPreviousTrackedValue(value);
```
daniel-stoian-lgp, line 380: I see a problem here:
The new useRef based hook returns the immediately-prior render's value instead.

Example: values a, b, then a re-render with value still b:

Old: returns a, a, a (holds the pre-change value)
New: returns a, a, b (ref overwritten to b on render 2, so render 3 reads b)

packages/core/CHANGELOG.md
```diff
@@ -2,6 +2,24 @@
...
+
+- `core/kind` to cache `defaultPropKeys` and apply defaults via `applyDefaultProps` on the functional render path
+- `core/kind.functionalKind` to use React `use()` for context
+- `core/usePublicClassNames` to use `normalizePublicClassNames`
+- `core/util.usePrevious` to use `useRef` instead of updating state during render
+- `core/util.setDefaultProps` to delegate to `applyDefaultProps`
+
+### Fixed
+
+- `core/kind` `inline` handler injection to use a props snapshot so each handler receives the same base props
```
daniel-stoian-lgp: make sure to add only public api changes in the changelog

packages/core/kind/functionalKind.js
```diff
@@ -145,9 +146,12 @@ const functionalKind = (config) => {
 		return useRender(props, context); // eslint-disable-line react-hooks/rules-of-hooks
 	};
 
+	const defaultPropKeys = defaultProps ? Object.keys(defaultProps) : null;
+	const handlerKeys     = handlers     ? Object.keys(handlers)     : null;
+
 	const Component = function (props) {
 		// Hooks must always be called unconditionally and in the same order.
-		const ctx = useContext(contextType);
+		const ctx = use(contextType);
```
daniel-stoian-lgp: @alexandrumorariu , i remember we had problems with using  use() hook previously in kind. do you remember the exact problem?

daniel-stoian-lgp (changes requested):
see my comments.

packages/core/kind/functionalKind.js
dan-ichim-lgp: @daniel-stoian-lgp  in kind is already use() hook.
https://github.com/enactjs/enact/blob/master/packages/core/kind/kind.js#L159
This change can be reverted; I made it for consistency.

alexandrumorariu: Here is the reason for useContext: 
use() is not the same as useContext(). The React docs are explicit: use() can read a Context or a Promise. When passed a Promise, it suspends — and suspension during SSR with renderToString does not work the same way as in a streaming (renderToPipeableStream) environment.

The risk surfaces like this:
js

// If contextType is accidentally a Promise-like object, or if
// use() is passed something that triggers Suspense...
const ctx = use(contextType); // throws a Promise → breaks renderToString

With useContext(contextType) this can't happen — it only accepts a Context object and never suspends.

The original class branch also used static contextType which is immune to this entirely.
The fix

Replace use with useContext in the rewritten component:
js

// Before
import {createContext, use} from 'react';
// ...
const ctx = use(contextType);

// After
import {createContext, useContext} from 'react';
// ...

alexandrumorariu (approved):
LGTM

daniel-stoian-lgp (approved):

PR #3443, opened 2026-07-09
NXT-14658: Changes for functionalKind to work with react compiler

alexandrumorariu:
### Checklist

* [X] I have run automated testing and it is passed

I made changes for functionalKind to work with react compiler.

### Resolution
I disabled react compiler for functinalKind.

### Additional Considerations
We should recreate functional kind to work with react compiler.

### Links
NXT-14658

### DIFF
--- a/packages/core/kind/functionalKind.js
+++ b/packages/core/kind/functionalKind.js
@@ -150,6 +150,8 @@ const functionalKind = (config) => {
 	const handlerKeys     = handlers     ? Object.keys(handlers)     : null;
 
 	const Component = function (props) {
+		"use no memo"; // Added to remove react-compiler optimizations
+
 		// Hooks must always be called unconditionally and in the same order.
 		// useContext only accepts Context and never suspends (unlike use(), which can suspend on Promises and break SSR).
 		const ctx = useContext(contextType);
--- a/packages/core/kind/functionalKind.js
+++ b/packages/core/kind/functionalKind.js
@@ -143,8 +143,10 @@ const functionalKind = (config) => {
 		if (renderStyles)   props = renderStyles(props, context);
 		if (renderComputed) props = renderComputed(props, context);
 
-		return props;
+		return useRender(props, context); // eslint-disable-line react-hooks/rules-of-hooks
 	};
+
+	const useRenderKind = (props, context) => renderKind(props, context);
 
 	const defaultPropKeys = defaultProps ? Object.keys(defaultProps) : null;
 	const handlerKeys     = handlers     ? Object.keys(handlers)     : null;
@@ -154,6 +156,12 @@ const functionalKind = (config) => {
 		// useContext only accepts Context and never suspends (unlike use(), which can suspend on Promises and break SSR).
 		const ctx = useContext(contextType);
 		const boundHandlers = useHandlers(handlers, props, ctx);
+
+		// The functional branch below renders through this wrapper rather than `renderKind` directly.
+		// `render` is user-supplied and may call hooks, so the render path taken during an actual React
+		// render must itself be a hook. The `use` prefix is what tells React Compiler (and the
+		// rules-of-hooks lint) that this call may run hooks.
+
 
 		// Merge incoming props with bound handlers.
 		let merged = {
@@ -167,8 +175,7 @@ const functionalKind = (config) => {
 
 		checkPropTypes(Component, merged);
 
-		renderKind(merged, ctx);
-		return useRender(merged, ctx);
+		return useRenderKind(merged, ctx);
 	};
 
 	if (name)         Component.displayName = name;
@@ -183,9 +190,8 @@ const functionalKind = (config) => {
 	// of the React render cycle (e.g. in tests or server-side utilities).
 	Component.inline = (props, context) => {
 		const updated = applyDefaultProps({...props}, defaultProps, defaultPropKeys);
-		const merged = renderKind(bindInlineHandlers(updated, handlers, handlerKeys, context), context);
 
-		return useRender(merged, context); // eslint-disable-line react-hooks/rules-of-hooks
+		return renderKind(bindInlineHandlers(updated, handlers, handlerKeys, context), context);
 	};
 
 	return Component;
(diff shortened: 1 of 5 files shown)


packages/core/kind/functionalKind.js
```diff
@@ -155,6 +157,12 @@ const functionalKind = (config) => {
 		const ctx = useContext(contextType);
 		const boundHandlers = useHandlers(handlers, props, ctx);
 
+		// The functional branch below renders through this wrapper rather than `renderKind` directly.
```
dan-ichim-lgp: This comment looks copied from kind.js and does not apply here: there is no “functional branch”, the user entry is useRender not render, and it sits in the middle of Component instead of next to useRenderKind. Please fix or drop it.

packages/core/kind/functionalKind.js
```diff
@@ -146,6 +146,8 @@ const functionalKind = (config) => {
 		return useRender(props, context); // eslint-disable-line react-hooks/rules-of-hooks
 	};
 
+	const useRenderKind = (props, context) => renderKind(props, context);
```
dan-ichim-lgp: The use* prefix on this wrapper is only a naming hint. useRender (and any hooks inside it) still run from renderKind, which is not a hook — same Rules of Hooks violation as the eslint-disable on that call.

This only works if the compiler inlines renderKind into useRenderKind. That is not guaranteed, because renderKind is also used by Component.inline.

Please split prep from render so the React path is a real hook:

const prepareKindProps = (props, context) => {
  if (renderStyles) props = renderStyles(props, context);
  if (renderComputed) props = renderComputed(props, context);
  return props;
};

const useRenderKind = (props, context) => useRender(prepareKindProps(props, context), context);

// Component:
return useRenderKind(merged, ctx);

// inline (hook-free contract unchanged):
return useRender(prepareKindProps(...), context);

Same change belongs in kind.js for functional: true.

An earlier commit on this branch was closer (call useRender from Component), but it discarded renderKind's return value. Styles/computed mutate and return props — keep the return value.

packages/ui/ComponentOverride/ComponentOverride.js
```diff
@@ -57,6 +57,10 @@ import {isValidElementType} from 'react-is';
  * @public
  */
 const ComponentOverride = ({component: Component, ...props}) => {
+	// Despite its capitalized, component-like signature, this is also called directly as a plain
+	// function, including from the render of a class `kind`. React Compiler would otherwise compile it as a component and emit a `useMemoCache`
+	// call, which throws "Invalid hook call" on those direct-call paths.
+	'use no memo';
```
dan-ichim-lgp: The diagnosis is right: PascalCase + JSX means infer mode compiles this and emits useMemoCache, and it is also called as a plain function from class kind render (LabeledIcon, Card/ImageItem via ImageOverride). That is a real invalid hook call.

`'use no memo'` is documented as a temporary escape hatch, not a library API. For a PoC it can stay if the comment makes that explicit and the directive is the first statement.

A stable fix is to stop looking like a component when used as a helper:

function componentOverride({component: Component, ...props}) { /* ... */ }
export default componentOverride;
export {componentOverride as ComponentOverride};

camelCase + JSX is not compiled in infer mode, and existing ComponentOverride imports keep working.

dan-ichim-lgp:
Request changes.

Intent is right: functionalKind should work with React Compiler, including hooks in useRender, without opting every component out via `'use no memo'`.

The useRenderKind wrapper is not enough on its own. useRender still runs from renderKind (not a hook), and renderKind is also used by Component.inline, so the compiler may not treat this as a valid hook path.

Please address the inline comments, then confirm it was run with React Compiler enabled:
- functionalKind with hooks in useRender
- kind({ functional: true })
- a class kind that calls ComponentOverride as a function (e.g. LabeledIcon)

dan-ichim-lgp (approved):

packages/ui/ComponentOverride/ComponentOverride.js
```diff
@@ -57,6 +57,11 @@ import {isValidElementType} from 'react-is';
  * @public
  */
 const ComponentOverride = ({component: Component, ...props}) => {
+	// Despite its capitalized, component-like signature, this is also called directly as a plain
+	// function, including from the render of a class `kind`. React Compiler would otherwise compile it as a component and emit a `useMemoCache`
+	// call, which throws "Invalid hook call" on those direct-call paths.
+	// TODO: rewrite this component to work properly with babel-plugin-react-compiler
+	'use no memo';
```
ion-andrusciac-lgp:
```suggestion
	return getComponentOverride(props);
```

packages/ui/ComponentOverride/ComponentOverride.js
ion-andrusciac-lgp:
```suggestion
```

ion-andrusciac-lgp:
```suggestion

const getComponentOverride = ({component: Component, ...props}) => {
	return Component && (
		isValidElementType(Component) && (
			<Component {...props} />
		) || isValidElement(Component) && (
			cloneElement(Component, props)
		)
	) || null;
};

```

ion-andrusciac-lgp, line 75:
```suggestion
	getComponentOverride,
	ComponentOverride
```

ion-andrusciac-lgp: We split this logic to safely differentiate the usage:
Use `<ComponentOverride {...props} />` when you need a React Function Component.
Use `getComponentOverride(Component, props)` when you need to call it inline as a regular function (e.g., inside class components) to avoid React Compiler/Hook errors.

packages/ui/ComponentOverride/ComponentOverride.js
```diff
@@ -57,6 +57,11 @@ import {isValidElementType} from 'react-is';
  * @public
  */
 const ComponentOverride = ({component: Component, ...props}) => {
```
ion-andrusciac-lgp:
```suggestion

const getComponentOverride = ({component: Component, ...props}) => {
	return Component && (
		isValidElementType(Component) && (
			<Component {...props} />
		) || isValidElement(Component) && (
			cloneElement(Component, props)
		)
	) || null;
};

const ComponentOverride = (props) => {
```

ion-andrusciac-lgp (approved):

daniel-stoian-lgp (approved):

PR #3446, opened 2026-07-29
NXT-570: Added background option for transparent Image - #405

alexandrumorariu:
### Checklist

Changed Image component to be able to change the background for transparent images.

### Resolution

### Additional Considerations

### Links
NXT-570

### DIFF
--- a/packages/ui/Image/Image.js
+++ b/packages/ui/Image/Image.js
@@ -70,10 +70,10 @@ const ImageBase = kind({
 		'aria-label': PropTypes.string,
 
 		/**
-		 *	A color to render behind the image.
-		 *	Accepts any valid CSS `background-color` value (named color, hex, `rgb()`, etc.).
-		 *	Gradients are not supported here; they are CSS images, not colors; use
-		 *	`backgroundSrc` for an image background instead.
+		 * A color to render behind the image.
+		 * Accepts any valid CSS `background-color` value (named color, hex, `rgb()`, etc.).
+		 * Gradients are not supported here; they are CSS images, not colors; use
+		 * `backgroundSrc` for an image background instead.
 		 *
 		 * @type {String}
 		 * @public
@@ -198,7 +198,6 @@ const ImageBase = kind({
 			if (imageSrc) layers.push(`url("${imageSrc}")`);
 			if (backgroundImageSrc) layers.push(`url("${backgroundImageSrc}")`);
 			if (placeholder && imageSrc !== placeholder) layers.push(`url("${placeholder}")`);
-
 
 			return layers.length ? layers.join(', ') : null;
 		},
--- a/packages/ui/Image/Image.js
+++ b/packages/ui/Image/Image.js
@@ -197,8 +197,9 @@ Gradients are not supported here; they are CSS images, not colors; use
 
 			const layers = [];
 			if (imageSrc) layers.push(`url("${imageSrc}")`);
-			if (placeholder && imageSrc !== placeholder) layers.push(`url("${placeholder}")`);
 			if (backgroundImageSrc) layers.push(`url("${backgroundImageSrc}")`);
+			if (placeholder && imageSrc !== placeholder) layers.push(`url("${placeholder}")`);
+
 
 			return layers.length ? layers.join(', ') : null;
 		},
@@ -220,7 +221,7 @@ Gradients are not supported here; they are CSS images, not colors; use
 				{...rest}
 				aria-label={ariaLabel || alt}
 				ref={componentRef}
-				style={{...style, backgroundColor, backgroundImage: bgImage}}
+				style={{...style, ...(backgroundColor ? {backgroundColor} : null), backgroundImage: bgImage}}
 			>
 				{children}
 				<img className={css.img} src={imgSrc} alt={alt} onLoad={onLoad} onError={onError} />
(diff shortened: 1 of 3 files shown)


packages/ui/Image/Image.js
```diff
@@ -155,6 +181,7 @@ const ImageBase = kind({
 	},
 
 	defaultProps: {
+		backgroundColor: 'transparent',
```
daniel-stoian-lgp: This silently overrides existing apps' background colors, Because the prop now defaults to 'transparent' and is applied after the style spread, every existing ui/Image (and every themed Image) now renders with inline background-color: transparent. 
That overrides both style={{backgroundColor: ...}} from callers and any background-color set via CSS class. No theme CSS currently sets it on the Image root, but app code using the style prop, a documented pattern for this component, regresses silently. 
Fix: drop the default and apply conditionally.

packages/ui/Image/Image.js
```diff
@@ -166,25 +193,38 @@ const ImageBase = kind({
...
-			warning(imageSrc, 'Image requires that either the "placeholder" or "src" props be specified.');
-			if (!imageSrc) return null;
-			return placeholder && imageSrc !== placeholder ? `url("${imageSrc}"), url("${placeholder}")` : `url("${imageSrc}")`;
+			const backgroundImageSrc = selectSrc(backgroundSrc);
+			warning(imageSrc || backgroundImageSrc, 'Image requires that either the "src", "placeholder", or "backgroundSrc" props be specified.');
+
+			const layers = [];
+			if (imageSrc) layers.push(`url("${imageSrc}")`);
+			if (placeholder && imageSrc !== placeholder) layers.push(`url("${placeholder}")`);
+			if (backgroundImageSrc) layers.push(`url("${backgroundImageSrc}")`);
```
daniel-stoian-lgp, line 199: Do not apply the sugestion on Gh, but do it from IDE, so lint is checked

 CSS paints the first background layer on top, so the order is src (top), backgroundSrc, placeholder (bottom). The placeholder must sit below backgroundSrc so it doesn't show through transparent regions of src after load.

```suggestion
			const layers = [];
			if (imageSrc) layers.push(`url("${imageSrc}")`);
			if (backgroundImageSrc) layers.push(`url("${backgroundImageSrc}")`);
			if (placeholder && imageSrc !== placeholder) layers.push(`url("${placeholder}")`);

```

packages/ui/Image/Image.js
```diff
@@ -166,25 +193,38 @@ const ImageBase = kind({
...
+		delete rest.backgroundSrc;
 
 		return (
-			<div role="img" {...rest} aria-label={ariaLabel || alt} ref={componentRef} style={{...style, backgroundImage: bgImage}}>
+			<div
+				role="img"
+				{...rest}
+				aria-label={ariaLabel || alt}
+				ref={componentRef}
+				style={{...style, backgroundColor, backgroundImage: bgImage}}
```
daniel-stoian-lgp:
```suggestion
				style={{...style, ...(backgroundColor ? {backgroundColor} : null), backgroundImage: bgImage}}
```

packages/ui/Image/Image.js
```diff
@@ -69,6 +69,32 @@ const ImageBase = kind({
...
+		 * @public
+		 */
+		backgroundColor: PropTypes.string,
+
+		/**
+		 * A second image to render behind the main `src` image.
+		 *
+		 * This is useful when `src` points to an image with transparent areas — this image
+		 * will show through those transparent regions, layered above `backgroundColor` (if
+		 * both are set). Accepts the same string or screen-size-keyed object format as `src`.
```
daniel-stoian-lgp:
```suggestion
		 * This is useful when `src` points to an image with transparent areas — this image
		 * will show through those transparent regions, layered above `backgroundColor` and
		 * `placeholder` but below `src`. Accepts the same string or screen-size-keyed object format as `src`.
```

packages/ui/Image/Image.js
```diff
@@ -166,25 +193,38 @@ const ImageBase = kind({
...
 		},
 		imgSrc: ({src}) => selectSrc(src) || null
 	},
 
-	render: ({alt, 'aria-label': ariaLabel, bgImage, children, componentRef, css, imgSrc, onError, onLoad, style, ...rest}) => {
+	render: ({alt, 'aria-label': ariaLabel, backgroundColor, bgImage, children, componentRef, css, imgSrc, onError, onLoad, style, ...rest}) => {
 		delete rest.placeholder;
 		delete rest.sizing;
 		delete rest.src;
+		delete rest.backgroundSrc;
```
daniel-stoian-lgp: please sort the props on deletion

packages/ui/Image/Image.js
```diff
@@ -69,6 +69,32 @@ const ImageBase = kind({
...
 		'aria-label': PropTypes.string,
 
+		/**
+		 * A color (or any valid CSS `background-color` value, including gradients) to render
+		 * behind the image.
+		 *
+		 * This is useful when `src` (or `backgroundSrc`) points to an image with transparent
+		 * areas — set this to change what shows through those transparent regions. Changing
+		 * this value dynamically (e.g. via state) will update what's visible behind the image
+		 * without affecting the image itself.
```
daniel-stoian-lgp: Please rephrase jsdoc, gradients are not shupported



```suggestion
A color to render behind the image.

Accepts any valid CSS `background-color` value (named color, hex, `rgb()`, etc.).
Gradients are not supported here; they are CSS images, not colors; use
`backgroundSrc` for an image background instead.
```

daniel-stoian-lgp (approved):

dan-ichim-lgp (approved):

PR #3448, opened 2026-08-07
NXT-18124: Migrated `enact/core` to TypeScript

ion-andrusciac-lgp:
### Checklist

Migrated `enact/core` to TypeScript

### Resolution

### Additional Considerations

### Links
NXT-18124

### Comments


packages/core/package.json
daniel-stoian-lgp:
```suggestion
  "main": "index.ts",
```

packages/core/kind/kind.ts
```diff
@@ -144,31 +146,35 @@ const kind = (config) => {
...
 	const renderComputed = cfgComputed ? computed(cfgComputed) : false;
-	const renderKind = (props, context) => {
-		if (renderStyles) props = renderStyles(props, context);
-		if (renderComputed) props = renderComputed(props, context);
+	const renderKind = (props: CallbackObject, context: Context<any>): ReactElement | null => {
+		if (renderStyles && typeof renderStyles === 'function') props = renderStyles(props, context);
+		if (renderComputed && typeof renderComputed === 'function') props = renderComputed(props, context);
 
-		return render(props, context);
+		return (render as RenderFunction)(props, context);
```
daniel-stoian-lgp: KindConfig.render is declared optional (render?: RenderFunction in kind/types.ts:38), but this file's own JSDoc @typedef KindConfig documents render without brackets. This line resolves the contradiction with an unchecked cast instead of a real guard:

return (render as RenderFunction)(props, context);
Since the type now says render is optional, kind({name: 'Foo', ...}) (no render) typechecks cleanly and then throws TypeError: render is not a function on first render.


Either make render required in KindConfig (matching the JSDoc) and drop the cast, or add an explicit runtime guard here (e.g. invariant(typeof render === 'function', 'kind() requires a render function');) if there's a legitimate case for omitting it.

Suggested fix if render really is always required:

	invariant(typeof render === 'function', 'kind() requires a `render` function');
	const renderKind = (props: CallbackObject, context: Context<any>): ReactElement | null => {
		if (renderStyles && typeof renderStyles === 'function') props = renderStyles(props, context);
		if (renderComputed && typeof renderComputed === 'function') props = renderComputed(props, context);
		return render(props, context);
	};

packages/core/hoc/hoc.ts
```diff
@@ -60,25 +62,25 @@ const mergeFn = (key, defaultValue, userValue) => {
  * @memberof core/hoc
  * @public
  */
-const hoc = (defaultConfig, hawk) => {
+const hoc = (defaultConfig: CallbackObject, hawk: Callback) => {
```
daniel-stoian-lgp: Both parameters are required here, but this function's own JSDoc (* an instance config only and return a decorator...) and the very next lines of runtime logic explicitly support calling hoc(factory) with a single argument:

// normalize arguments to allow defaultConfig to be omitted
let factory = hawk;
let defaults: CallbackObject | null = defaultConfig;
if (!factory && typeof defaultConfig === 'function') {
	factory = defaultConfig as Callback;
	defaults = null;
}


I suggest adding an overload for the 1-arg form rather than loosening hawk to optional (which would also make the 2-arg call's second parameter incorrectly appear optional):

function hoc(defaultConfig: CallbackObject, hawk: Callback): ReturnType<typeof hoc>;
function hoc(hawk: Callback): ReturnType<typeof hoc>;
function hoc(defaultConfig: CallbackObject | Callback, hawk?: Callback) {

packages/core/kind/types.ts
```diff
@@ -0,0 +1,40 @@
...
+	computed?: CallbackObject;
+	defaultProps?: CallbackObject;
+	inline?: ComputedPropFunction;
+};
+
+export interface ComputedPropFunction {
+	(props: CallbackObject, context: Context<any>): any;
+}
+
+export interface HandlerFunction {
```
daniel-stoian-lgp: There are 3 incompatible HandlerFunction types in the same package:
enact/packages/core/handle/handle.ts:96:
enact/packages/core/kind/types.ts:21: 
enact/packages/core/types/handlerFunction.type.ts:3.

Suggestion: pick handle.ts's definition as canonical (delete the kind/types.ts copy, import instead), have types/handlerFunction.type.ts and re-export it rather than redeclare it, and consider making it generic over the context value type (HandlerFunction<C = unknown>) so context gets real typing instead of a Context<any> that resolves to any in practice.

alexandrumorariu (approved):
LGTM

PR #3450, opened 2026-08-13
NXT-18125: Migration of enact/spotlight to Typescript

dan-ichim-lgp:
### Checklist

Migrated @enact/spotlight to TypeScript (source and unit tests), following the same approach used for @enact/core (NXT-18124) and @enact/i18n (NXT-18126).

Also corrected a few inaccurate TypeScript typings in @enact/core so spotlight can consume core types directly (no local ambient stubs).

### Resolution

@enact/spotlight was converted from JavaScript to TypeScript: modules under src/, decorators/hooks (Spottable, SpotlightContainerDecorator, SpotlightRootDecorator, Pause, Accelerator), shared types under types/, and package tsconfig.json. Existing unit tests were migrated to TypeScript (.ts/.tsx) and continue to pass.

Follow-up review cleanup included:

- Removed types/enact-core.d.ts and use real @enact/core TypeScript exports
- Tightened spotlight types (ContainerNavigationEvent, ExtSelector, decorator config via Pick, required getSpotRef, etc.)
- Replaced unsafe as ContainerConfig casts with getContainerConfigOrThrow
- Removed redundant allocations in getRect / getIntersectionRect
- Cleaned remaining production/test casts
- Fixed SpotlightContainerDecorator default restrict ('self-first')
- Dropped redundant @testing-library/* deps from spotlight (provided by @enact/cli)

@enact/core typing fixes (no intentional runtime behavior change):

- HandlerFunction / useHandlers context typing
- returnsTrue overloads
- coerceArray accepts T | T[]
- platform typed as PlatformDescription

Runtime behavior is intended to remain unchanged; this is a type-safety / maintainability migration, not a functional redesign.

### Additional Considerations

### Links
NXT-18125

### Comments
Enact-DCO-1.0-Signed-off-by: Dan Ichim([dan.ichim@lgepartner.com](mailto:dan.ichim@lgepartner.com))


alexandrumorariu (approved):
LGTM

packages/spotlight/types/enact-core.d.ts
```diff
@@ -0,0 +1,151 @@
+/*
```
daniel-stoian-lgp: this whole file duplicates types @enact/core
since enact/core is now in typescript, please remove it and use directly the types from enact/core

packages/spotlight/Spottable/useSpottable.ts
```diff
@@ -73,7 +81,7 @@ const useSpottable = ({emulateMouse, getSpotRef, selectionKeys = [ENTER_KEY, REM
 	hook.setPropsAndContext({selectionKeys, spotlightDisabled, ...props}, context); // eslint-disable-line react-hooks/refs
 
 	useLayoutEffect(() => {
-		hook.load(getSpotRef() || null);
+		hook.load(getSpotRef?.() || null);
```
daniel-stoian-lgp: What was the reason for this change?
The original JS called getSpotRef() unconditionally. Omitting it threw TypeError: getSpotRef is not a function immediately. The optional chaining here makes that same omission silently load null instead, and UseSpottableConfig.getSpotRef is itself typed optional (types/SpottableProps.ts:31), so TypeScript doesn't catch the omission at any call site either.

packages/spotlight/Pause/Pause.ts
```diff
@@ -43,23 +43,25 @@
...
+type PausedState = false | true | Pause;
+
+let paused: PausedState = false;
 
 // Private, exported methods used by Spotlight to set and query the pause state from its public API
 
-function getPausedInstance () {
-	return paused === false ? null : paused.name;
+function getPausedInstance (): string | null {
+	return paused === false || paused === true ? null : paused.name;
```
daniel-stoian-lgp: what was the reason for the changed condition?

packages/spotlight/src/container.ts
```diff
@@ -1037,13 +1072,14 @@ function isWithinOverflowContainer (target, containerIds = getContainersForNode(
...
  */
-function notifyLeaveContainer (direction, current, currentContainerIds, next, nextContainerIds) {
+function notifyLeaveContainer (direction: Direction, current: Document | Element | null | undefined, currentContainerIds: string[], next: Document | Element | null | undefined, nextContainerIds: string[]): void {
 	currentContainerIds.forEach(containerId => {
 		if (!nextContainerIds.includes(containerId)) {
 			const config = getContainerConfig(containerId);
 
 			if (config && config.onLeaveContainer) {
-				config.onLeaveContainer({
+				const onLeaveContainer = config.onLeaveContainer as unknown as (ev: ContainerNavigationEvent) => void;
```
daniel-stoian-lgp: ContainerConfig.onLeaveContainer / onLeaveContainerFail / onEnterContainer are typed:

onEnterContainer: ((direction: Direction) => void) | null;
onLeaveContainer: ((direction: Direction) => void) | null;
onLeaveContainerFail: ((direction: Direction) => void) | null;
but are actually invoked with an event-shaped object at all 3 real call sites, e.g.:

```
const onLeaveContainer = config.onLeaveContainer as unknown as (ev: ContainerNavigationEvent) => void;
onLeaveContainer({type: 'onLeaveContainer', direction, target: current, relatedTarget: next});
```
as unknown as disables type checking on the cast value entirely, at all 3 sites (notifyLeaveContainer, notifyLeaveContainerFail, notifyEnterContainer) , so a future signature change to these callbacks, or an actual bug in how they're invoked, won't be caught by the compiler anywhere.

Suggested fix: type them correctly once at the source and drop all 3 casts:

```
	onEnterContainer: ((ev: ContainerNavigationEvent) => void) | null;
	onLeaveContainer: ((ev: ContainerNavigationEvent) => void) | null;
	onLeaveContainerFail: ((ev: Omit<ContainerNavigationEvent, 'relatedTarget'>) => void) | null;
```

packages/spotlight/src/container.ts
```diff
@@ -1062,12 +1098,13 @@ function notifyLeaveContainer (direction, current, currentContainerIds, next, ne
...
  * @private
  */
-function notifyLeaveContainerFail (direction, current, currentContainerIds) {
+function notifyLeaveContainerFail (direction: Direction, current: Document | Element | null | undefined, currentContainerIds: string[]): void {
 	currentContainerIds.forEach(containerId => {
 		const config = getContainerConfig(containerId);
 
 		if (config && config.onLeaveContainerFail) {
-			config.onLeaveContainerFail({
+			const onLeaveContainerFail = config.onLeaveContainerFail as unknown as (ev: Omit<ContainerNavigationEvent, 'relatedTarget'>) => void;
```
daniel-stoian-lgp: same here

packages/spotlight/src/container.ts
```diff
@@ -1086,13 +1123,14 @@ function notifyLeaveContainerFail (direction, current, currentContainerIds) {
...
  */
-function notifyEnterContainer (direction, previous, previousContainerIds, current, currentContainerIds) {
+function notifyEnterContainer (direction: Direction, previous: Document | Element | null | undefined, previousContainerIds: string[], current: Document | Element | null | undefined, currentContainerIds: string[]): void {
 	currentContainerIds.forEach(containerId => {
 		if (!previousContainerIds.includes(containerId)) {
 			const config = getContainerConfig(containerId);
 
 			if (config && config.onEnterContainer) {
-				config.onEnterContainer({
+				const onEnterContainer = config.onEnterContainer as unknown as (ev: ContainerNavigationEvent) => void;
```
daniel-stoian-lgp: same here

packages/spotlight/src/container.ts
```diff
@@ -529,9 +557,9 @@ const configureContainer = (...args) => {
  * @memberof spotlight/container
  * @public
  */
-const addContainer = (...args) => {
+const addContainer = (...args: (string | ContainerConfigUpdate)[]): string => {
 	const containerId = configureContainer(...args);
-	const config = getContainerConfig(containerId);
+	const config = getContainerConfig(containerId) as ContainerConfig;
```
daniel-stoian-lgp: getContainerConfig()'s | undefined is cast away at 5 call sites instead of guarded once

```
const getContainerConfig = (id: string): ContainerConfig | undefined => {
	return containerConfigs.get(id);
};
```

is force-cast with as ContainerConfig at 5 separate call sites across container.ts (lines 562, 752, 777) and spotlight.ts. (lines 293, 455) Each cast silently asserts the config exists without a runtime check: if any caller passes a stale or removed container id, the result is a runtime TypeError on the next property access rather than a compile-time signal.

The pattern repeating identically across 2 files suggests the API shape doesn't fit how it's actually used (callers "know" the container was just created/verified, but the type doesn't reflect that guarantee). 

I suggest a companion helper for call sites that have already established the container exists, so the guarantee is expressed once:

```
const getContainerConfigOrThrow = (id: string): ContainerConfig => {
	const config = getContainerConfig(id);
	invariant(config, `No container registered for id "${id}"`);
	return config;
};
```

and swap the 5 `as ContainerConfig` casts for calls to this instead. Same call-site guarantee, but backed by an actual runtime check instead of a compiler suppression.

packages/spotlight/types/SpottableProps.ts
```diff
@@ -0,0 +1,67 @@
...
+
+/**
+ * Configuration for {@link spotlight/Spottable.useSpottable}.
+ */
+export interface UseSpottableConfig extends SpottableCoreProps {
+	emulateMouse?: boolean;
+	getSpotRef?: () => HTMLElement | null | undefined;
+}
+
+type SpottableHandler = (ev: any, props?: any, context?: any) => any;
```
daniel-stoian-lgp: types/SpottableProps.ts, line 34 and types/SpotlightContainerProps.ts, line 4  have the same handler-function shape declared independently 2  times. this will be solved anturally after using enact/core which is uses typescript
import HandlerFunction from @enact/core's real types (or, if spotlight's usage is meaningfully different, declare one shared EventHandler type under packages/spotlight/types/ and import it in the other two files instead of redeclaring).

packages/spotlight/src/utils.ts
```diff
@@ -73,15 +84,21 @@ const testIntersection = (type, containerRect, elementRect) => {
...
+const contains = curry((containerRect: Rect, elementRect: Rect): boolean => {
 	return testIntersection('contains', containerRect, elementRect);
-});
+}) as {
+	(containerRect: Rect, elementRect: Rect): boolean;
+	(containerRect: Rect): (elementRect: Rect) => boolean;
+};
 
-function getIntersectionRect (container, element) {
+function getIntersectionRect (container: Element, element: Element): Rect {
```
daniel-stoian-lgp, line 101: this applies for both getIntersectionRect and getRect

Double allocation on the hot navigation path
Both functions build a full Rect literal with a throwaway all-zero center placeholder (to satisfy the now-required Rect.center field), then immediately overwrite it with a second, fully-computed object:

const rect: Rect = {
	left: cr.left, top: cr.top, width: cr.width, height: cr.height,
	right: cr.left + cr.width, bottom: cr.top + cr.height, element: elem,
	center: {x: 0, y: 0, left: 0, right: 0, top: 0, bottom: 0}   // <- discarded immediately
};
rect.center = {
	x: rect.left + Math.floor(rect.width / 2),
	// ...real values
};

getRect/getIntersectionRect run once per spottable element on every directional key press (getRects(...).map(getRect) in target.ts), so a screen with N focusable items does N extra small-object allocations per key press for no reason.

The sibling function getPointRect, a few lines below in this same file, already shows the cheap pattern: compute the real center once, then build the Rect literal with it:

`function getRect (elem: Element): Rect {
	const cr = elem.getBoundingClientRect();
	const left = cr.left;
	const top = cr.top;
	const width = cr.width;
	const height = cr.height;
	const centerX = left + Math.floor(width / 2);
	const centerY = top + Math.floor(height / 2);
	const center: RectCenter = {x: centerX, y: centerY, left: centerX, right: centerX, top: centerY, bottom: centerY};
	return {left, top, width, height, right: left + width, bottom: top + height, element: elem, center};
}`

packages/spotlight/src/navigate.ts
```diff
@@ -1,12 +1,58 @@
...
+import type {NavigableConfig} from '../types/NavigableConfig';
+import type {Rect, RectCenter} from '../types/Rect';
+
 const obliqueMinDistance = 1;
 const straightMinDistance = 0;
 
-const calcGroupId = ({x, y}) => y * 3 + x;
+type DistanceFunction = (rect: Rect) => number;
+
+interface GridPosition {
```
daniel-stoian-lgp: this interface is byte-identical to types/Position.ts's exported Position. Please import that

packages/spotlight/types/SpotlightContainerConfig.ts
```diff
@@ -0,0 +1,51 @@
...
+	containerConfig?: SpotlightContainerRuntimeConfig;
+	navigableFilter?: NavigableFilter;
+	preserveId?: boolean;
+}
+
+/**
+ * Default configuration for
+ * {@link spotlight/SpotlightContainerDecorator.SpotlightContainerDecorator}.
+ */
+export interface SpotlightContainerDecoratorConfig {
```
daniel-stoian-lgp: This interface hand-copies fields the sibling type SpotlightContainerRuntimeConfig  just above derives via Pick

`SpotlightContainerRuntimeConfig` (just above) correctly derives from ContainerConfig via Pick, so it auto-updates if ContainerConfig changes. `SpotlightContainerDecoratorConfig` hand-copies 6 of those same field types instead. If ContainerConfig changes one of them, this interface silently falls out of sync.

packages/spotlight/package.json
```diff
@@ -24,13 +24,31 @@
...
     "ramda": "^0.32.0",
     "react": "^19.2.7",
     "react-dom": "^19.2.7",
+    "typescript": "^7.0.2",
     "warning": "^4.0.3"
+  },
+  "devDependencies": {
+    "@testing-library/jest-dom": "^6.9.1",
+    "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
```
daniel-stoian-lgp: these 3 are not necesary. testing is handled via enact/cli

packages/spotlight/types/test-globals.d.ts
```diff
@@ -0,0 +1,18 @@
+// / <reference types="@testing-library/jest-dom" />
+
+declare module '*.less' {
+	const classes: Record<string, string>;
+	export default classes;
+}
+
+interface Element {
+	closest?: (selectors: string) => Element | null;
```
daniel-stoian-lgp: this breaks tsc --noEmit for unrelated production code


This is meant to support the JSDOM polyfill test helper, but since it's a global ambient merge (not scoped to a module), it widens Element.closest's type for the entire compile, including SpotlightContainer.ts:139, unrelated production code,

Removing this block makes tsc --noEmit pass with zero errors package-wide, and the polyfill file it's meant to support still type-checks fine without it (it already does its own local as Partial<Element> cast where it actually needs one, for the delete statement).

packages/core/types/handlerFunction.type.ts
```diff
@@ -10,4 +6,5 @@ import {CallbackObject} from './callback.type';
  * @param {Object<string, any>} props
  * @param {Object<string, any>} context
  */
-export type HandlerFunction<C = unknown> = (event: Event, props: CallbackObject, context: Context<C>) => any;
+// Parameters stay `any` so consumers can pass narrower event/prop shapes into handle().
+export type HandlerFunction = (event?: any, props?: any, context?: any) => any;
```
daniel-stoian-lgp: This was loosened to fix spotlight's SpottableProps.ts/SpotlightContainerProps.ts, which had 3 independently-hand-declared any-typed local aliases for this exact shape and needed to import the real one instead.

The problem: HandlerFunction is  consumed within core itself, and by every downstream package built on handle()/kind() 
I verified this is load-bearing, not theoretical: reverting this line to its strict form and re-running tsc --noEmit on spotlight alone produces 6 real errors. With the type loosened, none of that gets caught . Every handle()/kind() chain in every consuming package silently loses argument-shape checking.

Suggested direction: restore the strict, generic form here, and fix the 3 spotlight call sites that actually had the wrong/loose shapes instead

`export type HandlerFunction<C = unknown> = (event: Event, props: CallbackObject, context: Context<C>) => any;`

(would need Context/CallbackObject imports restored here and in packages/core/useHandlers/useHandlers.ts, which had its own Context<any> → any downgrade as part of the same commit.)

packages/spotlight/Spottable/useSpottable.ts
```diff
@@ -54,11 +56,17 @@ const REMOTE_OK_KEY = 16777221;
...
+const EMPTY_ATTRIBUTES: Record<string, string> = {};
 
-const useSpottable = ({emulateMouse, getSpotRef, selectionKeys = [ENTER_KEY, REMOTE_OK_KEY], spotlightDisabled, ...props} = {}) => {
+const useSpottable = ({
+	emulateMouse,
+	getSpotRef,
+	selectionKeys = [ENTER_KEY, REMOTE_OK_KEY],
+	spotlightDisabled,
+	...props
+}: UseSpottableConfig): UseSpottableResult => {
```
daniel-stoian-lgp: useSpottable's config parameter lost its = {} default .
is this intentional?

packages/spotlight/src/target.ts
```diff
@@ -473,45 +500,48 @@ function getTargetByDirectionFromPosition (direction, position, containerId) {
...
-function getLeaveForTarget (containerId, direction) {
+function getLeaveForTarget (containerId: string, direction: Direction): ContainerTarget | boolean | null {
 	const config = getContainerConfig(containerId);
 
 	if (config) {
-		const target = config.restrict !== 'self-only' && config.leaveFor && config.leaveFor[direction];
+		// `leaveFor` is documented as an <extSelector> per direction but, in practice, is only
+		// ever a CSS selector string or a single Element.
+		const target = (config.restrict !== 'self-only' && config.leaveFor && config.leaveFor[direction]) as
+			string | Element | false | null | undefined;
```
daniel-stoian-lgp: ExtSelector/LeaveFor's declared type still doesn't match what this code actually handles

types/ExtSelector.ts was touched in this PR, but in the opposite direction . the array member was widened from Element[] to (string | Element)[] for parseSelector/getContainerDefaultElement's legitimate use, which doesn't help here since LeaveFor's value type still claims to support arrays/NodeLists that this function silently can't handle.

Since it sounds like leaveFor genuinely never receives an array/NodeList in practice, could we narrow the type at the source instead of casting around it at the call site?


```
type LeaveForTarget = string | Element | null | undefined;
export type LeaveFor = Partial<Record<Direction, LeaveForTarget>> | null;
```

 then this cast (and the comment explaining why it's needed) goes away entirely, and the type honestly reflects what's supported.

packages/spotlight/src/target.ts
```diff
@@ -28,8 +28,20 @@ import {
...
+import type {Direction} from '../types/Direction';
+import type {PreferredEnterTo} from '../types/PreferredEnterTo';
+import type {Position} from '../types/Position';
+import type {Rect} from '../types/Rect';
+
+/** A candidate target discovered while searching for the nearest element in a direction. */
+interface DirectionalTarget {
+	direction: Direction;
+	target: Element;
+	distance?: number;
```
daniel-stoian-lgp: DirectionalTarget.distance is still optional. then mutated, then cast at line 630

**distance** is built via push then mutate instead of being computed at construction, so the type has to be optional, and the sort comparator casts that optionality away instead of the compiler being able to guarantee it's always set.

Suggested fix — compute distance inline where each target is built, and make the field required:

`interface DirectionalTarget {
	direction: Direction;
	target: Element;
	distance: number;
}`

`const getNearestTargetsInContainerFromPosition = (position: Position, containerId: string): DirectionalTarget[] => {
	return (['up', 'left', 'right', 'down'] as Direction[]).reduce((result: DirectionalTarget[], direction) => {
		const target = getTargetByDirectionFromPosition(direction, position, containerId);
		if (target) {
			result.push({
				direction,
				target,
				distance: getDistanceToTargetFromPosition(direction, position, getRect(target))
			});
		}
		return result;
	}, []);
};
const getNearestTargetInContainerFromPosition = (position: Position, containerId: string): Element | undefined => {
	const targets = getNearestTargetsInContainerFromPosition(position, containerId);
	if (!targets.length) {
		return;
	}
	targets.sort((a, b) => a.distance - b.distance);
	return targets[0].target;
};`

packages/spotlight/src/utils.ts
```diff
@@ -0,0 +1,289 @@
...
+	} catch {
+		warning(true, `parseSelector failed for selector: ${selector}`);
+	}
+
+	return result;
+}
+
+type IntersectionType = 'intersects' | 'contains';
+
+type RectEdges = Pick<Rect, 'left' | 'right' | 'top' | 'bottom'>;
```
daniel-stoian-lgp: New RectEdges type duplicates one already in navigate.ts

Both files already import type {Rect} from '../types/Rect', so a shared export from types/Rect.ts was the natural spot The same pattern is already correctly applied elsewhere in this PR for Position or HandlerFunction

packages/spotlight/src/navigate.ts
```diff
@@ -88,16 +130,18 @@ function prioritize (priorities, targetEdge) {
...
-	destGroup.sort(function (a, b) {
+	destGroup.sort(function (a: DestCandidate, b: DestCandidate) {
 		return a.distance - b.distance;
 	});
 
 	return destGroup;
 }
 
-function partition (rects, targetRect, straightOverlapThreshold, getGroupId) {
+function partition (rects: Rect[], targetRect: PartitionTarget, straightOverlapThreshold: number, getGroupId: GroupIdFunction): Rect[][] {
```
daniel-stoian-lgp: partition() casts a RectCenter to Rect, silently losing width/height on one call path

PartitionTarget = Rect | RectCenter, and partition is called at line ~283 with targetRect.center — a RectCenter, which has no width/height. In that call, targetWidth/targetHeight are undefined, so every corner-group reclassification check below (rect.left <= targetRect.right - targetWidth * straightOverlapThreshold, etc.) becomes a NaN-based comparison that's always false, silently defeating the corner-group boundary logic for that call path.

This predates this migration (same untyped property access existed in the original JS), so it's not a regression from this PR , but the migration formalized the dual-shape contract via PartitionTarget and then cast straight through it rather than branching on it, which is exactly the kind of thing typing this file was supposed to catch. Worth a follow-up to actually branch on which variant was received:

packages/spotlight/src/navigate.ts
```diff
@@ -88,16 +130,18 @@ function prioritize (priorities, targetEdge) {
...
+	const targetWidth = (targetRect as Rect).width;
+	const targetHeight = (targetRect as Rect).height;
```
daniel-stoian-lgp: partition() casts a RectCenter to Rect, silently losing width/height on one call path

PartitionTarget = Rect | RectCenter, and partition is called at line ~283 with targetRect.center,  a RectCenter, which has no width/height. In that call, targetWidth/targetHeight are undefined, so every corner-group reclassification check below (rect.left <= targetRect.right - targetWidth * straightOverlapThreshold, etc.) becomes a NaN-based comparison that's always false, silently defeating the corner-group boundary logic for that call path.

This predates this migration (same untyped property access existed in the original JS), so it's not a regression from this PR , but the migration formalized the dual-shape contract via PartitionTarget and then cast straight through it rather than branching on it, which is exactly the kind of thing typing this file was supposed to catch. 
Maybe we can setup a fallback value?:


`const targetWidth = 'width' in targetRect ? targetRect.width : 0;

const targetHeight = 'height' in targetRect ? targetRect.height : 0;`

packages/spotlight/Spottable/useSpottable.ts
dan-ichim-lgp: Yes

daniel-stoian-lgp (approved):

PR #3452, opened 2026-08-19
NXT-18690: migrate tests core, i18n to TS

dan-ichim-lgp:
### Checklist

Migrate @enact/core and @enact/i18n unit tests from JavaScript to TypeScript, and align Testing Library typing setup across core, i18n, and spotlight for feature/typescript.

### Resolution

- Renamed and typed unit specs: 21 in core and 9 in i18n (*-specs.js → *-specs.ts / *-specs.tsx).
- Added @types/jest and set tsconfig types to ["jest", "node"].
- Added @testing-library/react, @testing-library/jest-dom, and @testing-library/user-event as devDependencies (same versions as @enact/cli) so tsc uses real types instead of ambient shims.
- Removed **/tests/** from the i18n tsconfig exclude so migrated tests are typechecked.
- Added Jest collectCoverageFrom for TS in core.
- Slimmed spotlight test-globals.d.ts to a jest-dom reference + *.less (matchers come from the real package).

### Additional Considerations

### Links
NXT-18690

### Comments


packages/spotlight/package.json
```diff
@@ -41,6 +41,9 @@
     "warning": "^4.0.3"
   },
   "devDependencies": {
+    "@testing-library/jest-dom": "^6.9.1",
+    "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
```
daniel-stoian-lgp: these are handled through enact/cli

packages/core/internal/prop-types/tests/prop-types-specs.tsx
```diff
@@ -240,14 +240,14 @@ describe('prop-types', () => {
 		test('should call console.warn when prop value is given for EnactPropTypes.deprecated', () => {
 			const ref = createRef();
 
-			class DeprecatedComponent extends Component {
+			class DeprecatedComponent extends Component<any> {
 				static displayName = 'DeprecatedComponent';
 
 				static propTypes = {
-					typeDeprecated: EnactPropTypes.deprecated(EnactPropTypes.ref)
+					typeDeprecated: (EnactPropTypes.deprecated as any)(EnactPropTypes.ref)
```
daniel-stoian-lgp: this test doesn't actually test the deprecation contract

(EnactPropTypes.deprecated as any)(EnactPropTypes.ref) casts the whole callee to any to call it with only 1 of its 2 required arguments. The real signature is deprecated(base: Callback, config: CallbackObject) (prop-types.ts:63).

With config undefined, deprecate(() => true, config) takes the "config omitted" branch, treats the anonymous validator as the config object, and warns once at class-declaration time , before render() ever runs , then returns the validator completely unwrapped (base is never called). Both tests here pass because a warning fires for the wrong reason, not because the prop was set. The deprecation-on-first-use contract is never actually exercised.

`					typeDeprecated: EnactPropTypes.deprecated(EnactPropTypes.ref, {name: 'typeDeprecated', replacedBy: 'ref'})`

(Adjust the config shape to whatever deprecate()'s message formatting actually expects . the point is supplying a real config object so TS's "expected 2 arguments" error doesn't get cast away.)

packages/core/internal/prop-types/tests/prop-types-specs.tsx
```diff
@@ -259,14 +259,14 @@ describe('prop-types', () => {
 		});
 
 		test('should call console.warn even when prop value is not given for EnactPropTypes.deprecated', () => {
-			class DeprecatedComponent extends Component {
+			class DeprecatedComponent extends Component<any> {
 				static displayName = 'DeprecatedComponent';
 
 				static propTypes = {
-					typeDeprecated: EnactPropTypes.deprecated(EnactPropTypes.ref)
+					typeDeprecated: (EnactPropTypes.deprecated as any)(EnactPropTypes.ref)
```
daniel-stoian-lgp: same here

packages/core/dispatcher/tests/dispatcher-specs.ts
```diff
@@ -3,7 +3,7 @@ import {off, on, once} from '../dispatcher';
 describe('dispatcher', () => {
 	test('should register handlers on target', () => {
 		const handler = jest.fn();
-		on('localechange', handler, window);
+		on('localechange', handler, window as any);
```
daniel-stoian-lgp: not a bug in the tests, buy please check dispatcher.ts: 111

Type is wrong at the source .  target excludes Window

on/off default target = getDefaultTarget(), typed Node | Boolean. window is a legitimate target , already used in production (packages/ui/Cancelable/modalHandler.js) and already worked around with as unknown as Node in packages/i18n/I18nDecorator/I18n.ts and windowFocus.ts. The new dispatcher-specs.ts adds a 3rd/10th workaround site, this time with a bare as any (weaker than the as unknown as Node pattern used elsewhere).

`const on = function (name: string, fn: Callback, target: Node | Window | Boolean = getDefaultTarget()): undefined {`

Please Widen target's type here (and on off) to include Window/EventTarget, and drop the as any/as unknown as Node casts at all three call sites once this lands.

packages/core/handle/tests/handle-specs.tsx
```diff
@@ -3,24 +3,54 @@ import {Component} from 'react';
...
 	preventDefault,
 	stop
 } from '../handle';
+import type {Callback, CallbackObject} from '../../types';
+
+// The published types require a DOM `Event` and a props object on every handler while the runtime
+// contract accepts partial event payloads, non-functions, and omitted arguments. These tests
+// exercise that runtime contract, so the entry points are retyped with the looser signature rather
+// than casting at each of the call sites below.
+type TestHandler = {
```
daniel-stoian-lgp: Part of this is already stale .HandlerFunction's event/context were fixed to Event | null/context?: C in an earlier round, so the shadow types' looseness there is now redundant. But a real gap remains: EventHandler (handle.ts:106) declares only named/bindAs, not .finally() , even though handle() attaches .finally at runtime and this file calls it 4+ times.

```
export interface EventHandler extends HandlerFunction {
	named: (name: string) => HandlerFunction;
	bindAs: (obj: CallbackObject, name?: string) => HandlerFunction;
	finally: (cleanup: Callback) => EventHandler;
}
```

With that in place, most of TestHandler/TestHandlerFactory/TestCustomForwarder here can likely be dropped in favor of the real EventHandler type.

packages/core/usePublicClassNames/tests/usePublicClassNames-specs.ts
daniel-stoian-lgp, line 3: Please check main component file: PublicClassNames marks fields required that the hook treats as optional

```
export interface PublicClassNames {
	componentCss?: {[key: string]: string},
	customCss?: {[key: string]: string},
	publicClassNames?: boolean | string | string[],
}
```

The hook's own body immediately does if (!componentCss || !customCss) return componentCss;, and the JSDoc already marks all three params [optional]. This is why the migrated test needs customCss: null as any and {componentCss, customCss} as any . With the interface fixed, both casts become unnecessary.

packages/core/util/tests/Job-specs.ts
```diff
@@ -1,8 +1,11 @@
 import {act} from '@testing-library/react';
 
-import Job from '../Job';
+import JobBase from '../Job';
 
 describe('Job', () => {
+	// `timeout` is optional at runtime
+	const Job = JobBase as unknown as new (fn: (...args: any[]) => any, timeout?: number) => JobBase;
```
daniel-stoian-lgp: Job constructor re-derived and double-cast through unknown

`const Job = JobBase as unknown as new (fn: Callback, timeout?: number) => JobBase;`

Smaller fix: use the already-exported Callback type (from ../../types) instead of hand-rolling (...args: any[]) => any . at minimum keeps this in sync with Job's real fn param type if it changes.

packages/core/util/tests/util-specs.tsx
```diff
@@ -243,9 +243,12 @@ describe('util', () => {
 	});
 
 	describe('mapAndFilterChildren', () => {
+		// the `filter` argument is optional at runtime
+		const mapAndFilter = mapAndFilterChildren as (children: any, callback: (value: any, index?: number) => any, filter?: (value: any) => boolean) => any;
```
daniel-stoian-lgp: mapAndFilterChildren cast to an all-any signature

`const mapAndFilter = mapAndFilterChildren as (children: ReactNode, callback: Callback, filter?: FilterCallback<any>) => ReactNode;`

The current cast (children: any, callback: (value: any, index?: number) => any, filter?: (value: any) => boolean) => any) throws away type checking on every argument just to make filter optional. Casting to the real parameter types with only filter loosened keeps the same convenience without losing coverage on the other three positions.

packages/i18n/I18nDecorator/tests/useI18n-specs.tsx
```diff
@@ -6,6 +6,12 @@ import {updateLocale} from '../../locale';
 import {I18n} from '../I18n';
 import useI18n from '../useI18n';
 
+interface ComponentProps {
```
daniel-stoian-lgp: Local ComponentProps duplicates the already-exported UseI18nConfig

`import useI18n, {type UseI18nConfig} from '../useI18n';`

The local interface ComponentProps (3 fields) is a near-but-not-exact subset of UseI18nConfig. Missing resources/sync, and typing the array fields without | null where the real type allows it. Importing the real type keeps this fixture honest if the hook's config contract changes.

packages/core/package.json
```diff
@@ -43,7 +49,11 @@
     "warning": "^4.0.3"
   },
   "devDependencies": {
+    "@testing-library/jest-dom": "^6.9.1",
+    "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
```
daniel-stoian-lgp: these should be handled via enact/cli. pleased remove them

packages/i18n/package.json
```diff
@@ -57,6 +57,10 @@
     "ilib": "^14.21.0 || ^14.21.0-webos1"
   },
   "devDependencies": {
+    "@testing-library/jest-dom": "^6.9.1",
+    "@testing-library/react": "^16.3.2",
+    "@testing-library/user-event": "^14.6.1",
```
daniel-stoian-lgp: these are handled via enact/cli. please remove

packages/core/internal/Registry/tests/Registry-specs.tsx
```diff
@@ -5,22 +5,24 @@ import userEvent from '@testing-library/user-event';
 
 import Registry from '../Registry';
 
-const SomeContext = createContext();
+const SomeContext = createContext<any>(null);
```
daniel-stoian-lgp: Can you retype NotifiesTree/HandlesNotification's fixtures using Registry.ts's real exports instead of any? Right now SomeContext is createContext<any>(null), this.registry is registry?: any, and both components extend Component<any> — so nothing here actually gets checked against Registry's real contract.

```
import Registry from '../Registry';
import type {RegisterFunction, RegistryController, RegistryEvent} from '../Registry';

const SomeContext = createContext<RegisterFunction | null>(null);
```

Then registry: RegistryHandle (a new type worth adding , see the comment on Registry.ts) on NotifiesTree, registry?: RegistryController on HandlesNotification, and Component<ComponentPropsWithoutRef<'button'>> / Component<ComponentPropsWithoutRef<'div'>, {number: number}> instead of Component<any>.

One heads-up: the obvious way to type this.context here is declare context: ContextType<typeof SomeContext>;, but our Jest/Babel pipeline doesn't support TS declare class fields (tsc is fine with it, but the Jest transform throws TypeScript 'declare' fields must first be transformed by @babel/plugin-transform-typescript). A plain (non-declare) field would actually initialize this.context at construction and risk clobbering React's own context assignment, so don't add a field for it , just cast at the 3 call sites instead, e.g. this.context as RegisterFunction | null.

daniel-stoian-lgp: This comment applies to the main file **packages/core/internal/Registry/Registry.ts:3**

Once the test above is retyped, tsc turns up two real mismatches here that are worth fixing while we're at it:

1. RegistryController.notify is typed optional but never actually is.

```
export interface RegistryController {
	notify: (ev: RegistryEvent) => void;
	unregister: () => void;
}
```

The object register() returns implements notify(ev: RegistryEvent) (required), and it forwards straight into RegistryHandler, whose ev param is also required. The ? was just wrong.

2. Registry.create's return type is inferred through Object.freeze({...}) → Readonly<T>, which makes the parent setter statically unassignable even though setters remain fully callable on a frozen object at runtime (Object.freeze blocks reconfiguring/adding properties, not invoking an existing accessor). Can you add an explicit return type instead of letting it flow through Object.freeze's typing?

```
export interface RegistryHandle {
	parent: RegisterFunction | null;
	notify: (ev?: RegistryEvent, exclude?: (instance: RegistryInstance) => boolean) => void;
	register: RegisterFunction;
}
	create: (handler: RegistryHandler): RegistryHandle => {
		const instances: RegistryInstance[] = [];
		let currentParent: RegistryController;

		const registry = {
			set parent (register: RegisterFunction | null) {
```

...and at the end, freeze without going through the typed variable:

```
		};

		Object.freeze(registry);

		return registry as RegistryHandle;
	}
```

daniel-stoian-lgp (commented):
check my commemnts

daniel-stoian-lgp (approved):

ion-andrusciac-lgp (approved):

PR #3453, opened 2026-08-26
NXT-18688: Migrate ui package to TypeScript part 1

alexandrumorariu:
### Checklist

* [X] I have run automated testing and it is passed

Migrates the packages/ui package from JavaScript to TypeScript. This covers all 27 sub-packages found under ui/ (Layout, Button, Card, Cancelable, Changeable, FloatingLayer, Marquee, Media, Placeholder, Resizable, resolution, Routable, and the rest), including their internal helpers under ui/internal/ and their existing unit test suites — roughly 60 source files and 40 test files converted from .js/.jsx to .ts/.tsx.

The goal was a behavior-preserving migration: no runtime logic changes, only the addition of types and the minimum structural changes needed to satisfy a strict TypeScript configuration (strict: true, noImplicitAny: true, isolatedModules: true) and this repo's ESLint rules.

### Resolution

- Every migrated file type-checks cleanly against the project's tsconfig.json with zero errors, and passes both tsc --noEmit and an actual Babel transform pass (mirroring the Jest pipeline) with zero failures.

- Added types/enact.d.ts: ambient module declarations for the @enact/* internals and other third-party packages (invariant, warning, ramda/src/*, react-is, *.module.less, etc.) that ship no types of their own. Their internal callback signatures (e.g. kind()'s render/computed/handlers) are typed with explicit any parameters rather than left implicit, so consuming object literals get contextual typing instead of tripping noImplicitAny file-by-file.

- Added types/jsx-intrinsics.d.ts as a separate ambient file augmenting React's JSX.IntrinsicElements (for the <icon> slot pseudo-element used by Slottable-decorated components like LabeledIcon). This has to live in its own file with an export {}, because module-augmentation semantics require the declaring file to itself be a module — but the other ambient declarations (declare module '@enact/...') require the opposite (a plain script file, no top-level import/export) to be picked up correctly under isolatedModules + moduleResolution: bundler. Splitting them was the only way to satisfy both. 

- Public-facing component APIs (Layout, Cell, resolution, Router, ResolutionDecorator, etc.) got real, precise interface/type definitions rather than blanket any.

- For the framework-glue layer — classes generated inside hoc()/kind() factories — used Record<string, any> generics on Component/PureComponent plus a documented [key: string]: any index signature. These classes legitimately mix React state with a large set of dynamically-assigned instance fields (timers, DOM node caches, measurement state) that aren't part of the public contract and weren't worth hand-declaring one by one; the index signature is scoped to the internal implementation, while propTypes/defaultProps remain the authoritative typed contract for consumers. These fields/signatures are ordered to satisfy react/sort-comp (after lifecycle methods, before custom methods and render).

- Two places needed context: any = null; overrides (Marquee/MarqueeDecorator.tsx, Placeholder/PlaceholderDecorator.tsx) to type this.context when a class sets static contextType — implemented as a real initialized field rather than declare context: any;, since declare class fields need a specific Babel plugin ordering this project's Babel config doesn't have and broke test runs with a syntax error.

- Preserved the original codebase's intentional inner/outer name-shadowing pattern (const X = hoc(..., () => { function X (props) {...} })) along with its existing eslint-disable comments, updated to disable both no-shadow and @typescript-eslint/no-shadow together, since this repo's lint config runs both rules simultaneously.
Removed a useless empty constructor () {} from a test mock class in Marquee-specs.tsx flagged by @typescript-eslint/no-useless-constructor.

- Updated every package's package.json main field to point at the new .ts/.tsx filename (these were still pointing at the old .js names, which broke Jest's module resolution for directory-style imports like import X from '../ComponentOverride').

- Removed a couple of hardcoded .js extensions in import specifiers (from '../resolution.js', from '../index.js') that pointed at files which no longer exist under that name — tsc doesn't flag these under moduleResolution: bundler, but Jest's resolver does.

- Converted internal/localized-fonts/localized-fonts.js, which used CommonJS module.exports, to standard ESM export/export default for consistency with the rest of the codebase.

- Fixed one incidental logic bug surfaced during typing in internal/localized-fonts/localized-fonts.ts: a !config && !config.name condition that was checking the wrong operator (should likely have been ||); left the original behavior intact but calling it out for review.

### Additional Considerations
How to test: Run the existing unit test suite (jest) and lint (enact lint -s) for the ui package. All existing test files were migrated as part of this PR and should pass unchanged, since no test assertions or component behavior were modified, only types and import paths.

Outstanding questions / follow-ups:

- Button.tsx, Item.tsx, and LabeledIcon.tsx still import the sibling ui/Touchable and ui/Slottable packages, which are not migrated in this PR and still exist as .js in this repo. Since there's no .d.ts alongside them, TypeScript will flag those two imports as implicit any under noImplicitAny. Suggest either: (a) migrating Touchable/Slottable in a follow-up PR, or (b) adding minimal sibling declaration files (e.g. Touchable/Touchable.d.ts with declare const Touchable: any; export default Touchable;) as a stopgap until they're migrated.

- The internal/localized-fonts/localized-fonts.ts conditional bug mentioned above (!config && !config.name) is preserved as-is to avoid changing behavior in this PR — worth a follow-up fix in a separate change.

- Test files reference Jest globals (describe, test, jest, etc.) via a local types/jest-globals.d.ts ambient file since the provided tsconfig.json restricts types to ["node"]. If the project has (or plans to add) @types/jest globally, this file can likely be removed in favor of that.

- @types/react/@types/react-dom v19 were used in the validation environment to support the use() hook already present in Resizable/Marquee; confirm this matches the actual React version this package targets.

Side effects: None intended — this is a types-and-file-extension-only change. The one behavioral wrinkle worth double-checking in review is the context field initializer pattern described above (functionally equivalent to the previous behavior, but touches two class constructors).

### Links
NXT-18688

### Comments
Given the size (27 sub-packages, ~100 files), this may be easier to review folder-by-folder rather than as one large diff. Happy to split into smaller PRs per package if preferred.


packages/ui/types/enact.d.ts
```diff
@@ -0,0 +1,221 @@
+/**
```
daniel-stoian-lgp, line 1: This ambient file shadows real, already-migrated modules instead of augmenting them

Delete the module blocks for every already-migrated package (@enact/core/*, @enact/i18n/util) and wire packages/ui's tsconfig to consume core/i18n's real types directly

packages/ui/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "module": "esnext",
+    "moduleResolution": "bundler",
+    "noEmit": true,
+    "noFallthroughCasesInSwitch": true,
+    "noImplicitAny": true,
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "strict": true,
+    "target": "es2022",
+    "types": ["node"]
```
daniel-stoian-lgp, line 20: @types/node is required but not installed.  tsc fails before checking anything

@types/node isn't a devDependency of packages/ui

packages/ui/FloatingLayer/FloatingLayerContainer.ts
```diff
@@ -1,7 +1,15 @@
-import Registry from '@enact/core/internal/Registry';
+import Registry, {type RegistryInstance} from '@enact/core/internal/Registry';
+
+interface FloatingLayerContainerConfig {
+	floatLayerId?: string;
+}
 
 class FloatingLayerContainer {
-	constructor (config) {
+	registry: RegistryInstance;
```
daniel-stoian-lgp: This only compiles at this because enact.d.ts's shadow redefines RegistryInstance as a fictional object shape ({notify, register, unregister}). The real RegistryInstance (packages/core/internal/Registry/Registry.ts) is a callback function type, (ev?: RegistryEvent) => void , completely different from what Registry.create() actually returns (an object: {parent, notify, register}, exported from core as RegistryHandle).

packages/ui/Marquee/MarqueeBase.tsx
```diff
@@ -172,8 +172,8 @@ const MarqueeBase = kind({
 
 			const root = node.parentNode;
 			new global.IntersectionObserver(function (entries, observer) {
-				const {left, right} = entries[0].boundingClientRect;
-				const {left: rootLeft, right: rootRight} = entries[0].rootBounds;
+				const {left, right} = entries[0].boundingClientRect as DOMRectReadOnly;
+				const {left: rootLeft, right: rootRight} = entries[0].rootBounds as DOMRectReadOnly;
```
daniel-stoian-lgp: IntersectionObserverEntry.rootBounds is null whenever the intersection root has no layout box: display: none, not yet attached, a zero-size root, or when the root isn't visible. When that happens inside this observer callback, the destructure throws Cannot read properties of null instead of gracefully bailing 
(Pick whatever fallback is semantically right here : a 0,0 fallback, or an early return before entering the block)

```suggestion
				const {left: rootLeft, right: rootRight} = entries[0].rootBounds ?? {left: 0, right: 0} as DOMRectReadOnly;
```

packages/ui/Placeholder/PlaceholderControllerDecorator.tsx
```diff
@@ -93,13 +95,13 @@ const PlaceholderControllerDecorator = hoc(defaultConfig, (config, Wrapped) => {
 			} else {
 				this.node = this.nodeRef.current;
 				this.bounds = {
-					height: this.node.offsetHeight,
-					width: this.node.offsetWidth
+					height: this.node!.offsetHeight,
+					width: this.node!.offsetWidth
```
daniel-stoian-lgp: Unguarded non-null assertion on a ref that can legitimately be null


this.node comes from this.nodeRef.current, populated only if the wrapped component (WithRef(Wrapped)) actually forwards a DOM ref. Since this HOC is generic over any Wrapped, a function component that doesn't forward refs leaves nodeRef.current null, and this.node!.offsetHeight throws during componentDidMount.


					height: this.node?.offsetHeight ?? 0,
					width: this.node?.offsetWidth ?? 0

packages/ui/Measurable/Measurable.tsx
```diff
@@ -60,15 +60,15 @@ const useMeasurable = () => {
 				resizeObserver.observe(refToMeasure);
 
 				return () => {
-					resizeObserver.disconnect(refToMeasure);
+					(resizeObserver as any).disconnect(refToMeasure);
```
daniel-stoian-lgp, line 63: `as any` masks a real arity mismatch instead of fixing the call

ResizeObserver.prototype.disconnect() takes no arguments, passing refToMeasure is harmless today (native engines ignore extra args) but the type checker correctly flagged it, and the fix was to cast to `any` rather than drop the unnecessary argument.
Would be better to drop the argument. WDYT?

packages/ui/Routable/useLink.ts
```diff
@@ -1,15 +1,15 @@
...
-	const {path: currentPath, navigate} = use(RouteContext) || {};
+	const {path: currentPath, navigate} = use(RouteContext) || ({} as Partial<RouteContextValue>);
 	const handleNavigate = useCallback(
-		({path}) => {
+		({path}: {path: string}) => {
 			if (!navigate) return;
 
 			navigate({
-				path: resolve(currentPath, path)
+				path: resolve(currentPath, path) as string
```
daniel-stoian-lgp: resolve() (packages/ui/Routable/util.ts) is explicitly typed `string | undefined` and can genuinely return undefined. 
Today this is absorbed because the only navigate in RouteContext (Routable.tsx's handleNavigate) re-resolves and gates on `if (path) handler({path})` , but the cast here removes the compiler's ability to catch a future direct consumer of useLink()/RouteContext that skips that check, letting path: undefined flow downstream as if it were valid.


`const resolvedPath = resolve(currentPath, path);
			if (resolvedPath) navigate({path: resolvedPath});`

daniel-stoian-lgp:
packages/ui/Cancelable/Cancelable.tsx:158 (and 8 other HOC-generated classes: AnnounceDecorator/Announce.tsx:77, Changeable/Changeable.tsx:175, FloatingLayer/FloatingLayer.tsx:186, Marquee/MarqueeDecorator.tsx:470, Media/Media.tsx:123, Placeholder/PlaceholderControllerDecorator.tsx:84, Placeholder/PlaceholderDecorator.tsx:82, Routable/Router.tsx:21)


The `[key: string]: any` instance-field index signature is stamped on 9 HOC-generated classes across this diff. Only MarqueeDecorator.tsx carries the "This class is generated inside a hoc() factory..." justification comment defending it, and it's duplicated within that one file (lines 178 and 464), not copy-pasted across files.

Every case from this comments would require a different solution

packages/ui/tsconfig.json
alexandrumorariu: Done

daniel-stoian-lgp (approved):

PR #3454, opened 2026-08-28
NXT-18689: Migrated `ui` package to TypeScript

ion-andrusciac-lgp:
### Checklist

* [ ] I have run automated testing and it is passed

Migrated `ui` package to TypeScript

### Resolution

### Additional Considerations
- `kind()` is a single factory (`kind<P, C, D>(config)`) shared by every component built with it, so `render`/`computed` can only be typed against the caller's real prop shape if that shape (`P`) flows through as a generic parameter. TypeScript can't infer `P` from `render`'s parameter alone — a generic can't be inferred from a callback's parameter position, only from a value actually passed in. `_propTypes: {} as MyProps` exists solely to give TypeScript that value: a property whose sole purpose is to anchor `P` for inference, never read at runtime. Once `P` is known, two more generics ride along: `D` (inferred from the `defaultProps` object literal) and `C` (inferred from `computed`'s function return types). `ApplyDefaults<P, D>` then reconciles them — props with a default are optional in `P` (a caller doesn't have to pass them) but the type guarantees them non-optional inside `computed`/`render`, since `kind()` has already merged the defaults in by the time those functions run. `render`'s final signature (`ApplyDefaults<P, D> & C`) mirrors exactly what `renderKind()` builds at runtime: defaulted props, then computed values merged on top.

- `handle()`'s signature also threads a `context` argument through the whole chain: `prepareHandleArgs(ev, props, context)` passes `context` (typed as React's `Context<any>`, matching what `kind()` resolves via `contextType`/`use(contextType)`) down to every handler function in the chain, alongside `event` and `props` — this is why `HandlerFunction<C>`'s third parameter exists (`context?: C`). It lets a handler bound via `handle()` react to context values the same way `computed`/`render` can, without each handler having to re-derive context itself. 

- since we moved to typescript, there were changes needed to be done in docs-utils

- removed root tsconfig.json since it is not needed

### Links
NXT-18689

### Comments


packages/ui/Slider/PositionDecorator.tsx
```diff
@@ -84,6 +100,7 @@ const PositionDecorator = hoc((config, Wrapped) => {
 			}
 
 			const onChangeObj = {
+				color: {},
```
daniel-stoian-lgp: An empty object is truthy in JS, so any consumer doing if (ev.color) to distinguish a color-picker change from a regular slider change will now always see true, even on plain sliders. 
Fix: only add color to the payload inside the if (colorPicker) branch, as before (via a conditional spread)

packages/ui/Scroller/Scroller.tsx
```diff
@@ -6,90 +6,20 @@
...
-				{isHorizontalScrollbarVisible ? <Scrollbar {...horizontalScrollbarProps} /> : null}
-			</div>
-		</ResizeContext>
-	);
-};
-
-Scroller.displayName = 'ui:Scroller';
-
-Scroller.propTypes = /** @lends ui/Scroller.Scroller.prototype */ {
+export interface ScrollerProps {
```
daniel-stoian-lgp, line 23: let's just define the interface, with no jsdocs, and keep proptypes as-is

Removing protypes affects the scenarios when the consumers are on JS and send worng values for some props (checking proptypes would not be done)

Please do this fix in all 18 files where proptypes were removed. Also check ui/Slider.knob.tsx for a solution for kind components

daniel-stoian-lgp (approved):

packages/ui/package.json
```diff
@@ -34,6 +34,8 @@
   "dependencies": {
     "@enact/core": "^5.6.0",
     "@enact/i18n": "^5.6.0",
+    "@types/ramda": "^0.32.0",
```
dan-ichim-lgp: `@types/ramda` and `@types/warning` (also `typescript` on line 47) are under `dependencies`. Consumers of `@enact/ui` should not install type packages or the compiler at runtime. Move all three to `devDependencies` — they are already duplicated there.

packages/ui/package.json
```diff
@@ -42,11 +44,13 @@
     "react": "^19.2.8",
     "react-dom": "^19.2.8",
     "react-is": "^19.2.8",
+    "typescript": "^7.0.2",
```
dan-ichim-lgp: `typescript` `^7.0.2` is a production dependency. Enact CLI already compiles `.ts`; this package does not need TypeScript at runtime. Move it to `devDependencies`.

packages/ui/Touchable/Touchable.tsx
```diff
@@ -8,17 +8,53 @@
...
+import {holdConfigPropType, HoldConfigPropType} from './Hold';
+import {pinchConfigPropType, PinchConfigPropType} from './Pinch';
+import useTouch, {useTouchConfig} from './useTouch';
+import {Callback} from '../types';
+import PropTypes from 'prop-types';
 
-const selectProps = (props) => {
+export interface TouchableProps {
+	disabled?: boolean
+	dragConfig: DragConfigPropType
```
dan-ichim-lgp: `dragConfig`, `flickConfig`, `holdConfig`, and `pinchConfig` (line 49) are required on `TouchableProps`, but the PropTypes and docs treat them as optional instance overrides. Please mark them optional (`?`).

packages/ui/Touchable/Touchable.tsx
```diff
@@ -8,17 +8,53 @@
...
+	flickConfig: FlickConfigPropType
+	holdConfig: HoldConfigPropType
+	noResume?: boolean
+	onDown?: Callback<void, PointerEvent<HTMLElement>>
+	onDrag?: Callback<any, {type: 'onDrag', x: number, y: number}>
+	onDragEnd?: Callback<any, {type: 'onDragEnd'}>
+	onDragStart: Callback<any, {type: 'onDragStart', x: number, y: number}>
```
dan-ichim-lgp: `onDragStart` is required while `onDrag` / `onDragEnd` are optional. Gesture callbacks all have no default usage requirement — `onDragStart` should be optional too.

packages/ui/VirtualList/VirtualList.tsx
```diff
@@ -10,17 +10,47 @@
...
+export interface ScrollToProps {
+	align?: 'left' | 'right' |'top' | 'bottom' | 'topleft' | 'topright' | 'bottomleft' | 'bottomright';
+	animate?: boolean;
+	focus?: boolean;
+	index?: number;
+	node?: HTMLElement;
+	position?: {x: number, y: number};
+}
+
+export interface VirtualListProps {
```
dan-ichim-lgp, line 31: Every field on `VirtualListProps` is required, but `virtualListDefaultProps` supplies defaults for all of them (`cbScrollTo`, `direction`, `onScroll*`, scrollbar visibility, `role`, `scrollMode`, …). TypeScript consumers would have to pass every defaulted prop. Please mark these optional (or `Partial<>` the public props and keep the defaults object fully required).

packages/core/kind/types.ts
```diff
@@ -1,43 +1,143 @@
...
+ */
+export type PropsWithDOM<P> = P & Omit<HTMLAttributes<HTMLDivElement>, keyof P>;
+
+/**
+ * Configuration for CSS class name mapping.
+ */
 export type StylesBlock = {
-	css: CallbackObject<string>;
+	/** The CSS module map or stylesheet object of the component. */
+	css: CSSProperties;
```
dan-ichim-lgp: `StylesBlock.css` is typed as `CSSProperties`. A CSS module map is `Record<string, string>` (keys like `button`, `layout`), not inline styles. This will reject real `import css from './X.module.less'` maps.

packages/core/kind/types.ts
```diff
@@ -1,43 +1,143 @@
...
+	/** The static local class name to apply to the root element. */
 	className: string;
```
dan-ichim-lgp: `className` and `publicClassNames` are required, but JSDoc and existing `kind()` callers treat them as optional. Callers currently add dummy `publicClassNames: false` or `as StylesBlock` to satisfy tsc. Please make both optional.

packages/ui/Toggleable/Toggleable.tsx
```diff
@@ -11,11 +11,80 @@ import useHandlers from '@enact/core/useHandlers';
...
-import {useRef} from 'react';
+import {ElementType, SyntheticEvent, useRef} from 'react';
 import warning from 'warning';
 
+import {CallbackObject} from '../types';
+
 import {useToggle} from './useToggle';
 
+export interface ToggleableConfig {
+	activate: string;
```
dan-ichim-lgp: `activate` and `deactivate` are typed `string`, but `defaultConfig` sets them to `null` (around line 115). Use `string | null` and make the rest of `ToggleableConfig` optional so it matches the defaults.

packages/ui/Toggleable/Toggleable.tsx
```diff
@@ -11,11 +11,80 @@ import useHandlers from '@enact/core/useHandlers';
...
+	 * Default toggled state applied at construction when the toggled prop is `undefined` or
+	 * `null`.
+	 *
+	 * @name defaultSelected
+	 * @memberof ui/Toggleable.Toggleable.prototype
+	 * @type {Boolean}
+	 * @default false
+	 * @public
+	 */
+	& {[K in PropKey]?: boolean}
```
dan-ichim-lgp, line 58: JSDoc `@name defaultSelected` sits on `{[K in PropKey]}` (`selected` by default). The `selected` docs on line 70 sit on `{[K in DefaultPropKey]}`. The mapped types themselves are correct; the comments are swapped.

packages/core/kind/types.ts
```diff
@@ -1,43 +1,143 @@
...
+	computed?: {
+		[K in keyof C]: (
+			props: ComputedCallbackObject<ApplyDefaults<P, D>>,
+			context: any
+		) => C[K]
+	};
+	/**
+	 * The final render function. It receives a heavily strictly-typed props object
+	 * containing your base props, applied defaults, and extracted computed values.
+	 */
```
dan-ichim-lgp, line 139: `kind()` is now `kind<P, C, D>`. `functionalKind` still takes unparameterized `FunctionalKindConfig` and does not return `KindComponent<P, D>`. Please give it the same generics.

packages/ui/types/callback.type.ts
```diff
@@ -0,0 +1,3 @@
+export type CallbackObject<T = any> = Record<string, T>;
```
dan-ichim-lgp: This file (and `handlerFunction.type.ts`) copies `@enact/core/types`, and names `Callback`’s second param `C` vs `P` in core. Please re-export from core instead of duplicating.

packages/core/kind/types.ts
```diff
@@ -1,43 +1,143 @@
-import {ComponentType, Context, ReactElement} from 'react';
+import {WeakValidationMap} from 'prop-types';
+import {ComponentType, Context, CSSProperties, HTMLAttributes, ReactElement} from 'react';
 
-import {CallbackObject, HandlerFunction} from '../types';
+import {Callback, CallbackObject, HandlerFunction} from '../types';
```
dan-ichim-lgp, line 5: The comment still says `KindComponent` omits the props generic. This PR added `KindComponent<P, DP>`. Please update or delete the workaround type now that `kind()` is generic.

dan-ichim-lgp (approved):

PR #3455, opened 2026-09-01
NXT-19483: Migrate webos to TS

dan-ichim-lgp:
### Checklist

Migrated `@enact/webos` to TypeScript, following the same approach used for `@enact/core` (NXT-18124), `@enact/i18n` (NXT-18126), and `@enact/spotlight` (NXT-18125).

### Resolution

`@enact/webos` was converted from JavaScript to TypeScript: `application`, `deviceinfo`, `keyboard`, `lastInputType`, `LS2Request`, `platform`, `pmloglib`, `speech`, and `signLang` (HOCs as `.tsx`), shared types under `types/`, and package `tsconfig.json`. Nested `package.json` `main` fields now point at the TypeScript sources. Existing unit tests were migrated to TypeScript (`.ts`) and continue to pass.
`tsc --noEmit` and `enact lint --strict` pass. Runtime behavior is intended to remain unchanged; this is a type-safety / maintainability migration, not a functional redesign. JSDoc and PropTypes are preserved.
`tsconfig.json` maps `@enact/webos/*` to `./*` so in-package imports such as `@enact/webos/pmloglib` type-check without installing the package on itself.

### Additional Considerations

### Links
NXT-19483

### Comments
Enact-DCO-1.0-Signed-off-by: Dan Ichim (dan.ichim@lgepartner.com)


packages/webos/deviceinfo/deviceinfo.ts
```diff
@@ -36,11 +37,11 @@ const device = {};
  * @memberof webos/deviceinfo
  * @public
  */
-const deviceinfo = (callback) => {
+const deviceinfo = (callback: (info: DeviceInfo) => void) => {
 	if (Object.keys(device).length === 0) {
 		try {
 			const webOSSystem = window.webOSSystem ?? window.PalmSystem;
-			const info = JSON.parse(webOSSystem.deviceInfo);
+			const info = JSON.parse(webOSSystem!.deviceInfo as string);
```
ion-andrusciac-lgp: Maybe it would be safer to extend the Window interface globally so TypeScript understands these properties, and use optional chaining (?.) (similar to i18n or spotlight)?

packages/webos/deviceinfo/deviceinfo.ts
```diff
@@ -71,12 +72,12 @@ const deviceinfo = (callback) => {
 					}
 					if (response.firmwareVersion) {
 						device.version = response.firmwareVersion;
-						const segments = device.version.split('.');
-						const keys = ['versionMajor', 'versionMinor', 'versionDot'];
+						const segments = (device.version as string).split('.');
```
ion-andrusciac-lgp, line 75: Maybe add an optional `firmwareVersion` prop to the `LS2Response` type to avoid casting as a string?

packages/webos/platform/platform.ts
```diff
@@ -27,30 +29,36 @@ const webOSVersion = [
...
-	const platformInfo = {webos: false};
+	const platformInfo: WebOSPlatform = {webos: false};
 
 	if (userAgent.indexOf('SmartTV') > -1) {
 		platformInfo.tv = true;
 	} else {
 		const webOSSystem = window.webOSSystem ?? window.PalmSystem;
 		try {
-			let legacyInfo = JSON.parse(webOSSystem.deviceInfo || '{}');
+			let legacyInfo = JSON.parse(webOSSystem!.deviceInfo || '{}');
```
ion-andrusciac-lgp: I think if the Window interface is extended globally, then here we can use the optional operator, which will be safer

packages/webos/signLang/SignLangDecorator.tsx
```diff
@@ -79,49 +89,49 @@ const SignLangDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
-			this.signLangDelayId = null;
 		}
 
-		componentDidUpdate (prevProps) {
+		componentDidUpdate (prevProps: SignLangDecoratorProps) {
 			checkPropTypes(this, this.props, prevProps);
 		}
 
-		requestSignLang = (active) => {
+		signLangDelayId: ReturnType<typeof setTimeout> | null = null;
```
ion-andrusciac-lgp, line 101:
```suggestion
		signLangDelayId: ReturnType<typeof setTimeout> | undefined;
```

packages/webos/signLang/SignLangDecorator.tsx
```diff
@@ -79,49 +89,49 @@ const SignLangDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
+		requestSignLang = (active: boolean) => {
 			const {signLangId = '', signLangOption = {}} = this.props;
 
 			if (signLangId.length > 0) {
 				if (active) {
 					if (signLangDelay === 0) {
 						startSignLang(signLangId, signLangOption);
 					} else {
-						clearTimeout(this.signLangDelayId);
+						clearTimeout(this.signLangDelayId as unknown as number);
```
ion-andrusciac-lgp, line 111:
```suggestion
						clearTimeout(this.signLangDelayId);
```

packages/webos/signLang/SignLangDecorator.tsx
```diff
@@ -79,49 +89,49 @@ const SignLangDecorator = hoc(defaultConfig, (config, Wrapped) => {
...
 						this.signLangDelayId = setTimeout(() => {
 							startSignLang(signLangId, signLangOption);
 						}, signLangDelay);
 					}
 				} else {
-					if (signLangDelay > 0) clearTimeout(this.signLangDelayId);
+					if (signLangDelay > 0) clearTimeout(this.signLangDelayId as unknown as number);
```
ion-andrusciac-lgp, line 117:
```suggestion
					if (signLangDelay > 0) clearTimeout(this.signLangDelayId);
```

packages/webos/deviceinfo/deviceinfo.ts
ion-andrusciac-lgp, line 44:
```suggestion
			const info = JSON.parse(webOSSystem?.deviceInfo || '{}');
```

packages/webos/platform/platform.ts
ion-andrusciac-lgp, line 55:
```suggestion
			let legacyInfo = JSON.parse(webOSSystem?.deviceInfo || '{}');
```

ion-andrusciac-lgp (approved):

packages/webos/deviceinfo/deviceinfo.ts
daniel-stoian-lgp: I think we need to revert this to the previous code Dan commited.

Original JS was `JSON.parse(webOSSystem.deviceInfo)`. When neither `window.webOSSystem` nor `window.PalmSystem` exists, that line threw error, which was deliberately caught by the surrounding try/catch to set the fallback:
} catch {
    device.modelName = device.modelNameAscii = 'webOS Device';
}

The ?. + || '{}' rewrite makes JSON.parse('{}') succeed instead of throwing, so the catch block  never runs anymore. device.modelName/modelNameAscii now silently come back undefined instead of 'webOS Device'.

